PR #13, opened 2019-08-15
auth: add ldap support

rchincha:


hallyn:
Looks like a good start, thanks.

pkg/api/auth.go
```diff
@@ -20,46 +25,89 @@ func authFail(w http.ResponseWriter, realm string, delay int) {
...
 }
 
+// nolint (gocyclo) - we use closure making this a complex subroutine
 func BasicAuthHandler(c *Controller) mux.MiddlewareFunc {
 	realm := c.Config.HTTP.Realm
 	if realm == "" {
 		realm = "Authorization Required"
 	}
 	realm = "Basic realm=" + strconv.Quote(realm)
-	delay := c.Config.HTTP.Auth.FailDelay
```
hallyn: Just to make sure, was this also intended (switching it for '5')?

hallyn: Oh, duh.  ignore.

pkg/api/auth.go
```diff
@@ -20,46 +25,89 @@ func authFail(w http.ResponseWriter, realm string, delay int) {
...
 				}
-
 				// Process request
 				next.ServeHTTP(w, r)
 			})
 		}
 	}
 
 	credMap := make(map[string]string)
+	delay := c.Config.HTTP.Auth.FailDelay
```
hallyn: Ok but based on the above logic, note that c.Config.HTTP.Auth.FailDelay could still be nil?

pkg/api/auth.go
```diff
@@ -20,46 +25,89 @@ func authFail(w http.ResponseWriter, realm string, delay int) {
...
 	realm := c.Config.HTTP.Realm
 	if realm == "" {
 		realm = "Authorization Required"
 	}
 	realm = "Basic realm=" + strconv.Quote(realm)
-	delay := c.Config.HTTP.Auth.FailDelay
 
-	if c.Config.HTTP.Auth.HTPasswd.Path == "" {
+	// no password based authN, if neither LDAP nor HTTP BASIC is enabled
+	if c.Config.HTTP.Auth == nil || (c.Config.HTTP.Auth.HTPasswd.Path == "" && c.Config.HTTP.Auth.LDAP == nil) {
```
rchincha: @hallyn either c.Config.HTTP.Auth is nil which means it enters and returns, or FailDelay is not specified and defaults to 0 (nil value for int type) but it is otherwise configurable.

pkg/api/auth.go
rchincha: See comment above

PR #20, opened 2019-08-28
docs: fix Makefile

rchincha:


tych0:
It would be nice to have a way to build the binary without swag, even if some of the stuff (e.g. docs or whatever) wouldn't work.

rchincha:
> 
> 
> It would be nice to have a way to build the binary without swag, even if some of the stuff (e.g. docs or whatever) wouldn't work.

The bits in the docs/ dir are autogenerated. We can probably fix this with bazel.

tych0:
Are you sure they don't? per my comment above, I think the PHONY target indirection will cause them to be rebuilt.

rchincha:
> 
> 
> Are you sure they don't? per my comment above, I think the PHONY target indirection will cause them to be rebuilt.

PHONY(doc) depends on real(docs/doc.go), so yes that target is evaluated, but real target will not be rebuilt. Best to try out this patch.

tych0:
On Thu, Aug 29, 2019 at 11:58:37AM -0700, Ramkumar Chinchani wrote:
> > 
> > 
> > Are you sure they don't? per my comment above, I think the PHONY target indirection will cause them to be rebuilt.
> 
> PHONY(doc) depends on real(docs/doc.go), so yes that target is evaluated, but real target will not be rebuilt. Best to try out this patch.

Ah, yep, I see. It still would be nice to have a way to build the
binary without having to install "swag", but otherwise this LGTM.

PR #25, opened 2019-09-16
version: add commit id to binary during build

rchincha:


Makefile
```diff
@@ -1,5 +1,6 @@
 export GO111MODULE=on
 TOP_LEVEL=$(shell git rev-parse --show-toplevel)
+GIT_COMMIT=$(shell git rev-list -1 HEAD)
```
tych0: We use,

COMMIT_HASH=$(shell git describe --always --tags --long)
COMMIT=$(if $(shell git status --porcelain --untracked-files=no),$(COMMIT_HASH)-dirty,$(COMMIT_HASH))

Which I think gives a little better output than this. Feel free to change it or not :)

rchincha: Fixed.

PR #27, opened 2019-09-20
auth: add LDAP support

### LINKED ISSUE #23: add LDAP authentication support
Currently, zot supports only mutual TLS and htpasswd authN. Would be useful to add LDAP support so that can hook up to Windows AD (for example)

rchincha:
fixes #23


rchincha:
RFC 4513 section 5.1.2

Unauthenticated Bind operations can have significant security issues
 (see Section 6.3.1).  In particular, users intending to perform
 Name/Password Authentication may inadvertently provide an empty
 password and thus cause poorly implemented clients to request
 Unauthenticated access.  Clients SHOULD be implemented to require
 user selection of the Unauthenticated Authentication Mechanism by
 means other than user input of an empty password.  Clients SHOULD
 disallow an empty password input to a Name/Password Authentication
 user interface.  Additionally, Servers SHOULD by default fail
 Unauthenticated Bind requests with a resultCode of
 unwillingToPerform.

PR #35, opened 2019-11-08
ldap: setup a new ldap conn on bind failure

rchincha:


pkg/api/ldap.go
```diff
@@ -85,6 +85,7 @@ func (lc *LDAPClient) Authenticate(username, password string) (bool, map[string]
 	if lc.BindDN != "" && lc.BindPassword != "" {
 		err := lc.Conn.Bind(lc.BindDN, lc.BindPassword)
 		if err != nil {
+			lc.Conn = nil
```
hallyn: Is there any kind of lc.Conn.Close() you need to do?

rchincha: Likely to fail in this particular case, but we can attempt it.

tych0: Even if part of it fails, it still may release some resources, so worth doing I think.

PR #36, opened 2019-11-18
ldap: improve recovery when connection failures

rchincha:


pkg/api/ldap.go
```diff
@@ -69,28 +72,49 @@ func (lc *LDAPClient) Connect() error {
 	return nil
 }
 
+func sleepAndRetry(retries, maxRetries int) bool {
+	if retries > maxRetries {
+		panic("invalid retries")
```
hallyn: is panicking ever the right thing?

rchincha: Well, only when the function behavior is "undefined" because some invariant got violated.
https://kernelnewbies.org/FAQ/BUG

hallyn: But that doesn't seem to be the case.  The loop from which you call sleepAndRetry() doesn't prevent calling it when retries hits maxRetries.  So if connection continues to fail, then you are guaranteed to hit this check.  In which case, in the prior version you would return an error rather than panic().
The kernel bug is intended to be used for conditions which are never intended to be hit.  This looks like it's intended to be hit on persistent ldap server error.

rchincha: ah! :) pushed an updated commit.

hallyn:
Great - thanks.

PR #37, opened 2019-11-25
log: improve logging

rchincha:


pkg/api/routes.go
```diff
@@ -213,13 +214,14 @@ func (rh *RouteHandler) GetManifest(w http.ResponseWriter, r *http.Request) {
 		case errors.ErrManifestNotFound:
 			WriteJSON(w, http.StatusNotFound, NewError(MANIFEST_UNKNOWN, map[string]string{"reference": reference}))
 		default:
+			rh.c.Log.Error().Err(err).Msg("unexpected error")
 			w.WriteHeader(http.StatusInternalServerError)
 		}
 		return
 	}
 
-	WriteData(w, http.StatusOK, mediaType, content)
```
rchincha: This is likely the root cause of the bug that we are chasing.

hallyn: Sorry, I looked at this earlier and was confused.  Now I see why.  Shouldn't this be a separate, 3-line patch?

rchincha: ok, let's split this.

PR #38, opened 2019-11-26
reorder header and data writes

rchincha:


rchincha:
> 
> 
> Thanks. Looks good, other than empty commit message :)

```
Nov 25 12:33:34 aci-zot zot[8368]: 2019/11/25 12:33:34 http: superfluous response.WriteHeader call from github.com/anuvu/zot/pkg/api.(*statusWriter).WriteHeader (log.go:40)
```

This patch tries to get rid of this msg.

PR #39, opened 2019-12-11
refactor: make zot 'go get'-able

rchincha:
1) github.com/jtblin/go-ldap-client is an abandoned project and not keeping
up with go modules, so pull that code in (single file, BSD license)

2) minor refactor ldap as a result of 1)

3) commit docs which should change only with dist-spec changes

4) update go.mod


tych0:
Might be nice to have these as separate commits since they're separate logical changes.

Other than that, if the LDAP library is abandoned, perhaps we should look into switching to a different one?

rchincha:
> 
> 
> Might be nice to have these as separate commits since they're separate logical changes.
> 
> Other than that, if the LDAP library is abandoned, perhaps we should look into switching to a different one?

This LDAP library was the easiest one to use in terms of interfaces. We could explore that.

tych0:
> Ironically, it is the LDAP library that is breaking the "go get" behavior we want, so a cascade effect :\

Ok, but can't we split the commits? I don't see how it's related.

rchincha:
> 
> 
> > Ironically, it is the LDAP library that is breaking the "go get" behavior we want, so a cascade effect :\
> 
> Ok, but can't we split the commits? I don't see how it's related.

I can probably do two splits - 1) ldap + go.mod changes, and 2) the doc changes.

rchincha:
> 
> 
> > > Ironically, it is the LDAP library that is breaking the "go get" behavior we want, so a cascade effect :\
> > 
> > 
> > Ok, but can't we split the commits? I don't see how it's related.
> 
> I can probably do two splits - 1) ldap + go.mod changes, and 2) the doc changes.

Do you want separate PRs for these though?

tych0:
Nah, this lgtm!

PR #41, opened 2019-12-12
update .gitignore

rchamarthy:
Signed-off-by: Ravi Chamarthy <ravchama@cisco.com>


rchincha:
https://github.com/anuvu/zot/issues/15
This needs fixing first and htpasswd file needs to be generated before we can gitignore test/data

PR #55, opened 2020-01-06
routes: be more paranoid

rchincha:
once we have an non-EOF error in this path, not clear if we can really
recover.


pkg/api/routes.go
```diff
@@ -921,6 +921,7 @@ func WriteDataFromReader(w http.ResponseWriter, status int, length int64, mediaT
 		} else if err != nil {
 			// other kinds of intermittent errors can occur, e.g, io.ErrShortWrite
 			logger.Error().Err(err).Msg("copying data into http response")
+			panic(err)
```
rchamarthy: Isn’t there a log.Panic?

PR #57, opened 2020-01-07
go.mod: update json-iterator

rchincha:
older versions of json-iterator appear to have an issue with maps

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/gofrs/uuid v3.2.0+incompatible
 	github.com/gorilla/handlers v1.4.2
 	github.com/gorilla/mux v1.7.3
-	github.com/json-iterator/go v1.1.8
+	github.com/json-iterator/go v1.1.9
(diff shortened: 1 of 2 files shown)


tych0:
Can you describe the issue? Is there a bug report somewhere, or a stack trace?

rchincha:
> 
> 
> Can you describe the issue? Is there a bug report somewhere, or a stack trace?

We are most certainly hitting this panic.
https://github.com/anuvu/zot/blob/master/pkg/api/routes.go#L898

On the json-iterator/go side,
https://github.com/json-iterator/go/commit/acfec88f7a0d5140ace3dcdbee10184e3684a9e1
https://github.com/json-iterator/go/issues/388

Note that this fix is still under test.

tych0:
Ok, this would all be good to have in the commit message.

rchincha:
> 
> 
> Ok, this would all be good to have in the commit message.

Updated the PR. However, let's wait for your tests to complete.

PR #59, opened 2020-01-15
authN: first try local htpasswd then LDAP

rchincha:
We are noticing that LDAP connectivity issues and timeouts can affect
authN behavior from client side (which can timeout as well).

Instead, put local authN first so at least we have a reliable authN
method.

But, the caveat is that it is best if the local and LDAP user list
doesn't overlap.


tych0:
Cool, thanks :)

PR #61, opened 2020-01-16
compliance: be compliant with dist-spec compliance tests

rchincha:
dist-spec compliance tests are now becoming a part of dist-spec repo
itself - we want to be compliant

pkg/api/regex.go:
	* revert uppercasing in repository names

pkg/api/routes.go:
	* ListTags() should support the URL params 'n' and 'last'
	  for pagination

	* s/uuid/session_id/g to use the dist-spec's naming

	* Fix off-by-one error in GetBlobUpload()'s http response "Range" header

	* DeleteManifest() success status code is 202

	* Fix PatchBlobUpload() to account for "streamed" use case
	  where neither "Content-Length" nor "Content-Range" headers are set

pkg/storage/storage.go:
	* Add a "streamed" version of PutBlobChunk() called PutBlobChunkStreamed()

pkg/compliance/v1_0_0/check.go:
	* fix unit tests to account for changed response status codes

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -17,6 +17,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"path"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -35,6 +36,7 @@ const (
 	DistContentDigestKey = "Docker-Content-Digest"
 	BlobUploadUUID       = "Blob-Upload-UUID"
 	DefaultMediaType     = "application/json"
+	BinaryMediaType      = "application/octet-stream"
 )
 
 type RouteHandler struct {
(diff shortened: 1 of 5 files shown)


rchincha:
This PR supersedes https://github.com/anuvu/zot/pull/60

pkg/storage/storage.go
```diff
@@ -548,6 +559,39 @@ func (is *ImageStore) GetBlobUpload(repo string, uuid string) (int64, error) {
...
+		blobUploadPath,
+		os.O_WRONLY|os.O_CREATE,
+		0600,
+	)
+	if err != nil {
+		is.log.Fatal().Err(err).Msg("failed to open file")
+	}
+	defer file.Close()
+
+	if _, err := file.Seek(0, 2); err != nil {
```
tych0: Can we use os.SEEK_END here?

pkg/api/routes.go
```diff
@@ -124,12 +129,86 @@ func (rh *RouteHandler) ListTags(w http.ResponseWriter, r *http.Request) {
...
+
+	if ok {
+		if len(nQuery) != 1 {
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+
+		var n1 int64
+
+		if n1, err = strconv.ParseInt(nQuery[0], 10, 0); err != nil {
```
tych0: I think you can just use Atoi() if you want.

pkg/storage/storage.go
rchincha: os.SEEK* deprecated apparently, use io.Seek*

pushing an updated PR

pkg/api/routes.go
rchincha: I suppose, but staying consistent with the usage in other places in code.

pkg/storage/storage.go
tych0: Oh, right, thanks!

PR #64, opened 2020-01-24
Dockerfile for running zot server

### LINKED ISSUE #63: Add Dockerfile for running zot server
There is a Dockerfile in the repo, but it is only used for development purposes

jdolitsky:
Resolves #63


jdolitsky:
@hallyn thanks for review. 

Is the confusion with the line in README
>By default, the server will use no auth, listen on port 5000, and store content at `/var/lib/registry`.

? Maybe better stated as 
> ... stores content at `/var/lib/registry` (inside the container)

?

hallyn:
> @hallyn thanks for review.
> 
> Is the confusion with the line in README
> 
> > By default, the server will use no auth, listen on port 5000, and store content at `/var/lib/registry`.
> 
> ? Maybe better stated as
> 
> > ... stores content at `/var/lib/registry` (inside the container)
> 
> ?

As a user, the important thing to me would be that it's at ./registry on the host.  Container can do what it wants.  So I would say something like "stores content at ./registry (bind mounted to /var/lib/registry in the container)", or just drop the last part and say "stores the content at the current path under registry".  But again I'm probably abnormal, so pick the one you like best - or leave it - and let me know and I'll merge :)

thanks.

jdolitsky:
@hallyn - I've just pushed again to try to clarify. Feel free to merge then edit yourself, I will not be offended :)

hallyn:
Thanks!

PR #65, opened 2020-01-24
Add support for bearer/token auth

### LINKED ISSUE #24: add token-based authentication
Currently, supports only "BASIC", need "BEARER token" support as well.
Also, so we don't clobber LDAP server if configured as authN backend.

pmengelbert:
New options added to configuration file to reference a public key used
to validate authorization tokens signed by an auth server with
corresponding private key.

Resolves #24

Signed-off-by: Peter Engelbert


rchincha:
@pmengelbert - thanks a lot for the PR

pmengelbert:
> @pmengelbert - thanks a lot for the PR

Least we can do!  Thanks for the support and inspiration on the conformance testing

pkg/api/controller_test.go
```diff
@@ -782,3 +806,165 @@ func TestBasicAuthWithLDAP(t *testing.T) {
...
+
+		resp, err := resty.R().Post(BaseURL3 + "/v2/" + AuthorizedNamespace + "/blobs/uploads/")
+		So(err, ShouldBeNil)
+		So(resp, ShouldNotBeNil)
+		So(resp.StatusCode(), ShouldEqual, 401)
+
+		authorizationHeader := parseBearerAuthHeader(resp.Header().Get("Www-Authenticate"))
+		resp, err = resty.R().
+			SetQueryParam("service", authorizationHeader.Service).
+			SetQueryParam("scope", "repository:"+AuthorizedNamespace+":push,pull").
```
jdolitsky: I believe if the zot auth layer is working properly, this should just be
```
SetQueryParam("scope", authorizationHeader.Scope)
```

(and below also). May need to also get another token fro the GET request (may have different scopes, push vs. pull)

pkg/api/auth.go
```diff
@@ -13,19 +13,65 @@ import (
...
+
+func bearerAuthHandler(c *Controller) mux.MiddlewareFunc {
+	authorizer, err := auth.NewAuthorizer(&auth.AuthorizerOptions{
+		Realm:           c.Config.HTTP.Auth.Bearer.Realm,
+		Service:         c.Config.HTTP.Auth.Bearer.Service,
+		PublicKeyPath:   c.Config.HTTP.Auth.Bearer.Cert,
+		AccessEntryType: bearerAuthDefaultAccessEntryType,
+	})
+	if err != nil {
+		panic(err)
```
rchincha: c.Log.Panic().Err(err).Msg("...")

pkg/api/auth.go
```diff
@@ -13,19 +13,65 @@ import (
...
-func authFail(w http.ResponseWriter, realm string, delay int) {
-	time.Sleep(time.Duration(delay) * time.Second)
-	w.Header().Set("WWW-Authenticate", realm)
-	w.Header().Set("Content-Type", "application/json")
-	WriteJSON(w, http.StatusUnauthorized, NewError(UNAUTHORIZED))
+const (
+	bearerAuthDefaultAccessEntryType = "repository"
+)
+
+func AuthHandler(c *Controller) mux.MiddlewareFunc {
```
rchincha: Does this we can only "bearer" or "basic" but never together?

pkg/api/auth.go
pmengelbert: Noted and changed locally.  Will push an update once we have addressed the other comments raised

PR #76, opened 2020-02-24
dedupe: use hard links to dedupe blobs

rchincha:
As the number of repos and layers increases, the greater the probability
that layers are duplicated. We dedupe using hard links when content is
the same. This is intended to be purely a storage layer optimization.
Access control when available is orthogonal this optimization.

Also, note that we designed dedupe to make slow and steady progress in the
background so that zot can continue to serve layers/images while dedupe
is happening.

Add a durable cache to help speed up layer lookups.

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -37,22 +37,48 @@ type ImageStore struct {
 	rootDir     string
 	lock        *sync.RWMutex
 	blobUploads map[string]BlobUpload
+	cache       *Cache
 	log         zerolog.Logger
 }
 
 // NewImageStore returns a new image store backed by a file storage.
 func NewImageStore(rootDir string, log zlog.Logger) *ImageStore {
-	is := &ImageStore{rootDir: rootDir,
+	if _, err := os.Stat(rootDir); os.IsNotExist(err) {
+		if err := os.MkdirAll(rootDir, 0700); err != nil {
+			log.Error().Err(err).Str("rootDir", rootDir).Msg("unable to create root dir")
+			return nil
+		}
+	}
+
+	is := &ImageStore{
+		rootDir:     rootDir,
 		lock:        &sync.RWMutex{},
 		blobUploads: make(map[string]BlobUpload),
+		cache:       NewCache(rootDir, "cache", log),
 		log:         log.With().Caller().Logger(),
 	}
 
-	if _, err := os.Stat(rootDir); os.IsNotExist(err) {
-		_ = os.MkdirAll(rootDir, 0700)
-	}
-
 	return is
+}
+
+// RLock read-lock
+func (is *ImageStore) RLock() {
+	is.lock.RLock()
+}
+
+// RUnlock read-unlock
+func (is *ImageStore) RUnlock() {
+	is.lock.RUnlock()
+}
+
+// Lock write-lock
+func (is *ImageStore) Lock() {
+	is.lock.Lock()
+}
+
+// Unlock write-unlock
+func (is *ImageStore) Unlock() {
+	is.lock.Unlock()
 }
 
 // InitRepo creates an image repository under this store.
@@ -63,16 +89,10 @@ func (is *ImageStore) InitRepo(name string) error {
 		return nil
 	}
 
-	// create repo dir
-	ensureDir(repoDir)
-
 	// create "blobs" subdir
-	dir := path.Join(repoDir, "blobs")
-	ensureDir(dir)
-
+	ensureDir(path.Join(repoDir, "blobs"), is.log)
 	// create BlobUploadDir subdir
-	dir = path.Join(repoDir, BlobUploadDir)
-	ensureDir(dir)
+	ensureDir(path.Join(repoDir, BlobUploadDir), is.log)
 
 	// "oci-layout" file - create if it doesn't exist
 	ilPath := path.Join(repoDir, ispec.ImageLayoutFile)
@@ -85,7 +105,8 @@ func (is *ImageStore) InitRepo(name string) error {
 		}
 
 		if err := ioutil.WriteFile(ilPath, buf, 0644); err != nil {
-			is.log.Panic().Err(err).Str("file", ilPath).Msg("unable to write file")
+			is.log.Error().Err(err).Str("file", ilPath).Msg("unable to write file")
+			return err
 		}
 	}
 
@@ -101,7 +122,8 @@ func (is *ImageStore) InitRepo(name string) error {
 		}
 
 		if err := ioutil.WriteFile(indexPath, buf, 0644); err != nil {
-			is.log.Panic().Err(err).Str("file", indexPath).Msg("unable to write file")
+			is.log.Error().Err(err).Str("file", indexPath).Msg("unable to write file")
+			return err
 		}
 	}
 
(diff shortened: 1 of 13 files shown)


pkg/cache/cache.go
```diff
@@ -0,0 +1,90 @@
...
+
+// Blob is a blob record
+type Blob struct {
+	Path string
+}
+
+func NewCache(rootDir string, name string, log zlog.Logger) *Cache {
+	db, err := bolt.Open(path.Join(rootDir, name+".db"), 0600, nil)
+	if err != nil {
+		panic(err)
```
tych0: Let's return an error here (and elsewhere) instead; that way calling code can decide what to do instead of just panic.

pkg/cache/cache.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	return &Cache{db: db, log: log}
+}
+
+func (c *Cache) PutBlob(digest string, path string) {
+	if err := c.db.Update(func(tx *bolt.Tx) error {
+		root := tx.Bucket([]byte(BlobsCache))
+		if root == nil {
+			panic("")
```
tych0: Seems like we should say *something* useful here :)

pkg/cache/cache.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	if len(blobPath.String()) == 0 {
+		return "", nil
+	}
+
+	return blobPath.String(), nil
+}
+
+func (c *Cache) DeleteBlob(digest string, path string) {
```
tych0: Unused?

pkg/storage/storage.go
```diff
@@ -140,12 +147,13 @@ func (is *ImageStore) ValidateRepo(name string) (bool, error) {
 
 		found[file.Name()] = true
 	}
-
-	for k, v := range found {
-		if !v && k != BlobUploadDir {
-			return false, nil
+	/*
+		for k, v := range found {
```
tych0: We can drop this I guess?

pkg/storage/storage.go
```diff
@@ -865,3 +885,84 @@ func ensureDir(dir string) {
...
+			// validate blob digest
+			f, err := os.Open(blobPath)
+			if err != nil {
+				is.log.Error().Err(err).Str("blob", blobPath).Msg("failed to open blob")
+				return errors.ErrUploadNotFound
+			}
+
+			blobDigest, err := godigest.FromReader(f)
+			if err != nil {
+				is.log.Panic().Err(err).Str("blobPath", blobPath).Msg("unable to calculate hash")
```
tych0: I think we leak f here.

hallyn:
So as part of dedup'ing you are re-shasumming all the files.  While in general that may sometimes be a good thing to do to verify there's been no corruptoin, it seems like it should be done as part of a different process, and opt-outable.
All we should really need to do here is check the filenames.  We can cache based on (filename, size), and for good measure we can go ahead - ONCE we find a duplicate - that both have the same shasum, before deduping.
But doing this now for every dedup run seems like an insane amount of unnecessary disk i/o.

(If you feel strongly about this, please do argue;  I might be thinking wrongly)

pkg/storage/storage.go
```diff
@@ -767,18 +788,46 @@ func (is *ImageStore) FullBlobUpload(repo string, body io.Reader, digest string)
...
+		is.log.Panic().Err(err).Str("blobPath", dst).Msg("unable to lookup blob record")
+	}
+
+	if dstRecord == "" {
+		if err := is.cache.PutBlob(dstDigest.String(), dst); err != nil {
+			is.log.Panic().Err(err).Str("blobPath", dst).Msg("unable to insert blob record")
+		}
+
+		// move the blob from uploads to final dest
+		_ = os.Rename(src, dst)
```
hallyn: If we risk actually losing blobs here, logging the error seems worthwhile.

pkg/storage/storage.go
```diff
@@ -124,7 +147,7 @@ func (is *ImageStore) ValidateRepo(name string) (bool, error) {
 		return false, errors.ErrRepoNotFound
 	}
 
-	if len(files) != 4 {
+	if len(files) < 3 {
```
hallyn: I think I understand why s/!= 4/< 3/, but the comment at the top of this function should be updated to reflect this change, as it still says we expect exactly 4 entries.

pkg/storage/storage.go
```diff
@@ -195,7 +212,7 @@ func (is *ImageStore) GetRepositories() ([]string, error) {
 			return nil
 		}
 
-		is.log.Debug().Str("dir", path).Str("name", info.Name()).Msg("found image store")
+		//is.log.Debug().Str("dir", path).Str("name", info.Name()).Msg("found image store")
```
hallyn: Are you intending to uncomment this in the future?

pkg/cache/cache.go
rchincha: Left some panic()s in coz that is assured to not happen as per the package. But safeguards us in case package versions change behavior.

pkg/storage/storage.go
rchincha: Yes, once we deploy and learn about how stable the new code is, may have to bring this back in.

pkg/storage/storage.go
rchincha: Handling the return error here and one other place.

hallyn:
@rchincha - what changed with the new commits since I approved it?

@tych0 how are you feeling about this?

rchincha:
> 
> 
> @rchincha - what changed with the new commits since I approved it?
> 
> @tych0 how are you feeling about this?


@codecov
codecov/patch — 65.33% of diff hit (target 68.48%) 

Trying to hit that target with more unit tests. Getting there.
The code itself hasn't changed.

tych0:
I still feel like all these panic()s are the wrong thing. Take the new function that is the point of this PR: DedupeBlob(). If something goes wrong in there, the user doesn't see a nice 500, they just get the connection closed on them.

rchincha:
> 
> 
> I still feel like all these panic()s are the wrong thing. Take the new function that is the point of this PR: DedupeBlob(). If something goes wrong in there, the user doesn't see a nice 500, they just get the connection closed on them.

A failure here is catastrophic. Returning 500s is ok as long as any operations on the affected blobs continues to return 500s.

rchincha:
I am updating the PR mainly to improve coverage and let codecov allow me to pass.

tych0:
> A failure here is catastrophic.

Consider how this looks as a user: someone is copying something over, and then zot closes, and the kernel closes the TCP connection. Then they more data to that connection (i.e. an ack), and they get an RST, because they're in the reset state. That manifests as "connection reset", which then triggers them to ping you in IRC, and we all look for network errors for a while. A panic() here is simply incorrect. It's not a matter of taste.

hallyn:
So in particular, certainly inability by bolt to create a DB which is only used for dedup, is not fatal to zot.

Likewise, any failure by boltdb should not be a fatal error.

A failure in an image write path, like mkdir for new repo, should probably place zot into a RO mode.

rchincha:
> 
> 
> So in particular, certainly inability by bolt to create a DB which is only used for dedup, is not fatal to zot.
> 
> Likewise, any failure by boltdb should not be a fatal error.
> 
> A failure in an image write path, like mkdir for new repo, should probably place zot into a RO mode.

A fallback-readonly-mode is becoming very useful (disk-full, etc also), although we could still get 500s. Let me make this change as well.

hallyn:
> > So in particular, certainly inability by bolt to create a DB which is only used for dedup, is not fatal to zot.
> > Likewise, any failure by boltdb should not be a fatal error.
> > A failure in an image write path, like mkdir for new repo, should probably place zot into a RO mode.
> 
> A fallback-readonly-mode is becoming very useful (disk-full, etc also), although we could still get 500s. Let me make this change as well.

Honestly a fallback readonly mode can be a lower priority separate PR.  For now just making sure that we always just return errors, don't panic, and don't stop running is good enough.  At worst, it'll mean a lot of repeated attempts to write which return failure.

hallyn:
As Tycho pointed out yesterday, a separate RO mode would just, through extra state, increase the likelyhood of a bug, so let's instead stick to return an error to the caller when write fails.  de facto ro.

I still see a panic or two, once those are gone we can merge.

PR #77, opened 2020-03-05
zot: run GC after manifest removal

tych0:
Clients today expect the repo to clean up if there are unused blobs, not to
manually delete things they think are unused. Let's do that, and use
umoci's code to do it since it's tested and works.

Signed-off-by: Tycho Andersen <tycho@tycho.ws>


hallyn:
Now is that run locked against other actions on the same repo?  I.e. if someone else starts adding an image a moment after you've started the umoci.Clean for your delete, will you risk removing blobs which the add is expecting to be there (bc it had done a HEAD)?

hallyn:
(Note that the distribution spec specifically says you should push the blobs first, then push the manifest referencing them, so the blobs could be GC'd in the meantime)

pkg/storage/storage_test.go
```diff
@@ -135,8 +135,6 @@ func TestAPIs(t *testing.T) {
 					So(err, ShouldNotBeNil)
 				})
 
-				err = il.DeleteBlob("test", d.String())
-				So(err, ShouldBeNil)
```
rchincha: Since the GC is happening inline, this should actually be a non-nil error?

pkg/storage/storage.go
```diff
@@ -502,6 +504,16 @@ func (is *ImageStore) DeleteImageManifest(repo string, reference string) error {
 		return err
 	}
 
+	oci, err := umoci.OpenLayout(dir)
+	if err != nil {
+		return err
+	}
+	defer oci.Close()
+
+	if err := oci.GC(context.Background()); err != nil {
```
rchincha: With dedupe, delete should be very fast since we are removing links for most part.

tych0:
I think you're right that there are some races here. The only thing I can think is to lock while GC is running, and stop any parallel blob Put()s?

hallyn:
So longer term, maybe the way to go about this is (a) keep all the blobs in one shared directory, hardlinking them into repo blob dirs as needed.  (b) when we gc, take all shared blobs which have nlink=1 and move them into purgatory/ ,  (c) when a blob in purgatory is checked for (HEAD) , move it back to shared, (d) when a manifest is uploaded, any blobs it refers to that are in purgatory are moved back, and then (d) after some time (an hour?) rm -rf jigoku; mv purgatory/ jigoku/ .  That way we won't keep any blobs around too long but avoid ever losing something that's in the process of being uploaded.

Shorter term what you (Tycho) have might work, I need to re-read.

hallyn:
No, I don't think what you have works.  The problem is that we'd have to lock across the *client's* 'upload' routine, which from our point of view i just a bunch of HEAD gets followed by the manifest PUT.  Just locking those individually does not prevent the case of:

```
HEAD b1 -> ok
HEAD b2 -> absent
POST b2
HEAD b3 -> ok
[gc runs, b1 gets deleted]
POST manifest
```

tych0:
Posting a manifest with non-present blobs will be rejected, won't it? So I
don't think this scenario is possible.

hallyn:
> Posting a manifest with non-present blobs will be rejected, won't it? So I don't think this scenario is possible.

All right, yes, I guess that pkg/storage/storage.go:PutImageManifest() will return an ErrBlobNotFound in that case.  That's #notnice, and I don't think any client will not then assume that the repo must be busted, so the skopeo copy will fail - but at least the repo will be ok, so I guess it's sufficient for now.

hallyn:
```
pkg/storage/storage.go:442:3: not enough arguments to return

	have (error)

	want (string, error)

pkg/storage/storage.go:447:3: not enough arguments to return

	have (error)

	want (string, error)
```

pkg/api/routes.go
```diff
@@ -40,16 +41,25 @@ const (
 )
 
 type RouteHandler struct {
-	c *Controller
+	c        *Controller
+	blobLock sync.Mutex
```
rchincha: yup, this is what I have in my PR as well, except just did a defer unlock() pattern.
A manifestLock is also needed coz multiple clients can potentially update/delete a manifest, and we need to decide a lock order in that case - manifest lock followed by blob lock.

rchincha:
https://github.com/tych0/zot/pull/1

@tych0 to fix bazel build failures against this PR.

pkg/storage/storage_test.go
rchincha: at least validates, GC did happen

pkg/api/routes.go
tych0: I'm not sure what defer unlock() does in terms of helping us out minimizing the critical section?

tych0:
Hrm. The test failure looks legit; the GC is deleting the blob it's failing to open right before it. I must not understand how something in zot works: seems like the manifest should have a tag in the sub-oci-repo for that thing, right?

rchincha:
This PR superseded by https://github.com/anuvu/zot/commit/bbed38cfa0c5481a2d91c2dee075ffe3ca151ad9.

PR #79, opened 2020-03-20
Tycho run gc after manifest removal

rchincha:
Multi-author patch set.

Tycho did the original garbage-collection patch.
I did the errata and enhancements.
Individual commits should have their own explanations.

This patch also needed umoci GC path to be fixed (patch for that merged) hence the updated go.mod.


tych0:
LGTM, modulo the CI failures; thanks for getting this going!

PR #82, opened 2020-04-03
build: fix stacker build

rchincha:
copy sources into the container rather than use bind-mount


rchincha:
Fixed with https://github.com/anuvu/zot/pull/116

PR #84, opened 2020-04-16
Zot search

shimish2:
1. Integrated Zot Search using GraphQl 
2. Added one subroute as /v2/query


pkg/api/controller.go
```diff
@@ -12,20 +12,27 @@ import (
...
+	DBFilePath = "./data/db/ZotSearch.db"
+)
+
 type Controller struct {
 	Config     *Config
 	Router     *mux.Router
 	ImageStore *storage.ImageStore
 	Log        log.Logger
 	Server     *http.Server
+	DBPath     string
```
rchamarthy, line 30: I recommend this to go to separate process of its own. In real deployment we can front this with a reverse proxy.

pkg/extensions/search/utils/db.go
```diff
@@ -0,0 +1,229 @@
...
+	// NameDB ...
+	NameDB = "NvdPkgName"
+	// NameverDB ...
+	NameverDB = "NvdPkgNameVer"
+	// NvdmetaDB ...
+	NvdmetaDB = "NvdMeta"
+)
+
+/*Conn ... Create a database connection to */
+func Conn(dbPath string) *bolt.DB {
```
rchamarthy, line 25: Connect?

pkg/extensions/search/utils/db.go
```diff
@@ -0,0 +1,229 @@
...
+	uerr = updateNVDPkg(NameDB, mapList[1], db)
+	if uerr != nil {
+		fmt.Println("Unable to Update Name Package Bucket")
+
+		return false
+	}
+
+	uerr = updateNVDPkg(NameverDB, mapList[2], db)
+	if uerr != nil {
+		fmt.Println("Unable to Update Name-Version Package Bucket")
```
rchamarthy, line 83: these need to be logs

pkg/storage/storage.go
```diff
@@ -357,7 +357,7 @@ func (is *ImageStore) PutImageManifest(repo string, reference string, mediaType
 		is.log.Error().Err(err).Msg("unable to unmarshal JSON")
 		return "", errors.ErrBadManifest
 	}
-
+	// nolint (gomnd)
 	if m.SchemaVersion != 2 {
```
rchamarthy, line 361: make schema version a package level constant.

pkg/api/controller.go
```diff
@@ -12,20 +12,27 @@ import (
...
+	DB         *bolt.DB
```
tych0, line 31: Rather than make people Close() the DB explicitly, can we make a wrapper Close() method? This way if we add more things that need to be closed to this structure, we don't have to go update all the callers again. It also has a nice benifit of preventing an abstraction leak. Unrelated, but does most of this stuff need to be public?

pkg/extensions/search/utils/db.go
```diff
@@ -0,0 +1,229 @@
...
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	})
+
+	uerr := updateNVDPkg(VendorDB, mapList[0], db)
+	if uerr != nil {
+		fmt.Println("Unable to Update Vendor Package Bucket")
```
tych0, line 70: Why not return an error here?

pkg/extensions/search/utils/db.go
```diff
@@ -0,0 +1,229 @@
...
+				err := json.Unmarshal(v, &dbcveidlist)
+				if err != nil {
+					fmt.Println("Unable to Unmarshal Data from PkgVendor Bucket")
+
+					return err
+				}
+
+				cveidlist = append(cveidlist, dbcveidlist...)
+
+				encode, _ := json.Marshal(cveidlist)
```
tych0, line 118: Seems like we should check this error.

pkg/extensions/search/utils/db.go
```diff
@@ -0,0 +1,229 @@
...
+	err := db.View(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(NvdmetaDB))
+
+		v = b.Get([]byte(filename))
+
+		return nil
+	})
+	if err != nil {
+		fmt.Println("Not able to search")
+		fmt.Println(err)
```
tych0, line 164: Looks like all these functions should return an error.

pkg/extensions/search/utils/util.go
```diff
@@ -0,0 +1,532 @@
+package utils
```
tych0, line 1: This seems like it could be better than "utils". "types", at least? "cve-info"?

rchincha:
Would expose a config subsection under pkg/api/config.go called Extensions
This would help us keep separation between dist-spec and our own innovations

pkg/extensions/search/utils/db.go
shimish2: I am returning false here if there is error in updating. Can return error but have to change method return type.

pkg/extensions/search/utils/db.go
shimish2: Yes, will do

pkg/extensions/search/utils/db.go
shimish2: Ok, I will change all method return type as error and update the code accordingly.

pkg/extensions/search/utils/util.go
shimish2: Yes, I will change this

pkg/api/controller.go
shimish2: Yes, as Ram suggested below

pkg/extensions/search/utils/db.go
shimish2: I will change all the information messages/errors  in form of logs.

pkg/storage/storage.go
shimish2: Ok

rchincha:
Also, pls pay attention to coverage - too much of a drop.

pkg/api/controller.go
shimish2: Yes, it does not need to be public as I am using in same package. I will also add a wrapper Close() method.

pkg/extensions/search/utils/db.go
shimish2: Yes, I will change it to Connect

PR #85, opened 2020-04-21
Add graphQL search support

shimish2:
1. Work on review changes.
2. Added more unit test case to increase code coverage.


pkg/api/ldap.go
```diff
@@ -10,6 +10,7 @@ import (
 	"fmt"
 
 	"github.com/anuvu/zot/errors"
+        // nolint (stylecheck)
```
tych0, line 13: Indentation looks off here, but other than that lgtm.

examples/config-test.json
```diff
@@ -9,5 +9,8 @@
   },
   "log":{
     "level":"debug"
+  },
+  "extension":{
+    "dbPath":"./data/db/ZotSearch.db"
```
rchincha, line 14: "extensions": {
"search": {
"dbPath": ...
}
}

pkg/api/config.go
```diff
@@ -67,21 +69,37 @@ type LogConfig struct {
 	Output string
 }
 
+type ExtensionConfig struct {
+	Cve    *CveConfig
+	DBPath string
```
rchincha, line 74: Make this part of SearchConfig

examples/config-test.json
shimish2: Ok

pkg/api/config.go
shimish2: Ok, I will change that

pkg/api/controller.go
```diff
@@ -16,6 +16,10 @@ import (
 	"github.com/gorilla/mux"
 )
 
+const (
+	DBFilePath = "./data/db/ZotSearch.db"
```
rchincha, line 20: Is this a separate directory to store the db? zot already has a cache.db it puts at top dir

pkg/api/routes.go
```diff
@@ -102,8 +106,9 @@ func (rh *RouteHandler) SetupRoutes() {
 			rh.blobLockWrapper(rh.DeleteBlobUpload)).Methods("DELETE")
 		g.HandleFunc("/_catalog",
 			rh.ListRepositories).Methods("GET")
-		g.HandleFunc("/",
-			rh.CheckVersionSupport).Methods("GET")
+		g.HandleFunc("/", rh.CheckVersionSupport).Methods("GET")
```
rchincha, line 109: don't change the indent here - it follows convention above

pkg/api/routes.go
```diff
@@ -102,8 +106,9 @@ func (rh *RouteHandler) SetupRoutes() {
...
+		//nolint (lll)
+		g.HandleFunc("/query", gqlHandler.GraphQL(search.NewExecutableSchema(search.Config{Resolvers: &search.Resolver{DB: rh.c.Config.Extension.Cve.cvedb, Cve: &cveinfo.CveInfo{Log: rh.c.Log}}})))
```
rchincha, line 111: break this into several lines? not possible? use nolint sparingly

rchincha, line 111: also, is this URI path protected with authentication or anyone can access this route?

pkg/cli/root.go
```diff
@@ -47,6 +47,7 @@ func NewRootCmd() *cobra.Command {
 					panic(errors.ErrBadConfig)
 				}
 			}
+			config.Extension.Cve = api.NewCveConfig(config.Extension.DBPath, config.Log)
```
rchincha, line 50: Why not make this a part of NewConfig() itself?

pkg/extensions/search/utils/db.go
```diff
@@ -0,0 +1,244 @@
+package cveinfo
```
rchincha, line 1: this is a search/cve extension so call it cve.go - that it uses a db is a separate thing

pkg/api/controller.go
shimish2: Yes, I made that to store the data, if it has to be cache.db I can change that.

pkg/api/routes.go
shimish2: Ok, I will follow that

pkg/api/routes.go
shimish2: Ok, I will do that

pkg/api/controller.go
rchincha: No, call it search.db for now.

pkg/storage/storage.go
```diff
@@ -152,7 +153,7 @@ func (is *ImageStore) ValidateRepo(name string) (bool, error) {
 		is.log.Error().Err(err).Str("dir", dir).Msg("unable to read directory")
 		return false, errors.ErrRepoNotFound
 	}
-
+	// nolint (gomnd)
```
rchincha, line 156: what is this nolint? you are not even touching this file right?

pkg/extensions/search/utils/db.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// nolint (interfacer)
+func Close(db *bbolt.DB) error {
+	err := db.Close()
+	return err
+}
+
+//CreateDB ...
+func (cve CveInfo) CreateDB(dbname string, db *bbolt.DB) bool {
```
rchincha, line 42: Where is this being used in real code above?

pkg/api/routes.go
shimish2: I think it is authenticated since it is a part of same subrouter/route prefix.

rchincha: Pls. cover it with a test case

pkg/storage/storage.go
shimish2: Yes I am not touching this file but I got linting error during building, so I tried to fix that

pkg/extensions/search/utils/db.go
shimish2: Used in util.go InitSearch method

pkg/extensions/search/utils/db.go
shimish2: Ok, I will change that

pkg/api/routes.go
shimish2: Ok, I will do that

pkg/cli/root.go
shimish2: I thought our plan is to make CveSearch as a different service, its good to have different method for CveConfig and use wherever we need as of now.

PR #89, opened 2020-05-19
compat: when in "world-readable" mode, return the WWW-Authenticate

rchincha:
header

containers/image is the dominant client library to interact with
registries.

It detects which authentication to use based on the WWW-Authenticate
header returned when pinging "/v2/" end-point. If we didn't return this
header, then creds are not used for other write-protected end-points.
Hence, the compatibility fix.


tych0:
lgtm

PR #93, opened 2020-05-27
dedupe: record relative path for cache entries

rchincha:
In a production use case we found that the actual rootdir can be moved.
Currently, cache entries for dedupe record the full blob path which
doesn't work in the move use case.

Only for dedupe cache entries, record relative blob paths.

### DIFF
--- a/pkg/storage/cache.go
+++ b/pkg/storage/cache.go
@@ -2,6 +2,7 @@ package storage
 
 import (
 	"path"
+	"path/filepath"
 	"strings"
 
 	"github.com/anuvu/zot/errors"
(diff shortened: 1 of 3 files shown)


hallyn:
Can you give an example, file gets moved from X to Y ?

rchincha:
> 
> 
> Can you give an example, file gets moved from X to Y ?

Before:

zot storage pointing to /data/production/ would lead cache entries of the form "/data/production/<repo>/blobs/sha256/<blob>". The move could be due to data corruption or copy /data/production over to a remote host for debugging but not at /data/production, etc.

Now, if we rename /data/production to something else then the cache is broken. Not a ideal situation. We now still construct absolute paths but as rootDir (runtime config param) and relative path recorded in the cache.

PR #97, opened 2020-06-06
Add CLI commands for searching CVE

tsnaik:
Adding commands for searching CVE (Common Vulnerabilities and Exposures) in images hosted on Zot

It supports two types of queries
- Searching images affected by a CVE given a CVE ID
	```
	zot search images --cve-id <cveID> --url <zot_server_url>
	zot search images -c <cveID> --url <zot_server_url>
	```
- Searching by an image name to see the list of CVEs affecting that image
	```
	zot search cve --image-name <image_name> --url <zot_server_url>
	zot search cve -I <image_name> --url <zot_server_url>
	```
Add `--user <username:password>` to any command if credentials are needed for access. Like `--user someuser:somepassword`


tsnaik:
splitting cve search and core search into separate branches.. so that core search cli can be implemented independently without any dependency on cve search cli or backend

PR #98, opened 2020-06-11
fix: the bug when htpasswd has multiple creds

tsnaik:
earlier, when you had more than one creds in htpasswd file separated by newline, it used to only read the first cred in the file and ignore the rest.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ test/data/
 coverage.html
 tags
 vendor/
+.vscode/
--- a/pkg/api/auth.go
+++ b/pkg/api/auth.go
@@ -141,15 +141,16 @@ func basicAuthHandler(c *Controller) mux.MiddlewareFunc {
 			if err != nil {
 				panic(err)
 			}
+			defer f.Close()
 
-			for {
-				r := bufio.NewReader(f)
-				line, err := r.ReadString('\n')
-				if err != nil {
-					break
+			scanner := bufio.NewScanner(f)
+
+			for scanner.Scan() {
+				line := scanner.Text()
+				if strings.Contains(line, ":") {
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -19,4 +19,5 @@ test/data/
 .idea/
 coverage.html
 tags
-vendor/
```
rchincha, line 22: make this a separate commit

pkg/api/auth.go
```diff
@@ -141,15 +141,16 @@ func basicAuthHandler(c *Controller) mux.MiddlewareFunc {
 			if err != nil {
 				panic(err)
 			}
+			defer f.Close()
 
-			for {
```
rchincha, line 145: Can you add a test case to prove this path works?
1. Single entry with a newline at end of file
2. Single entry without a newline at end of file
3. Multiple entries with a newline at end of file
4. Multiple entries without a newline at end of file

tsnaik: done

pkg/api/controller_test.go
```diff
@@ -79,6 +93,131 @@ func TestNew(t *testing.T) {
...
+				So(resp.StatusCode(), ShouldEqual, 401)
+			}()
+		}
+	})
+}
+
+func TestHtpasswdTwoCreds(t *testing.T) {
+	Convey("Two creds", t, func() {
+		twoCredtests :=
+			[]string{"alice:$2y$12$wc6M7d7vTyFD3m9J2hTWPuM3q1gd1lM.8Epnpd8YheZCh6hx/vxma\n" +
```
rchincha, line 157: Instead of static creds here, can we just something on the fly?

PR #107, opened 2020-06-25
Upgraded build pipeline

shimish2:
Build system was using go 1.12.x. Needed version upgrade of go to integrate with trivy, trivy is using 1.13. Updated the travis CI configuration file to use latest go version i.e 1.14.x.  

Golangci-lint used was 1.21.0, upgraded golangci-lint to 1.26.0. 

Build system was using bazel 0.26.1, upgraded it to 3.0.0 

Bazel rules_go version changed to 0.23.3 and go version to 1.14.4

Bazel gazelle version changed to v0.21.0

Bazel build tools version changed to 0.25.1

Bazel skylib version changed to 1.0.2


rchincha:
So this is need to pull in trivy?

shimish2:
> So this is needed to pull in trivy?

Yes, tried with our present build system and was not able to build https://travis-ci.org/github/anuvu/zot/jobs/701811410

pkg/storage/storage.go
```diff
@@ -111,7 +112,7 @@ func (is *ImageStore) InitRepo(name string) error {
 			is.log.Panic().Err(err).Msg("unable to marshal JSON")
 		}
 
-		if err := ioutil.WriteFile(ilPath, buf, 0644); err != nil {
+		if err := ioutil.WriteFile(ilPath, buf, 0600); err != nil {
```
tych0, line 115: What's the reason for this change? This just makes it more annoying to introspect things as non-zot, which I expect we'll want to do at some point. Is there a threat model or something?

shimish2: After upgrading golangci-lint and go version, gosec linter highlighted this and suggested the permission should be 0600 or less, so i changed it to 0600, if a permission should be 0644 i can add nolint to it.

tych0: I think it should be 644. Linters (especially go ones) are often not correct :)

shimish2: Ok, I will do the necessary changes.

rchincha: Well, could really be writable only on first create - 0400 would be fine also if the subsequent access is only needed for zot daemon and via HTTP.

shimish2: Ok, should I change it to 0400 ?

rchincha: No, leave it at 644 for now. We can address it in a separate PR if needed.
This is the gosec error.
G306: Poor file permissions used when writing to a new file

tych0: > access is only needed for zot daemon

I don't think that's true, though.

shimish2: Changed back to 0644

rchincha:
If there are no objections, will merge this. About time we update our deps.

PR #108, opened 2020-06-25
systemd: add a systemd service example file

rchincha:
Copy this file into /etc/systemd/system, and

\# systemctl enable zot
\# systemctl start zot


examples/zot.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=OCI Distribution Registry
+Documentation=https://github.com/anuvu/zot
+After=network.target auditd.service local-fs.target
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/zot serve /etc/zot/config.json
```
tych0, line 8: I wonder if this should just be `zot`?

rchincha: Would the user of this to tweak, but basically not to be left to the devices of PATH env var.

PR #111, opened 2020-06-26
Zot search extension

shimish2:


pkg/extensions/search/generated.go
```diff
@@ -0,0 +1,3031 @@
+// Code generated by github.com/99designs/gqlgen, DO NOT EDIT.
```
tych0, line 1: Why commit all this generated code, why not generate it as part of the build system?

rchincha: IINM, because 'go get' doesn't do this autogeneration and this file is not expected to change very often. Same argument we used for swagger docs.

tych0: Is there a reason to worry about what "go get" does? Can't we just build it with the makefile?

I think committing the swagger docs is probably wrong too :)

rchincha: Basically convenience - so this, until we have a proper dockerhub way of distributing zot (_docker run zot:x.y.z_)

tych0: That's not really distributing it, it still requires you to have a go compiler installed. If you have a go compiler installed, isn't it possible to just do 'make' and generate all this stuff?

PR #112, opened 2020-07-06
gc: add a policy to skip garbage collecting new blobs

rchincha:
We perform inline garbage collection of orphan blobs. However, the
dist-spec poses a problem because blobs begin their life as orphan blobs
and then a manifest is add which refers to these blobs.

We use umoci's GC() to perform garbage collection and policy support
has been added recently which can control whether a blob can be skipped
for GC.

In this patch, we use a time-based policy to skip blobs.


rchincha:
Updated the PR to take of CI/CD failures. No code changes.

PR #113, opened 2020-07-09
stacker: fix stacker build

rchincha:


rchincha:
"make test" fails because we are running as root and directories can really 'mkdir -p' anywhere.

hallyn:
Fails for me with:
https://paste.ubuntu.com/p/mb7hZq9qVV/

hallyn:
With very latest stacker, i get https://paste.ubuntu.com/p/7kdGRhBvX8/

tych0:
I think the underlying problem here is that go is searching the roots/ bind mounted directory and failing looking at its own filesystem. If you see the error message:

```
go: directory roots/temp-snapshot-build-498088494/rootfs/go/pkg/mod/gopkg.in/ini.v1@v1.51.0 outside available modules
```

it doesn't need to be looking in roots/ at all. Maybe try playing with where --roots-dir lives?

hallyn:
This does not build.  I'd say go ahead and merge it anyway, but did you read Tycho's comment above?  With this patch:

```
serge@sl:~/src/zot$ git diff Makefile
diff --git a/Makefile b/Makefile
index ce7d754..335d4b7 100644
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ binary-container:

 .PHONY: binary-stacker
 binary-stacker:
-       stacker build --substitute PWD=$$PWD
+       stacker --roots-dir=../zot-roots build --substitute PWD=$$PWD

 .PHONY: image
 image:
```

It builds for me.

PR #114, opened 2020-07-10
auth: support a read-only mode

rchincha:
This is useful if we want to roll out experimental versions of zot
pointing to some storage shared with another zot instance.

Also, when under storage full conditions, will be useful to turn on this
flag to prevent further writes.


tych0:
@rchincha should we merge this? can we disable code coverage if we don't care about it?

PR #115, opened 2020-07-14
cli: move client-only code out of the server flow

tsnaik:
earlier, some of the client exclusive code was being run on zot server
instance too. That caused an error when server doesn't have env variables required by client.

This PR moves that code out of the way of server


hallyn:
lgtm thanks

PR #117, opened 2020-07-16
gc: perform garbage collection based on changed-time not mod-time

rchincha:
Dedupe code path uses hard links to optimize storage. However, creating
hard links doesn't update mod-time, which can cause the following
scenario.

1) upload blob with hash1 as repo1/hash1

2) create a manifest with repo1/hash1 so repo/hash1 is no longer an
orphan

3) wait until after GC timeout

4) upload blob with same hash1 as repo2/hash1, but note that mod-time of
hash1 is not updated

5) upload blob with a different hash2 as repo2/hash2

6) create a manifest with repo2/hash2 which will GC repo2/hash1

7) now the manifest with repo2/hash1 will fail because repo2/hash1 was GC'ed in step 6)

Use change-time instead. However, note that the downside is if hard
links to a particular blob happen too frequently it may never be garbage
collected.

### DIFF
--- a/pkg/storage/storage_fs.go
+++ b/pkg/storage/storage_fs.go
@@ -26,6 +26,7 @@ import (
 	"github.com/opencontainers/umoci/oci/casext"
 	artifactspec "github.com/oras-project/artifacts-spec/specs-go/v1"
 	"github.com/rs/zerolog"
+	"golang.org/x/sys/unix"
 	zerr "zotregistry.io/zot/errors"
 	"zotregistry.io/zot/pkg/extensions/monitoring"
 	zlog "zotregistry.io/zot/pkg/log"
@@ -1203,7 +1204,7 @@ retry:
 				return err
 			}
 
-			is.log.Debug().Str("blobPath", dst).Msg("dedupe: creating hard link")
+			is.log.Debug().Str("blobPath", dst).Str("dstRecord", dstRecord).Msg("dedupe: creating hard link")
 
 			if err := os.Link(dstRecord, dst); err != nil {
 				is.log.Error().Err(err).Str("blobPath", dst).Str("link", dstRecord).Msg("dedupe: unable to hard link")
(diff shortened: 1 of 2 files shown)


hallyn:
In step 5, did you mean repo2/hash1 ?

rchincha:
> 
> 
> In step 5, did you mean repo2/hash1 ?

Updated my earlier comment, hopefully clear now. We should be able to write test case to demo this.

mikemccracken:
> > In step 5, did you mean repo2/hash1 ?
> 
> Updated my earlier comment, hopefully clear now. We should be able to write test case to demo this.

Is there a test case that we can add to this PR?

rchincha:
Here is the evidence about the different gc/dedupe behavior

With this change, gc is skipped (as expected):

```
{"level":"debug","message":"skipping garbage collection of blob sha256:a186000422feab857329c684e9fe91412b1a5db084100b37a98cfc95b62aa867 because of policy 0"}
{"level":"debug","message":"garbage collected 0 blobs"}
```

Without this change:

```
{"level":"info","digest":"sha256:a186000422feab857329c684e9fe91412b1a5db084100b37a98cfc95b62aa867","blobPath":"/tmp/oci-gc-test925382854/gc2/blobs/sha256/a186000422feab857329c684e9fe91412b1a5db084100b37a98cfc95b62aa867","caller":"/local/rchincha/go/src/github.com/anuvu/zot/pkg/storage/storage_fs.go:1618","message":"perform GC on blob"}
{"level":"debug","message":"garbage collecting blob: sha256:a186000422feab857329c684e9fe91412b1a5db084100b37a98cfc95b62aa867"}
{"level":"debug","message":"garbage collected 1 blobs"}
```

but net effect is if a blob is not and is found in the dedupe cache, it is copied back.

hallyn:
> note that the downside is if hard
> links to a particular blob happen too frequently it may never be garbage
> collected.

But that means they're being constantly pushed so we don't want to garbage collect them, right?

hallyn:
Oh, but, removing a hard link also updates change-time.

PR #118, opened 2020-07-17
cli: add option to ignore TLS verification

tsnaik:
adds a property in config : "verify-tls". It allows user to make unsecured HTTPS connections by skipping TLS certificate verification

```console
$ zot config add local https://localhost:8080         
$ zot config local -l                       
url = http://localhost:8080
showspinner = true
verify-tls = true
$ zot config local verify-tls false
$ zot config local -l                       
url = https://localhost:8080
showspinner = true
verify-tls = false
```

### DIFF
--- a/pkg/cli/client.go
+++ b/pkg/cli/client.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"io/ioutil"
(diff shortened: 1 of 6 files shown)


mikemccracken:
this looks fine to me, it'd be convenient for my usual use case if there was a corresponding cmdline option so I don't have to separately set config...

tsnaik:
> this looks fine to me, it'd be convenient for my usual use case if there was a corresponding cmdline option so I don't have to separately set config...

@mikemccracken, we have discussed that option but chose to go not go with that for UX simplicity. @rchincha can add perhaps

rchincha:
> 
> 
> this looks fine to me, it'd be convenient for my usual use case if there was a corresponding cmdline option so I don't have to separately set config...

config modeled after git-style config and is one-time, and keeping one way is less confusing.

mikemccracken:
> > this looks fine to me, it'd be convenient for my usual use case if there was a corresponding cmdline option so I don't have to separately set config...
> 
> config modeled after git-style config and is one-time, and keeping one way is less confusing.

ok, it's your call :)

PR #124, opened 2020-08-03
Add support to scan images for CVEs

shimish2:
Following features are added in this PR:- 

1. Support to scan images for vulnerability scanning (CVEs). 
2. Integrated with trivy to support vulnerability scanning.
3. Periodically updating trivy db.
4. Added following graphql api endpoints for image scanning:- 
        CVEListForImage:- given an image returns list of CVEs
        ImageListForCVE:- given a CVE id returns list of image:tag 
        ImageListWithCVEFixed:- given an image and CVE id return the list of tags in an image not affected by that CVE
5. Image scanning feature is added as an extension.


errors/errors.go
```diff
@@ -34,4 +34,8 @@ var (
 	ErrConfigNotFound          = errors.New("cli: config with the given name does not exist")
 	ErrNoURLProvided           = errors.New("cli: no URL provided in argument or via config. see 'zot config -h'")
 	ErrIllegalConfigKey        = errors.New("cli: given config key is not allowed")
+	ErrNotSupported            = errors.New("search: squashfs image scanning not supported")
```
rchincha, line 37: ErrScanNotSupported?

errors/errors.go
```diff
@@ -34,4 +34,8 @@ var (
...
+	ErrJSONNotFound            = errors.New("repository: index.json not found")
```
rchincha, line 38: Where was this needed?

examples/config-cve.json
```diff
@@ -0,0 +1,20 @@
...
+        "rootDirectory": "/tmp/zot"
+    },
+    "http": {
+        "address": "127.0.0.1",
+        "port": "8080"
+    },
+    "log": {
+        "level": "debug"
+    },
+    "extensions": {
```
rchincha, line 13: +1

errors/errors.go
shimish2: I will change that

errors/errors.go
shimish2: I used this errors message while checking valid image format to scan in cve.go

errors/errors.go
```diff
@@ -34,4 +34,8 @@ var (
 	ErrConfigNotFound          = errors.New("cli: config with the given name does not exist")
 	ErrNoURLProvided           = errors.New("cli: no URL provided in argument or via config. see 'zot config -h'")
 	ErrIllegalConfigKey        = errors.New("cli: given config key is not allowed")
+	ErrScanNotSupported        = errors.New("search: squashfs image scanning not supported")
+	ErrJSONNotFound            = errors.New("repository: index.json not found")
```
rchincha, line 38: https://github.com/anuvu/zot/blob/master/pkg/storage/storage.go#L264
we already have an error for this case, reuse it

examples/config-cve.json
```diff
@@ -0,0 +1,20 @@
...
+        "search": {
+            "cve": {
+                "periodicupdates": true
```
rchincha, line 16: Where in the code is this being used?

examples/config-cve.json
```diff
@@ -0,0 +1,20 @@
...
+            }
+        }
+    }
+}
```
rchincha, line 20: Make sure you put newlines at the end of all json, go files

pkg/api/config.go
```diff
@@ -68,21 +72,36 @@ type LogConfig struct {
...
+	Search *SearchConfig
+}
+
+type SearchConfig struct {
+	// CVE search
+	CVE *CVEConfig
+}
+
+type CVEConfig struct {
+	UpdateInterval time.Duration // should be 2 hours or more, if not specified default be kept as 24 hours
```
rchincha, line 85: So the config param is called "UpdateInterval" not "PeriodicUpdates"? and it is a duration value, not bool.

pkg/api/config.go
```diff
@@ -68,21 +72,36 @@ type LogConfig struct {
...
-	Commit  string
-	Storage StorageConfig
-	HTTP    HTTPConfig
-	Log     *LogConfig
+	Version    string
+	Commit     string
+	Storage    StorageConfig
+	HTTP       HTTPConfig
+	Log        *LogConfig
+	Extensions *ExtensionConfig
```
rchincha, line 94: +1
This keeps the separation between core functionality and extensions.

examples/config-cve.json
rchincha: This example field is not correct. Replace with UpdateInterval and test pls.

pkg/api/controller.go
```diff
@@ -49,6 +50,18 @@ func (c *Controller) Run() error {
 		os.Exit(1)
 	}
 
+	// Updating the CVE Database
+	if c.Config != nil && c.Config.Extensions != nil && c.Config.Extensions.Search != nil &&
+		c.Config.Extensions.Search.CVE.UpdateInterval > 1 {
+		go func() {
+			err := cveinfo.UpdateCVEDb(c.Config.Storage.RootDirectory, c.Log,
+				c.Config.Extensions.Search.CVE.UpdateInterval, false)
```
rchincha, line 58: What happens when queries arrive when this update is in progress? Is something serializing it?

pkg/api/controller.go
```diff
@@ -49,6 +50,18 @@ func (c *Controller) Run() error {
...
+			if err != nil {
```
rchincha, line 59: Add a log to indicate that an update has started.
Add a log in this path for errors.
Also add a log that update was successful etc, so we know what is going on.

pkg/api/config.go
```diff
@@ -68,21 +72,36 @@ type LogConfig struct {
...
-		Commit:  Commit,
-		Storage: StorageConfig{GC: true, Dedupe: true},
-		HTTP:    HTTPConfig{Address: "127.0.0.1", Port: "8080"},
-		Log:     &LogConfig{Level: "debug"},
+		Version:    dspec.Version,
+		Commit:     Commit,
+		Storage:    StorageConfig{GC: true, Dedupe: true},
+		HTTP:       HTTPConfig{Address: "127.0.0.1", Port: "8080"},
+		Log:        &LogConfig{Level: "debug"},
+		Extensions: &ExtensionConfig{&SearchConfig{CVE: &CVEConfig{UpdateInterval: updateInterval}}},
```
rchincha, line 104: Extensions are strictly optional, so they cannot have default values like this.
Right now, whether we like it or not, trivy scan db is always enabled, which is not right.

pkg/api/routes.go
```diff
@@ -104,9 +112,22 @@ func (rh *RouteHandler) SetupRoutes() {
...
+		g.HandleFunc("/debug/pprof/symbol/",
+			pprof.Symbol).Methods("GET")
+		g.HandleFunc("/debug/pprof/trace",
+			pprof.Trace).Methods("GET")
 	}
 	// swagger docs "/swagger/v2/index.html"
 	rh.c.Router.PathPrefix("/swagger/v2/").Methods("GET").Handler(httpSwagger.WrapHandler)
+	// Zot Search Extension Router
+	rh.c.Router.PathPrefix("/query").Methods("GET", "POST").Handler(rh.searchHandler())
+	rh.c.Router.PathPrefix("/debug/pprof/heap").Methods("GET").Handler(pprof.Handler("heap"))
```
rchincha, line 130: Remove this, or add a "profiling: bool" Extensions param.
And add a if ... that

pkg/api/routes.go
```diff
@@ -104,9 +112,22 @@ func (rh *RouteHandler) SetupRoutes() {
...
+			pprof.Profile).Methods("GET")
+		g.HandleFunc("/debug/pprof/symbol/",
+			pprof.Symbol).Methods("GET")
+		g.HandleFunc("/debug/pprof/trace",
+			pprof.Trace).Methods("GET")
 	}
 	// swagger docs "/swagger/v2/index.html"
 	rh.c.Router.PathPrefix("/swagger/v2/").Methods("GET").Handler(httpSwagger.WrapHandler)
+	// Zot Search Extension Router
+	rh.c.Router.PathPrefix("/query").Methods("GET", "POST").Handler(rh.searchHandler())
```
rchincha, line 129: Since extensions are optional, this should also be wrapped in some if <cond>

pkg/extensions/search/cve/cve.go
```diff
@@ -0,0 +1,234 @@
...
+
+			return false, errors.ErrJSONNotFound
+		}
+
+		cveinfo.Log.Error().Err(err).Msg("Unable to open index.json")
+
+		return false, errors.ErrInvalidJSON
+	}
+
+	var index ispec.Index
```
rchincha, line 79: Really? the linter is forcing these many whitespaces between lines?

pkg/extensions/search/cve/cve.go
```diff
@@ -0,0 +1,234 @@
...
+
+		imageLayers := blobManifest.Layers
+
+		for _, imageLayer := range imageLayers {
+			switch imageLayer.MediaType {
+			case types.OCILayer, types.DockerLayer:
+				return true, nil
+
+			default:
+				return false, nil
```
rchincha, line 115: Add a "skipping" DEBUG level msg.

pkg/extensions/search/cve/testdata/zot-dis-test/blobs/sha256/5de64447fc6d0caa31843bd7dd43b2fdde02475403d1ea016a9e512e02790166
```diff
@@ -0,0 +1 @@
+{"created":"1970-01-01T00:00:00Z","author":"Bazel","architecture":"amd64","os":"linux","config":{"Env":["PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin","SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt"]},"rootfs":{"type":"layers","diff_ids":["sha256:59839d326e715c1ff37d9bf6ce1186a4ab85f3dfd6fcfee9a614e7e2a136fc97","sha256:b335a3ef21dea5a9b17da691ae44b7a344e6a1fa649c7aa5ae61216ca42b1f46"]},"history":[{"created":"1970-01-01T00:00:00Z","created_by":"bazel build ...","author":"Bazel"},{"created":"1970-01-01T00:00:00Z","created_by":"bazel build ...","author":"Bazel"}]}
```
rchincha, line 1: Did we forget to remove some files? What are all these blob thingies?

pkg/api/routes.go
rchincha: And these extensions are protected behind authentication, so make sure the test cases ensure that.

errors/errors.go
shimish2: I added this error because ErrJSONNotFound gives more detailed reason "repository: index.json not found" instead of "repository: not found".

examples/config-cve.json
shimish2: Yes, I forget to update this file with latest changes that I had done regarding the CveConfig. I will replace with UpdateInterval.

I have added this config-cve file if we want to start the zot with cve extension also.

examples/config-cve.json
shimish2: Ok

pkg/api/config.go
shimish2: Yes

pkg/api/controller.go
shimish2: There is a log in UpdateCVEDb method, I will add more for more clarifications.

pkg/api/config.go
shimish2: Ok, I will change it like that.

pkg/api/routes.go
shimish2: Yes, I added for analysing the memory leak. I will remove this endpoints.

pkg/api/routes.go
shimish2: Yes, I have already added test cases for it

shimish2: > Since extensions are optional, this should also be wrapped in some if

Ok

pkg/extensions/search/cve/cve.go
shimish2: Yes

pkg/extensions/search/cve/cve.go
shimish2: Ok

pkg/extensions/search/cve/testdata/zot-dis-test/blobs/sha256/5de64447fc6d0caa31843bd7dd43b2fdde02475403d1ea016a9e512e02790166
shimish2: This are the test files I added to scan an image.

errors/errors.go
rchincha: Without an index.json, not really a valid repo for us. So just reuse that same pattern, and add logs.

pkg/api/controller.go
rchincha: But we want to track when this goroutine got invoked and is done.

errors/errors.go
shimish2: Ok, I will do

pkg/api/controller.go
shimish2: Queries has to wait when they are trying to read from boltdb.

pkg/api/controller.go
shimish2: Yes, i will add the log for that

pkg/api/controller.go
```diff
@@ -49,6 +51,25 @@ func (c *Controller) Run() error {
 		os.Exit(1)
 	}
 
+	// Updating the CVE Database
+	if c.Config != nil && c.Config.Extensions != nil && c.Config.Extensions.Search != nil &&
+		c.Config.Extensions.Search.CVE.UpdateInterval > 1 {
```
mikemccracken, line 56: If we misconfigure the update interval to be <= 1, it looks like it'll just silently never start the updater, right?
Might be friendlier to have that as a separate 'if' so you can log a message that says e.g. "update interval config <= 1, not scheduling CVE DB updates.

In fact - unless this is covered elsewhere in the logging, It'd also be nice to have an else at the end to just log that it's not scheduling any CVE updates. It's nice to be explicit about this kind of thing in logs, so you don't have to look at code and think hard about what it was doing based on logs that should be there but aren't.

shimish2: 1. I have added a log message where this IsValidImageFormat is called. 
2. Every time I add a new line to file looks like VS code removes it upon saving, I will add it now.
3. I will add that separate if condition and log a message but do we need to proceed further if updates are not scheduled.

mikemccracken: ok to 1 & 2 - for 3, I'm not sure I follow.
I was suggesting something roughly like this (but you can probably think of ways to improve this logic)

Also, these aren't blockers, if you don't see the value in adding these logs that's fine by me.

```go
// Updating the CVE Database
 	if c.Config != nil && c.Config.Extensions != nil && c.Config.Extensions.Search != nil {
 		if c.Config.Extensions.Search.CVE.UpdateInterval <= 1 {
				c.Log.Info().Msg("CVE update interval set to too-short interval <= 1, not scheduling updates.")
               }else{
		go func() {
 			for {
 				c.Log.Info().Msg("Updating the CVE database")
  				err := cveinfo.UpdateCVEDb(c.Config.Storage.RootDirectory, c.Log)
 				if err != nil {
 					panic(err)
 				}

  				c.Log.Info().Str("Db update completed, next update scheduled after ", c.Config.Extensions.Search.CVE.UpdateInterval.String()).Msg("hours") //nolint: lll
  				time.Sleep(c.Config.Extensions.Search.CVE.UpdateInterval * time.Hour)
 			}
 		}()
 	    }
    }else{
            c.Log.Info().Msg("No CVE config specified, not scheduling CVE updates")
}
```

So it doesn't proceed with scheduling anything if the interval is too short, it just logs it and moves on...

pkg/api/config.go
mikemccracken: Should it be a Duration? That's supposed to be nanoseconds, right? It looks like you're using it as an integer number of hours, should we just make it an int and avoid confusion here?

When you use it later you're multiplying by time.Hour in `time.Sleep(UpdateInterval * time.Hour)` anyway

pkg/api/controller.go
shimish2: I will update if-else logic as above and add this log messages.

pkg/api/config.go
shimish2: Ok I will make that as an integer and use it like that time.Sleep(time.Duration(c.Config.Extensions.Search.CVE.UpdateInterval) * time.Hour)

rchincha: No, leave this as time.Duration and in JSON config it should be "24h"
https://golang.org/pkg/time/#ParseDuration

If it is less that 2h, then default to 2h and throw a warning.

mikemccracken: > No, leave this as time.Duration and in JSON config it should be "24h"
> https://golang.org/pkg/time/#ParseDuration
> 
> If it is less that 2h, then default to 2h and throw a warning.

that makes sense too. didn't realize the json was supposed to be a string. @shimish2 sorry for the extra work there :)

rchincha: @shimish2 pls. test and verify just the same

shimish2: > > No, leave this as time.Duration and in JSON config it should be "24h"
> > https://golang.org/pkg/time/#ParseDuration
> > If it is less that 2h, then default to 2h and throw a warning.
> 
> that makes sense too. didn't realize the json was supposed to be a string. @shimish2 sorry for the extra work there :)

No problem @mikemccracken

PR #126, opened 2020-08-10
ci/cd: install skopeo

rchincha:
zot trivy extensions test code needs an oci layout as test data.
Install skopeo to help with that.


shimish2:
build passing after installing skopeo and using skopeo copy command to download image locally, looks good to me.

PR #141, opened 2020-10-01
routes: refactor locks to handle large file uploads

rchincha:
The storage layer is protected with read-write locks.
However, we may be holding the locks over unnecessarily large critical
sections.

The typical workflow is that a blob is first uploaded via a per-client
private session-id meaning the blob is not publicly visible yet. When
the blob being uploaded is very large, the transfer takes a long time
while holding the lock.

Private session-id based uploads don't really need locks, and hold locks
only when blobs are published after the upload is complete.

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -57,26 +57,6 @@ func (rh *RouteHandler) searchHandler() *gqlHandler.Server {
 	return gqlHandler.NewDefaultServer(search.NewExecutableSchema(resConfig))
 }
 
-// blobRLockWrapper calls the real handler with read-lock held.
-func (rh *RouteHandler) blobRLockWrapper(f func(w http.ResponseWriter,
-	r *http.Request)) func(w http.ResponseWriter, r *http.Request) {
-	return func(w http.ResponseWriter, r *http.Request) {
-		rh.c.ImageStore.RLock()
-		f(w, r)
-		rh.c.ImageStore.RUnlock()
-	}
-}
-
-// blobLockWrapper calls the real handler with write-lock held.
-func (rh *RouteHandler) blobLockWrapper(f func(w http.ResponseWriter,
-	r *http.Request)) func(w http.ResponseWriter, r *http.Request) {
-	return func(w http.ResponseWriter, r *http.Request) {
-		rh.c.ImageStore.Lock()
-		f(w, r)
-		rh.c.ImageStore.Unlock()
-	}
-}
-
 func (rh *RouteHandler) SetupRoutes() {
 	rh.c.Router.Use(AuthHandler(rh.c))
 	g := rh.c.Router.PathPrefix(RoutePrefix).Subrouter()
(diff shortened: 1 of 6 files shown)


pkg/api/routes.go
```diff
@@ -98,15 +98,15 @@ func (rh *RouteHandler) SetupRoutes() {
 		g.HandleFunc(fmt.Sprintf("/{name:%s}/blobs/{digest}", NameRegexp.String()),
 			rh.blobLockWrapper(rh.DeleteBlob)).Methods("DELETE")
 		g.HandleFunc(fmt.Sprintf("/{name:%s}/blobs/uploads/", NameRegexp.String()),
-			rh.blobLockWrapper(rh.CreateBlobUpload)).Methods("POST")
```
shimish2, line 101: Locks are removed from CreateBlobUpload, PatchBlobUpload, UpdateBlobUpload, all this handler calls https://github.com/anuvu/zot/blob/ea0ef61a654b47e06ec978a53f176419dd5f0277/pkg/storage/storage.go#L111 internally and this method initializes repository, may be we need a lock here also (inside InitRepo method).

rchincha: > Locks are removed from CreateBlobUpload, PatchBlobUpload, UpdateBlobUpload, all this handler calls 
Not removed, they have been moved into pkg/storage/storage.go

shimish2: Yes but InitRepo is called everywhere before lock is acquired, I think we also need a lock there.

mikemccracken:
@shimish2 this last commit doesn't seem to fit with the PR, did you mean to include it here?

shimish2:
> @shimish2 this last commit doesn't seem to fit with the PR, did you mean to include it here?

I added this commit to pass build because tests were failing due to API rate limit of trivy db, I can create a separate PR to merge this commit in master.

mikemccracken:
> > @shimish2 this last commit doesn't seem to fit with the PR, did you mean to include it here?
> 
> I added this commit to pass build because tests were failing due to API rate limit of trivy db, I can create a separate PR to merge this commit in master.

ok, makes sense. no need to worry about separate PRs - since in the end the list of commits is what's important to keep clean. it's ok if PRs have a bunch of commits, as long as it's understandable for reviewers.
maybe just in the future add a comment about why the seemingly-unrelated commit is added.

shimish2:
> > > @shimish2 this last commit doesn't seem to fit with the PR, did you mean to include it here?
> > 
> > 
> > I added this commit to pass build because tests were failing due to API rate limit of trivy db, I can create a separate PR to merge this commit in master.
> 
> ok, makes sense. no need to worry about separate PRs - since in the end the list of commits is what's important to keep clean. it's ok if PRs have a bunch of commits, as long as it's understandable for reviewers.
> maybe just in the future add a comment about why the seemingly-unrelated commit is added.

Sure, I will keep in mind.

PR #142, opened 2020-10-01
server: add read and write timeout in server configuration

shimish2:
Currently when we create a server we do not set timeout, it's a good practice to set both read and write timeouts and it helps our server to be more accessible.


rchincha:
Don't we need to take care of idleTimeout as well?

shimish2:
@rchincha Do we need this ?

PR #145, opened 2020-10-15
customizable binaries

shimish2:
This PR is about creating a "zot-minimal" build target which only does dist-spec serving.

Added two build tags minimal and extended, minimal will create a binary that only serves dist-spec serving while extended will include CVE scan also.

Added one build rule (minimal-binary) also in Makefile, this rule will generate a minimal-binary while existing binary rule will now generate full binary.

Note:- 
 **commit "build: increase wait timeout for travis bazel build process" bazel is taking more than 20 minutes to complete build process.**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,20 +7,24 @@ PATH := bin:$(PATH)
 TMPDIR := $(shell mktemp -d)
 
 .PHONY: all
-all: doc binary debug test check
+all: doc binary binary-minimal debug test check
+
+.PHONY: binary-minimal
+binary-minimal: doc
+	go build -tags minimal -v  -ldflags "-X  github.com/anuvu/zot/pkg/api.Commit=${COMMIT}" -o bin/zot-minimal ./cmd/zot
 
 .PHONY: binary
 binary: doc
-	go build -v -ldflags "-X  github.com/anuvu/zot/pkg/api.Commit=${COMMIT}" -o bin/zot -tags=jsoniter ./cmd/zot
+	go build -tags extended -v -ldflags "-X  github.com/anuvu/zot/pkg/api.Commit=${COMMIT}" -o bin/zot ./cmd/zot
 
 .PHONY: debug
 debug: doc
-	go build -v -gcflags all='-N -l' -ldflags "-X  github.com/anuvu/zot/pkg/api.Commit=${COMMIT}" -o bin/zot-debug -tags=jsoniter ./cmd/zot
+	go build -tags extended -v -gcflags all='-N -l' -ldflags "-X  github.com/anuvu/zot/pkg/api.Commit=${COMMIT}" -o bin/zot-debug ./cmd/zot
 
 .PHONY: test
 test:
 	$(shell mkdir -p test/data;  cd test/data; ../scripts/gen_certs.sh; cd ${TOP_LEVEL}; sudo skopeo --insecure-policy copy -q docker://centos:latest oci:${TOP_LEVEL}/test/data/zot-test:0.0.1;sudo skopeo --insecure-policy copy -q docker://centos:8 oci:${TOP_LEVEL}/test/data/zot-cve-test:0.0.1)
-	go test -v -race -cover -coverpkg ./... -coverprofile=coverage.txt -covermode=atomic ./...
+	go test -tags extended -v -race -cover -coverpkg ./... -coverprofile=coverage.txt -covermode=atomic ./...
 
 .PHONY: covhtml
 covhtml:
@@ -29,7 +33,7 @@ covhtml:
 .PHONY: check
 check: .bazel/golangcilint.yaml
 	golangci-lint --version || curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s v1.26.0
-	golangci-lint --config .bazel/golangcilint.yaml run --enable-all ./cmd/... ./pkg/...
+	golangci-lint --config .bazel/golangcilint.yaml run --enable-all --build-tags extended ./cmd/... ./pkg/...
 
 docs/docs.go: 
 	swag -v || go install github.com/swaggo/swag/cmd/swag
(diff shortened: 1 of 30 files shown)


hallyn (approved):

pkg/extensions/extension_minimal.go
```diff
@@ -0,0 +1,26 @@
...
+	"net/http"
+	"time"
+
+	"github.com/anuvu/zot/pkg/log"
+	"github.com/anuvu/zot/pkg/storage"
+)
+
+// ExtensionServer ...
+type ExtensionServer struct {
+	GraphqlServer http.Handler
```
rchincha, line 15: Is this needed at all?

Makefile
```diff
@@ -7,20 +7,24 @@ PATH := bin:$(PATH)
 TMPDIR := $(shell mktemp -d)
 
 .PHONY: all
-all: doc binary debug test check
+all: doc minimal-binary debug test check
+
+.PHONY: minimal-binary
+minimal-binary: doc
```
rchincha, line 13: binary-minimal

cmd/zot/BUILD.bazel
```diff
@@ -9,9 +9,10 @@ go_library(
 )
 
 go_binary(
-    name = "zot",
+    name = "zot-extended",
```
rchincha, line 12: Just call the full thing just zot

Makefile
rchincha: Also, does "minimal" remove the cli as well?

Makefile
```diff
@@ -7,20 +7,24 @@ PATH := bin:$(PATH)
...
+	go build -tags minimal -v  -ldflags "-X  github.com/anuvu/zot/pkg/api.Commit=${COMMIT}" -o bin/zot-minimal ./cmd/zot
 
 .PHONY: binary
 binary: doc
-	go build -v -ldflags "-X  github.com/anuvu/zot/pkg/api.Commit=${COMMIT}" -o bin/zot -tags=jsoniter ./cmd/zot
+	go build -tags extended -v -ldflags "-X  github.com/anuvu/zot/pkg/api.Commit=${COMMIT}" -o bin/zot ./cmd/zot
```
rchincha, line 18: s/extended/extensions/g

pkg/extensions/extension_minimal.go
shimish2: Yes, see routes.go we need graphql server when we enable cve scan and when we want to generate new minimal binary 
ExtensionServer need to be passed.

Makefile
shimish2: No, its not removing cli

cmd/zot/BUILD.bazel
shimish2: ok

Makefile
shimish2: Should I used tags as extended/extensions ??

pkg/api/controller.go
```diff
@@ -52,31 +51,8 @@ func (c *Controller) Run() error {
 	}
 
 	// Updating the CVE Database
```
rchincha, line 53: Remove this comment
And also add another comment which says,
// Enable extensions based on configuration

pkg/api/controller.go
```diff
@@ -52,31 +51,8 @@ func (c *Controller) Run() error {
...
-
-				c.Log.Info().Str("Db update completed, next update scheduled after", c.Config.Extensions.Search.CVE.UpdateInterval.String()).Msg("") //nolint: lll
-
-				time.Sleep(c.Config.Extensions.Search.CVE.UpdateInterval)
-			}
-		}()
-	} else {
-		c.Log.Info().Msg("Cve config not provided, skipping cve update")
+	if c.Config != nil {
+		ext.EnableExtension(c.Config.Extensions, c.Log, c.Config.Storage.RootDirectory)
```
rchincha, line 55: EnableExtensions (plural)

pkg/api/routes.go
```diff
@@ -96,7 +90,7 @@ func (rh *RouteHandler) SetupRoutes() {
 	rh.c.Router.PathPrefix("/swagger/v2/").Methods("GET").Handler(httpSwagger.WrapHandler)
 	// Zot Search Extension Router
```
rchincha, line 91: Remove this comment and say
// Setup routes for extensions based on configuration

pkg/extensions/minimal.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/gorilla/mux"
+)
+
+// DownloadTrivyDB ...
+func DownloadTrivyDB(dbDir string, log log.Logger, updateInterval time.Duration) error {
+	return nil
+}
+
+func EnableExtension(extension *ExtensionConfig, log log.Logger, rootDir string) {
+	log.Info().Msg("given zot binary doesn't support any extensions, please build zot full binary for this feature")
```
rchincha, line 19: Good log msg.

pkg/extensions/minimal.go
```diff
@@ -0,0 +1,23 @@
...
+}
+
+func SetupRoutes(router *mux.Router, rootDir string, imgStore *storage.ImageStore, log log.Logger) {
+}
```
rchincha, line 23: Add a similar log saying routes were not setup because extensions are not enabled.

pkg/extensions/minimal.go
mikemccracken: Yes, helpful - maybe also this should be a `log.Warn()` so it'll stand out a bit?

pkg/api/controller.go
shimish2: Ok

pkg/api/controller.go
shimish2: Ok

pkg/extensions/minimal.go
shimish2: Sure, I think its good to log as a warning because this part of code will run when user starts server providing extension configuration.

shimish2: Changed log message also, please review.

rchincha (approved):
lgtm

PR #146, opened 2020-10-15
routes: refactor locks to handle large file uploads

shimish2:
The storage layer is protected with read-write locks.
However, we may be holding the locks over unnecessarily large critical
sections.

The typical workflow is that a blob is first uploaded via a per-client
private session-id meaning the blob is not publicly visible yet. When
the blob being uploaded is very large, the transfer takes a long time
while holding the lock.

Private session-id based uploads don't really need locks, and hold locks
only when blobs are published after the upload is complete.

Note:- 
1. commit: "test: minimize trivy db download tests to avoid api rate limit" added to avoid api rate limit during build process.
2. Since I added multiple parallel request tests now, make test is taking longer time to execute and that's why we need to add wait during make process also in travis, commit: "build: add wait time during travis make build process" serve this purpose.

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -111,6 +111,9 @@ func (is *ImageStore) Unlock() {
 func (is *ImageStore) InitRepo(name string) error {
 	repoDir := path.Join(is.rootDir, name)
 
+	is.Lock()
+	defer is.Unlock()
+
 	if fi, err := os.Stat(repoDir); err == nil && fi.IsDir() {
 		return nil
 	}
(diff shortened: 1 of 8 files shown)


pkg/storage/storage.go
```diff
@@ -525,6 +540,9 @@ func (is *ImageStore) DeleteImageManifest(repo string, reference string) error {
 		return errors.ErrRepoNotFound
 	}
 
+	is.Lock()
```
rchincha, line 543: move this under Parse below

rchincha (approved):
lgtm

PR #147, opened 2020-10-21
build: update docker and stacker files

rchincha:
Update docker and stacker builds

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,23 @@
 # ---
 # Stage 1: Install certs, build binary, create default config file
 # ---
-FROM docker.io/golang:1.13.6-alpine3.11 AS builder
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,7 +1,7 @@
 # ---
 # Stage 1: Install certs, build binary, create default config file
 # ---
-FROM docker.io/golang:1.13.6-alpine3.11 AS builder
+FROM docker.io/golang:1.15.3-alpine3.12 AS builder
```
shimish2, line 4: Currently go.mod, bazel and .travis uses go 1.14, I think we should first verify on 1.15 before changing it here.

shimish2 (approved):
LGTM

Issue #161, opened 2020-12-02
Third-Party Dependency and License List


McCodeman:
As a user, I want a list of all the project's direct software dependencies and their respective licenses.

Acceptance Criteria:
* [x] Use a tool to automate collection of direct dependencies and their licenses. For example, the following tools do this:
  * https://github.com/google/go-licenses
  * https://github.com/go-enry/go-license-detector
* [x] Publish list of third-party dependencies and licenses to THIRD-PARTY-LICENSES.md (or asciidoc).

A good example output is https://github.com/kubernetes/cloud-provider-aws/blob/master/THIRD-PARTY-LICENSES.

rchincha:
Fixed as part of https://github.com/anuvu/zot/pull/165

PR #165, opened 2020-12-03
doc: add third-party software deps list

rchincha:
Add a THIRD-PARTY-LICENSES.md to list all known deps.

Fixes issue #161


THIRD-PARTY-LICENSES.md
```diff
@@ -0,0 +1,146 @@
+github.com/99designs/gqlgen                          MIT License
```
tych0, line 1: I'm not so convinced this is a good idea -- it seems destined to get out of date without at least CI checking that it's still up to date.

What about a "make show-licenses" target instead which just renders this from the existing go.mod?

McCodeman (approved):
LGTM

PR #168, opened 2020-12-16
build: remove bazel

rchincha:
The idea initially was to use bazel to do our builds, however golang
build system is now good enough and our code base is entirely go.
It is also slowing down our travis ci/cd pipeline.


shimish2 (commented):
We can definitely remove bazel but reason travis build taking too long looks different to me, we already removed bazel build previously but still travis build terminates.

rchincha:
Superseded by PR#169

PR #172, opened 2021-01-22
added graphql api to return repository list with latest tag

shimish2:


examples/config-cve.json
```diff
@@ -4,8 +4,8 @@
         "rootDirectory": "/tmp/zot"
     },
     "http": {
-        "address": "127.0.0.1",
-        "port": "8080"
+        "address": "0.0.0.0",
+        "port": "5000"
```
rchincha, line 8: why change this example? committed by mistake?

examples/config-minimal.json
```diff
@@ -4,7 +4,7 @@
         "rootDirectory": "/tmp/zot"
     },
     "http": {
-        "address": "127.0.0.1",
+        "address": "0.0.0.0",
```
rchincha, line 7: same here?

examples/config-cve.json
shimish2: Yes by mistake

pkg/extensions/extensions.go
```diff
@@ -56,11 +56,11 @@ func EnableExtensions(extension *ExtensionConfig, log log.Logger, rootDir string
 
 // SetupRoutes ...
 func SetupRoutes(extension *ExtensionConfig, router *mux.Router, storeController storage.StoreController,
-	log log.Logger) {
+	log log.Logger, enableCVE bool) {
```
rchincha, line 59: In future, lets make this is a ... ExtensionOption

pkg/api/routes.go
```diff
@@ -97,7 +97,11 @@ func (rh *RouteHandler) SetupRoutes() {
 	rh.c.Router.PathPrefix("/swagger/v2/").Methods("GET").Handler(httpSwagger.WrapHandler)
 	// Setup Extensions Routes
 	if rh.c.Config != nil && rh.c.Config.Extensions != nil {
```
rchincha, line 99: also check rh.c.Config.Extensions.Search.Enable

andaaron: This code was agnostic of what extensions we have implemented. Do we want it to be aware of what extensions we have?
The check for the extensions themselves just enabled the graphql API. And what the API can provide is controlled from inside ext.SetupRoutes.

I believe we should leave this code as is was on master, and modify  the code in https://github.com/anuvu/zot/pull/172/files#diff-86b8c7ae55e63b332b1b61b1e50093ca23babbdf0ad78ff1f74bcbec501cc188R58 to read this configuration.
The configuration is already available in rh.c.Config.Extensions, we don't need to pass it as a separate parameter.

pkg/extensions/extensions.go
andaaron: I don't believe we need this parameter. We should already have what we need inside extension

pkg/storage/storage.go
```diff
@@ -268,6 +268,7 @@ func (is *ImageStore) ValidateRepo(name string) (bool, error) {
 func (is *ImageStore) GetRepositories() ([]string, error) {
 	dir := is.rootDir
 
+	fmt.Println(dir)
```
andaaron, line 271: Is this temporary?

andaaron (commented):
See also my replies to Ram's comments.
Looks like the PR also needs a rebase after the go.mod changes.

pkg/storage/storage.go
shimish2: Yes, I will remove this

pkg/extensions/search/common/oci_layout.go
```diff
@@ -3,52 +3,31 @@ package common
...
+	"github.com/google/go-containerregistry/pkg/v1/types"
 	godigest "github.com/opencontainers/go-digest"
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
-// CveInfo ...
+// OciLayoutInfo ...
 type OciLayoutUtils struct {
-	Log             log.Logger
-	StoreController storage.StoreController
```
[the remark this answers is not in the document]
shimish2: I removed storage because this functions can be utilize just by passing blob or index.json path and there should not be any dependency to create a StoreController to use this method.

pkg/extensions/search/common/common.go
```diff
@@ -0,0 +1,155 @@
...
+	LabelAnnotationLicenses    = "org.label-schema.license"
+)
+
+type TagInfo struct {
+	Name      string
+	Digest    string
+	Timestamp time.Time
+}
+
+func GetImageRepoPath(image string, storeController storage.StoreController) string {
```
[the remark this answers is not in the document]
shimish2: I move it to Common because it is common method and should not be limited to use by OCI.

pkg/api/routes.go
shimish2: Sure, I will add a check there

pkg/extensions/extensions.go
shimish2: Yes, I will remove that

pkg/extensions/search/common/common.go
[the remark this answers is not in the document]
andaaron: But that only works for local file system storage.

andaaron (dismissed):

go.mod
```diff
@@ -3,32 +3,28 @@ module github.com/anuvu/zot
...
-	github.com/AdaLogics/go-fuzz-headers v0.0.0-20210921152813-f50b76b2163b // indirect
+	github.com/99designs/gqlgen v0.13.0
 	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751
 	github.com/apex/log v1.9.0
 	github.com/aquasecurity/trivy v0.0.0-00010101000000-000000000000
-	github.com/briandowns/spinner v1.12.0
-	github.com/chartmuseum/auth v0.5.0
-	github.com/containerd/containerd v1.5.4 // indirect
+	github.com/briandowns/spinner v1.16.0
+	github.com/chartmuseum/auth v0.4.5
```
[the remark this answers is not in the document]
rchincha: For your commit, there should be no 'go.mod' changes right?

errors/errors.go
```diff
@@ -40,4 +41,5 @@ var (
 	ErrInvalidRoute            = errors.New("routes: invalid route prefix")
 	ErrImgStoreNotFound        = errors.New("routes: image store not found corresponding to given route")
 	ErrEmptyValue              = errors.New("cache: empty value")
+	ErrEmptyRepoList           = errors.New("extension: no repository found")
```
rchincha, line 44: just flag this as "search: no repo..." etc?

go.mod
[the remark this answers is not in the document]
rchincha: Only use whatever is on "master" now

pkg/extensions/search/cve/cve_test.go
```diff
@@ -457,90 +446,6 @@ func TestDownloadDB(t *testing.T) {
 	})
 }
 
-func TestImageFormat(t *testing.T) {
```
rchincha, line 460: Did this test case get removed? why so?

andaaron: https://github.com/anuvu/zot/pull/172/files#diff-b86e11fa5a3102b336caebec3b30a9d35e26af554dd8658f124dba2404b7d24aR156

go.mod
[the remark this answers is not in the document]
shimish2: Yes, there was conflict during master rebase. I will update it accordingly

andaaron (approved):

rchincha (approved):
lgtm

PR #180, opened 2021-04-06
enable zot to point to multiple storage locations

### LINKED ISSUE #181: single zot server instance should support multiple storage backends
At present a single zot instance have only one single storage i.e being specified in config file e.g 
``` 
"storage":{
    "rootDirectory":"/tmp/zot"
  }
```
but now we want zot to be scalable in terms of storage and restriction is to be done using single zot server instance.

shimish2:
Closes #181


errors/errors.go
```diff
@@ -37,4 +37,6 @@ var (
 	ErrScanNotSupported        = errors.New("search: scanning of image media type not supported")
 	ErrCLITimeout              = errors.New("cli: Query timed out while waiting for results")
 	ErrDuplicateConfigName     = errors.New("cli: cli config name already added")
+	ErrInvalidRoute            = errors.New("route: invalid route prefix")
+	ErrImgStorageNotFound      = errors.New("routes: image storage not found corresponding to given route")
```
rchincha, line 41: "image store" not "...storage"?

examples/config-multiple.json
```diff
@@ -0,0 +1,22 @@
+{
+    "version": "0.1.0-dev",
+    "storage": {
+        "/a": {
+            "rootDirectory": "/tmp/zot1"
```
rchincha, line 5: you should probably add at least a "dedupe" flag per entry because let's say we have "dedupe" turned on globally, how does one know if a particular volume/filesystem will not fail on hard links?

errors/errors.go
shimish2: Ok, i will chage to this  ```ErrImgStoreNotFound  = errors.New("routes: image store not found corresponding to given route")```

examples/config-multiple.json
shimish2: could you please explain what "dedupe" global option means because now 
``` json 
"storage": {
       "RootDirectory": "/tmp/zot",
       "Dedupe": true
    },

File: examples/config-multiple.json [hunk 3]
```diff
@@ -0,0 +1,30 @@
+{
+    "version": "0.1.0-dev",
+    "storage": {
+        "rootDirectory": "/tmp/zot",
+        "dedupe": true,
+        "gc": true,
```
rchamarthy, line 6: what is the default value for gc/dedupe? I think default should be "true". I guess the only way we would say false here is if the underlying storage is actually shared.

pkg/api/controller.go
```diff
@@ -43,20 +43,37 @@ func (c *Controller) Run() error {
...
+	if c.Config.Storage.RootDirectory != "" {
+		c.ImageStore["/"] = storage.NewImageStore(c.Config.Storage.RootDirectory,
+			c.Config.Storage.GC, c.Config.Storage.Dedupe, c.Log)
+		// Enable extensions if extension config is provided
+		if c.Config != nil && c.Config.Extensions != nil {
+			ext.EnableExtensions(c.Config.Extensions, c.Log, c.Config.Storage.RootDirectory)
+		}
+	} else {
+		// we can't proceed without global storage
 		os.Exit(1)
```
rchamarthy, line 61: shouldnt we panic here with a proper log message?

examples/config-multiple.json
```diff
@@ -0,0 +1,30 @@
+{
+    "version": "0.1.0-dev",
+    "storage": {
+        "rootDirectory": "/tmp/zot",
```
rchamarthy, line 4: shouldn't this simply be "/" in the subpaths section? The subpath is obviously the longest prefix, right? This will allow us to operationally move subpaths into different volumes on a running server in the future.

pkg/api/controller.go
```diff
@@ -43,20 +43,37 @@ func (c *Controller) Run() error {
...
 
-	c.ImageStore = storage.NewImageStore(c.Config.Storage.RootDirectory, c.Config.Storage.GC,
-		c.Config.Storage.Dedupe, c.Log)
-	if c.ImageStore == nil {
-		// we can't proceed without at least a image store
+	c.Router = engine
+	c.Router.UseEncodedPath()
+
+	// Assign global level image storage
+	c.ImageStore = make(map[string]*storage.ImageStore)
```
rchamarthy, line 50: make this a Trie.

rchamarthy (changes requested):

pkg/api/controller.go
shimish2: Yes, we should. I will add panic and log.

examples/config-multiple.json
shimish2: I didn't understand your point, could you please explain in detail ?

pkg/api/controller.go
shimish2: Why do we need Trie here instead of map ?

rchincha (commented):
Add these as gentle addendums to existing model. 
So the additional paths and just more imageStores referenced under some repo path.
By default would stick to dist-spec behavior - dedupe and GC are all nice-to-have which can be turned on.

examples/config-multiple.json
rchamarthy: Ignore it. longest prefix has issues.

pkg/api/controller.go
rchamarthy: ignore this as well.

README.md
```diff
@@ -22,6 +22,9 @@ https://anuvu.github.io/zot/
 * Storage optimizations:
   * Automatic garbage collection of orphaned blobs
   * Layer deduplication using hard links when content is identical
+* Multiple storage backend using single zot server:
```
rchincha, line 25: Just say,
* Serve [multiple storage paths](example-config) (and backends) using a single zot server

errors/errors.go
```diff
@@ -37,4 +37,6 @@ var (
 	ErrScanNotSupported        = errors.New("search: scanning of image media type not supported")
 	ErrCLITimeout              = errors.New("cli: Query timed out while waiting for results")
 	ErrDuplicateConfigName     = errors.New("cli: cli config name already added")
+	ErrInvalidRoute            = errors.New("route: invalid route prefix")
```
rchincha, line 40: make both these changes read as ... "routes:"

examples/config-multiple.json
rchincha: The approach so far has been zot always default to dist-spec, and extensions need to enabled explicitly.

pkg/api/config.go
```diff
@@ -69,10 +69,17 @@ type LogConfig struct {
 	Output string
 }
 
+type GlobalStorageConfig struct {
```
rchincha, line 72: Yup, the right model IMO.

errors/errors.go
shimish2: Sure

README.md
shimish2: Sure

README.md
```diff
@@ -22,6 +22,9 @@ https://anuvu.github.io/zot/
...
+  * Added subpath sections in storage config, see example config [file](./examples/config-multiple.json)
```
rchincha, line 26: config-multiple.json <- where is this file, not in this PR?

pkg/storage/storage.go
```diff
@@ -37,6 +37,11 @@ type BlobUpload struct {
 	ID        string
 }
 
+type StoreController struct {
+	SingleStore *ImageStore
```
rchincha, line 41: Call this "DefaultStore"?

README.md
shimish2: Its in the PR

pkg/storage/storage.go
shimish2: Yes we can call it as DefaultStore

pkg/api/controller_test.go
```diff
@@ -1896,7 +2026,7 @@ func TestParallelRequests(t *testing.T) {
 	config.HTTP.Port = SecurePort1
 	htpasswdPath := makeHtpasswdFileFromString(getCredString(username, passphrase))
 
-	// defer os.Remove(htpasswdPath)
+	//defer os.Remove(htpasswdPath)
```
hallyn, line 2029: several commented-out defers here?

shimish2: Yes, because some tests are running in separate go routine and defer invokes before their completion

hallyn: That would explain *removing* them.

shimish2: Yes I forgot to remove, removed now.

rchincha (approved):
lgtm

PR #185, opened 2021-04-22
server: add idle timeout in http server configuration

shimish2:
Set idle timeout value to 15 minutes 

@rchincha please let me know if we want it to be changed.

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"time"
 
 	"github.com/anuvu/zot/errors"
 	ext "github.com/anuvu/zot/pkg/extensions"
@@ -15,6 +16,10 @@ import (
 	"github.com/anuvu/zot/pkg/storage"
 	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
+)
+
+const (
+	idleTimeout = 120 * time.Second
 )
 
 type Controller struct {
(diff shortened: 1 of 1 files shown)


pkg/api/controller.go
```diff
@@ -17,6 +18,10 @@ import (
 	"github.com/gorilla/mux"
 )
 
+const (
+	ioTimeout = 15
```
rchincha, line 22: call this idleTimeout and set it to the full timeout value
IdleTimeout =   120 * time.Second,

pkg/api/controller.go
```diff
@@ -17,6 +18,10 @@ import (
 	"github.com/gorilla/mux"
 )
 
+const (
+	idleTimeout = 120
```
rchincha, line 22: idleTimeout = 120 * time.Second,

pkg/api/controller.go
```diff
@@ -60,7 +65,11 @@ func (c *Controller) Run() error {
 	_ = NewRouteHandler(c)
 
 	addr := fmt.Sprintf("%s:%s", c.Config.HTTP.Address, c.Config.HTTP.Port)
-	server := &http.Server{Addr: addr, Handler: c.Router}
+	server := &http.Server{
+		Addr:        addr,
+		Handler:     c.Router,
+		IdleTimeout: idleTimeout * time.Second,
```
rchincha, line 71: and make this IdleTimeout: idleTimeout

rchincha (approved):
lgtm

PR #189, opened 2021-05-04
build: remove bazel files

shimish2:
Since we no longer support bazel builds, removing all files related to bazel.


rchincha (approved):
lgtm

PR #191, opened 2021-05-05
controller: add default headers

rchincha:
Add support for adding some default headers, CORS in this initial commit.
(experimental patch only)


shimish2 (approved):
hosted binary consisting of this change and now UI is able to get response.

Issue #51, opened 2019-12-23
add access control support


rchincha:
zot currently lacks access control support. A needed feature if a single instance needs to be shared across multiple tenants.

rchincha:
Some packages under consideration:

https://github.com/casbin/casbin
https://github.com/ory/ladon
https://github.com/mikespook/gorbac
https://github.com/zpatrick/rbac

ad-m:
https://www.openpolicyagent.org/

PR #192, opened 2021-05-13
add identity-based access control

peusebiu:
Hello guys, what do you say about this structure for authorization config?

#51

### DIFF
--- a/examples/config-policy.json
+++ b/examples/config-policy.json
@@ -1,0 +1,54 @@
+{
+  "version": "0.1.0-dev",
+  "storage": {
+    "rootDirectory": "/tmp/zot"
+  },
+  "http": {
+    "address": "127.0.0.1",
+    "port": "8080",
+    "realm": "zot",
+    "auth": {
+      "htpasswd": {
+        "path": "test/data/htpasswd"
+      },
+      "failDelay": 1
+    },
+    "accessControl": {
+      "repos1/repo": {
+        "policies": [
+          {
+            "users": ["alice", "bob"],
+            "actions": ["create", "read", "update", "delete"]
+          },
+          {
+            "users": ["mallory"],
+            "actions": ["create", "read"]
+          }
+        ],
+        "defaultPolicy": ["read"]
+      },
+      "repos2/repo": {
+        "policies": [
+          {
+            "users": ["bob"],
+            "actions": ["read", "create"]
+          },
+          {
+            "users": ["mallory"],
+            "actions": ["create", "read"]
+          }
+        ],
+        "defaultPolicy": ["read"]
+      },
+      "adminPolicy": {
+          "users": ["admin"],
+          "actions": ["read", "create", "update", "delete"]
+      }
+    }
+  },
+  "log": {
+    "level": "debug",
+    "output": "/tmp/zot.log"
+  }
+}
+
--- a/pkg/api/authz.go
+++ b/pkg/api/authz.go
@@ -1,0 +1,236 @@
+package api
+
+import (
+	"context"
+	"encoding/base64"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/anuvu/zot/pkg/log"
+	"github.com/gorilla/mux"
+)
+
+type contextKey int
+
+const (
+	// actions
+	CREATE = "create"
+	READ   = "read"
+	UPDATE = "update"
+	DELETE = "delete"
+
+	// request-local context key
+	authzCtxKey contextKey = 0
+)
+
+type AccessControlConfig struct {
+	Repositories Repositories
+	AdminPolicy  Policy
+}
+
+type Repositories map[string]PolicyGroup
+
+type PolicyGroup struct {
+	Policies      []Policy
+	DefaultPolicy []string
+}
+
+type Policy struct {
+	Users   []string
+	Actions []string
+}
+
+// AccessController authorizes users to act on resources.
+type AccessController struct {
+	Config *AccessControlConfig
+	Log    log.Logger
+}
+
+// AccessControlContext context passed down to http.Handlers.
+type AccessControlContext struct {
+	userAllowedRepos []string
+	isAdmin          bool
+}
+
+func NewAccessController(config *Config) *AccessController {
+	return &AccessController{
+		Config: config.AccessControl,
+		Log:    log.NewLogger(config.Log.Level, config.Log.Output),
+	}
+}
+
+// getReadRepos get repositories from config file that the user has READ perms.
+func (ac *AccessController) getReadRepos(username string) []string {
+	var repos []string
+
+	for r, pg := range ac.Config.Repositories {
+		for _, p := range pg.Policies {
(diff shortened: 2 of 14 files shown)


andaaron (commented):
Looks good.
Just to make the behavior clear. Would the policies be applied in the order in which they are present in the list? Meaning the first one which would match the user name would be applied, and following policies would be ignored.

examples/config-policy.json
```diff
@@ -0,0 +1,76 @@
...
+        "skipVerify":false,
+        "subtreeSearch":true
+      },
+      "htpasswd": {
+        "path": "test/data/htpasswd"
+      },
+      "failDelay": 5
+    },
+    "accessControl": {
+      "defaults": [
```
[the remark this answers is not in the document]
peusebiu: > Looks good.
> Just to make the behavior clear. Would the policies be applied in the order in which they are present in the list? Meaning the first one which would match the user name would be applied, and following policies would be ignored.

Yes, that's what I'm thinking. If no user found then fallback to defaults.

examples/config-policy.json
```diff
@@ -0,0 +1,76 @@
...
+        {
+          "users": ["alice", "bob"],
+          "actions": ["read", "write"]
+        },
+        {
+          "users": ["*"],
+          "actions": ["read"]
+        }
+      ],
+      "repositories": [
```
shimish2, line 42: can we add everything on repositories level, if repo is present in list then it will follow those rules and if repo not present in list it will follow whatever default we decides. 

``` json
 "repositories":
  [
       {
 		"name": "repos/repo1",
 		"actions": {
 			"read": ["alice", "bob"],
 			"write": ["alice", "bob"]
 		}
 	},
 	{
 		"name": "repos/repo2",
 		"actions": {
 			"read": ["*"],
 			"write": ["carlos"]
 		}
 	},
 	 {
 		"name": "repos/repo3",
 		"actions": {
 			"read": ["alice", "bob"],
 		}
 	 }
 ]
  ```

**Comment by rchincha** [comment]
Just to capture requirements, 

1) Site-wide system administrators who can add/delete/update images

2) Per repo administrators who can add/delete/update images
2.1) Per repo users can who read images
2.2) Per repo users who can add images but never delete or update them

**Comment by rchincha** [comment]
```
{
	"accessControl": {
		"/repos/repo1": {
			"policies": [
						{ 	
							"users": ["alice", "bob"],
							"actions": ["create", "read", "update", "delete"]
						},
						{
							"users": ["mallory"],
							"actions": ["create", "read"]
						}
			],
			"defaultPolicy": ["read"]
		},
		"/repos/repo1": {
			"policies": [
						{ 	
							"users": ["alice", "bob"],
							"actions": ["create", "read", "update", "delete"]
						},
						{
							"users": ["mallory"],
							"actions": ["create", "read"]
						}
			],
			"defaultPolicy": ["read"]
		},
		"defaultPolicy": ["read"]
	}
}

peusebiu:
@rchincha Seems like your example doesn't cover 1. Site-wide system administrators who can add/delete/update images.
Adding a policies list just under the "accessControl" would solve this.

rchincha (commented):
"defaultPolicy" at level of "accessControl" covers that case. There should be no other reason it falls back to that case other than site admins. Am I mistaken?

peusebiu:
Hello guys, can you review it now please?

Question, If the clients tries to modify the config (adding users, actions) and the config is not valid should I throw an error, or just log the error and let the user fix his configuration?

rchincha:
> 
> 
> Hello guys, can you review it now please?
> 
> Question, If the a client try to modify the config (adding users, actions) and the config is not valid should I throw an error, or just log the error and let the user fix his configuration and try again?

Yes, throw an error indicating why it is wrong and exit.
We should probably add a cli option to test/verify a config to make life easier for site admins.

examples/config-policy.json
```diff
@@ -0,0 +1,57 @@
...
+            }
+          ],
+          "defaultPolicy": ["read"]
+        }
+      },
+      "adminPolicy": {
+          "users": ["admin"],
+          "actions": ["create", "update", "delete"]
+        },
+      "defaultAdminPolicy": ["read"]
```
peusebiu, line 49: do we really need this? I don't have a special case where I use defaultAdminPolicy, in the end it's just a union between actions and defaultAdminPolicy.
So why do we need it?

pkg/api/authz.go
```diff
@@ -0,0 +1,222 @@
...
+
+	return repos
+}
+
+// can verify if a user can do action on repository.
+func (a *AccessController) can(username, action, repository string) bool {
+	can := false
+	// check repo based policy
+	for r, pg := range a.Config.Repositories {
+		if strings.Contains(repository, r) {
```
shimish2, line 87: If a particular user have write permission for this repo:- /infra/ubuntu and he tries to request to write ubuntu (a new repo e.g ```skopeo copy docker://ubuntu:latest docker://<zot-address>/ubuntu:latest). I think user will be able to write on our root directory since we are using strings.Contains(repository, r) ?

pkg/api/authz.go
```diff
@@ -0,0 +1,222 @@
...
+
+func NewAccessController(config *Config) *AccessController {
+	return &AccessController{
+		Config: config.AccessControl,
+		Log:    log.NewLogger(config.Log.Level, config.Log.Output),
+	}
+}
+
+// getReadRepos get repositories from config file that the user has READ perms.
+func (a *AccessController) getReadRepos(username string) []string {
```
shimish2, line 67: This method result will be used to search a repo only, if yes can we make its return type hashset ?

peusebiu: I'm not sure I understand.
This method is used to search all repos on which the user has READ perm.
The result is used to filter out zot's repos by username's rights.

Can you detail a bit?

pkg/api/authz.go
peusebiu: That's right, thank you, I'll fix it.

pkg/api/config.go
```diff
@@ -134,3 +139,39 @@ func (c *Config) Validate(log log.Logger) error {
 
 	return nil
 }
+
+// LoadAccessControlConfig populates config.AccessControl struct with values from config.
+func (c *Config) LoadAccessControlConfig() error {
```
peusebiu, line 144: I don't like this one.
Because I removed the repositories key, I Unmarshall into c.HTTP.RawAccessControl and then manually populate c.AccessControl.

Another approach I used was with a global variable, but that was even uglier.
If only viper supported Custom Unmarshall on specific fields...
If you have a better idea, please let me know.

Thank you.

rchincha:
@peusebiu can you rebase and push this PR again?

rchincha:
@peusebiu, can you squash your commits into one. This looks good.

peusebiu:
> @peusebiu, can you squash your commits into one. This looks good.

done, Thank you!

rchincha (approved):
lgtm

PR #197, opened 2021-05-20
api: add healthz route

### LINKED ISSUE #196: implement api similar to healthz to verfiy liveness and readiness of zot server

shimish2:
Added /v2/healthz route to verify liveness and readiness probe of zot server

Successful request to /v2/healthz route will return status code 200

Closes #196

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -85,6 +85,8 @@ func (rh *RouteHandler) SetupRoutes() {
 			rh.ListRepositories).Methods("GET")
 		g.HandleFunc("/",
 			rh.CheckVersionSupport).Methods("GET")
+		g.HandleFunc("/healthz",
+			rh.CheckHealth).Methods("GET")
 	}
 	// swagger docs "/swagger/v2/index.html"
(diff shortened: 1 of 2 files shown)


pkg/api/routes.go
```diff
@@ -96,6 +98,17 @@ func (rh *RouteHandler) SetupRoutes() {
...
 
+// CheckHealth godoc
+// @Summary Check Health API to check server health
+// @Description Check if zot server is up or not
+// @Router /v2/ [get]
+// @Accept json
+// @Accept json
+// @Success 200 {string} string "ok".
+func (rh *RouteHandler) CheckHealth(w http.ResponseWriter, r *http.Request) {
+	WriteData(w, http.StatusOK, "application/json", []byte{})
```
rchamarthy, line 109: what is the actual health check here? Just that http router and the port is working?

shimish2: Yes

shimish2: Do you want me to add more ?

mikemccracken: I think this is enough for starters, if we can't respond to this then an orchestrator should definitely restart us.

PR #198, opened 2021-05-21
[DRAFT] go.mod: update module versions

rchincha:
graphql generation code has changed, that still needs fixing.


shimish2:
Can we update golang to 1.16 also with this PR ?

shimish2:
We need to update .travis.yml also to use go 1.16 for checks.

PR #199, opened 2021-05-25
logs: add an audit log for API calls

roxanaN:
The format in the zot-audit.log file:

```
{"level":"info","subject":"test","action":"POST","object":"/v2/test/centos/blobs/uploads/","status":202,"time":"2021-05-25T11:33:25.7515376+03:00","message":"HTTP API"}
```

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -26,11 +26,24 @@ type Controller struct {
 	Router          *mux.Router
 	StoreController storage.StoreController
 	Log             log.Logger
+	Audit           *log.Logger
 	Server          *http.Server
 }
 
 func NewController(config *Config) *Controller {
-	return &Controller{Config: config, Log: log.NewLogger(config.Log.Level, config.Log.Output)}
+	var controller Controller
+
+	logger := log.NewLogger(config.Log.Level, config.Log.Output)
+
+	controller.Config = config
+	controller.Log = logger
+
+	if config.Log.Audit != "" {
+		audit := log.NewAuditLogger(config.Log.Level, config.Log.Audit)
+		controller.Audit = audit
+	}
+
+	return &controller
 }
 
 func DefaultHeaders() mux.MiddlewareFunc {
@@ -61,6 +74,10 @@ func (c *Controller) Run() error {
 		log.SessionLogger(c.Log),
 		handlers.RecoveryHandler(handlers.RecoveryLogger(c.Log),
 			handlers.PrintRecoveryStack(false)))
+
+	if c.Audit != nil {
+		engine.Use(log.SessionAuditLogger(c.Audit))
+	}
 
 	c.Router = engine
 	c.Router.UseEncodedPath()
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -1,8 +1,10 @@
 package log
 
 import (
+	"encoding/base64"
 	"net/http"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/gorilla/mux"
@@ -43,6 +45,28 @@ func NewLogger(level string, output string) Logger {
 	return Logger{Logger: log.With().Caller().Timestamp().Logger()}
 }
 
+func NewAuditLogger(level string, audit string) *Logger {
+	zerolog.TimeFieldFormat = time.RFC3339Nano
+	lvl, err := zerolog.ParseLevel(level)
+
+	if err != nil {
+		panic(err)
+	}
+
+	zerolog.SetGlobalLevel(lvl)
+
+	var auditLog zerolog.Logger
+
+	auditFile, err := os.OpenFile(audit, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0600)
+	if err != nil {
+		panic(err)
+	}
+
+	auditLog = zerolog.New(auditFile)
+
+	return &Logger{Logger: auditLog.With().Timestamp().Logger()}
+}
+
 type statusWriter struct {
 	http.ResponseWriter
 	status int
@@ -65,6 +89,7 @@ func (w *statusWriter) Write(b []byte) (int, error) {
 	return n, err
 }
 
+// SessionLogger logs session details.
 func SessionLogger(log Logger) mux.MiddlewareFunc {
 	l := log.With().Str("module", "http").Logger()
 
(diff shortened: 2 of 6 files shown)


pkg/log/log.go
```diff
@@ -29,6 +32,7 @@ func NewLogger(level string, output string) Logger {
 	zerolog.SetGlobalLevel(lvl)
 
 	var log zerolog.Logger
+	var audit_log zerolog.Logger
```
peusebiu, line 35: don't use underscore for variables, use camelCase.

eg: auditLog

roxanaN: I will change that. Thank you, @peusebiu

rchincha:
Couple of comments:

Pls. add test cases to verify that the log is indeed being populated.
Also, note that this audit file is "optional", so the rest of the code shouldn't break if this entry is missing.

pkg/log/log.go
rchincha: Did the linter not catch this?

pkg/log/log.go
```diff
@@ -140,7 +140,7 @@ func SessionLogger(log Logger, audit Logger) mux.MiddlewareFunc {
 				Msg("HTTP API")
 
 			if method == "PUT" || method == "POST" || method == "PATCH" || method == "DELETE" {
```
rchincha, line 142: Use http.MethodPost etc. 
https://golang.org/src/net/http/method.go

pkg/log/log.go
```diff
@@ -140,7 +140,7 @@ func SessionLogger(log Logger, audit Logger) mux.MiddlewareFunc {
...
-				if statusCode == 200 || statusCode == 201 || statusCode == 202 || statusCode == 203 {
+				if statusCode == 200 || statusCode == 201 || statusCode == 202 {
```
rchincha, line 143: Instead of "status", use "statusCode" as in the log msg above.
```
Int("statusCode", statusCode).
```

rchincha (commented):
Pls don't forget to add unit tests to cover this path.

rchincha:
Branch has merge conflicts

pkg/api/controller.go
```diff
@@ -27,11 +27,13 @@ type Controller struct {
 	Router     *mux.Router
 	ImageStore *storage.ImageStore
 	Log        log.Logger
+	Audit      log.Logger
 	Server     *http.Server
 }
 
 func NewController(config *Config) *Controller {
-	return &Controller{Config: config, Log: log.NewLogger(config.Log.Level, config.Log.Output)}
+	var Log, Audit = log.NewLogger(config.Log.Level, config.Log.Output, config.Log.Audit)
```
shimish2, line 35: log,audit := log.NewLogger(config.Log.Level, config.Log.Output, config.Log.Audit) instead of 
var Log, Audit = log.NewLogger(config.Log.Level, config.Log.Output, config.Log.Audit)

shimish2 (commented):
I see total three commits in the PR with two commits having same message, can we squash those commits into one ?

pkg/log/log_test.go
```diff
@@ -0,0 +1,203 @@
...
+				}
+
+				So(auditLog.Subject, ShouldEqual, username)
+				So(auditLog.Action, ShouldEqual, http.MethodPost)
+				So(auditLog.Status, ShouldEqual, http.StatusAccepted)
+				So(auditLog.Object, ShouldEqual, path)
+			})
+		})
+
+		err = os.Remove("./zot.log")
```
andaaron, line 193: Why not call os.Remove using defer,  similar to https://github.com/anuvu/zot/pull/199/files#diff-c15ccbf2169a85a01173cece538ff8583ee6905d27c8ffbdedc4ec2d953bb50eR111. That would make sure they are deleted.

pkg/log/log_test.go
```diff
@@ -0,0 +1,203 @@
...
+		if err != nil {
+			panic(err)
+		}
+		c.Config.Storage.RootDirectory = dir
+		go func() {
+			// this blocks
+			if err := c.Run(); err != nil {
+				return
+			}
+		}()
```
andaaron, line 134: This is long shot but I think reusing the c variable over here and at line 147 is triggering the race detector in the test case failure.
I see the other tests are passing if they pass the controller as a parameter to the functions running in these goroutines.

pkg/log/log_test.go
```diff
@@ -0,0 +1,203 @@
...
+					Get(BaseURL1 + "/v2/")
+				So(err, ShouldBeNil)
+				So(resp, ShouldNotBeNil)
+				So(resp.StatusCode(), ShouldEqual, 200)
+
+				byteValue, _ := ioutil.ReadAll(jsonFile)
+				So(len(byteValue), ShouldEqual, 0)
+			})
+
+			Convey("Test POST request", func() {
```
shimish2, line 169: Please add similar testcases for other http methods also e.g delete, patch

pkg/log/log.go
```diff
@@ -102,15 +136,26 @@ func SessionLogger(log Logger) mux.MiddlewareFunc {
...
+			log.Str("clientIP", clientIP).
 				Str("method", method).
 				Str("path", path).
 				Int("statusCode", statusCode).
 				Str("latency", latency.String()).
 				Int("bodySize", bodySize).
 				Interface("headers", headers).
 				Msg("HTTP API")
+
+			if method == http.MethodPost || method == http.MethodPut ||
```
rchincha, line 148: combine this into a single
if (..||..) && (statusCode == ...) {
}
?

andaaron (approved):

andaaron:
@rchincha @shimish2 shall we merge this?

shimish2 (dismissed):
lgtm

rchincha:
Quick question, what happens if the "audit" field is missing in the config?

pkg/log/log.go
```diff
@@ -40,7 +43,23 @@ func NewLogger(level string, output string) Logger {
...
+
+		auditLog = zerolog.New(auditFile)
+
+		return Logger{Logger: log.With().Caller().Timestamp().Logger()},
+			Logger{Logger: auditLog.With().Timestamp().Logger()}
+	}
+
+	// if audit file does not exist, redirect the output to /dev/null
+	return Logger{Logger: log.With().Caller().Timestamp().Logger()},
+		Logger{Logger: zerolog.New(ioutil.Discard)}
```
roxanaN, line 62: @rchincha, if the "audit" field is missing in the config, the output is redirected to **/dev/null**

pkg/api/controller.go
```diff
@@ -58,7 +60,7 @@ func (c *Controller) Run() error {
 
 	engine := mux.NewRouter()
 	engine.Use(DefaultHeaders(),
```
rchincha, line 62: Why don't we do this?
if c.Audit {
engine.Use(AuditLogger)
}

roxanaN: Do you mean having 2 separate middleware functions?

rchincha: Yes, wouldn't that be cleaner?
Sorry it is taking a little longer with this PR, but let's fix it cleanly.

pkg/log/log.go
```diff
@@ -102,15 +136,25 @@ func SessionLogger(log Logger) mux.MiddlewareFunc {
...
 				Str("latency", latency.String()).
 				Int("bodySize", bodySize).
 				Interface("headers", headers).
 				Msg("HTTP API")
+
+			if (method == http.MethodPost || method == http.MethodPut ||
+				method == http.MethodPatch || method == http.MethodDelete) &&
+				(statusCode == http.StatusOK || statusCode == http.StatusCreated || statusCode == http.StatusAccepted) {
+				audit.Info().
+					Str("subject", username).
```
rchincha, line 152: Another issue is that we may not have a "username" if Bearer AuthN is used. So maybe we should also add ClientIP (above) into this record.

rchincha (approved):
lgtm

PR #201, opened 2021-05-25
Implement a way to search for an image by manifest, config or layer digest

### LINKED ISSUE #175: cli: given a layer hash, return a list of images that contain that layer
Currently, we support listing/searching images within a repo, but we have found it useful to also retrieve a list of images given a layer hash.

andaaron:
```
Usage:
  zot images [config-name] [flags]

Flags:
  -d, --digest string   List images containing a specific manifest, config, or layer digest
[...]
```

Resolves #175

### DIFF
--- a/pkg/extensions/extensions.go
+++ b/pkg/extensions/extensions.go
@@ -33,7 +33,7 @@ func downloadTrivyDB(dbDir string, log log.Logger, updateInterval time.Duration)
 
 // EnableExtensions ...
 func EnableExtensions(extension *ExtensionConfig, log log.Logger, rootDir string) {
-	if extension.Search != nil && extension.Search.CVE != nil {
+	if extension.Search != nil && extension.Search.Enable && extension.Search.CVE != nil {
 		defaultUpdateInterval, _ := time.ParseDuration("2h")
 
 		if extension.Search.CVE.UpdateInterval < defaultUpdateInterval {
@@ -46,7 +46,7 @@ func EnableExtensions(extension *ExtensionConfig, log log.Logger, rootDir string
 			err := downloadTrivyDB(rootDir, log,
 				extension.Search.CVE.UpdateInterval)
 			if err != nil {
-				panic(err)
+				log.Error().Err(err).Msg("error while downloading TrivyDB")
 			}
 		}()
 	} else {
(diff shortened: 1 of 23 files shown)


rchincha:
@andaaron, couple of comments:
1) the original images-by-name search now returns config, layers, etc
I would keep the original output but add a -vvv type option for more details

2) what is the expected argument format for -d option? the full digest like "sha256:abcd...."? We should also allow partial matches

andaaron:
> @andaaron, couple of comments:
> 
> 1. the original images-by-name search now returns config, layers, etc
>    I would keep the original output but add a -vvv type option for more details
> 2. what is the expected argument format for -d option? the full digest like "sha256:abcd...."? We should also allow partial matches

1. Sounds good
2. Currently it's any substring that matches the digest, including the 'sha256:' part.

pkg/extensions/extensions.go
```diff
@@ -55,9 +55,13 @@ func EnableExtensions(extension *ExtensionConfig, log log.Logger, rootDir string
...
 // SetupRoutes ...
-func SetupRoutes(router *mux.Router, storeController storage.StoreController, log log.Logger) {
+func SetupRoutes(extension *ExtensionConfig, router *mux.Router, storeController storage.StoreController,
+	log log.Logger) {
 	log.Info().Msg("setting up extensions routes")
-	resConfig := search.GetResolverConfig(log, storeController)
-	router.PathPrefix("/query").Methods("GET", "POST").
-		Handler(gqlHandler.NewDefaultServer(search.NewExecutableSchema(resConfig)))
+
+	if extension.Search != nil && extension.Search.Enable {
```
shimish2, line 62: I think we are missing one case here, if user did not set enable to true but provided cve config, we will not setup /query routes but we will still download cve database. 

pkg/extensions/extensions.go 
``` 
// Enable extensions if extension config is provided		if c.Config != nil &amp;&amp; c.Config.Extensions != nil {			ext.EnableExtensions(c.Config.Extensions, c.Log, c.Config.Storage.RootDirectory)		}
```

and  https://github.com/andaaron/zot/blob/cli1/pkg/extensions/extensions.go#L35

andaaron: Updated: https://github.com/anuvu/zot/pull/201/files#diff-86b8c7ae55e63b332b1b61b1e50093ca23babbdf0ad78ff1f74bcbec501cc188R36

rchincha (approved):
lgtm

PR #203, opened 2021-06-18
Added new extension "sync"

peusebiu:
Periodically poll registries and pull images according to sync's config


Makefile
```diff
@@ -16,16 +16,16 @@ binary-minimal: doc
 
 .PHONY: binary
 binary: doc
-	go build -tags extended -v -ldflags "-X  github.com/anuvu/zot/pkg/api.Commit=${COMMIT} -X github.com/anuvu/zot/pkg/api.BinaryType=extended" -o bin/zot ./cmd/zot
+	go build -tags extended,containers_image_openpgp -v -ldflags "-X  github.com/anuvu/zot/pkg/api.Commit=${COMMIT} -X github.com/anuvu/zot/pkg/api.BinaryType=extended" -o bin/zot ./cmd/zot
```
rchincha, line 19: why was this "containers_image_openpgp" needed?

peusebiu: I had build errors on both conformance and travis CI:

```
# github.com/containers/image/v5/signature
/go/pkg/mod/github.com/containers/image/v5@v5.12.0/signature/mechanism_gpgme.go:16:16: undefined: gpgme.Context
```
Found this tag in an issue which was talking about same error.

examples/config-sync.json
```diff
@@ -0,0 +1,45 @@
...
+		"address":"127.0.0.1",
+		"port":"8080"
+	},
+	"log":{
+		"level":"debug"
+	},
+	"extensions":{
+		"sync":[
+			{
+				"url":"registry1",
```
rchamarthy, line 16: we will need proxy information and any certs that server may publish for trust in this configuration.

pkg/extensions/sync/sync.go
```diff
@@ -0,0 +1,363 @@
...
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	_, _ = w.Write([]byte{})
+}
+
+// getCatalog gets all repos from a registry.
+func getCatalog(registryName string) (catalog, error) {
+	var c catalog
+
+	registryCatalogURL := fmt.Sprintf("%s%s%s", "https://", registryName, "/v2/_catalog")
```
rchamarthy, line 61: Its best if we take the entire URL as config instead of assuming https.

pkg/extensions/sync/sync.go
```diff
@@ -0,0 +1,363 @@
...
+	client := resty.New()
```
rchamarthy, line 62: we need proxy support else this whole sync feature is un-usable.

peusebiu: I've tested it, and it reads the proxy config from env variables. Do you think we still need the config?

peusebiu:
I can not further improve the code coverage in sync... I'm testing all the lines, but the errors in containers/image.
If there is a way please let me know.

pkg/extensions/sync/sync.go
```diff
@@ -0,0 +1,436 @@
...
+	registryName := strings.Replace(strings.Replace(regCfg.URL, "http://", "", 1), "https://", "", 1)
+
+	repos := filterRepos(catalog.Repositories, regCfg.Content)
+
+	images, err := imagesToCopyFromRepos(registryName, repos, srcCtx, log)
+	if err != nil {
+		return err
+	}
+
+	err = filterImagesByTagRegex(&images, regCfg.Content, log)
```
shimish2, line 340: Is it possible to call filterImagesByTagRegex and filterImagesBySemver inside imagesToCopyFromRepos https://github.com/peusebiu/zot/blob/sync/pkg/extensions/sync/sync.go#L292 ?

shimish2 (commented):
Added my comments and you can increase code coverage by adding more negative cases e.g  deep copy fails due to invalid config, providing invalid cert path etc.

pkg/extensions/sync/sync.go
```diff
@@ -0,0 +1,436 @@
...
+				tagged := getTagFromRef(refs[i], log)
+				if tagged == nil {
+					continue
+				}
+
+				_, ok := semver.NewVersion(tagged.Tag())
+				if ok != nil {
+					log.Info().Msgf("tag %s is not semver, skipping", tagged.Tag())
+
+					refs = append(refs[:i], refs[i+1:]...)
```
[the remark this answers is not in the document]
peusebiu: I'm deleting refs (inplace) which are not semver compliant.
Because my for loops until "i"<len(refs) I need to decrement "i" when I'm deleting a ref. Otherwise  this line:
tagged := getTagFromRef(refs[i], log)
will get index error.

pkg/extensions/sync/sync.go
peusebiu: Sure.

pkg/extensions/sync/sync.go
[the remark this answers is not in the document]
shimish2: and refs[i+1] will never give out of range ?

pkg/extensions/sync/sync.go
```diff
@@ -0,0 +1,521 @@
...
+		if err != nil {
+			return c, err
+		}
+
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(caCert)
+
+		client.SetTLSClientConfig(&tls.Config{RootCAs: caCertPool})
+
+		cert, err := tls.LoadX509KeyPair(clientCert, clientKey)
```
shimish2, line 126: to get a coverage for line 128, If we will pass invalid key pair e.g (different public key corresponding to private key) it should give error.

pkg/extensions/sync/sync.go
```diff
@@ -0,0 +1,521 @@
...
+}
+
+// parseRepositoryReference parses input into a reference.Named, and verifies that it names a repository, not an image.
+func parseRepositoryReference(input string) (reference.Named, error) {
+	ref, err := reference.ParseNormalizedNamed(input)
+	if err != nil {
+		return nil, err
+	}
+
+	if !reference.IsNameOnly(ref) {
```
shimish2, line 173: 174 can be covered if we pass image ?

pkg/extensions/sync/sync_test.go
```diff
@@ -0,0 +1,1367 @@
+package sync_test
```
shimish2, line 1: by changing package name to sync, you can add lot more tests that will cover all internal functions.

andaaron: In case you need to test internal functions, add a separate package 'sync_internal_test', that should work without moving the tests inside 'sync'.

pkg/extensions/sync/sync.go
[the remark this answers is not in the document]
peusebiu: https://play.golang.org/p/WUfd-ENZfFT

I played with the code, I guess I'm not missing anything.

pkg/extensions/sync/sync.go
```diff
@@ -0,0 +1,524 @@
...
+		// schedule each registry sync
+		ticker = time.NewTicker(regCfg.PollInterval)
+
+		go func(cfg RegistryConfig) {
+			defer os.RemoveAll(certsDir)
+			defer policyContext.Destroy() //nolint: errcheck
+			// run sync first, then run on interval
+			if err := syncRegistry(cfg, log, destRegistry, destCtx, policyContext); err != nil {
+				log.Err(err).Msg("Sync exited with error, stopping it...")
+				ticker.Stop()
```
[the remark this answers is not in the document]
peusebiu: I use a retry package which retries 3 times, on 5 minutes intervals, only if the error is retriable(network issues), otherwise it won't.

pkg/extensions/sync/sync.go
```diff
@@ -0,0 +1,524 @@
...
+	var catalog catalog
+
+	if err = retry.RetryIfNecessary(context.Background(), func() error {
+		catalog, err = getCatalog(regCfg)
+		return err
+	}, retryOptions); err != nil {
+		return err
+	}
+
+	registryName := strings.Replace(strings.Replace(regCfg.URL, "http://", "", 1), "https://", "", 1)
```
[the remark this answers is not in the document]
peusebiu: I need the url to registry to get the catalog, but the registryName which I pass to containers/image package doesn't need the protocol, so I just strip it after I get the catalog.

pkg/extensions/extensions.go
```diff
@@ -31,8 +34,7 @@ func downloadTrivyDB(dbDir string, log log.Logger, updateInterval time.Duration)
 	}
 }
 
-// EnableExtensions ...
-func EnableExtensions(extension *ExtensionConfig, log log.Logger, rootDir string) {
+func EnableExtensions(extension *extConf.ExtensionConfig, log log.Logger, rootDir string, config *config.Config) {
```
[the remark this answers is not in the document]
peusebiu: Yes, and it passes, I don't understand, what is wrong with that line?

pkg/api/config/config.go
```diff
@@ -120,6 +121,28 @@ func (c *Config) Sanitize() *Config {
...
+			if regCfg.Credentials != (types.DockerAuthConfig{}) {
+				regCfg.Credentials.Password = "******"
+			}
+
+			regCfgs[id] = regCfg
+		}
+
+		s.Extensions.Sync = &regCfgs
+
+		return s
```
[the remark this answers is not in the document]
peusebiu: Ok, I missed this one, thanks.

.github/workflows/ci-cd.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           go get -u github.com/swaggo/swag/cmd/swag
           go mod download
           sudo apt-get update
-          sudo apt-get install rpm
+          sudo apt-get install rpm libgpgme-dev libassuan-dev libbtrfs-dev libdevmapper-dev
```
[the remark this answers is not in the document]
peusebiu: I need to check this one.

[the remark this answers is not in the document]
peusebiu: Not needed anymore.

pkg/api/config/config.go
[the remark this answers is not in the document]
peusebiu: Fixed

errors/errors.go
```diff
@@ -39,4 +39,5 @@ var (
 	ErrDuplicateConfigName     = errors.New("cli: cli config name already added")
 	ErrInvalidRoute            = errors.New("routes: invalid route prefix")
 	ErrImgStoreNotFound        = errors.New("routes: image store not found corresponding to given route")
+	ErrInvalidRepositoryName   = errors.New("input names a reference, not a repository")
```
rchincha, line 42: Make this "routes: not a repository name"

Makefile
```diff
@@ -34,7 +34,7 @@ covhtml:
 .PHONY: check
 check: ./golangcilint.yaml
 	golangci-lint --version || curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s v1.26.0
-	golangci-lint --config ./golangcilint.yaml run --enable-all --build-tags extended ./cmd/... ./pkg/...
```
rchincha, line 37: containers_image_openpgp? what is this used for?

README.md
```diff
@@ -221,6 +222,19 @@ c3/openjdk-dev                    commit-2674e8a-squashfs   b545b8ba  321MB
 c3/openjdk-dev                    commit-d5024ec-squashfs   cd45f8cf  321MB
 ```
 
+# Sync
+Periodically synchronize images between zot registries.
+The synchronization is achieved by copying all the images found at source to destination.
+To use it see [sync-config](examples/config-sync.json)
+Supports:
+  - TLS verifying
```
rchincha, line 230: "TLS verification"

README.md
```diff
@@ -221,6 +222,19 @@ c3/openjdk-dev                    commit-2674e8a-squashfs   b545b8ba  321MB
...
+  - Prefix filtering (can contain multiple repos, eg repo1/repoX/repoZ)
+  - Tags regex filtering
+  - Tags semver compliance filtering (the 'v' prefix is optional)
+  - Basic auth
```
rchincha, line 234: "BASIC"

Makefile
peusebiu: see https://github.com/containers/image#supported-build-tags.

I couldn't build it otherwise.

README.md
peusebiu: fixed, thank you.

README.md
peusebiu: fixed, thank you.

README.md
```diff
@@ -31,6 +31,7 @@ https://anuvu.github.io/zot/
   * Automatic garbage collection of orphaned blobs
   * Layer deduplication using hard links when content is identical
 * Serve [multiple storage paths (and backends)](./examples/config-multiple.json) using a single zot server
+* Synchronize with other zot registries [sync](#sync)
```
rchincha, line 34: "Pull and synchronize ..."

README.md
```diff
@@ -221,6 +222,19 @@ c3/openjdk-dev                    commit-2674e8a-squashfs   b545b8ba  321MB
 c3/openjdk-dev                    commit-d5024ec-squashfs   cd45f8cf  321MB
 ```
 
+# Sync
+Periodically synchronize images between zot registries.
```
rchincha, line 226: "Periodically pull and synchronize ..."

cmd/zot/main_test.go
```diff
@@ -4,14 +4,15 @@ import (
...
+	"github.com/anuvu/zot/pkg/api/config"
 	"github.com/anuvu/zot/pkg/cli"
 	. "github.com/smartystreets/goconvey/convey"
 )
 
 func TestIntegration(t *testing.T) {
 	Convey("Make a new controller", t, func() {
-		config := api.NewConfig()
-		c := api.NewController(config)
+		conf := config.New()
```
rchincha, line 14: why this refactor?

pkg/extensions/extensions.go
```diff
@@ -52,16 +54,72 @@ func EnableExtensions(extension *ExtensionConfig, log log.Logger, rootDir string
...
+			}
+		}
+
+		var serverCert string
+
+		var serverKey string
+
+		var CACert string
+
+		if config.HTTP.TLS != nil {
```
rchincha, line 74: What are you trying to achieve here? I see that you are copying these certs. But anything that is needed by the extension should be in the ExtensionConfig itself.

rchincha:
Is this "sync" also on-demand? Let's say we sync only ever 1 hour. We do a GET image at downstream zot which will fail, now we wait for an hour before this call succeeds? Maybe we should trigger a sync immediately, even better, do the sync in the context of the original call.

cmd/zot/main_test.go
peusebiu: I needed to refactor config because I needed certs inside sync package, and because config was part of api package which imports extensions => circular import.

pkg/extensions/extensions.go
peusebiu: I thought that repeating information in config was bad, but if you agree to repeat this information in the ExtensionConfig, then I don't need the config refactoring above.

peusebiu:
> Is this "sync" also on-demand? Let's say we sync only ever 1 hour. We do a GET image at downstream zot which will fail, now we wait for an hour before this call succeeds? Maybe we should trigger a sync immediately, even better, do the sync in the context of the original call.

It's also on demand yes, it's listening on a POST call on /sync, see https://github.com/anuvu/zot/pull/203/files/00bd7dd7fbceca0927cd69ca6ac98b81fb70ed2d#diff-86b8c7ae55e63b332b1b61b1e50093ca23babbdf0ad78ff1f74bcbec501cc188R123

README.md
peusebiu: Fixed

README.md
peusebiu: fixed

peusebiu:
Added sync on demand and it works as follows:
1) if the image/manifest for that image can not be found when a client GETs a manifest, sync that image from one of the registry in the sync configuration and then return the manifest to the client.
2) Otherwise return errors (repo not found/manifest not found)

@rchincha 
We need to come up with a way to specify dockerhub in the configuration file as a source registry for the sync-on-demand feature. Right now sync-on-demand works only with the registries specified in the configuration file.

errors/errors.go
```diff
@@ -40,4 +40,6 @@ var (
 	ErrInvalidRoute            = errors.New("routes: invalid route prefix")
 	ErrImgStoreNotFound        = errors.New("routes: image store not found corresponding to given route")
 	ErrEmptyValue              = errors.New("cache: empty value")
+	ErrInvalidRepositoryName   = errors.New("routes: not a repository name")
+	ErrSyncRegistry            = errors.New("sync: couldn't fetch registry's catalog")
```
rchincha, line 44: ErrSyncMissingCatalog

examples/config-sync.json
```diff
@@ -0,0 +1,52 @@
...
+	"log":{
+		"level":"debug"
+	},
+	"extensions":{
+		"sync":[
+			{
+				"url": "https://registry1:5000",
+				"pollInterval": "6h",
+				"tlsVerify": true,
+				"credentials": {
```
rchincha, line 19: Would consider "credentials" to point to a separate file instead, so that we can align with kubernetes secrets [1] which can be mounted as a file also.

[1] https://kubernetes.io/docs/concepts/configuration/secret/

examples/config-sync.json
```diff
@@ -0,0 +1,52 @@
...
+					"username": "user",
+					"password": "pass"
+				},
+				"certDir": "/home/user/certs",
+				"content":[
+					{
+						"prefix":"repo1/repo",
+						"tags":{
+							"regex":"4.*",
```
rchincha, line 28: So this means get me all semver tags starting with 4.*

examples/config-sync.json
```diff
@@ -0,0 +1,52 @@
...
+							"semver":true
```
rchincha, line 29: What happens when semver is false?

peusebiu: It will exclude all the tags filtered earlier by "4.*", so right now filtering by tags is an intersection.
Would it be ok if I make it an union?

eg: Sync all tags with regex 4.x and all tags which are not semver.

examples/config-sync.json
peusebiu: Yes. it's an intersection.

examples/config-sync.json
peusebiu: Ok, added credentials file along with plain text

errors/errors.go
peusebiu: fixed

pkg/extensions/sync/sync.go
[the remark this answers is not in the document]
andaaron: Note in your example you always skip the first element.

An other interesting perspective here: https://zetcode.com/golang/filter-slice/

examples/config-sync.json
andaaron: I find it counterintuitive to apply both filters one after the other. @rchincha what do you think?

peusebiu: I haven't touch it yet, I will make it a union, if you all agree.

pkg/extensions/extensions.go
```diff
@@ -52,16 +54,101 @@ func EnableExtensions(extension *ExtensionConfig, log log.Logger, rootDir string
...
+			Cfg:        *extension.Sync,
+			Log:        log,
+		}
+
+		router.HandleFunc("/sync", postSyncer.Handler).Methods("POST")
+	}
+}
+
+// SyncOneImage ...
+func SyncOneImage(config *config.Config, log log.Logger, repoName, reference string) (bool, error) {
```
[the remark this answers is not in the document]
peusebiu: What do you mean about different implementation w/o extensions?
SyncOneImage will not work if extensions.Sync is not enabled.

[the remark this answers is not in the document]
andaaron: I mean currently you have two implementations of SyncOneImage, one in minimal.go and one here.

rchincha:
> 
> 
> This part is not clear to me in terms of intended behavior:
> 
> > One question though is for OnDemand sync of public repos.
> > GET docker.io/image:tag to sync from docker and put it locally under docker.io/image:tag similarly other public repos. Would be awesome if we could add a unit test for this also.
> 
> Let me try to see if I get it right. Do you mean:
> 
> > GET `zot.mydomain.com/docker.io/image:tag` to sync `docker.io/image:tag` from docker and put it locally under `zot.mydomain.com/docker.io/image:tag` similarly other public repos. Would be awesome if we could add a unit test for this also.
> 
> ?
> 
> This is inconsistent because the image if synced using sync on demand would be stored under ` zot.mydomain.com/docker.io/image:tag` and the same image if synced by the logic in the policy would be stored under `zot.mydomain.com/image:tag` And we would end up with 2 different copies of the same image.
> 
> Or am I missing something?

Let's say I have busybox:1.0 and two sync upstreams

GET zot.mydomain.com/busybox:latest, which upstream do you get it from? What do you do if both upstreams have it and how do you know that both upstreams have the same meaning for "busybox" image repo. I am ok with what you are proposing that as long as:
1. we assume as long as ANY of the upstreams have it, we can pull it in,
2. priority is list order as specified in configuration,
3.  and also "docker.io" etc are special repos? how do we specify this in configuration

rchincha:
> > > Let's say I have busybox:1.0 and two sync upstreams
> > GET zot.mydomain.com/busybox:latest, which upstream do you get it from? What do you do if both upstreams have it and how do you know that both upstreams have the same meaning for "busybox" image repo. I am ok with what you are proposing that as long as:
> > 
> > 1. we assume as long as ANY of the upstreams have it, we can pull it in,
> > 2. priority is list order as specified in configuration,
> > 3. and also "docker.io" etc are special repos? how do we specify this in configuration
> 
> I wasn't proposing anything. I'm trying to figure out what your proposal is, and what the implications are. To my understanding right now the non-on-demand sync doesn't take the original host into account when computing the local image path. But your comment mentions the on-demand version should.
> 
> Shouldn't the sync behave the same way in both cases? Or are you proposing to prefix the images with the original upstream hostname in all cases?
> 
> Also I understand your points 1 and 2, but why would "docker.io" be special in your point 3?

For the first iteration of this feature, let's keep it simple. 
1.  Priority order is configuration specification order
2. "docker.io", "quay.io" etc are just another sync upstreams, nothing special
3. We will try getting "image:tag" from upstreams in order or step 1 above.

PR #205, opened 2021-06-29
api: use blob cache path while making hard link

shimish2:
previously mount blob will look for blob that is provided in http request and try to hard link that path but ideally we should look for path from ourcache and do the hard link of that particular path, this commit does the same.

### DIFF
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -1801,6 +1801,7 @@ func TestCrossRepoMount(t *testing.T) {
 		if err != nil {
 			panic(err)
 		}
+
 		defer os.RemoveAll(dir)
 
 		c.Config.Storage.RootDirectory = dir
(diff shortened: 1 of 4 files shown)


pkg/api/controller_test.go
```diff
@@ -1907,6 +1909,27 @@ func TestCrossRepoMount(t *testing.T) {
 		So(err, ShouldBeNil)
 		So(postResponse.StatusCode(), ShouldEqual, 201)
 
+		// We have uploaded a blob and since we have provided digest it should be full blob upload and there should be entry
+		// in cache, now try mount blob request status and it should be 201 because now blob is present in cache
+		// and it should do hard link.
+
+		params["mount"] = digest
```
rchincha, line 1916: remove this "mount" param?

shimish2: I am setting params["mount"] = "sha:" some lines above, so before making next request i am assigning valid digest.

PR #209, opened 2021-07-06
TLS certs in CLI client

roxanaN:


pkg/cli/client.go
```diff
@@ -74,6 +77,37 @@ func doHTTPRequest(req *http.Request, verifyTLS bool, resultsPtr interface{}) (h
 		httpClient = createHTTPClient(verifyTLS)
 	}
 
+	// TODO check if /etc/con...etc exists
+
+	clientCert := "/etc/containers/certs.d/" + req.Host + "/client.cert"
```
shimish2, line 82: use fmt.Sprintf()

pkg/cli/client_test.go
```diff
@@ -0,0 +1,169 @@
...
+			args := []string{"--url", HOST}
+			configPath := makeConfigFile(`{"configs":[{"_name":"imagetest","showspinner":false}]}`)
+			defer os.Remove(configPath)
+			imageCmd := NewImageCommand(new(mockService))
+			imageBuff := bytes.NewBufferString("")
+			imageCmd.SetOut(imageBuff)
+			imageCmd.SetErr(ioutil.Discard)
+			imageCmd.SetArgs(args)
+			err := imageCmd.Execute()
+			// So(err, ShouldEqual, "cli: invalid URL format")
```
shimish2, line 94: looks like args should be args := []string{"imagetest","--url", HOST} and 
compare equivalent objects, error to error not error to string.

pkg/cli/client.go
roxanaN: I plan to use [path join](https://golang.org/pkg/path/#example_Join).
Is that ok?

shimish2: Yes

rchincha:
* /home/runner/work/zot/zot/pkg/cli/client_test.go 
  Line 242:
  Expected: nil
  Actual:   'Get "https://127.0.0.1:8088/v2/_catalog": x509: certificate signed by unknown authority'
  goroutine 40 [running]:

roxanaN:
> go test -tags extended -race -v -run TestTLSWithoutAuth ./pkg/cli/...
> 
> * /home/runner/work/zot/zot/pkg/cli/client_test.go
>   Line 242:
>   Expected: nil
>   Actual:   'Get "https://127.0.0.1:8088/v2/_catalog": x509: certificate signed by unknown authority'
>   goroutine 40 [running]:

pkg/cli/client.go
```diff
@@ -19,8 +22,16 @@ import (
 )
 
 var httpClient *http.Client //nolint: gochecknoglobals
-
-const httpTimeout = 5 * time.Minute
+var mux sync.Mutex          //nolint: gochecknoglobals
```
andaaron, line 25: Please rename the variable to suggest what it's used for.

@rchincha @shimish2 we need this because the client transport is updated with certificates (which come from the http.Request object), and each request could have different host destinations, each needing a different CA or client certificate.

This does mean that all our requests would be done in a series.

Was there a specific reason why http.Client is a global? If it were local to doHTTPRequest we wouldn't have this problem.

rchincha: So we are going to be holding that mutex for the entire time we do a http request/response? Doesn't look right!

andaaron: Well yes, unless we make the assumption the client will connect to the same server every time and reuse the transport object with the certificates set the first time.

Or we make the client a local variable.

andaaron: I think setting the certificates once would be clean in: https://github.com/anuvu/zot/pull/209/files#diff-6661d8b960c79657b52bd4fd22680ca8b3608faa4d91c4e10ad52db3e6bf4300R36

shimish2: Why we need lock here, is it because of http client ?

andaaron: There's a data race here https://github.com/anuvu/zot/pull/209/files#diff-6661d8b960c79657b52bd4fd22680ca8b3608faa4d91c4e10ad52db3e6bf4300R129, and tests are failing.

pkg/cli/client.go
```diff
@@ -6,32 +6,84 @@ import (
...
+			return nil, err
+		}
+
+		tlsConfig = returnedTLSConfig
+	} else {
+		// Check if the /etc/containers/certs.d/$IP:$PORT dir exists
+		clientCertsDir := filepath.Join(certsPath, host)
+		if dirExists(clientCertsDir) {
+			returnedTLSConfig, err := getTLSConfig(clientCertsDir, caCertPool, tlsConfig)
+			if err != nil && !os.IsPermission(err) {
```
andaaron, line 68: Why treat the permissions error differently?
I suggest at least logging permission errors.

pkg/cli/client.go
```diff
@@ -70,8 +122,21 @@ func makeGraphQLRequest(url, query, username,
 }
 
 func doHTTPRequest(req *http.Request, verifyTLS bool, resultsPtr interface{}) (http.Header, error) {
-	if httpClient == nil {
-		httpClient = createHTTPClient(verifyTLS)
+	var httpClient *http.Client
+
+	host := req.Host
+
+	if httpClientsMap[host] == nil {
```
andaaron, line 129: I believe you should still use a mutex while reading/updating the dictionary.
There should not be a performance hit, most of the time waiting is for the request itself.

Makefile
```diff
@@ -25,8 +25,15 @@ debug: doc
 .PHONY: test
 test:
 	$(shell mkdir -p test/data;  cd test/data; ../scripts/gen_certs.sh; cd ${TOP_LEVEL}; sudo skopeo --insecure-policy copy -q docker://public.ecr.aws/t0x7q1g8/centos:7 oci:${TOP_LEVEL}/test/data/zot-test:0.0.1;sudo skopeo --insecure-policy copy -q docker://public.ecr.aws/t0x7q1g8/centos:8 oci:${TOP_LEVEL}/test/data/zot-cve-test:0.0.1)
+	$(shell sudo mkdir -p /etc/containers/certs.d/127.0.0.1:8088/; sudo cp test/data/client.* /etc/containers/certs.d/127.0.0.1:8088/; sudo cp test/data/ca.* /etc/containers/certs.d/127.0.0.1:8088/;)
+	$(shell sudo chmod a=rwx /etc/containers/certs.d/127.0.0.1:8088/*.key)
 	go test -tags extended -v -race -cover -coverpkg ./... -coverprofile=coverage.txt -covermode=atomic ./...
 
+.PHONY: test-cleaner
```
rchincha, line 32: Make this a part of "clean" target itself, why create a new one

pkg/cli/client.go
```diff
@@ -6,32 +6,83 @@ import (
...
+				return nil, err
+			} else if err == nil {
+				tlsConfig = returnedTLSConfig
+			}
+		}
 	}
 
+	tlsConfig.BuildNameToCertificate() // nolint: staticcheck
+
+	tr := http.DefaultTransport.(*http.Transport).Clone()
```
rchincha, line 79: A roundabout way to do:
tr := &http.Transport{TLSClientConfig: tlsConf}

pkg/cli/client.go
```diff
@@ -6,32 +6,83 @@ import (
...
+		insecureSkipVerify = true
+	}
+
+	tlsConfig := &tls.Config{
+		RootCAs:            caCertPool,
+		InsecureSkipVerify: insecureSkipVerify, //nolint: gosec
+	}
+
+	// Check if the /home/user/.config/containers/certs.d/$IP:$PORT dir exists
+	home := os.Getenv("HOME")
```
rchincha, line 54: if "verifyTLS" == false, why even go down this path?

pkg/cli/client.go
```diff
@@ -6,32 +6,83 @@ import (
...
 	"strings"
 	"sync"
 	"time"
 
 	zotErrors "github.com/anuvu/zot/errors"
 )
 
-var httpClient *http.Client //nolint: gochecknoglobals
+var httpClientsMap = make(map[string]*http.Client) //nolint: gochecknoglobals
+var accessControl sync.Mutex                       //nolint: gochecknoglobals
```
rchincha, line 25: call this "httpClientLock"

pkg/cli/client.go
```diff
@@ -70,8 +121,24 @@ func makeGraphQLRequest(url, query, username,
...
+		if err != nil {
+			return nil, err
+		}
+
+		accessControl.Lock()
+		defer accessControl.Unlock()
+
+		httpClientsMap[host] = httpClient
+	} else {
+		httpClient = httpClientsMap[host]
```
rchincha, line 141: This is not correct. This must also be under lock for read.

Makefile
roxanaN: I think it should belong to the test side and for this reason I put it separately.

rchincha:
@roxanaN a couple of fixes/cleanup of your PR. Pls see this commit:
https://github.com/rchincha/zot/commit/f4222efd789ceec632fdb825535c7d8b33d6ea60

andaaron (approved):

PR #212, opened 2021-07-20
cache fixes

shimish2:
- dedupe: stat blob path before creating link

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -39,4 +39,5 @@ var (
 	ErrDuplicateConfigName     = errors.New("cli: cli config name already added")
 	ErrInvalidRoute            = errors.New("routes: invalid route prefix")
 	ErrImgStoreNotFound        = errors.New("routes: image store not found corresponding to given route")
+	ErrEmptyValue              = errors.New("cache: empty value")
 )
--- a/pkg/storage/cache.go
+++ b/pkg/storage/cache.go
@@ -51,6 +51,12 @@ func NewCache(rootDir string, name string, log zlog.Logger) *Cache {
 }
 
 func (c *Cache) PutBlob(digest string, path string) error {
+	if path == "" {
+		c.log.Error().Err(errors.ErrEmptyValue).Str("digest", digest).Msg("empty path provided")
+
+		return errors.ErrEmptyValue
+	}
+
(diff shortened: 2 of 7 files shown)


pkg/storage/storage.go
```diff
@@ -1084,6 +1091,13 @@ func (is *ImageStore) checkCacheBlob(digest string) (string, error) {
 		return "", err
 	}
 
+	// there is an entry of digest in cache with empty value.
+	if dstRecord == "" {
```
rchincha, line 1095: empty record means "no entry found", would enforce that check in cache.PutBlob() that we don't ever write an empty record

shimish2: Yes we can add additional check there also

errors/errors.go
```diff
@@ -39,4 +39,5 @@ var (
 	ErrDuplicateConfigName     = errors.New("cli: cli config name already added")
 	ErrInvalidRoute            = errors.New("routes: invalid route prefix")
 	ErrImgStoreNotFound        = errors.New("routes: image store not found corresponding to given route")
+	ErrEmptyPath               = errors.New("cache: empty path")
```
rchincha, line 42: Call this ErrEmptyValue = errors.New("cache: empty value")

pkg/storage/cache.go
```diff
@@ -51,6 +51,12 @@ func NewCache(rootDir string, name string, log zlog.Logger) *Cache {
 }
 
 func (c *Cache) PutBlob(digest string, path string) error {
+	if path == "" {
+		c.log.Error().Err(errors.ErrEmptyPath).Msg("empty path provided")
```
rchincha, line 55: Include the "digest" in this log

pkg/storage/storage.go
```diff
@@ -985,7 +985,18 @@ retry:
 			}
 			goto retry
 		}
+
 		dstFi, err := os.Stat(dst)
+		if err == nil {
+			is.log.Debug().Str("blobPath", dst).Msg("dedupe: blob exists")
+			err = os.Remove(src)
```
rchincha, line 992: Why is this needed?

rchincha (approved):
lgtm

PR #214, opened 2021-07-23
routes: add Content-Type header for CHECK blob

rchincha:
With recent docker client-side changes, on 'docker pull' we see:
"Error response from daemon: missing or empty Content-Type header"

Hence, set Content-Type header.


rchincha:
For reference: https://github.com/goharbor/harbor/issues/13740

shimish2 (commented):
unit test to just verify header is set with correct value ?

shimish2 (approved):
lgtm

PR #216, opened 2021-08-09
storage: add s3 backend support (without GC and dedupe)

peusebiu:
using: https://github.com/distribution/distribution/tree/main/registry/storage/driver

Supports a lot of storages types.

### DIFF
--- a/examples/config-s3.json
+++ b/examples/config-s3.json
@@ -1,0 +1,53 @@
+{
+    "version": "0.1.0-dev",
+    "storage": {
+        "rootDirectory": "/zot",
+        "storageDriver": {
+            "name": "s3",
+            "region": "us-east-2",
+            "bucket": "zot-storage",
+            "secure": true,
+            "skipverify": false
+        },
+        "subPaths": {
+            "/a": {
+                "rootDirectory": "/zot-a",
+                "storageDriver": {
+                    "name": "s3",
+                    "region": "us-east-2",
+                    "bucket": "zot-storage",
+                    "secure": true,
+                    "skipverify": false
+                }
+            },
+            "/b": {
+                "rootDirectory": "/zot-b",
+                "storageDriver": {
+                    "name": "s3",
+                    "region": "us-east-2",
+                    "bucket": "zot-storage",
+                    "secure": true,
+                    "skipverify": false
+                }
+            },
+            "/c": {
+                "rootDirectory": "/zot-c",
+                "storageDriver": {
+                    "name": "s3",
+                    "region": "us-east-2",
+                    "bucket": "zot-storage",
+                    "secure": false,
+                    "skipverify": false
+                }
+            }
+        }
+    },
+    "http": {
+        "address": "127.0.0.1",
+        "port": "8080",
+        "ReadOnly": false
+    },
+    "log": {
+        "level": "debug"
+    }
+}
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -15,8 +15,11 @@ import (
 	"github.com/anuvu/zot/pkg/extensions/monitoring"
 	"github.com/anuvu/zot/pkg/log"
 	"github.com/anuvu/zot/pkg/storage"
+	"github.com/anuvu/zot/pkg/storage/s3"
 	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
+
+	"github.com/docker/distribution/registry/storage/driver/factory"
 )
 
 const (
@@ -62,6 +65,7 @@ func DefaultHeaders() mux.MiddlewareFunc {
 	}
 }
 
+// nolint: gocyclo
 func (c *Controller) Run() error {
 	// validate configuration
 	if err := c.Config.Validate(c.Log); err != nil {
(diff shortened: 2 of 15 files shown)


examples/config-s3.json
```diff
@@ -0,0 +1,23 @@
+{
+    "version": "0.1.0-dev",
+    "storage": {
+        "rootDirectory": "/zot",
+        "objectStoreParams": {
```
rchincha, line 5: let's review this part carefully.

pkg/api/controller.go
```diff
@@ -95,8 +98,23 @@ func (c *Controller) Run() error {
...
-			c.Config.Storage.GC, c.Config.Storage.Dedupe, c.Log)
+		var defaultStore storage.ImageStore
+		if len(c.Config.Storage.ObjectStoreParams) == 0 {
+			defaultStore = storage.NewImageStore(c.Config.Storage.RootDirectory,
+				c.Config.Storage.GC, c.Config.Storage.Dedupe, c.Log)
+		} else {
+			storeName := fmt.Sprintf("%v", c.Config.Storage.ObjectStoreParams["name"])
+
+			// Init a Storager from connection string.
+			store, err := factory.Create(storeName, c.Config.Storage.ObjectStoreParams)
```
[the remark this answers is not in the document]
peusebiu: I don;t understand the question.

You can do that already if you want with regionendpoint parameter.

```
                "storageDriver": {
                    "name": "s3",
                    "region": "us-east-2",
                    "regionendpoint": "localhost:4566",
                    "bucket": "zot-storage",
                    "secure": false,
                    "skipverify": false
                }
                
```

localhost:4566 is https://github.com/localstack/localstack

pkg/api/controller.go
```diff
@@ -96,8 +99,23 @@ func (c *Controller) Run() error {
...
+			storeName := fmt.Sprintf("%v", c.Config.Storage.StorageDriver["name"])
+
+			// Init a Storager from connection string.
+			store, err := factory.Create(storeName, c.Config.Storage.StorageDriver)
+			if err != nil {
+				c.Log.Error().Err(err).Str("rootDir", c.Config.Storage.RootDirectory).Msg("Unable to create s3 service")
+				return err
+			}
+
+			defaultStore = s3.NewImageStore(c.Config.Storage.RootDirectory,
```
rchincha, line 116: This can't be right. Don't we need to be told, it is a "s3 driver" before doing this. And drivers we don't support, 'zot verify' should reject and so should 'zot serve'

pkg/api/controller.go
```diff
@@ -96,8 +99,23 @@ func (c *Controller) Run() error {
...
 		}
 
-		defaultStore := storage.NewImageStore(c.Config.Storage.RootDirectory,
-			c.Config.Storage.GC, c.Config.Storage.Dedupe, c.Log)
+		var defaultStore storage.ImageStore
+		if len(c.Config.Storage.StorageDriver) == 0 {
+			defaultStore = storage.NewImageStore(c.Config.Storage.RootDirectory,
+				c.Config.Storage.GC, c.Config.Storage.Dedupe, c.Log)
+		} else {
+			storeName := fmt.Sprintf("%v", c.Config.Storage.StorageDriver["name"])
```
rchincha, line 107: Check and enforce it is "s3" driver only until we add support for others.

peusebiu: Fixed

pkg/api/controller.go
peusebiu: Ok, fixed, thanks.

pkg/api/controller.go
rchincha: One more comment. Move this "s3" into a pkg/storage/storage.go as some constant so we don't compare strings directly. Something like this.


```
const (
S3StorageDriverName = "s3"
)
```

pkg/api/controller.go
```diff
@@ -96,8 +100,25 @@ func (c *Controller) Run() error {
...
+			if storeName != "s3" {
+				c.Log.Fatal().Err(errors.ErrBadConfig).Msg("zot supports only s3 storage driver")
```
rchincha, line 110: Don't use project names directly. Just say "unsupported storage driver %s"

pkg/api/controller_test.go
```diff
@@ -131,6 +137,126 @@ func TestNew(t *testing.T) {
...
+		baseURL := getBaseURL(port, false)
+		conf := config.New()
+		conf.HTTP.Port = port
+
+		bucket := "zot-storage-test"
+		endpoint := os.Getenv("S3MOCK_ENDPOINT")
+
+		storageDriverParams := map[string]interface{}{
+			"rootDir":        "zot",
+			"name":           "s3",
```
rchincha, line 220: Make this S3StorageDriverName

rchincha: and other places too

pkg/api/controller_test.go
```diff
@@ -131,6 +137,126 @@ func TestNew(t *testing.T) {
...
 
+func TestObjectStorageController(t *testing.T) {
+	skipIt(t)
+	Convey("Negative make a new object storage controller", t, func() {
+		port := getFreePort()
+		conf := config.New()
+		conf.HTTP.Port = port
+		storageDriverParams := map[string]interface{}{
+			"rootDir": "zot",
+			"name":    "s3",
```
rchincha, line 148: Change this to the string constant.

PR #219, opened 2021-08-20
api: implement delete by tag

shimish2:
- Added image deletion by tag feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ https://anuvu.github.io/zot/
 * Uses [OCI image layout](https://github.com/opencontainers/image-spec/blob/master/image-layout.md) for image storage
   * Can serve any OCI image layout as a registry 
 * Supports [helm charts](https://helm.sh/docs/topics/registries/)
+* Supports image deletion by tag
 * Currently suitable for on-prem deployments (e.g. colocated with Kubernetes)
 * Compatible with ecosystem tools such as [skopeo](#skopeo) and [cri-o](#cri-o)
 * [Vulnerability scanning of images](#Scanning-images-for-known-vulnerabilities)
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -588,11 +588,14 @@ func (is *ImageStore) DeleteImageManifest(repo string, reference string) error {
 		return errors.ErrRepoNotFound
 	}
 
-	// as per spec "reference" can only be a digest and not a tag
+	isTag := false
+
+	// as per spec "reference" can be a digest and a tag
 	digest, err := godigest.Parse(reference)
 	if err != nil {
-		is.log.Error().Err(err).Msg("invalid reference")
-		return errors.ErrBadManifest
+		is.log.Debug().Str("invalid digest: ", reference).Msg("storage: assuming tag")
+
+		isTag = true
 	}
 
 	is.Lock()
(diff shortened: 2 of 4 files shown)


pkg/storage/storage.go
```diff
@@ -588,11 +588,14 @@ func (is *ImageStore) DeleteImageManifest(repo string, reference string) error {
...
 	}
 
+	isTag := false
+
 	// as per spec "reference" can only be a digest and not a tag
 	digest, err := godigest.Parse(reference)
 	if err != nil {
-		is.log.Error().Err(err).Msg("invalid reference")
-		return errors.ErrBadManifest
+		is.log.Debug().Str("invalid digest: ", reference).Msg("storage: assuming tag")
```
darshilgit, line 596: Is there a good way to see if it's a tag?

should we at least check if it matches a pattern? If not well defined than something as simple as having characters around `:`

pkg/storage/storage.go
```diff
@@ -588,11 +588,14 @@ func (is *ImageStore) DeleteImageManifest(repo string, reference string) error {
 		return errors.ErrRepoNotFound
 	}
 
+	isTag := false
+
 	// as per spec "reference" can only be a digest and not a tag
```
darshilgit, line 593: remove this

pkg/storage/storage.go
shimish2: we can enforce a check to look for specific set of tags following this rule https://github.com/opencontainers/image-spec/blob/main/annotations.md#rules but as of now we do not set any restriction on tag during upload so if we want to enforce this check we should do same for upload also.

pkg/storage/storage.go
shimish2: Sure will remove.

README.md
```diff
@@ -17,6 +17,7 @@ https://anuvu.github.io/zot/
 * Uses [OCI image layout](https://github.com/opencontainers/image-spec/blob/master/image-layout.md) for image storage
   * Can serve any OCI image layout as a registry 
 * Supports [helm charts](https://helm.sh/docs/topics/registries/)
+* Supports deletion by tag also.
```
rchincha, line 20: "Support image deletion by tag"

pkg/storage/storage.go
```diff
@@ -620,7 +623,19 @@ func (is *ImageStore) DeleteImageManifest(repo string, reference string) error {
...
+		if isTag {
+			tag, ok := m.Annotations[ispec.AnnotationRefName]
+			if ok && tag == reference {
+				is.log.Debug().Str("deleting tag", tag).Msg("")
+
+				digest = m.Digest
+
+				found = true
+
+				continue
```
rchincha, line 635: We really cannot have the same tag repeat in an index.json? or can we?

shimish2: No this loop is appending those manifests which we want to keep so we can't break from here .

README.md
```diff
@@ -17,6 +17,7 @@ https://anuvu.github.io/zot/
 * Uses [OCI image layout](https://github.com/opencontainers/image-spec/blob/master/image-layout.md) for image storage
   * Can serve any OCI image layout as a registry 
 * Supports [helm charts](https://helm.sh/docs/topics/registries/)
+* Supports image deletion by tag.
```
rchincha, line 20: remove the trailing period, none of the above statements have it

shimish2: Ok

rchincha (approved):
lgtm

PR #223, opened 2021-09-13
[Identity-based Authorization] Add an option to specify a policy for …

peusebiu:
…all repositories

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -186,40 +186,68 @@ identities. An additional per-repository default policy can be specified for
 identities not in the whitelist. Furthermore, a global admin policy can also be
 specified which can override per-repository policies.
 
+Glob patterns can also be used as repository paths.
+
+Authorization is granted based on the longest path matched.
+For example repos2/repo repository will match both "**" and "repos2/repo" keys,
+in such case repos2/repo policy will be used because it's longer.
+
+Because we use longest path matching we need a way to specify a global policy to override all the other policies.
+For example, we can specify a global policy with "**" (will match all repos), but any other policy will overwrite it,
+because it will be longer. So that's why we have the option to specify an adminPolicy.
+
+Basically '**' means repositories not matched by any other per-repository policy.
+
+create/update/delete can not be used without 'read' action, make sure read is always included in policies!
+
 ```
 "accessControl": {
-    "repos1/repo": {
-        "policies": [
+    "**": {                                                    # matches all repos (which are not matched by any other per-repository policy)
+      "policies": [                                            # user based policies
         {
-            "users": ["alice", "bob"],
-            "actions": ["create", "read", "update", "delete"]
-        },
-        {
-            "users": ["mallory"],
-            "actions": ["create", "read"]
+          "users": ["charlie"],
+          "actions": ["read", "create", "update"]
         }
+      ],
+      "defaultPolicy": ["read", "create"]                      # default policy which is applied for all users => so all users can read/create repositories
+    },
+    "tmp/**": {                                                # matches all repos under tmp/ recursively
+      "defaultPolicy": ["read", "create", "update"]            # so all users have read/create/update on all repos under tmp/ eg: tmp/infra/repo
+    },
+    "infra/*": {                                               # matches all repos directly under infra/ (not recursively)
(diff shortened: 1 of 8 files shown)


shimish2 (commented):
How we are dealing with this particular scenario:- 

Suppose for user **charlie** we specified an all repository policy and allowed action is **create** only and for another repository e.g repo1 a policy contains **charlie** user and permitted action is ["create","update","delete","read"].

 Are we allowing other actions apart from create  for user **charlie**  for repo1 ? 

e.g config
``` json 
"accessControl": {
    "/": {
      "policies": [
        {
          "users": ["charlie"],
          "actions": ["create"]
        }
      ],
      "defaultPolicy": ["read"]
    },
    "repos1/repo": {
        "policies": [
        {
            "users": ["alice", "bob","charlie"],
            "actions": ["create", "read", "update", "delete"]
        },
        {
            "users": ["mallory"],
            "actions": ["create", "read"]
        }
        ],
        "defaultPolicy": ["read"]
    }
  ```

**Comment by shimish2** [comment]
> Yes, that's what I thought about, in the current implementation we indeed allow charlie to write/read/update/delete on repo1.
> Basically we just make a union with all the policies.
> 
> 1. If user has perm "X" on a repo from all-repositories based policy then allow otherwise 2
> 2. if not 1 then check repository based policy then allow otherwise 3
> 3. if not 2 then check admin based policy then allow otherwise don't allow
> 
> Is this ok?

@rchincha should we do intersection of all policies or current implementation is fine.

**Comment by peusebiu** [comment]
> Another comment is that /v2/ must always be available to folks without authn or authz because rest of the bootstrapping will happen based on that. Is this true? Is there a test case for it?

Yeah, it's covered for authz but not for authn. I had a testcase but I was assuming that clients would give username:password on /v2/ route.
I will update to also cover authn logic.

File: pkg/api/controller_test.go [hunk 1]
```diff
@@ -1666,6 +1666,15 @@ func TestAuthorizationWithBasicAuth(t *testing.T) {
 					},
 					DefaultPolicy: []string{},
 				},
+				"globalpolicy": api.PolicyGroup{
```
rchincha, line 1669: typo "globalpolicy"

pkg/api/authz.go
```diff
@@ -79,12 +79,20 @@ func (ac *AccessController) getReadRepos(username string) []string {
 // can verifies if a user can do action on repository.
 func (ac *AccessController) can(username, action, repository string) bool {
 	can := false
-	// check repo based policy
-	pg, ok := ac.Config.Repositories[repository]
+
+	pg, ok := ac.Config.Repositories["globalpolicy"]
```
rchincha, line 83: Make "globalpolicy" a string constant

rchincha (commented):
let me evaluate whether we should just have "/" section or this "globalpolicy". Also whether these paths should be regexps, and we have a "*' or a "/*"

pkg/api/controller_test.go
peusebiu: it's not, viper is not case sensitive.

peusebiu:
@rchincha  rebased, ready to be reviewed.

examples/README.md
```diff
@@ -186,9 +186,24 @@ identities. An additional per-repository default policy can be specified for
...
+Regexes can also be used as repository paths.
+
+Authorization is granted based on the longest path matched.
+For example repos2/repo repository will match both ".*" and "repos2/repo" keys,
+in such case repos2/repo policy will be used.
+
 ```
 "accessControl": {
-    "repos1/repo": {
+    ".*": {
```
rchincha, line 197: Can we use the same pattern we use for "sync" feature?
\* and ** for subdirs?

rchincha (commented):
Here are the test cases we need to cover:
1) developer test repos - any user can create a repo under a repo prefix
2) production repos - admin users need to first create a repo and then some users can write to this repo, but cannot delete
3) admin users - some admin users can do anything

examples/README.md
peusebiu: Added glob patterns

rchincha (approved):
lgtm

Issue #224, opened 2021-09-17
ldap: rate-limit bind/search else may swamp LDAP server


rchincha:
```
{"level":"error","error":"LDAP Result Code 51 \"Busy\": 00002024: LdapErr: DSID-0C0607F8, comment: No other operations may be performed on the connection while a bind is outstanding., data 0, v2580\u0000"

PR #226, opened 2021-09-18
authn: serialize ldap authn calls

rchincha:
Some LDAP servers are not MT-safe in that when searches happen with binds
in flight leads to errors such as:
"comment: No other operations may be performed on the connection while a
bind is outstanding"

Add goroutine-id in logs to help debug MT bugs.

Fixes issue #224


shimish2 (approved):
lgtm

PR #229, opened 2021-09-23
go.mod: update deps to address dependabot alerts

rchincha:

### DIFF
--- a/Dockerfile-conformance
+++ b/Dockerfile-conformance
@@ -1,7 +1,7 @@
 # ---
 # Stage 1: Install certs, build binary, create default config file
 # ---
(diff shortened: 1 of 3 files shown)


Dockerfile-conformance
```diff
@@ -1,7 +1,7 @@
 # ---
 # Stage 1: Install certs, build binary, create default config file
 # ---
-FROM docker.io/golang:1.13.6-alpine3.11 AS builder
+FROM docker.io/golang:1.17-alpine AS builder
```
shimish2, line 4: should we use 1.16 ?

rchincha: moved to 1.16

shimish2 (approved):

PR #231, opened 2021-09-27
Clean blob uploads when clients interrupts uploading, closes #225

### LINKED ISSUE #225: Bug: Storage repo increase in case an image upload request is interrupted
The temporary folder .uploads created when an image is uploaded is not deleted when the upload request is canceled.
This can result in running out of space when exploited.

peusebiu:


pkg/api/routes.go
```diff
@@ -904,6 +904,12 @@ func (rh *RouteHandler) PatchBlobUpload(w http.ResponseWriter, r *http.Request)
 
 	if err != nil {
 		switch err {
+		case io.ErrUnexpectedEOF:
```
adodon2go, line 907: Is it possible to be interrupted also in UpdateBlobUpload function? (see line 1015)

pkg/storage/storage_test.go
```diff
@@ -73,6 +73,39 @@ func TestAPIs(t *testing.T) {
...
+			So(err, ShouldBeNil)
+			So(b, ShouldEqual, l)
+
+			err = il.CleanBlobUpload("test-clean", v)
+			So(err, ShouldBeNil)
+
+			err = il.FinishBlobUpload("test-clean", v, buf, d.String())
+			So(err, ShouldNotBeNil)
+			So(b, ShouldEqual, l)
+
```
adodon2go, line 103: We can test also that BlobUploadInfo returns error here

pkg/storage/storage.go
```diff
@@ -1215,6 +1215,22 @@ func (is *ImageStore) DeleteBlob(repo string, digest string) error {
 	return nil
 }
 
+func (is *ImageStore) CleanBlobUpload(repo string, uuid string) error {
```
adodon2go, line 1218: Why we don't reuse DeleteBlobUpload ?

peusebiu: oh, I didn't see it...

pkg/api/controller_test.go
```diff
@@ -391,6 +391,94 @@ func TestBasicAuth(t *testing.T) {
...
+		ctx := context.Background()
+		timeout := 1 * time.Second
+		ctx, cancel := context.WithTimeout(ctx, timeout)
+
+		_, _ = resty.R().
+			SetHeader("Content-Length", fmt.Sprintf("%d", len(blob))).
+			SetHeader("Content-Type", "application/octet-stream").
+			SetQueryParam("digest", digest).
+			SetBody(blob).
+			SetContext(ctx).
```
peusebiu, line 452: I can not simulate a client interrupt... it behaves different on different machines.
For example this test passes all the time on my machine.

If you have any suggestion, let me know.

pkg/storage/storage_test.go
```diff
@@ -73,6 +73,39 @@ func TestAPIs(t *testing.T) {
...
+			b, err := il.PutBlobChunk("test-clean", v, 0, int64(l), buf)
+			So(err, ShouldBeNil)
+			So(b, ShouldEqual, l)
+
+			err = il.CleanBlobUpload("test-clean", v)
+			So(err, ShouldBeNil)
+
+			err = il.FinishBlobUpload("test-clean", v, buf, d.String())
+			So(err, ShouldNotBeNil)
+			So(b, ShouldEqual, l)
```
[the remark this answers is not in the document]
peusebiu: Because I need to interrupt exactly while PutBlobChunk function is running, see my last message.

pkg/api/controller_test.go
rchincha: Push a few blobs and go to sleep before it finishes and the other goroutine interrupts it.

pkg/storage/storage_test.go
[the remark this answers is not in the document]
rchincha: Push a few blobs and go to sleep before it finishes and the other goroutine interrupts it.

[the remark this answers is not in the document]
peusebiu: But sleep(1 second) might be sufficient on my machine, and at the same time be too long on a machine with more resources, or too short.
I can't even find a reliable logic for my machine, right now this test passes 90+% times.

I tried with context.WithCancel() like you say, but it's the same as context.WithTimeout().

I think it's impossible to test this or I'm missing something.

You say push a few blobs, but in testing I need to know which blob is interrupted so I can check that it got removed.

```
		go func(ctx context.Context) {
			_, _ = resty.R().
				SetHeader("Content-Length", fmt.Sprintf("%d", len(blob))).
				SetHeader("Content-Type", "application/octet-stream").
				SetQueryParam("digest", digest).
				SetBody(blob).
				SetContext(ctx).
				Patch(baseURL + loc)
		}(ctx)

		time.Sleep(1 * time.Second)
		cancel()

		resp, err = resty.R().Get(baseURL + "/v2/" + AuthorizedNamespace + "/blobs/uploads/" + sessionID)
		So(err, ShouldBeNil)
		So(resp, ShouldNotBeNil)
		So(resp.StatusCode(), ShouldEqual, 404)
```

This is what you proposed, this is the result on my machine:

```
--- PASS: TestClientInterrupts (1.88s)
PASS
coverage: 15.8% of statements
ok      github.com/anuvu/zot/pkg/api    2.234s  coverage: 15.8% of statements
peusebiu@PEUSEBIU-FC130:~/zot$ go test -v -race -coverprofile=coverage.txt -covermode=atomic  -tags=extended -timeout=360s  github.com/anuvu/zot/pkg/api -run TestClientInterrupts
=== RUN   TestClientInterrupts

  Interrupted blob upload {"level":"info","params":{"Version":"1.0.0-dev","Commit":"","BinaryType":"","AccessControl":null,"Storage":{"RootDirectory":"/tmp/oci-repo-test103928246","Dedupe":true,"GC":true,"SubPaths":null},"HTTP":{"Address":"127.0.0.1","Port":"37819","TLS":null,"Auth":null,"RawAccessControl":null,"Realm":"","AllowReadAccess":false,"ReadOnly":false},"Log":{"Level":"debug","Output":"","Audit":""},"Extensions":null},"goroutine":41,"caller":"/home/peusebiu/zot/pkg/api/controller.go:70","time":"2021-10-06T13:17:59.1163505+03:00","message":"configuration settings"}
{"level":"info","module":"http","clientIP":"127.0.0.1:56134","method":"POST","path":"/v2/everyone/isallowed/blobs/uploads/","statusCode":202,"latency":"1.0634ms","bodySize":0,"headers":{"Accept-Encoding":["gzip"],"Content-Length":["0"],"User-Agent":["go-resty/1.12.0 (https://github.com/go-resty/resty)"]},"goroutine":44,"caller":"/home/peusebiu/zot/pkg/log/log.go:153","time":"2021-10-06T13:17:59.622741+03:00","message":"HTTP API"}
✔✔✔{"level":"info","module":"http","clientIP":"127.0.0.1:56134","method":"GET","path":"/v2/everyone/isallowed/blobs/uploads/a4fa9741-693a-4a73-9f2d-5eba92ee4d15","statusCode":204,"latency":"88.3µs","bodySize":0,"headers":{"Accept-Encoding":["gzip"],"User-Agent":["go-resty/1.12.0 (https://github.com/go-resty/resty)"]},"goroutine":44,"caller":"/home/peusebiu/zot/pkg/log/log.go:153","time":"2021-10-06T13:18:00.6285624+03:00","message":"HTTP API"}
✔✔✘


Failures:

  * /home/peusebiu/zot/pkg/api/controller_test.go 
  Line 463:
  Expected: '404'
  Actual:   '204'
  (Should be equal)
  Diff:     '4204'
```

pkg/api/controller_test.go
peusebiu: Ok, I finally found a way to tell when the upload has started, and then interrupt. Thanks!

PR #233, opened 2021-09-29
[cloud usecase] zot cli should not download manifests when listing images

paulsterpu:

PR #234, opened 2021-09-30
Added storage interface

peusebiu:


pkg/storage/storage_test.go
```diff
@@ -147,17 +187,32 @@ func TestAPIs(t *testing.T) {
...
+
+					// same manifest for coverage
 					_, err = il.PutImageManifest("test", "1.0", ispec.MediaTypeImageManifest, mb)
 					So(err, ShouldBeNil)
 
 					_, err = il.PutImageManifest("test", "2.0", ispec.MediaTypeImageManifest, mb)
 					So(err, ShouldBeNil)
 
-					_, err = il.PutImageManifest("test", "3.0", ispec.MediaTypeImageManifest, mb)
+					_, err := il.PutImageManifest("test", "3.0", ispec.MediaTypeImageManifest, mb)
```
andaaron, line 210: Why is the colon needed?

andaaron (commented):
1 question, otherwise code looks good.

pkg/storage/storage_fs.go
```diff
@@ -0,0 +1,1340 @@
...
+	buf, err := ioutil.ReadFile(path.Join(dir, "index.json"))
+
+	if err != nil {
+		is.log.Error().Err(err).Str("dir", dir).Msg("failed to read index.json")
+
+		if os.IsNotExist(err) {
+			return nil, "", "", errors.ErrRepoNotFound
+		}
+
+		return nil, "", "", err
```
shimish2, line 369: try to cover this line by revoking all permission from index.json

pkg/storage/storage_fs.go
```diff
@@ -0,0 +1,1340 @@
...
+	if !found {
+		return nil, "", "", errors.ErrManifestNotFound
+	}
+
+	p := path.Join(dir, "blobs", digest.Algorithm().String(), digest.Encoded())
+
+	buf, err = ioutil.ReadFile(p)
+
+	if err != nil {
+		is.log.Error().Err(err).Str("blob", p).Msg("failed to read manifest")
```
shimish2, line 412: this can be covered by generating a fake image which does not contain manifest blob.

pkg/storage/storage_fs.go
```diff
@@ -0,0 +1,1340 @@
...
+
+		if os.IsNotExist(err) {
+			return nil, "", "", errors.ErrManifestNotFound
+		}
+
+		return nil, "", "", err
```
shimish2, line 418: this can be covered by revoking read permission of particular blob.

pkg/extensions/search/common/oci_layout.go
```diff
@@ -19,21 +20,22 @@ import (
...
-	Log log.Logger
+	Log             log.Logger
+	StoreController storage.StoreController
 }
 
 // NewOciLayoutUtils initializes a new OciLayoutUtils object.
-func NewOciLayoutUtils(log log.Logger) *OciLayoutUtils {
-	return &OciLayoutUtils{Log: log}
+func NewOciLayoutUtils(storeController storage.StoreController, log log.Logger) *OciLayoutUtils {
+	return &OciLayoutUtils{Log: log, StoreController: storeController}
```
shimish2, line 29: Do we need StoreController here ? 

I was thinking to keep oci layout utils independent of store controller and whoever wants to use  utils function they just need to provide correct path.

peusebiu: Yes, I saw your refactor, unfortunately I need to use imageStore.GetIndexContent() and imageStore.GetBlobContent() inside oci_layout.
If you think about another solution, let me know.

The only thing I can think about is passing storeController from resolver.go down to each function.

pkg/storage/storage_fs.go
peusebiu: Thank you very much for all the tips and tricks for testing!

pkg/extensions/search/common/oci_layout.go
paulsterpu: This is the approach I am using in #233 , passing storeController to each function.

pkg/storage/storage_fs.go
```diff
@@ -0,0 +1,1333 @@
...
+		}
+
+		return imgStore
+	}
+
+	return sc.DefaultStore
+}
+
+// NewImageStoreFS returns a new image store backed by a file storage.
+func NewImageStoreFS(rootDir string, gc bool, dedupe bool, log zlog.Logger) ImageStore {
```
rchincha, line 105: Why not just call this NewImageStore()? Understood, ImageStore is now an interface

andaaron: Because there's going to be a similar method for S3, and it should be clear which is for the currently supported FS and which is for S3.

rchincha: zot.NewImageStore() and s3.NewImageStore() should be two different things, right?

PR #236, opened 2021-10-05
Implement "scrub" to check manifest/blob integrity

Andreea-Lupu:
*  Adding "scrub" cli command and using umoci codebase as a library ( for "umoci stat" )

*  For the moment this command can be used only if the server is not running

*  Example of output for cli command:

```
$ bin/zot scrub examples/config-multiple.json  
IMAGE NAME                        TAG                       STATUS    ERROR         
postgres                          latest                    affected  stat: discard trailing "application/vnd.oci.image.config.v1+json" blob: expected sha256:c0616bde68243ffda5e0a535502d6173da5644562fd5560059a369ea6563793f not sha256:2dc11ae9a808dedf5a1cd8431231f6b0a5e47dbcd61e527a975a474421ddc0d8: verified reader digest mismatch  
a/busybox                         latest                    affected  blob: not found  
busybox                           latest                    ok                  
zot1/golang                       alpine                    ok                  
zot2/postgres                     latest                    ok          
 ```

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -69,7 +69,6 @@ func DefaultHeaders() mux.MiddlewareFunc {
 	}
 }
 
-// nolint: gocyclo
 func (c *Controller) Run() error {
 	// validate configuration
 	if err := c.Config.Validate(c.Log); err != nil {
@@ -102,6 +101,62 @@ func (c *Controller) Run() error {
 	}
 
 	c.Metrics = monitoring.NewMetricsServer(enabled, c.Log)
+
+	if err := c.InitImageStore(); err != nil {
+		return err
+	}
+
+	monitoring.SetServerInfo(c.Metrics, c.Config.Commit, c.Config.BinaryType, c.Config.GoVersion, c.Config.Version)
+	_ = NewRouteHandler(c)
+
+	addr := fmt.Sprintf("%s:%s", c.Config.HTTP.Address, c.Config.HTTP.Port)
+	server := &http.Server{
+		Addr:        addr,
+		Handler:     c.Router,
+		IdleTimeout: idleTimeout,
+	}
+	c.Server = server
+
+	// Create the listener
+	l, err := net.Listen("tcp", addr)
+	if err != nil {
+		return err
+	}
+
+	if c.Config.HTTP.TLS != nil && c.Config.HTTP.TLS.Key != "" && c.Config.HTTP.TLS.Cert != "" {
+		if c.Config.HTTP.TLS.CACert != "" {
+			clientAuth := tls.VerifyClientCertIfGiven
+			if (c.Config.HTTP.Auth == nil || c.Config.HTTP.Auth.HTPasswd.Path == "") && !c.Config.HTTP.AllowReadAccess {
+				clientAuth = tls.RequireAndVerifyClientCert
+			}
+
+			caCert, err := ioutil.ReadFile(c.Config.HTTP.TLS.CACert)
+			if err != nil {
+				panic(err)
+			}
+
+			caCertPool := x509.NewCertPool()
+
+			if !caCertPool.AppendCertsFromPEM(caCert) {
+				panic(errors.ErrBadCACert)
+			}
+
+			server.TLSConfig = &tls.Config{
+				ClientAuth:               clientAuth,
+				ClientCAs:                caCertPool,
+				PreferServerCipherSuites: true,
+				MinVersion:               tls.VersionTLS12,
+			}
+			server.TLSConfig.BuildNameToCertificate() // nolint: staticcheck
+		}
+
+		return server.ServeTLS(l, c.Config.HTTP.TLS.Cert, c.Config.HTTP.TLS.Key)
+	}
+
+	return server.Serve(l)
+}
+
+func (c *Controller) InitImageStore() error {
 	c.StoreController = storage.StoreController{}
 
 	if c.Config.Storage.RootDirectory != "" {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by peusebiu** [comment]
Should this command be run when the server is down?
If yes we should add a check and a warning if it's not down.

Otherwise we should use locks.

File: pkg/cli/root.go [hunk 1]
```diff
@@ -79,6 +82,46 @@ func NewRootCmd() *cobra.Command {
...
+
+			defer func(ln net.Listener) {
+				err = ln.Close()
+				if err != nil {
+					panic(err)
+				}
+			}(ln)
+
+			if err != nil {
+				fmt.Println("Server is running")
```
shimish2, line 108: Can we use Info log here instead of println ?

pkg/storage/scrub.go
```diff
@@ -0,0 +1,395 @@
...
+func (sc StoreController) CheckAllBlobsIntegrity(resultWriter io.Writer) error {
+	imageErr := make(chan stringResult)
+	ctx, cancel := context.WithCancel(context.Background())
+
+	var wg sync.WaitGroup
+
+	imagesList := []ImageStore{}
+	imagesList = append(imagesList, sc.DefaultStore)
+
+	for _, subStore := range sc.SubStore {
```
rchincha, line 172: This design may just be overkill. 
Why not just iterate over the repos in a simple tight loop as first implementation. I don't think you need anything more than that. When you make it online you can move that to a single goroutine and pls pay attention to the locks. Let's not optimize prematurely.

rchincha:
@Andreea-Lupu, can you pls rebase and re-push

rchincha:
Pls rebase this and resubmit

pkg/storage/scrub.go
```diff
@@ -0,0 +1,244 @@
...
+		imagesList = append(imagesList, subStore)
+	}
+
+	results := ScrubResults{}
+
+	for _, is := range imagesList {
```
rchincha, line 53: Lines 47-53 can be combined into a single loop?

pkg/storage/scrub.go
```diff
@@ -0,0 +1,244 @@
...
+	} else {
+		// check layers
+		for _, s := range stat.History {
+			layer := s.Layer
+			if layer == nil {
+				continue
+			}
+
+			// check if layer exists
+			bytes, err := ioutil.ReadFile(path.Join(dir, "blobs/sha256", layer.Digest.Hex()))
```
rchincha, line 136: You can use either layer.Digest.String() or layer.Digest.Encoded() depending on what you want to do here? Don't hardcode "sha256"

rchincha: What if the layer is 100GB? you will exhaust all memory. 
Use this pattern instead:
https://github.com/anuvu/zot/blob/main/pkg/storage/storage_fs.go#L870

pkg/storage/scrub.go
```diff
@@ -0,0 +1,244 @@
...
+			if err != nil {
+				imageRes = getResult(imageName, tagName, errors.ErrBlobNotFound)
+				break
+			}
+
+			// check layer content
+			expectedHash := fmt.Sprintf("%x", sha256.Sum256(bytes))
```
rchincha, line 143: Should not refer to any specific hash algorithm (sha256) here in this code directly.

rchincha (commented):
Pls take a look at the following comments.

pkg/api/controller.go
```diff
@@ -83,6 +83,60 @@ func (c *Controller) Run() error {
...
+		Addr:        addr,
+		Handler:     c.Router,
+		IdleTimeout: idleTimeout,
+	}
+	c.Server = server
+
+	// Create the listener
+	l, err := net.Listen("tcp", addr)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
Andreea-Lupu: Thanks for suggestion, I added a test for it.

pkg/cli/root.go
Andreea-Lupu: Done

pkg/storage/scrub.go
Andreea-Lupu: Ok. I changed the implementation and now it just iterates over the repos in a loop.

pkg/storage/scrub.go
Andreea-Lupu: I changed the code and now it's using the pattern you suggested. Thanks!:)

pkg/storage/scrub.go
Andreea-Lupu: Because I'm using the above pattern suggested by you for checking layers, I don't need anymore this part of code so I removed it.

pkg/storage/scrub.go
Andreea-Lupu: In order to combine this in a single loop this is what I did: First of all, I need to iterate through all repos from controller, this means the DefaultStore which has type ImageStore and also all values from SubStore which is a map[string]ImageStore. So to iterate through all of this in a single loop I created a map[string]ImageStore using SubStore's elements if SubStore is not nil, and after this I added to it a new entry with key  "" and value DefaultSore.

pkg/storage/scrub.go
```diff
@@ -0,0 +1,395 @@
...
+
+	"github.com/anuvu/zot/errors"
+	"github.com/olekukonko/tablewriter"
+	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/opencontainers/umoci"
+	"github.com/opencontainers/umoci/oci/casext"
+)
+
+const (
+	imageNameWidth = 32
```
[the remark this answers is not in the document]
Andreea-Lupu: I made some changes in my code and now only 4 constants from here can be found also in pkg/cli, so I don't know if it's really needed a new package only for these. I prefer to keep them here for the moment, but I will take in consideration this suggestion when I will work for the online part.

rchincha:
If config-file is not specified, print help/usage instead of panic'ing like this.

$ bin/zot scrub

{"level":"info","time":"2021-12-09T17:06:05Z","message":"The server is running, in order to perform the scrub command the server should be shut down"}                        panic: Error: server is running                                                                                                                                                                                                                                                                                                                             goroutine 1 [running]:                                                                                                                                                        zotregistry.io/zot/pkg/cli.NewRootCmd.func2(0xc00070ec80, {0x3de91a0, 0x0, 0x0})                                                                                                      zotregistry.io/zot/pkg/cli/root.go:105 +0x309

rchincha (approved):
lgtm

PR #271, opened 2021-11-05
stress test suite

roxanaN:
Stress test for pushing images

### DIFF
--- a/pkg/perf/images/create_images.go
+++ b/pkg/perf/images/create_images.go
@@ -1,0 +1,199 @@
+package images
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path"
+
+	godigest "github.com/opencontainers/go-digest"
+	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+func CreateImages(parallelImagesNumber int) error {
+	pushedImages := GetImagesName()
+	for i := 0; i < len(pushedImages); i++ {
+		imageDir := fmt.Sprintf(imageDirPath, pushedImages[i])
+		if _, err := os.Stat(imageDir); os.IsNotExist(err) {
+			if err := MakeImageDir(imageDir); err != nil {
+				return err
+			}
+
+			if err := CreateLayout(imageDir); err != nil {
+				log.Fatalln(err)
+			}
+
+			manifestName, manifestSize := CreateBlobs(size, imageDir)
+			CreateIndex(imageDir, manifestName, manifestSize)
+		}
+	}
+
+	for i := 1; i <= parallelImagesNumber; i++ {
+		imageDir := fmt.Sprintf(imageDirPath, fmt.Sprintf(parallelImagesName, i))
+		if _, err := os.Stat(imageDir); os.IsNotExist(err) {
+			if err := MakeImageDir(imageDir); err != nil {
+				log.Fatalln(err)
+			}
+
+			if err := CreateLayout(imageDir); err != nil {
+				log.Fatalln(err)
+			}
+
+			manifestName, manifestSize := CreateBlobs(i*size, imageDir)
+			CreateIndex(imageDir, manifestName, manifestSize)
+		}
+	}
+
+	return nil
+}
+
+func DeleteImages(parallelImagesNumber int) error {
+	pushedImages := GetImagesName()
+	for i := 0; i < len(pushedImages); i++ {
+		if err := os.RemoveAll(fmt.Sprintf(imageDirPath, pushedImages[i])); err != nil {
+			return err
+		}
+	}
+
+	for i := 1; i <= parallelImagesNumber; i++ {
+		if err := os.RemoveAll(fmt.Sprintf(imageDirPath, fmt.Sprintf(parallelImagesName, i))); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func CreateLayout(imageName string) error {
+	data := ispec.ImageLayout{
+		Version: ispec.ImageLayoutVersion,
+	}
+
+	file, err := json.MarshalIndent(data, "", " ")
+	if err != nil {
+		return err
+	}
+
+	err = ioutil.WriteFile(fmt.Sprintf("%s/%s", imageName, ispec.ImageLayoutFile), file, 0600)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func CreateBlobs(size int, imageName string) (string, int64) {
+	// create blob
+	filePath := fmt.Sprintf(pathFormat, imageName, filename)
+	bigBuff := make([]byte, size)
+
+	err := ioutil.WriteFile(filePath, bigBuff, 0600)
+	if err != nil {
+		log.Fatalln("Error creating file", err)
+	}
+
+	fileName, filePath, _ := GetFileAttribute(filename, filePath, imageName)
+
+	// compress blob
+	layerPath := fmt.Sprintf(pathFormat, imageName, tarName)
+	filesList := []string{filePath}
+
+	layer, err := os.Create(layerPath)
+	if err != nil {
+		log.Fatalln("Error writing archive", err)
+	}
+
+	err = CreateArchive(filesList, layer)
+	if err != nil {
+		log.Fatalln("Error creating archive", err)
+	}
+
+	layerName, _, layerSize :=
+		GetFileAttribute(tarName, layerPath, imageName)
+
+	// create config
+	RootFS := ispec.RootFS{
+		Type: "layers",
+		DiffIDs: []godigest.Digest{
+			godigest.Digest(fmt.Sprintf(digestFormat, fileName)),
+		},
+	}
+
+	config := ispec.Image{
+		Architecture: "amd64",
+		OS:           "linux",
+		RootFS:       RootFS,
+	}
+
+	configFile, _ := json.MarshalIndent(config, "", " ")
+	configPath := fmt.Sprintf(pathFormat, imageName, configFileName)
+
+	err = ioutil.WriteFile(configPath, configFile, 0600)
+	if err != nil {
+		log.Fatalln("Error creating config file", err)
+	}
+
+	configName, _, configSize := GetFileAttribute(configFileName, configPath, imageName)
+
+	// create manifest
+	annotationsMap := make(map[string]string)
+	annotationsMap[ispec.AnnotationRefName] = "1.0"
+	manifestConfig := ispec.Descriptor{
+		MediaType:   "application/vnd.oci.image.config.v1+json",
+		Digest:      godigest.Digest(fmt.Sprintf(digestFormat, configName)),
+		Size:        configSize,
+		Annotations: annotationsMap,
+	}
+
+	layersConfig := ispec.Descriptor{
+		MediaType: "application/vnd.oci.image.layer.v1.tar+gzip",
+		Digest:    godigest.Digest(fmt.Sprintf(digestFormat, layerName)),
+		Size:      layerSize,
+	}
+
+	manifest := ispec.Manifest{
+		Config: manifestConfig,
+		Layers: []ispec.Descriptor{
+			layersConfig,
+		},
+	}
+	manifest.SchemaVersion = 2
+
+	manifestFile, _ := json.MarshalIndent(manifest, "", " ")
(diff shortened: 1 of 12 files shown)


pkg/performance/images/structs.go
```diff
@@ -0,0 +1,53 @@
...
+	digestFormat     = "sha256:%s"
+	layoutFormat     = "%s/oci-layout"
+	filename         = "file"
+	tarName          = "file.tar.gz"
+	configFileName   = "configFile"
+	manifestFileName = "manifestFile"
+)
+
+var imageNameList = []string{
+	"zot-tests-single-images-dummy",
```
rchincha, line 14: pls add comments on what these workloads are.

rchincha (commented):
@roxanaN, is your PR in a shape to actually try out?

We should test with https://github.com/anuvu/zot/pull/243 enabled.

Also can you rename "performance" as "perf", too long.

pkg/perf/perf_suite_test.go
```diff
@@ -0,0 +1,154 @@
...
+				pushedImages = append(pushedImages, imageName)
+				arguments := setCopyCommand(op)
+
+				arguments = setCommandVariables(op, arguments, imageName, imageName, true)
+				commands = append(commands, strings.Join(arguments, " "))
+			}
+
+			Expect(runCommands(commands)).To(Equal(true))
+
+			for _, imageName := range pushedImages {
```
peusebiu, line 113: I see here you are removing images, but not in the rest of tests.

Maybe we should remove all of them after the stress test?

pkg/perf/images/helper.go
```diff
@@ -0,0 +1,64 @@
...
+	"os"
+	"os/exec"
+	"strings"
+)
+
+func getDigest(str string) string {
+	return strings.Fields(str)[0]
+}
+
+func encryptFile(path string) string {
```
peusebiu, line 15: getFileDigest

pkg/perf/perf_suite_test.go
roxanaN: This is my intention, to remove the images at the end of the tests. Apparently I missed this on one of the pushing tests.

rchincha:
@roxanaN can you post an example of the commandline on how to run/drive this stress test? I saw a 'ginkgo' driver earlier, not sure if that is what is still applicable.

rchincha:
@roxanaN pls rebase your PR

rchincha:
` zot config -l`
^ this will take care of authentication params etc

`zot stress aci-zot`
^ use the name in config above to drive the test.

andaaron:
@rchincha I thought the requirement is for these tests to work with any registry which is compliant with the oci dist spec. I doesn't  seem ok to add this functionality in the zot client.
That besides not being an actual zot functionality.

rchincha:
> 
> 
> @rchincha I thought the requirement is for these tests to work with any registry which is compliant with the oci dist spec. I doesn't seem ok to add this functionality in the zot client. That besides not being an actual zot functionality.

Question then is, should this stress tool get a different repo? Everything in zot world is dist-spec.

andaaron:
> > @rchincha I thought the requirement is for these tests to work with any registry which is compliant with the oci dist spec. I doesn't seem ok to add this functionality in the zot client. That besides not being an actual zot functionality.
> 
> Question then is, should this stress tool get a different repo? Everything in zot world is dist-spec.

I'm just saying I don't see a good reason to couple the zot client code with the test tooling from the point of view of their responsibilities. The test tooling should work with or without the zot client. And the zot client doesn't need the performance test tooling to be included.

Is having the test code already compiled a good enough reason? What other benefits are there besides this?

rchincha:
> 
> 
> > > @rchincha I thought the requirement is for these tests to work with any registry which is compliant with the oci dist spec. I doesn't seem ok to add this functionality in the zot client. That besides not being an actual zot functionality.
> > 
> > 
> > Question then is, should this stress tool get a different repo? Everything in zot world is dist-spec.
> 
> I'm just saying I don't see a good reason to couple the zot client code with the test tooling from the point of view of their responsibilities. The test tooling should work with or without the zot client. And the zot client doesn't need the performance test tooling to be included.
> 
> Is having the test code already compiled a good enough reason? What other benefits are there besides this?

From a user pov, how would one go about starting the test? git clone, build and run?

andaaron:
I was assuming the user 

> > > > @rchincha I thought the requirement is for these tests to work with any registry which is compliant with the oci dist spec. I doesn't seem ok to add this functionality in the zot client. That besides not being an actual zot functionality.
> > > 
> > > 
> > > Question then is, should this stress tool get a different repo? Everything in zot world is dist-spec.
> > 
> > 
> > I'm just saying I don't see a good reason to couple the zot client code with the test tooling from the point of view of their responsibilities. The test tooling should work with or without the zot client. And the zot client doesn't need the performance test tooling to be included.
> > Is having the test code already compiled a good enough reason? What other benefits are there besides this?
> 
> From a user pov, how would one go about starting the test? git clone, build and run?

I was assuming the user is a zot developer wanting to test a specific zot setup or collecting data to compare with a different registry. Are you thinking about a different user?

rchincha:
> > From a user pov, how would one go about starting the test? git clone, build and run?
> 
> I was assuming the user is a zot developer wanting to test a specific zot setup or collecting data to compare with a different registry. Are you thinking about a different user?

Possibly. Here are the use cases for this:
1. ci/cd, so that we run stress tests against releases/commits so we don't regress - this can be part of zot's own ci/cd
2. devs, so they validate any performance improvements they attempt
3. troubleshooting, perf tests in deployments to understand bottlenecks
4. benchmarking, perf tests against many different registry projects to understand where we stand

pkg/perf/images/create_images.go
```diff
@@ -0,0 +1,142 @@
...
+			log.Fatal(err)
+		}
+
+		createLayout(imageName)
+		manifestName, manifestSize := createBlobs(i*size, imageName)
+		createIndex(imageName, manifestName, manifestSize)
+	}
+}
+
+func createLayout(imageName string) {
```
andaaron, line 24: Per our offline discussion, please try to use the github.com/opencontainers/image-spec/specs-go/v1 and the github.com/opencontainers/umoci libraries. A lot of code in this PR is duplicating that functionality.

andaaron:
> > > From a user pov, how would one go about starting the test? git clone, build and run?
> > 
> > 
> > I was assuming the user is a zot developer wanting to test a specific zot setup or collecting data to compare with a different registry. Are you thinking about a different user?
> 
> Possibly. Here are the use cases for this:
> 
> 1. ci/cd, so that we run stress tests against releases/commits so we don't regress - this can be part of zot's own ci/cd
> 2. devs, so they validate any performance improvements they attempt
> 3. troubleshooting, perf tests in deployments to understand bottlenecks
> 4. benchmarking, perf tests against many different registry projects to understand where we stand

I guess it makes sense if non-developers like a QA team, a support team, or someone evaluating zot want to make comparisons.

rchincha:
Better not to use containers/image copy because it verifies blobs using sha256 before pushing which is unlikely to saturate CPU

rchincha:
Whatever images you plan to upload, you must pre-compute the SHA256 during test initialization phase and then run tests. We want zero latency from client side.

rchincha:
The problem I am seeing when using containers/image is that we do a streamed blob upload which needs to compute sha256 inline which adds significant delay.

rchincha:
@roxanaN https://github.com/project-zot/zot/pull/331
^ pls take a look at this on how this test should shape up.

pkg/perf/perf_suite_test.go
```diff
@@ -0,0 +1,203 @@
...
+	. "zotregistry.io/zot/pkg/perf"
+	. "zotregistry.io/zot/pkg/perf/images"
+)
+
+// nolint: gochecknoglobals
+var (
+	username                    string
+	password                    string
+	address                     string
+	repo                        string
```
rchincha, line 26: This is a good idea for allowing folks to specify, so we can actually test on a live server if need be.

pkg/perf/perf_suite_test.go
```diff
@@ -0,0 +1,203 @@
...
+		imageName := "zot-tests-dummy-push"
+
+		runtime := b.Time("runtime", func() {
+			Expect(PushImage(path.Join("images", imageName), username, password, address, repo, imageName)).To(BeNil())
+		})
+		Expect(runtime.Seconds()).To(BeNumerically("<", singleImagePushTime),
+			"Push oprations shouldn't take too long!")
+	}, times)
+
+	Measure("Performance measuring - parallel push", func(b Benchmarker) {
```
rchincha, line 126: No need to keep saying "Performance measuring" everywhere.
Just say "Parallel Monolithic Push" ... etc

rchincha (commented):
Noticed, "top" reports like so:
1060863 rchincha  20   0   11.0g   6.1g   7384 R 100.0   7.8   3:56.54 perf.test   

What is it really doing? with so much mem consumption too and stays stuck for a while.

Would be nice to have feedback.

Here is my server.json, rest is untouched.

```
$ cat server_config.json                                             {                                                                                                                                                                                 "address": "localhost:5000",                                                                                                                                                  "username": "",                                                                                                                                                               "password": "",                                                                                                                                                               "repo": "perf"                                                                                                                                                            }
```

pkg/perf/images/create_images.go
```diff
@@ -0,0 +1,199 @@
...
+func CreateImages(parallelImagesNumber int) error {
+	pushedImages := GetImagesName()
+	for i := 0; i < len(pushedImages); i++ {
+		imageDir := fmt.Sprintf(imageDirPath, pushedImages[i])
+		if _, err := os.Stat(imageDir); os.IsNotExist(err) {
+			if err := MakeImageDir(imageDir); err != nil {
+				return err
+			}
+
+			if err := CreateLayout(imageDir); err != nil {
```
rchincha, line 24: No need to create the full layout. Just create blobs of various sizes. Right now, if i understand your code correctly, you are only creating one size. You just need a small range of blob sizes which you can mix and match when uploading.

pkg/perf/images/create_images.go
```diff
@@ -0,0 +1,199 @@
...
+				log.Fatalln(err)
+			}
+
+			manifestName, manifestSize := CreateBlobs(size, imageDir)
```
rchincha, line 28: just need this, also "imageDir" must be a variable, so we can pick the fastest storage or even in some cases ramfs.

pkg/perf/images/create_images.go
rchincha: Oh wait! You are doing this so that you don't have to compute the shasums, sure, your layout possibilities are potentially very large. Would rather have a small set of blobs (so disk storage is minimal) and then precompute/store shasums and then create tests and combine on the fly.

pkg/perf/utils.go
```diff
@@ -0,0 +1,98 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"path"
+
+	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"gopkg.in/resty.v1"
+)
+
+func PushImage(imagePath string, username string, passphrase string,
```
rchincha, line 13: There are many types of image push, you should have tests for all of them.

https://github.com/opencontainers/distribution-spec/blob/main/spec.md#push

rchincha (commented):
Your test suite is assuming https, but we may want to run http also.

Also, we need percentiles of completion:
https://cloud.google.com/monitoring/api/v3/distribution-metrics

This is not enough.


```
  Ran 10 samples:                                                                                                                                                               runtime:                                                                                                                                                                        Fastest Time: 0.019s                                                                                                                                                          Slowest Time: 0.024s                                                                                                                                                          Average Time: 0.021s ± 0.002s      
  ```


Quick way to run your own private zot is like so:
$ docker run -p 5000:5000 ghcr.io/project-zot/zot:latest

rchincha:
You can also change the Makefile to create a "zot-perf" target/binary.

.PHONY: perf
perf: <deps>
   go test -c -gcflags "-N -l" -v -o bin/zot-stress ./pkg/perf/...

pkg/perf/perf_suite_test.go
rchincha: ginkgo also does a Before..() and After...() which may be cleaner to read.

Global Setup and Teardown: BeforeSuite and AfterSuite
https://onsi.github.io/ginkgo/

PR #277, opened 2021-11-15
go.mod: upgrade module deps

rchincha:


rchincha:
@peusebiu moved containers/image to upstream. We need to pass in flags to make squashfs work.

tych0:
You mention needing to pass flags, which patch actually passes those flags?

rchincha:
> 
> 
> You mention needing to pass flags, which patch actually passes those flags?

@peusebiu needs to fix his part of zot code to do that. So would ask to take ownership of this PR for that.

shimish2 (approved):
lgtm

PR #285, opened 2021-11-18
sync: skip tls verify on /v2/_catalog when a registry is configured w…

peusebiu:
…ith tls-verify false

sync: cleanup the orphaned private download dir on failure, closes 282

### DIFF
--- a/pkg/extensions/sync/sync.go
+++ b/pkg/extensions/sync/sync.go
@@ -106,6 +106,11 @@ func getUpstreamCatalog(regCfg *RegistryConfig, credentials Credentials, log log
 		client.SetCertificates(cert)
 	}
 
+	// nolint: gosec
+	if regCfg.TLSVerify != nil && !*regCfg.TLSVerify {
+		client.SetTLSClientConfig(&tls.Config{InsecureSkipVerify: true})
+	}
+
 	if credentials.Username != "" && credentials.Password != "" {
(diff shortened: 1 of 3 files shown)


rchincha (approved):
lgtm

pkg/extensions/sync/sync.go
```diff
@@ -377,6 +382,8 @@ func syncRegistry(regCfg RegistryConfig, storeController storage.StoreController
 			return err
 		}
 
+		defer os.RemoveAll(path.Join(imageStore.RootDir(), imageName, SyncBlobUploadDir, uuid))
```
adodon2go, line 385: Any particular reason we are suing a new var on line #372 (why not assigning directly on line #371 the upstreamRef if we are not using ref at all inside the for block)

adodon2go (approved):

pkg/extensions/sync/sync.go
peusebiu: Yes, because I'm using the variable in function literal which I'm passing to retry.RetryIfNecessary(). You can't use range scoped vars in function literals. 
`
func() error {
			_, err = copy.Image(context.Background(), policyCtx, localRef, upstreamRef, &options)
			return err
		}
`

peusebiu:
@rchincha  rebased, it can be merged

rchincha (approved):
lgtm

PR #296, opened 2021-11-24
Added S3 blobs dedupe functionality

peusebiu:
- now blobs are just files which contains a path to the real blob under rootDir/BLOBS/
- linked_blobs.json contains links between real blobs and the blobs which points to them.
eg: 
`{"links":{"sha256:801bfaa63ef2094d770c809815b9e2b9c1194728e5e754ef7bc764030e140cea":["/zot/alpine/blobs/sha256/801bfaa63ef2094d770c809815b9e2b9c1194728e5e754ef7bc764030e140cea"]}`

I'm using linked_blobs.json to check if I can safely delete real blobs.

Can not safely switch dedupe config value if zot already run with a given rootDir because of the entire different logic of dedupe vs non dedupe. I should add a check.

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -101,13 +101,15 @@ func (c *Controller) Run() error {
 	c.StoreController = storage.StoreController{}
 
 	if c.Config.Storage.RootDirectory != "" {
-		if c.Config.Storage.Dedupe {
-			err := storage.ValidateHardLink(c.Config.Storage.RootDirectory)
-			if err != nil {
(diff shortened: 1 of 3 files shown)


rchincha:
Since we have a boltdb, can we get away with a different table/bucket for backwards-compatibility

rchincha:
Dedupe is currently global right? So we do a lookup(sha256) -> full path relative to the rootdir. What makes s3 different?

rchincha:
This can be closed right?
We now have https://github.com/project-zot/zot/pull/504?

peusebiu:
closed because it was reimplemented in https://github.com/project-zot/zot/pull/504

PR #298, opened 2021-11-25
Sync prefix can be an exact match or a glob pattern, closes #297

### LINKED ISSUE #297: sync: prefix can be an exact match or a glob
prefix: /abc should download only /abc repo
prefix: /abc* should download all repos starting with and including /abc, eg,  [/abc, /abc1, /abc2 ...]

peusebiu:


pkg/extensions/sync/sync_internal_test.go
```diff
@@ -168,6 +168,37 @@ func TestSyncInternal(t *testing.T) {
...
+				Prefix: "repo",
+			},
+		}
+		filteredRepos, err := filterRepos(repos, contents)
+		So(err, ShouldBeNil)
+		So(filteredRepos[0], ShouldResemble, []string{"repo", "repo1", "repo2"})
+		So(filteredRepos[1], ShouldResemble, []string{"repo/repo2", "repo/repo2/repo3/repo4"})
+
+		// last prefix shouldn't match, because it was already matched on the first prefix (repo*)
+		So(len(filteredRepos), ShouldEqual, 2)
```
adodon2go, line 190: Maybe it makes sense to reorder prefixes and add new tests.
If I get it right, if we interchange lines 175 with 181 , than the length(filteredRepos) should be 3, right?

pkg/extensions/sync/sync.go
```diff
@@ -338,7 +338,11 @@ func syncRegistry(regCfg RegistryConfig, storeController storage.StoreController
 
 	log.Info().Msg("filtering repos based on sync prefixes")
 
-	repos := filterRepos(catalog.Repositories, regCfg.Content)
+	repos, err := filterRepos(catalog.Repositories, regCfg.Content)
+	if err != nil {
+		log.Error().Err(err).Msg("error while filtering repos based on prefixes")
+		return err
```
adodon2go, line 344: If we return this error here that will stop the sync on a Registry where the error occurs. Is this the desired behavior?
Maybe it is better to just log the error and continue. 
For me is something similar to stopping a server only because we have a bug somewhere in the client library.

peusebiu: Hmm, I can do that, but currently the server won't start if there is any error in sync config. so this shouldn't ever happen.

pkg/extensions/sync/sync_internal_test.go
peusebiu: Yes, that's right.

examples/README.md
```diff
@@ -362,14 +362,17 @@ Configure each registry sync:
...
-						"prefix":"/repo1/repo",         # pull all images under /repo1/repo
+						"prefix":"/repo1/repo",         # pull image repo1/repo
 						"tags":{                        # filter by tags
 							"regex":"4.*",                # filter tags by regex
 							"semver":true                 # filter tags by semver compliance
 						}
 					},
 					{
-						"prefix":"/repo2/repo"          # pull all images under /repo2/repo
+						"prefix":"/repo2/repo*"         # pull all images that matches repo2/repo.*
```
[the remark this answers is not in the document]
peusebiu: It's just glob pattern matching, I guess anyone can google that :D.

pkg/cli/root.go
```diff
@@ -188,6 +189,21 @@ func LoadConfiguration(config *config.Config, configPath string) {
...
 
+	// check glob patterns in sync are compilable
+	if config.Extensions != nil && config.Extensions.Sync != nil {
+		for _, regCfg := range config.Extensions.Sync.Registries {
+			if regCfg.Content != nil {
+				for _, content := range regCfg.Content {
+					ok := glob.ValidatePattern(content.Prefix)
+					if !ok {
+						log.Error().Err(glob.ErrBadPattern).Str("pattern", content.Prefix).Msg("pattern could not be compiled")
+						panic(errors.ErrBadConfig)
```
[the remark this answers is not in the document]
peusebiu: it's equivalent with /repo1/* it will match any image under /repo1/* but not recursively.

examples/README.md
```diff
@@ -396,3 +399,4 @@ Configure each registry sync:
 		}
 ```
 
+Prefixes can be strings that exactly matches repositories or they can be glob patterns.
```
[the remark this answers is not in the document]
peusebiu: Thank you! I will update. nice finding.

pkg/extensions/sync/sync.go
peusebiu: Ok, fixed that, thank you!

examples/README.md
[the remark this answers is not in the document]
peusebiu: Fixed, I will push an update.

pkg/extensions/extensions.go
```diff
@@ -72,10 +72,12 @@ func EnableSyncExtension(config *config.Config, log log.Logger, storeController
 	if config.Extensions.Sync != nil {
 		defaultPollInterval, _ := time.ParseDuration("1h")
 		for id, registryCfg := range config.Extensions.Sync.Registries {
-			if registryCfg.PollInterval < defaultPollInterval {
+			if registryCfg.Content != nil && registryCfg.PollInterval < defaultPollInterval {
```
adodon2go, line 75: I think we also need to add in this condition :
`len(registryCfg.Content)>0`

adodon2go (approved):
LGTM

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/apex/log v1.9.0
 	github.com/aquasecurity/trivy v0.0.0-00010101000000-000000000000
 	github.com/aquasecurity/trivy-db v0.0.0-20210916043317-726b7b72a47b
+	github.com/bmatcuk/doublestar/v4 v4.0.2
```
rchincha, line 13: Was this package needed for this functionality?
The downside of pulling in packages is we don't know what else they bring in.

rchincha (approved):
lgtm

PR #304, opened 2021-12-02
Fix race detected in sync test, closes #302

### LINKED ISSUE #302: test: race detected in sync tests
https://github.com/anuvu/zot/runs/4390432603?check_suite_focus=true

```
/sync/sync.go:483","time":"2021-12-02T03:52:37.182564946Z","caller":"github.com/anuvu/zot/pkg/extensions/sync/sync.go:483","time":"2021-12-02T03:52:37.182578047Z","message":"sync exited with error, stopping it..."}
==================
WARNING: DATA RACE
Write at 0x000004cb2490 by goroutine 66:
  github.com/anuvu/zot/pkg/log.NewLogger()
      github.com/anuvu/zot/pkg/log/log.go:23 +0x9a
  github.com/anuvu/zot/pkg/api.NewController()
  ```

### PR DESCRIPTION
**Pull request by peusebiu**

Because after a server.Shutdown sync goroutine could be still running, let's use waitgroup to wait for a sync to either return an error or to finish 1 run, otherwise a data race on zerolog could happen.

TODO: add comments
run build multiple times

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -1,12 +1,14 @@
 package api
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
 	"io/ioutil"
 	"net"
 	"net/http"
+	goSync "sync"
 	"time"
 
 	"github.com/gorilla/handlers"
@@ -34,6 +36,7 @@ type Controller struct {
 	Audit           *log.Logger
 	Server          *http.Server
 	Metrics         monitoring.MetricServer
+	wgShutDown      *goSync.WaitGroup // use it to gracefully shutdown goroutines
 }
 
 func NewController(config *config.Config) *Controller {
@@ -43,6 +46,7 @@ func NewController(config *config.Config) *Controller {
 
 	controller.Config = config
 	controller.Log = logger
+	controller.wgShutDown = new(goSync.WaitGroup)
 
 	if config.Log.Audit != "" {
 		audit := log.NewAuditLogger(config.Log.Level, config.Log.Audit)
@@ -195,7 +199,7 @@ func (c *Controller) Run() error {
 
 	// Enable extensions if extension config is provided
 	if c.Config.Extensions != nil && c.Config.Extensions.Sync != nil {
-		ext.EnableSyncExtension(c.Config, c.Log, c.StoreController)
+		ext.EnableSyncExtension(c.Config, c.wgShutDown, c.StoreController, c.Log)
 	}
 
 	monitoring.SetServerInfo(c.Metrics, c.Config.Commit, c.Config.BinaryType, c.Config.GoVersion, c.Config.Version)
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Review by rchincha — COMMENTED** [review]
@peusebiu instead of waiting for sync to finish, can we not cancel on the context.Context that we can pass all the way? Because sync can be stuck for a long long time.

**Comment by peusebiu** [comment]
> @peusebiu instead of waiting for sync to finish, can we not cancel on the context.Context that we can pass all the way? Because sync can be stuck for a long long time.

I don't get this, do you want to pass a context.Context and cancel that?

File: pkg/api/controller.go [hunk 1]
```diff
@@ -246,3 +250,10 @@ func (c *Controller) Run() error {
 
 	return server.Serve(l)
 }
+
+func (c *Controller) Shutdown() {
+	c.SyncWaitGroup.Wait()
```
rchincha, line 255: "wait gracefully"

rchincha: add this comment ^

pkg/api/controller.go
```diff
@@ -34,6 +36,7 @@ type Controller struct {
 	Audit           *log.Logger
 	Server          *http.Server
 	Metrics         monitoring.MetricServer
+	SyncWaitGroup   *goSync.WaitGroup
```
rchincha, line 39: rename var as "wgShutDown"

pkg/extensions/extensions.go
```diff
@@ -83,7 +85,7 @@ func EnableSyncExtension(config *config.Config, log log.Logger, storeController
 			}
 		}
 
-		if err := sync.Run(*config.Extensions.Sync, storeController, log); err != nil {
+		if err := sync.Run(*config.Extensions.Sync, storeController, log, wg); err != nil {
```
rchincha, line 88: Make "log" as last param

pkg/extensions/sync/sync.go
```diff
@@ -467,6 +468,8 @@ func Run(cfg Config, storeController storage.StoreController, logger log.Logger)
 			continue
 		}
 
+		wg.Add(1)
```
rchincha, line 471: // increment reference since will be busy, so shutdown has to wait

pkg/extensions/sync/sync.go
```diff
@@ -482,6 +485,9 @@ func Run(cfg Config, storeController storage.StoreController, logger log.Logger)
 					credentialsFile[upstreamRegistry], uuid.String()); err != nil {
 					l.Error().Err(err).Msg("sync exited with error, stopping it...")
 					ticker.Stop()
+					wg.Done()
+				} else {
+					wg.Done()
 				}
```
rchincha, line 491: move wg.Done() here in one place instead

rchincha:
Change commit msg:
"add wait group for graceful shutdown"

rchincha (approved):
lgtm

PR #326, opened 2021-12-08
Install notation as prerequisite for test Makefile target

adodon2go:
**What type of PR is this?**
bug fix

When running 'make test' the tests will fail if the user does not have notation tool installed.
We need a target that will check if notation is installed and if it is not- to install it before running the tests.

Failures:

  * zotregistry.io/zot/pkg/api_test/controller_test.go 
  Line 2971:
  Expected: nil
  Actual:   'exec: "notation": executable file not found in $PATH'

524 total assertions

--- FAIL: TestImageSignatures (4.50s)

'make test' passes regardless if you have notation tool installed or not

**Automation added to e2e**:
Not needed

No

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ TOOLSDIR := hack/tools
 PATH := bin:$(TOOLSDIR)/bin:$(PATH)
 STACKER := $(shell which stacker)
 GOLINTER := $(TOOLSDIR)/bin/golangci-lint
+NOTATION := $(TOOLSDIR)/bin/notation
 OS ?= linux
 ARCH ?= amd64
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -49,6 +49,10 @@ test:
 test-clean:
 	$(shell sudo rm -rf /etc/containers/certs.d/127.0.0.1:8089/)
 
+.PHONY: install-notation
+install-notation:
+	notation -v || (curl -Lo notation.tar.gz https://github.com/notaryproject/notation/releases/download/v0.7.1-alpha.1/notation_0.7.1-alpha.1_linux_amd64.tar.gz && sudo tar xvzf notation.tar.gz -C /usr/bin notation; rm notation.tar.gz)
```
rchincha, line 54: https://github.com/project-zot/zot/pull/269/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R52

^ try following this model, wherein I create a hack/tools/bin dir and download there, so becomes independent on that binaries are on the host system.

Makefile
```diff
@@ -6,7 +6,9 @@ COMMIT ?= $(if $(shell git status --porcelain --untracked-files=no),$(COMMIT_HAS
 CONTAINER_RUNTIME := $(shell command -v podman 2> /dev/null || echo docker)
 PATH := bin:$(PATH)
```
rchincha, line 7: move PATH down after the TOOLSDIR var and set PATH := bin:$(TOOLSDIR)/bin:$(PATH)

andaaron (approved):

peusebiu (approved):

PR #329, opened 2021-12-10
Verify that write location is under is.rootDir

rafaeljigau:
Signed-off-by: Jigau Iosif-Rafael <ijigau@cisco.com>

### DIFF
--- a/pkg/storage/storage_fs.go
+++ b/pkg/storage/storage_fs.go
@@ -159,8 +159,23 @@ func (is *ImageStoreFS) Unlock() {
 
 func (is *ImageStoreFS) initRepo(name string) error {
 	repoDir := path.Join(is.rootDir, name)
+
+	absRepoDir, err := filepath.Abs(repoDir)
+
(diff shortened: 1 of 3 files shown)


pkg/storage/storage_fs.go
```diff
@@ -159,8 +159,23 @@ func (is *ImageStoreFS) Unlock() {
...
+
+	absRepoDir, err := filepath.Abs(repoDir)
+
+	if err != nil {
+		is.log.Error().Err(err).Msg("error trying to get absolute path of repo")
+
+		return err
+	}
+
+	if (is.rootDir + "/" + name) != absRepoDir {
```
peusebiu, line 171: use path.Join(is.rootDir, name)

rafaeljigau: I could have used this in the first place and no changes would have been needed.

The problem is what if name = "../../abc" the join would be different from what we expect.

andaaron:
Will pick this up later.

Issue #202, opened 2021-06-02
add performance/stress test suite


rchincha:
Here are the use cases for this:

*   ci/cd, so that we run stress tests against releases/commits so we don't regress - this can be part of zot's own ci/cd
*   devs, so they validate any performance improvements they attempt
*   troubleshooting, perf tests in deployments to understand bottlenecks
*   benchmarking, perf tests against many different registry projects to understand where we stand

https://github.com/anuvu/zot/blob/master/pkg/api/controller_test.go#L1929
^ Can we enhance this?

rchincha:
https://docs.openstack.org/developer/performance-docs/test_results/container_repositories/registry2/index.html

andaaron:
1. Remove the calls to the skopeo binary - replace them with a go library @peusebiu will provide details.
2. We need to increase the the load by pushing/pulling a larger number of images on a larger number of threads/goroutines. (maybe add configuration for the number of images)

rchincha:
docker distribution perf results
https://docs.openstack.org/developer/performance-docs/test_results/container_repositories/registry2/index.html

rchincha:
https://gist.github.com/ervinb/224359235ee046c68fbc9456d6c4662b

rchincha:
https://atlarge-research.com/pdfs/Petar_Galic_Master_Thesis_14_12_2020.pdf

PR #331, opened 2021-12-10
perf: add a 'zb' cmd for perf testing of dist-spec registries

rchincha:

**What type of PR is this?**
feature
performance benchmark tool

#202 

initial commit for a performance/benchmark tool


rchincha:
https://github.com/project-zot/zot/blob/main/pkg/compliance/v1_0_0/check.go
^ pls look at this for all the different API usages.

Issue #258, opened 2021-10-27
extensions: toggle each "extension" with enable=True/False option


rchincha:
The use case is:
1. we want to enable/disable only a particular extension without changing too much of the config
2. someone can prep a config and leave it disabled, and someone else can review and enable it

PR #339, opened 2021-12-20
Added config enable=true/false for extensions

rafaeljigau:
Signed-off-by: Jigau Iosif Rafael <ijigau@cisco.com>

**Which issue does this PR fix**: #258

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -255,7 +255,7 @@ func (c *Controller) InitImageStore() error {
 	}
 
 	// Enable extensions if extension config is provided
-	if c.Config.Extensions != nil && c.Config.Extensions.Sync != nil {
+	if c.Config.Extensions != nil && c.Config.Extensions.Sync != nil && c.Config.Extensions.Sync.Enable {
 		ext.EnableSyncExtension(c.Config, c.wgShutDown, c.StoreController, c.Log)
 	}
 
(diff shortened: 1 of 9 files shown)


adodon2go:
What happens in case sync is configured in disabled mode. Does it make sense to add a new test for this scenario?

rafaeljigau:
If zot is started with sync feature not enabled, I dont think there is something to test.

If the config it's changed from enable true to enable false, as I understand @peusebiu's sync code, the sync won't stop, so more logic needs to be added I think.

pkg/api/controller.go
```diff
@@ -253,7 +253,7 @@ func (c *Controller) InitImageStore() error {
 	}
 
 	// Enable extensions if extension config is provided
-	if c.Config.Extensions != nil && c.Config.Extensions.Sync != nil {
```
adodon2go, line 256: I think we also need to check if sync feature is enabled [here](https://github.com/project-zot/zot/blob/main/pkg/api/routes.go#L1271) and [here](https://github.com/project-zot/zot/blob/main/pkg/api/routes.go#L1283)
Also check with Sebi if changes in pkg/cli/root.go are needed.

peusebiu: Indeed, Alexei is right. 

Add a check here
https://github.com/project-zot/zot/blob/main/pkg/cli/root.go#L233

peusebiu: Also here: https://github.com/project-zot/zot/blob/main/pkg/extensions/extensions.go#L103

PR #341, opened 2021-12-21
coverage: add failure injection framework

rchincha:

**What type of PR is this?**

Feature

Provides a failure injection framework to inject and cover hard to reach code paths by artificially injecting failures.

### DIFF
--- a/pkg/test/inject_test.go
+++ b/pkg/test/inject_test.go
@@ -1,0 +1,120 @@
+//go:build dev
+// +build dev
+
+// This file should be linked only in **development** mode.
+
+package test_test
+
+import (
+	"errors"
+	"testing"
+
+	. "github.com/smartystreets/goconvey/convey"
+	"zotregistry.io/zot/pkg/test"
+)
+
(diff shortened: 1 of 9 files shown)


rchamarthy (dismissed):

pkg/test/inject_test.go
```diff
@@ -0,0 +1,103 @@
...
+	errNotZero = errors.New("not zero")
+	errCall1   = errors.New("call1 error")
+	errCall2   = errors.New("call2 error")
+)
+
+func foo() error {
+	fmap := map[string]string{"key1": "val1", "key2": "val2"}
+
+	_, ok := fmap["key1"] // should never fail
+	if !test.CheckOk(ok) {
```
andaaron, line 23: As an alternative approach, ideally test would be an interface with different implementations for test cases compared to the production code.

rchincha: Fixed that way now.

peusebiu (approved):

Issue #162, opened 2020-12-02
CONTRIBUTING.md (or asciidoc)


McCodeman:
As a prospective contributor, I want documentation explaining the process and expectations for contributing to this project.

Acceptance Criteria:
* [ ] Create CONTRIBUTING.md (or asciidoc) in root of project.
* [ ] Documentation should include:
  * [ ] How to get started including how to sign a Contributor License Agreement
  * [ ] Pointers to a project architecture document and other relevant information needed to familiarize a new contributor with project internals.
  * [ ] Link to how to setup a development environment.
  * [ ] Instructions on setting up any required accounts
  * [ ] GitHub workflow
  * [ ] How to request review of Issues and PRs
  * [ ] Issue and PR management including Issue Triage

Don't assume that the reader of this document has any background knowledge of the project, acronyms, etc. This document should enable self-service startup for new contributors.

rchincha:
https://github.com/project-zot/zot/blob/main/CONTRIBUTING.md
All other docs at:
https://zotregistry.io

PR #342, opened 2021-12-21
Added storage latency histogram metric closes #279

### LINKED ISSUE #279: metrics: add counters to track how long locks are being held
We also want to track how long locks are being held so that we get better visibility into contention and latency.

adodon2go:
**What type of PR is this?**
feature

#279 

We also want to track how long storage locks are being held so that we get better visibility into contention and latency.

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/s3/storage.go
+++ b/pkg/storage/s3/storage.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"time"
 
 	// Add s3 support.
 	"github.com/docker/distribution/registry/storage/driver"
@@ -27,6 +28,11 @@ import (
 	"zotregistry.io/zot/pkg/extensions/monitoring"
 	zlog "zotregistry.io/zot/pkg/log"
 	"zotregistry.io/zot/pkg/storage"
+)
+
+const (
+	RLOCK  = "RLock"
+	RWLOCK = "RWLock"
 )
 
 // ObjectStorage provides the image storage operations.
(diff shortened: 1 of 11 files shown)


pkg/storage/s3/storage.go
```diff
@@ -143,8 +149,16 @@ func (is *ObjectStorage) initRepo(name string) error {
 
 // InitRepo creates an image repository under this store.
 func (is *ObjectStorage) InitRepo(name string) error {
+	start := time.Now()
```
rchincha, line 152: why not move this into Lock()/Unlock() funcs itelf? Else all this code has to be aware of metrics!

rchincha (commented):
Push some non-trivial parallel traffic, and post some pics/data.

pkg/storage/s3/storage.go
adodon2go: The reason of not moving all the logic inside Lock/Unlock functions was the latency computation logic: we need to save somewhere the starting time (could be a global variable but that is not thread safe as we calculate the starting time before lock acquiring and there is nothing between Lock/Unlock operations so other goroutines can rewrite the global starting time) and then correlate later that specific starting time with the current goroutine that processes Unlock in a defer.
Alternatively we can add this info into a context. Did you prefer the context approach instead of the PR changes?

rchincha: var start time.Time // local var so doesn't need a lock

is.Lock(&start) => func Lock() { *start := time.Now() .... }

is.Unlock(start) => func Unlock() { ... end := time.Now(), latency = end.Sub(start), metrics)

rchincha: Of course, use better varnames, "lockTimestamp", "lockStart", etc

pkg/storage/s3/storage.go
```diff
@@ -143,8 +159,10 @@ func (is *ObjectStorage) initRepo(name string) error {
 
 // InitRepo creates an image repository under this store.
 func (is *ObjectStorage) InitRepo(name string) error {
+	lockStart := time.Now()
+
 	is.Lock()
```
rchincha, line 164: Maybe pass the "lockStart" into is.Lock(), otherwise api looks asymmetric.

rchincha, line 164: Move the time.Now() into is.Lock() {...} itself

So line #162 here becomes "var lockLatency time.Time"
and is.Lock() becomes "is.Lock(&lockLatency)"

pkg/storage/storage_fs.go
```diff
@@ -142,23 +144,31 @@ func NewImageStore(rootDir string, gc bool, dedupe bool, log zlog.Logger, metric
...
 	is.lock.RUnlock()
+
+	lockEnd := time.Now()
+	latency := lockEnd.Sub(*lockStart)
+	monitoring.ObserveStorageLockLatency(is.metrics, latency, is.RootDir(), RLOCK) // histogram
 }
 
 // Lock write-lock.
-func (is *ImageStoreFS) Lock() {
+func (is *ImageStoreFS) Lock(lockStart *time.Time) {
```
rchincha, line 161: move the "lockStart = time.Now()" into this func.

rchincha: and remove that from the callers.

rchincha (approved):
lgtm

Issue #344, opened 2021-12-22
Deduplicate parallel container downloads


vinjana:
**Is your feature request related to a problem? Please describe.**

Currently, if multiple clients request the same container from Zot and Zot pulls the container from a synchronized registry, it pulls the same container multiple times in parallel:

```
$ for i in $(seq 1 10); do (docker pull localhost:5000/cassandra &); done
$ ls /data/registry/cassandra/.sync
3c146850-f847-46b6-9b9d-6debbc75b79e
46e5414f-4f0d-4cf5-ac11-d0387dbcea1e
4987efe9-a489-46f4-9b71-9a82d43cfc10
6761bc89-2618-48cf-a460-2d01db11b098
6a4b5395-0d41-4055-8801-93187685b293
6e14ef54-7214-41c5-ae08-3ba48afe708e
b0b90fc6-4b77-4066-86b2-04529cf2a9bf
b54317d7-2950-4b1f-9d40-4f70edac51ef
d21d5adf-bea7-4915-b4f0-8fc068b642c6
```

This is problematic if the client has little control over the parallelism of requests, e.g. in a distributed environment, possibly with a large number of (human) users like in a cluster, and if containers are big. Effectively, the traffic is increased, and the individual pull is slowed down, which reduces the response time for the clients. Finally, if the number of parallel requests is very large, Zot may overload the upstream/synchronized server or run into its parallel request limits or traffic quotas.

Indeed, a primary use case we intend to use a local registry for is for caching containers -- to prevent the multiple download of the same container. Consequently, the benefit of using Zot as registry cache is reduced because of the current implementation.

Furthermore, even if the client has control over the container requests, each client has to work around the parallel pulls by Zot by implementing its own deduplication, e.g. by its own locking mechanism, by staging (first get the container then start container usages), or by random offsetting, etc.. Therefore, an infrastructure service like Zot that allows parallel requests can relief client applications from re-re-re-implementing the handling of parallelism, by doing it for them. 

**Describe the solution you'd like**

Zot should recognize that a container is currently being downloaded, and instead of starting the download again, wait for the first (and only) download to complete.

**Describe alternatives you've considered**

All other solutions that come to my mind are client-side solutions. 

**Additional context**

None.

rchincha:
Thanks @vinjana for filing this issue. Yes, we are aware of this behavior and a fix is in progress. The "sync" feature was merged fairly recently and we are iterating to make this better.

PR #345, opened 2021-12-22
sync on demand: in case of parallel requests pull image just once, cl…

peusebiu:
…oses #344

**What type of PR is this?**

#344 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -21,6 +21,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/gorilla/mux"
 	jsoniter "github.com/json-iterator/go"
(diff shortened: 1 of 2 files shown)


rchincha:
Why not solve this with per-image or per-repo channel? Do a lookup, if the image is not found, send the request to the downloader goroutine along with the callback context (http.response, image, ...)
Try not to add too many locks in the code, makes it unnecessarily complex and hard to debug going fwd.

pkg/api/routes.go
```diff
@@ -47,11 +48,12 @@ const (
 )
 
 type RouteHandler struct {
-	c *Controller
+	c              *Controller
+	demandedImages sync.Map
```
peusebiu, line 52: Thinking about it, not sure weather sync.Map is overkill. Maybe a normal map with an internal mutex would suffice.

peusebiu:
@rchincha is this better?

basically I have per image channel (map) where I store currently downloading images and their channel, if key is present then wait for sync to finish (for channel to be closed), otherwise sync on demand and remove key after and close the channel.

peusebiu:
closed in f89925fb274172b2664192d4243479c345c5b41b

Issue #306, opened 2021-12-03
perf: consider using sha_ni to speed up sha256 operations


rchincha:
https://github.com/minio/sha256-simd

rchincha:
golang already uses avx2 for sha256 acceleration. Once sha_ni is widely available, this becomes more attractive.

PR #347, opened 2021-12-24
storage: use sha256-simd from minio

rchincha:

**What type of PR is this?**
performance enchancement

#306 

Uses the sha_ni extensions for CPUs that support it.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,6 +26,7 @@ require (
 	github.com/gorilla/handlers v1.5.1
 	github.com/gorilla/mux v1.8.0
 	github.com/json-iterator/go v1.1.12
+	github.com/minio/sha256-simd v1.0.0
 	github.com/mitchellh/mapstructure v1.4.2
 	github.com/nmcclain/ldap v0.0.0-20210720162743-7f8d1e44eeba
 	github.com/notaryproject/notation v0.7.0-alpha.1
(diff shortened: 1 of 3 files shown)


rchincha:
Linux kernel already has included support for this.
http://lkml.iu.edu/hypermail/linux/kernel/1511.0/00383.html

adodon2go (approved):
LGTM
I did not have CPU support for this new feature on my VM (both locally and in the lab). The improvement is working only on bare metal? Did you have chance to try this on a machine that supports the feature and does the benchmarking shows a huge improvement for large blobs? Currently it takes between 3-4 sec to compute hash for blob with size 1000MB (includes the reading of the file which I guess takes a lot) using crypto/sha256 (fallback from sha256-simd without CPU SHA features support). Maybe there is still room for improvements here like caching the digest when the blob is first uploaded into .uploads (for systems that does not have CPU support for sha acceleration).

> adodon@adodon-cs-bld:/data/ssd/work/benchmark$ go test -bench=. -benchtime=60s
> goos: linux
> goarch: amd64
> pkg: github.com/benchmark
> cpu: Intel(R) Xeon(R) Gold 5218 CPU @ 2.30GHz
> BenchmarkNewSha100MB-20              205         350024972 ns/op
> BenchmarkNewSha1000MB-20              19        3503338046 ns/op
> PASS
> ok      github.com/benchmark    177.248s
> 
> adodon@adodon-cs-bld:/data/ssd/work/benchmark$ go test -bench=. -benchtime=60s
> goos: linux
> goarch: amd64
> pkg: github.com/benchmark
> cpu: Intel(R) Xeon(R) Gold 5218 CPU @ 2.30GHz
> BenchmarkNewSha100MB-20              205         350024972 ns/op
> BenchmarkNewSha1000MB-20              19        3503338046 ns/op
> PASS
> ok      github.com/benchmark    177.248s

rchincha:
https://github.com/project-zot/zot/pull/347#pullrequestreview-840786222
^ make sure "lscpu | grep sha_ni" is found.
Raw sha_ni improvement is huge (3-4X), but again the total time for zot use case will depend on whether network latency dominates.

PR #348, opened 2021-12-28
Added config enable=true/false for extensions closes #258

### LINKED ISSUE #258: extensions: toggle each "extension" with enable=True/False option

adodon2go:
Signed-off-by: Alexei Dodon <adodon@cisco.com>

Which issue does this PR fix: #258

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -45,7 +45,6 @@ func NewController(config *config.Config) *Controller {
 	var controller Controller
 
 	logger := log.NewLogger(config.Log.Level, config.Log.Output)
-
 	controller.Config = config
 	controller.Log = logger
 	controller.wgShutDown = new(goSync.WaitGroup)
@@ -137,7 +136,7 @@ func (c *Controller) Run() error {
 	if c.Config != nil &&
 		c.Config.Extensions != nil &&
 		c.Config.Extensions.Metrics != nil &&
-		c.Config.Extensions.Metrics.Enable {
+		*c.Config.Extensions.Metrics.Enable {
 		enabled = true
 	}
 
@@ -296,7 +295,7 @@ func (c *Controller) InitImageStore() error {
 	}
 
 	// Enable extensions if extension config is provided
-	if c.Config.Extensions != nil && c.Config.Extensions.Sync != nil {
+	if c.Config.Extensions != nil && c.Config.Extensions.Sync != nil && *c.Config.Extensions.Sync.Enable {
 		ext.EnableSyncExtension(c.Config, c.wgShutDown, c.StoreController, c.Log)
 	}
 
--- a/pkg/cli/root.go
+++ b/pkg/cli/root.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"time"
 
 	glob "github.com/bmatcuk/doublestar/v4"
 	"github.com/fsnotify/fsnotify"
@@ -15,6 +16,7 @@ import (
 	"zotregistry.io/zot/errors"
 	"zotregistry.io/zot/pkg/api"
 	"zotregistry.io/zot/pkg/api/config"
+	extconf "zotregistry.io/zot/pkg/extensions/config"
 	"zotregistry.io/zot/pkg/extensions/monitoring"
 	"zotregistry.io/zot/pkg/storage"
 )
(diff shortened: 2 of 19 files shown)


pkg/extensions/sync/sync_test.go
```diff
@@ -257,7 +257,7 @@ func TestSyncOnDemand(t *testing.T) {
 		}
 
 		syncConfig := &sync.Config{
-			Registries: []sync.RegistryConfig{syncRegistryConfig},
+			Registries: []sync.RegistryConfig{syncRegistryConfig}, Enable: true,
```
rchincha, line 260: Is there a test case, where "Enable=false" and sync is ignored?
How about for other extensions? "search", "cve"?

adodon2go: Added a new test case when sync is disabled.
For the rest of extensions, they already have option to enable/disable their respective features.

pkg/extensions/sync/sync_test.go
```diff
@@ -354,6 +354,74 @@ func TestSyncOnDemand(t *testing.T) {
...
+					Tags: &sync.Tags{
+						Regex:  &regex,
+						Semver: &semver,
+					},
+				},
+			},
+			URL:          srcBaseURL,
+			PollInterval: updateDuration,
+			TLSVerify:    &tlsVerify,
+			CertDir:      "",
```
peusebiu, line 385: Add OnDemand true, so we test both cases (on demand and periodically)

peusebiu (dismissed):
LGTM

pkg/extensions/config/config.go
```diff
@@ -15,15 +15,15 @@ type ExtensionConfig struct {
 type SearchConfig struct {
 	// CVE search
 	CVE    *CVEConfig
-	Enable bool
+	Enable *bool
```
rchincha, line 18: Ok, it can be "nil" (not specified, defaults to true) and "true/false"

pkg/api/controller.go
```diff
@@ -41,8 +42,28 @@ type Controller struct {
 func NewController(config *config.Config) *Controller {
 	var controller Controller
 
-	logger := log.NewLogger(config.Log.Level, config.Log.Output)
+	if config != nil && config.Extensions != nil {
+		defaultTrue := true
```
rchincha, line 46: "defaultTrue" usage here is weird? linter forcing you? what is the complaint?

pkg/api/controller.go
```diff
@@ -41,8 +42,28 @@ type Controller struct {
 func NewController(config *config.Config) *Controller {
 	var controller Controller
 
-	logger := log.NewLogger(config.Log.Level, config.Log.Output)
+	if config != nil && config.Extensions != nil {
```
rchincha, line 45: Why not move all this into LoadConfiguration() itself?

pkg/api/controller.go
adodon2go: will use 
`var defaultVal = true` 
instead

pkg/api/controller.go
rchincha: In that case, we definitely need one place to LoadConfiguration() even for server so that we avoid this mangling. 
I would LoadConfiguration() and mangle and then start various enabled features.

peusebiu (approved):

peusebiu (dismissed):
LGTM

pkg/cli/root.go
```diff
@@ -15,10 +16,16 @@ import (
 	"zotregistry.io/zot/errors"
 	"zotregistry.io/zot/pkg/api"
 	"zotregistry.io/zot/pkg/api/config"
+	extconf "zotregistry.io/zot/pkg/extensions/config"
 	"zotregistry.io/zot/pkg/extensions/monitoring"
 	"zotregistry.io/zot/pkg/storage"
 )
 
+const (
```
rchincha, line 24: No need for this. golang "time" package already has a time.Hour, so just make a new const called defaultCVEUpdate = 2*time.Hour

pkg/extensions/config/config.go
```diff
@@ -15,15 +15,15 @@ type ExtensionConfig struct {
...
 }
 
 type CVEConfig struct {
-	UpdateInterval time.Duration // should be 2 hours or more, if not specified default be kept as 24 hours
+	UpdateInterval time.Duration // should be 2 hours or more, if not specified default be kept as 2 hours
```
rchincha, line 22: default should be 24 hours for cve updates, should not go faster. At least this comment needs to be reverted.

pkg/cli/root.go
```diff
@@ -322,17 +382,9 @@ func LoadConfiguration(config *config.Config, configPath string) {
 		panic(err)
 	}
 
-	// various config checks
-	validateConfiguration(config)
-
 	// defaults
-	defaultTLSVerify := true
+	applyDefaultValues(config, viperInstance)
```
rchincha, line 386: Nice hook here.

rchincha: But where is this being used in the server/controller path?

adodon2go: For that part LoadConfiguration() is called and applyDefaultValues() inside it.

rchincha (approved):
lgtm

rchincha (approved):
lgtm

PR #365, opened 2022-01-13
ci/cd: add continuous benchmark action

rchincha:

**What type of PR is this?**

feature

Adds a performance benchmark in ci/cd so we don't regress across commits.

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/cmd/zb/main.go
+++ b/cmd/zb/main.go
@@ -13,7 +13,7 @@ import (
 func NewPerfRootCmd() *cobra.Command {
 	showVersion := false
 
-	var auth, workdir, repo, output string
+	var auth, workdir, repo, outFmt string
 
 	var concurrency, requests int
 
@@ -45,7 +45,7 @@ func NewPerfRootCmd() *cobra.Command {
 
 			requests = concurrency * (requests / concurrency)
 
-			Perf(workdir, url, auth, repo, concurrency, requests)
+			Perf(workdir, url, auth, repo, concurrency, requests, outFmt)
 		},
 	}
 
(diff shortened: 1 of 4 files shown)


cmd/zb/perf.go
```diff
@@ -198,7 +200,17 @@ func updateStats(summary *statsSummary, record statsRecord) {
 	summary.latencies = append(summary.latencies, record.latency)
 }
 
-func printStats(requests int, summary *statsSummary) {
+type cicdTestSummary struct {
```
andaaron, line 203: I wouldn't use 'cicd' in types or variable names. This is an assumption on how the user would use the feature rather than what the feature does. Why not use something like 'jsonTestSummary'?

cmd/zb/main.go
```diff
@@ -59,8 +59,8 @@ func NewPerfRootCmd() *cobra.Command {
 		"Number of multiple requests to make at a time")
 	rootCmd.Flags().IntVarP(&requests, "requests", "n", 1,
 		"Number of requests to perform")
-	rootCmd.Flags().StringVarP(&output, "output-format", "o", "",
-		"Output format of test results [default: stdout]")
+	rootCmd.Flags().StringVarP(&outFmt, "output-format", "o", "",
+		"Output format of test results: stdout (default), json, ci-cd")
```
andaaron, line 63: This is a bit confusing as there are not 3 options (stdout, json, ci-cd). And the stdout option is always enabled even if not specified.

cmd/zb/perf.go
```diff
@@ -198,7 +200,17 @@ func updateStats(summary *statsSummary, record statsRecord) {
...
+	Name  string      `json:"name"`
+	Unit  string      `json:"unit"`
+	Value interface{} `json:"value"`
+	Range string      `json:"range,omitempty"`
```
andaaron, line 207: How is this field used?

cmd/zb/main.go
rchincha: "ci-cd" is specific to the ci-cd we are using and only outputs "requests per sec" in the cicdSummary format.
json format instead would have everything in JSON format, which will be separate.

cmd/zb/perf.go
rchincha: "ci-cd" is specific to the ci-cd we are using and only outputs "requests per sec" in the cicdSummary format.
json format instead would have everything in JSON format, which will be separate.

cmd/zb/perf.go
rchincha: Specific to the ci-cd workflow we are using. Range says from the historical values +- how much it can deviate.

andaaron: Ok, now I understand

andaaron (approved):

PR #372, opened 2022-01-20
Add read-only tests, mixed read-only tests, mixed write-only tests and mixed read-write tests for zb binary

roxanaN:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/cmd/zb/helper.go
+++ b/cmd/zb/helper.go
@@ -1,0 +1,908 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+	"path"
+	"time"
+
+	"github.com/google/uuid"
+	imeta "github.com/opencontainers/image-spec/specs-go"
+	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"gopkg.in/resty.v1"
+	"zotregistry.io/zot/errors"
+	"zotregistry.io/zot/pkg/test"
+)
+
+func deleteTestRepo(repos []string, url string, client *resty.Client) error {
+	for _, repo := range repos {
+		resp, err := client.R().Delete((fmt.Sprintf("%s/v2/%s/", url, repo)))
+		if err != nil {
+			return err
+		}
+
+		// request specific check
+		statusCode := resp.StatusCode()
+		if statusCode != http.StatusAccepted {
+			return errors.ErrUnknownCode
+		}
+	}
+
+	return nil
+}
+
+func pullAndCollect(url string, repos []string, manifestItem manifestStruct,
+	config testConfig, client *resty.Client, statsCh chan statsRecord) []string {
+	manifestHash := manifestItem.manifestHash
+	manifestBySizeHash := manifestItem.manifestBySizeHash
+
+	func() {
+		start := time.Now()
+
+		var isConnFail, isErr bool
+
+		var statusCode int
+
+		var latency time.Duration
+
+		defer func() {
+			// send a stats record
+			statsCh <- statsRecord{
+				latency:    latency,
+				statusCode: statusCode,
+				isConnFail: isConnFail,
+				isErr:      isErr,
+			}
+		}()
+
+		if config.mixedSize {
+			smallSizeIdx := 0
+			mediumSizeIdx := 1
+			largeSizeIdx := 2
+
+			idx := flipFunc(config.probabilityRange)
+
+			switch idx {
+			case smallSizeIdx:
+				statusRequests["1MB"]++
+			case mediumSizeIdx:
+				statusRequests["10MB"]++
+			case largeSizeIdx:
+				statusRequests["100MB"]++
+			}
+
+			manifestHash = manifestBySizeHash[idx]
+		}
+
+		for repo, manifestTag := range manifestHash {
+			manifestLoc := fmt.Sprintf("%s/v2/%s/manifests/%s", url, repo, manifestTag)
+
+			// check manifest
+			resp, err := client.R().
+				SetHeader("Content-Type", "application/vnd.oci.image.manifest.v1+json").
+				Head(manifestLoc)
+
+			latency = time.Since(start)
+
+			if err != nil {
+				isConnFail = true
+
+				return
+			}
+
+			// request specific check
+			statusCode = resp.StatusCode()
+			if statusCode != http.StatusOK {
+				isErr = true
+
+				return
+			}
+
+			// send request and get the manifest
+			resp, err = client.R().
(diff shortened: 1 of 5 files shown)


cmd/zb/perf.go
```diff
@@ -622,6 +625,195 @@ func PushChunkStreamed(workdir, url, auth, trepo string, requests int,
...
+			// request specific check
+			statusCode = resp.StatusCode()
+			if statusCode != http.StatusOK {
+				isErr = true
+
+				return
+			}
+
+			manifestBody := resp.Body()
+			_, err = io.Copy(ioutil.Discard, bytes.NewReader(manifestBody))
```
rchincha, line 774: yup, ioutil.Discard also works

cmd/zb/perf.go
```diff
@@ -622,6 +625,195 @@ func PushChunkStreamed(workdir, url, auth, trepo string, requests int,
...
+
+				blobBody := resp.Body()
+				_, err = io.Copy(ioutil.Discard, bytes.NewReader(blobBody))
+			}
+
+			latency = time.Since(start)
+		}()
+	}
+
+	resp, err = client.R().Delete((fmt.Sprintf("%s/v2/%s/", url, repo)))
```
rchincha, line 812: Ok, not all registries may implement this DELETE, so makes sense to not check the status code here.

rchincha (commented):
Fix the ci/cd pipeline errors.
Looks ok.
Go ahead and do a mixed workload also (in another PR also ok) - read/write

cmd/zb/perf.go
andaaron: Do you think it would be safer to keep track of all created objects in a structure and delete one by one?

cmd/zb/perf.go
```diff
@@ -622,6 +625,215 @@ func PushChunkStreamed(workdir, url, auth, trepo string, requests int,
...
+			defer func() {
+				// send a stats record
+				statsCh <- statsRecord{
+					latency:    latency,
+					statusCode: statusCode,
+					isConnFail: isConnFail,
+					isErr:      isErr,
+				}
+			}()
+
```
rchincha, line 756: Some steps have been skipped.
Start with HEAD manifest (to check manifest/tag present), GET manifest, HEAD each blob and then GET each blob (this you have currently)

cmd/zb/perf.go
```diff
@@ -622,6 +625,215 @@ func PushChunkStreamed(workdir, url, auth, trepo string, requests int,
 	return nil
 }
 
+func PullMonolithStreamed(workdir, url, auth, trepo string, requests int,
```
rchincha, line 628: Call this simply Pull(), there aren't many variations like we do in Push.

cmd/zb/perf.go
rchincha: @roxanaN, if you like, you can move some code into pull.go and push.go - your choice

cmd/zb/perf.go
```diff
@@ -622,6 +625,215 @@ func PushChunkStreamed(workdir, url, auth, trepo string, requests int,
...
+				if err != nil {
+					log.Fatal(err)
+				}
+			}
+
+			latency = time.Since(start)
+		}()
+	}
+
+	resp, err = client.R().Delete((fmt.Sprintf("%s/v2/%s/", url, repo)))
```
rchincha, line 822: There is no DELETE of a top-level repo. You have to iterate over all the tags and delete each manifest entry.

https://github.com/opencontainers/distribution-spec/blob/main/spec.md#deleting-tags

cmd/zb/delete.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"fmt"
+	"net/http"
+
+	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"gopkg.in/resty.v1"
+	"zotregistry.io/zot/errors"
+)
+
+func deleteUploadedFiles(manifestHash map[string]string, configHash map[string]string,
```
rchincha, line 12: Why are we not deleting the entire repo after the test? Wouldn't that be quicker/cleaner?

cmd/zb/helper.go
```diff
@@ -0,0 +1,594 @@
...
+		statusCode := resp.StatusCode()
+		if statusCode != http.StatusAccepted {
+			return errors.ErrUnknownCode
+		}
+	}
+
+	return nil
+}
+
+func pushImage(workdir, url, trepo string, repos []string, size int,
```
rchincha, line 40: There are two types of push we are testing right? Which one is this? Monolith or Chunk?

cmd/zb/perf.go
```diff
@@ -36,6 +37,11 @@ const (
 	mediumBlob           = 10 * MiB
 	largeBlob            = 100 * MiB
 	cicdFmt              = "ci-cd"
+	smallSizeIdx         = 0
```
rchincha, line 40: Can we move these Idx into the test itself instead of making these global?

cmd/zb/perf.go
```diff
@@ -792,6 +805,57 @@ var testSuite = []testConfig{ // nolint:gochecknoglobals // used only in this te
...
+		tfunc: Pull,
+		size:  mediumBlob,
+	},
+	{
+		name:  "Pull 100MB",
+		tfunc: Pull,
+		size:  largeBlob,
+	},
+	{
+		name:             "Pull Mixed 28% 1MB, 70% 10MB, 2% 100MB",
```
rchincha, line 824: Make this "20% 1MB, 70% 10MB, 10% 100MB"

cmd/zb/perf.go
```diff
@@ -792,6 +805,57 @@ var testSuite = []testConfig{ // nolint:gochecknoglobals // used only in this te
...
+		tfunc:            Pull,
+		probabilityRange: normalizeProbabilityRange([]float64{0.28, 0.7, 0.02}),
+		mixedSize:        true,
+	},
+	{
+		name:             "Push Monolith Mixed 28% 1MB, 70% 10MB, 2% 100MB",
```
rchincha, line 830: Same here

rchincha:
@roxanaN, it may also be better if we display the read/write counts in the mixed test cases so that we confirm zb is doing the right thing.

For example,

Number of requests: 1000
- Pull 1M: 100
- Push 10M: 800

... etc

examples/config-bench.json
```diff
@@ -10,6 +10,6 @@
     },
     "log": {
         "level": "debug",
-        "output": "/dev/null"
```
rchincha, line 13: The idea behind doing /dev/null was to make sure that logging isn't affecting performance. Revert these config example changes?

rchincha (dismissed):
lgtm

rchincha (approved):
lgtm

rchincha (approved):
lgtm

Issue #359, opened 2022-01-10
bug storage: zot crashes if fd-limit reached


adodon2go:
**Describe the bug**
Zot crashes if file descriptor limit is reached

**To Reproduce**
1. Check the `Soft Limit" of zot process regarding file descriptors:

> midgard@ubuntu:~$ ps -ax | grep zot
>   49395 pts/1    Sl+    0:00 bin/zot serve examples/config-test.json
>   49484 pts/2    S+     0:00 grep --color=auto zot
> midgard@ubuntu:~$ cat /proc/49395/limits
> Limit                     Soft Limit           Hard Limit           Units     
> Max cpu time              unlimited            unlimited            seconds   
> Max file size             unlimited            unlimited            bytes     
> Max data size             unlimited            unlimited            bytes     
> Max stack size            8388608              unlimited            bytes     
> Max core file size        0                    unlimited            bytes     
> Max resident set          unlimited            unlimited            bytes     
> Max processes             95802                95802                processes 
> Max open files            1024                 1048576              files     
> Max locked memory         67108864             67108864             bytes     
> Max address space         unlimited            unlimited            bytes     
> Max file locks            unlimited            unlimited            locks     
> Max pending signals       95802                95802                signals   
> Max msgqueue size         819200               819200               bytes     
> Max nice priority         0                    0                    
> Max realtime priority     0                    0                    
> Max realtime timeout      unlimited            unlimited            us        
> midgard@ubuntu:~$ 

2. In above case the value is 1024.
3. Do some load test on zot (During 5 minutes interval a number of 1000 concurrent clients that uploads a small image like alpine:3.15)
4. Zot crashes
`
{"level":"error","error":"open /mnt/ramdisk/zot-extended/client62/.uploads/c0a605ed-ff0a-408e-a5d6-8672e2385458: too many open files","blob":"/mnt/ramdisk/zot-extended/client62/.uploads/c0a605ed-ff0a-408e-a5d6-8672e2385458","goroutine":132746,"caller":"zotregistry.io/zot/pkg/storage/storage_fs.go:936","time":"2022-01-10T22:19:51.773990827+02:00","caller":"zotregistry.io/zot/pkg/storage/storage_fs.go:936","message":"failed to open blob"}
2022/01/10 22:19:51 http: Accept error: accept tcp 127.0.0.1:5000: accept4: too many open files; retrying in 5ms
[...]

PR #374, opened 2022-01-21
storage: return errors instead of panics

rchincha:

**What type of PR is this?**

bug

#359 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/config/config.go
+++ b/pkg/api/config/config.go
@@ -47,6 +47,11 @@ type BearerConfig struct {
 	Cert    string
 }
 
+type RatelimitConfig struct {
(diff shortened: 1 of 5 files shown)


adodon2go (approved):
LGTM.
It will be useful to see some config example of this feature somewhere in examples/ . Otherwise you need to search in the repo code for how to use this.
In the context of https://github.com/project-zot/zot/issues/359 it behaves much better having this feature(with scrub I see less errors), although it can't fix issues automatically. Maybe scrub will do that periodically (plus cleaning up .uploads folder)

PR #387, opened 2022-01-27
Use InsecureSkipVerify only with https upstreams

peusebiu:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/sync/utils.go
+++ b/pkg/extensions/sync/utils.go
@@ -104,6 +104,13 @@ func getFileCredentials(filepath string) (CredentialsFile, error) {
 func getHTTPClient(regCfg *RegistryConfig, credentials Credentials, log log.Logger) (*resty.Client, error) {
 	client := resty.New()
 
+	registryURL, err := url.Parse(regCfg.URL)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/extensions/sync/utils.go
```diff
@@ -133,7 +133,7 @@ func getHTTPClient(regCfg *RegistryConfig, credentials Credentials, log log.Logg
 	}
 
 	// nolint: gosec
-	if regCfg.TLSVerify != nil && !*regCfg.TLSVerify {
+	if regCfg.TLSVerify != nil && !*regCfg.TLSVerify && strings.HasPrefix(regCfg.URL, "https") {
```
rchincha, line 136: What if "httpsabc.weirdprefix.com"?
Would actually parse the uri and then .Scheme

peusebiu: Ok, I'll use an url parser

andaaron (approved):

rchincha (approved):
lgtm

rchincha:
Rebase and resubmit

rchincha (approved):

PR #416, opened 2022-02-18
Added clustering github workflow

peusebiu:

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/.github/workflows/cluster.yaml
+++ b/.github/workflows/cluster.yaml
@@ -1,0 +1,162 @@
+name: "Clustering test"
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches: [main]
+  release:
+    types:
+      - published
+
+jobs:
+  client-tools:
+    name: Stateless zot with shared reliable storage
+    runs-on: ubuntu-latest
+    # services:
+    #   minio:
+    #     image: minio/minio:edge-cicd
+    #     env:
+    #       MINIO_ROOT_USER: minioadmin
+    #       MINIO_ROOT_PASSWORD: minioadmin
+    #     ports:
+    #       - 9000:9000
+    #     volumes:
+    #       - /tmp/data:/data
+    #     options: --name=minio --health-cmd "curl http://localhost:9000/minio/health/live"
+    steps:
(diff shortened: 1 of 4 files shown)


peusebiu:
need to change localstack with minio, as localstack seems to be very slow.

test/cluster/haproxy.cfg
```diff
@@ -0,0 +1,46 @@
+global
```
rchincha, line 1: move this file into examples/cluster/haproxy.cfg instead

rchincha: Maybe also test i/o with `zb` also.

rchincha: Also, give a separate name for this clustering setup - "stateless zot with shared reliable storage". There could be others and then we have to scramble.

peusebiu: fixed

.github/workflows/cluster.yaml
```diff
@@ -0,0 +1,134 @@
...
+          AWS_ACCESS_KEY_ID: minioadmin
+          AWS_SECRET_ACCESS_KEY: minioadmin
+      
+      - name: Run benchmark
+        run: |
+            ./bin/zot-linux-amd64 serve test/cluster/config-minio1.json &
+            ./bin/zot-linux-amd64 serve test/cluster/config-minio2.json &
+            ./bin/zot-linux-amd64 serve test/cluster/config-minio3.json &
+            sleep 10
+            make -o run-minimal BENCH_OUTPUT=ci-cd run-bench
```
rchincha, line 111: Why add a new Makefile target, just build bench and run it directly here.

rchincha:
Also,

```
Complete requests:    0
Failed requests:      100
```

.github/workflows/cluster.yaml
peusebiu: I don't want to run minimal zot, but s3.

rchincha:
@peusebiu, disable logs ("output": "/dev/null") so that the output is clean and readable.

rchincha (approved):
lgtm

PR #426, opened 2022-02-24
ui: add route handling for UI

rchincha:

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -34,6 +34,7 @@ import (
 	"zotregistry.io/zot/pkg/log"
 	"zotregistry.io/zot/pkg/storage"
 	"zotregistry.io/zot/pkg/test" // nolint: goimports
+
 	// as required by swaggo.
 	_ "zotregistry.io/zot/swagger"
 )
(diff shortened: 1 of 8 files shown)


pkg/api/routes.go
```diff
@@ -58,14 +58,14 @@ func NewRouteHandler(c *Controller) *RouteHandler {
 }
 
 func (rh *RouteHandler) SetupRoutes() {
-	rh.c.Router.Use(AuthHandler(rh.c))
+	prefixedRouter := rh.c.Router.PathPrefix(RoutePrefix).Subrouter()
```
shimish2, line 61: with this change I think we will be authorizing routes starting from /v2 only and /query will be skipped.

shimish2 (changes requested):

pkg/api/routes.go
rchincha: yes, currently a draft PR, still evaluating implications.

rchincha: This is fixed now. Pls take a look.

rchincha:
Superseded by PR #670

PR #434, opened 2022-03-01
tls: restrict tls cipher suites

rchincha:
golang will not turn them off by default:
https://github.com/golang/go/issues/13385

**What type of PR is this?**

bug

Restrict cipher suites to TLS 1.2 only.


shimish2 (approved):
lgtm

PR #436, opened 2022-03-02
s3: bugfix, use sync.Map instead of map for storing multi part upload…

peusebiu:
…s references

add storage lock in GetIndexContent

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/s3/storage.go
+++ b/pkg/storage/s3/storage.go
@@ -45,8 +45,8 @@ type ObjectStorage struct {
 	// We must keep track of multi part uploads to s3, because the lib
 	// which we are using doesn't cancel multiparts uploads
 	// see: https://github.com/distribution/distribution/blob/main/registry/storage/driver/s3-aws/s3.go#L545
-	isMultiPartUpload map[string]bool
-	metrics           monitoring.MetricServer
+	multiPartUploads sync.Map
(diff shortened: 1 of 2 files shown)


pkg/storage/s3/storage.go
```diff
@@ -35,6 +35,24 @@ const (
 	RWLOCK = "RWLock"
 )
 
+type multiPartUploads struct {
```
rchincha, line 38: Did you need this struct at all? Couldn't you have just used a sync.Map directly? It is nice to have the apis, but still.

pkg/storage/s3/storage.go
```diff
@@ -45,8 +63,8 @@ type ObjectStorage struct {
 	// We must keep track of multi part uploads to s3, because the lib
 	// which we are using doesn't cancel multiparts uploads
 	// see: https://github.com/distribution/distribution/blob/main/registry/storage/driver/s3-aws/s3.go#L545
-	isMultiPartUpload map[string]bool
-	metrics           monitoring.MetricServer
+	multiPartUploads multiPartUploads
```
rchincha, line 66: This could just be a sync.Map

rchincha (approved):
lgtm

Issue #376, opened 2022-01-21
scrub: make scrub inline and periodic


rchincha:
**Is your feature request related to a problem? Please describe.**

`zot scrub` scrubs the filesystem but as a offline tool.

**Describe the solution you'd like**

Would like the scrubbing to be continuous done to catch any filesystem errors.

adodon2go:
Do we want also for scrub to take some delete actions in case image:tag is being detected as not valid?
I'm thinking this will be useful in the context of https://github.com/project-zot/zot/pull/374/files & https://github.com/project-zot/zot/issues/359.  Also maybe cleanup in the .uploads folder will be useful.

andaaron:
In the first iteration scrib will be read-only. It would not try to fix the issue. We will fixing the issue separately.

andaaron:
Done in https://github.com/project-zot/zot/pull/440

PR #440, opened 2022-03-04
make scrub inline and periodic

Andreea-Lupu:

**What type of PR is this?**

#376 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/examples/config-allextensions.json
+++ b/examples/config-allextensions.json
@@ -41,6 +41,9 @@
 			"cve": {
 				"updateInterval": "2h"
 			}
+		},
+		"scrub": {
+			"interval": "24h"
 		}
 	}
 }
--- a/examples/config-scrub.json
+++ b/examples/config-scrub.json
@@ -1,0 +1,18 @@
+{
+    "distSpecVersion":"1.0.1",
+    "storage": {
+        "rootDirectory": "/tmp/zot"
+    },
+    "http": {
+        "address": "127.0.0.1",
+        "port": "8080"
+    },
+    "log": {
+        "level": "debug"
+    },
+    "extensions": {
+        "scrub": {
+            "interval": "24h"
+        }
+    }
+}
--- a/pkg/extensions/extensions.go
+++ b/pkg/extensions/extensions.go
@@ -12,6 +12,7 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"zotregistry.io/zot/pkg/api/config"
+	"zotregistry.io/zot/pkg/extensions/scrub"
 	"zotregistry.io/zot/pkg/extensions/search"
 	cveinfo "zotregistry.io/zot/pkg/extensions/search/cve"
 	"zotregistry.io/zot/pkg/extensions/sync"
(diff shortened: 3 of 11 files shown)


examples/config-scrub.json
```diff
@@ -0,0 +1,19 @@
...
+    },
+    "http": {
+        "address": "127.0.0.1",
+        "port": "8080"
+    },
+    "log": {
+        "level": "debug"
+    },
+    "extensions": {
+        "scrub": {
```
rchincha, line 14: Another option is to put this under the "storage" section. But let's keep it here for now.

pkg/extensions/extensions.go
```diff
@@ -34,6 +34,39 @@ func downloadTrivyDB(dbDir string, log log.Logger, updateInterval time.Duration)
 	}
 }
 
+// Scrub ...
+func scrub(log log.Logger, scrubInterval time.Duration, storeController storage.StoreController) error {
+	for {
+		log.Info().Msg("executing scrub to check manifest/blob integrity")
+
+		result, err := storeController.CheckAllBlobsIntegrity()
```
rchincha, line 42: The worry is whether this holds a lock for a long time. A simple test is to have many repos and then try to push an image when a scrub is ongoing, will that push hang? You need a generator pattern: https://wiki.python.org/moin/Generators

rchincha: Also don't you want to put his in scrub/scrub.go instead of here?

examples/config-allextensions.json
```diff
@@ -41,6 +41,9 @@
 			"cve": {
 				"updateInterval": "2h"
 			}
+		},
+		"scrub": {
+			"scrubInterval": "2h"
```
rchincha, line 46: Best to be set at "24h"

examples/config-scrub.json
```diff
@@ -0,0 +1,26 @@
+{
+    "distSpecVersion":"1.0.1",
+    "storage": {
+        "rootDirectory": "/tmp/zot",
+        "scrub": {
+            "enable": true,
```
rchincha, line 6: Do we need the "enable" here? just having scrub and its interval should suffice imo

pkg/storage/scrub.go
```diff
@@ -53,25 +56,38 @@ func (sc StoreController) CheckAllBlobsIntegrity() (ScrubResults, error) {
...
+	}
+
+	return results, nil
+}
 
-			results.ScrubResults = append(results.ScrubResults, imageResults...)
+func CheckImageStoreBlobsIntegrity(imgStore ImageStore, log *log.Logger) ([]ScrubImageResult, error) {
+	results := []ScrubImageResult{}
+
+	images, err := imgStore.GetRepositories()
```
rchincha, line 73: Call this "repos" instead

pkg/storage/scrub.go
```diff
@@ -53,25 +56,38 @@ func (sc StoreController) CheckAllBlobsIntegrity() (ScrubResults, error) {
...
+	if err != nil {
+		return results, err
+	}
+
+	for _, repo := range images {
+		imageResults, err := checkImage(repo, imgStore, log)
```
rchincha, line 79: So for each repo, we are lock/unlock and checking. This looks correct. Call this "checkRepo()" instead here.

examples/config-scrub.json
Andreea-Lupu: Done! Now if scrub interval is set then scrub will be enabled. If "scrub" field is  specified but no value provided for "interval" then scrub will not be enabled.

rchincha (dismissed):
lgtm

rchincha (approved):
lgtm

Issue #138, opened 2020-09-14
Root access for skopeo in Makefile


IonutCraciun:
Hello guys,

Why are you using sudo for skopeo operations?
Example: https://github.com/project-zot/zot/blob/main/Makefile#L50

shimish2:
Sorry for late reply, I did not got any notification when this issue was created and didn't check the Issues tab for a while. 

This was needed to run skopeo inside travis ci build and without sudo it returns this error 

Error: error running newgidmap: exec: "newgidmap": executable file not found in $PATH: 
ERRO[0000] error running newgidmap: : exec: "newgidmap": executable file not found in $PATH 
ERRO[0000] (unable to determine exit status)

andaaron:
Seems to be an issue with skopeo installed by https://github.com/project-zot/zot/blob/main/.github/workflows/ci-cd.yml#L51.
let's replace with skopeo installed using apt.

andaaron:
Done

PR #444, opened 2022-03-05
Proof of concept for running blackbox tests using bats test framework

andaaron:
**What type of PR is this?**
build

Lately the amount of shell code present in the makefile or the GH workflows has been growing, and the main reason is we need to do some black box testing using separate binaries (notation, openssl, curl, maybe zb).
Ideally we should organize this code into tests with a consistent architecture.

POC for using the bats test framework for black box testing using shell commands.

### DIFF
--- a/test/blackbox/upload.bats
+++ b/test/blackbox/upload.bats
@@ -1,0 +1,45 @@
+load helpers
+
+function setup_file() {
+    # Download test data
+    skopeo --insecure-policy copy docker://public.ecr.aws/t0x7q1g8/centos:7 oci:${TEST_DATA_DIR}/zot-test:0.0.1
+    skopeo --insecure-policy copy docker://public.ecr.aws/t0x7q1g8/centos:8 oci:${TEST_DATA_DIR}/zot-cve-test:0.0.1
+    # Setup zot server
+    local zot_root_dir=${BATS_FILE_TMPDIR}/zot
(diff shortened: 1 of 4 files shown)


test/blackbox/upload.bats
```diff
@@ -0,0 +1,45 @@
+load helpers
+
+function setup_file() {
+    # Download test data
+    skopeo --insecure-policy copy docker://public.ecr.aws/t0x7q1g8/centos:7 oci:${TEST_DATA_DIR}/zot-test:0.0.1
```
andaaron, line 5: Note skopeo installed via snap seems to need sudo, because `newuidmap`/`newgidmap` is not found by snap - must be a packaging issue or snap container issue.
Looks like skopeo installed via apt works fine without sudo.

So, as a side note related to the other make targets and https://github.com/project-zot/zot/issues/138, if we use the apt installed skopeo we may be able to remove some of the sudo usage in the Makefile.

rchincha:
So the proposal is to move all the shell scripting out of the GH workflows (which calls make) and the Makefile and here?

andaaron:
> So the proposal is to move all the shell scripting out of the GH workflows (which calls make) and the Makefile and here?

More of less. You would still need to install/configure the tooling in GH workflows. Maybe some setup needing sudo in the makefile as well.

andaaron:
Closing in favor of https://github.com/project-zot/zot/pull/467

PR #460, opened 2022-03-15
fix dependabot alert

rchincha:
https://github.com/project-zot/zot/security/dependabot/10

**What type of PR is this?**

bug

Fix a dependabot alert triggered by distribution spec version in go.mod


shimish2 (approved):
lgtm

Issue #466, opened 2022-03-17
gc: make garbage-collect periodic


rchincha:
**Is your feature request related to a problem? Please describe.**

Currently, gc is inline in the data path. But the issue is that if there are no uploads happening, gc will not be triggered.

**Describe the solution you'd like**

Instead, make gc periodic similar to scrub if gc=true and use the same gc policy to ignore blobs that are newer than gcDelay.

**Describe alternatives you've considered**

**Additional context**

PR #476, opened 2022-03-21
gc: make garbage-collect periodic

Andreea-Lupu:

**What type of PR is this?**

#466

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/examples/config-gc-periodic.json
+++ b/examples/config-gc-periodic.json
@@ -1,0 +1,24 @@
+{
+    "distSpecVersion":"1.0.1",
+    "storage": {
+        "rootDirectory": "/tmp/zot",
+        "gc": true,
+        "gcDelay": "1h",
+        "gcInterval": "24h",
+        "subPaths": {
+            "/a": {
+                "rootDirectory": "/tmp/zot1",
+                "gc": true,
+                "gcDelay": "1h",
+                "gcInterval": "24h"
+            }
+        }
+    },
+    "http": {
+        "address": "127.0.0.1",
+        "port": "8080"
+    },
+    "log": {
+        "level": "debug"
+    }
+}
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -259,11 +259,6 @@ func (c *Controller) InitImageStore(reloadCtx context.Context) error {
 		}
 
 		c.StoreController.DefaultStore = defaultStore
-
-		// Enable extensions if extension config is provided
-		if c.Config != nil && c.Config.Extensions != nil {
-			ext.EnableExtensions(c.Config, c.Log, c.Config.Storage.RootDirectory)
-		}
 	} else {
 		// we can't proceed without global storage
 		c.Log.Error().Err(errors.ErrImgStoreNotFound).Msg("controller: no storage config provided")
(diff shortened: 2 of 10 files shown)


pkg/storage/storage_fs.go
```diff
@@ -1641,3 +1643,58 @@ func DirExists(d string) bool {
...
+	if providedGCInterval < DefaultGCInterval {
+		log.Warn().Msg("GC interval set to too-short interval < 2h, changing GC duration to 2 hours and continuing.") // nolint: lll
+
+		return DefaultGCInterval
+	}
+
+	return providedGCInterval
+}
+
+func (is *ImageStoreFS) gcAllRepos() error {
```
peusebiu, line 1657: make this a function like ifOlderThan() and give it the imageStore as an argument

peusebiu: otherwise LGTM

examples/config-gc-periodic.json
```diff
@@ -0,0 +1,24 @@
+{
+    "distSpecVersion":"1.0.1",
+    "storage": {
+        "rootDirectory": "/tmp/zot",
+        "gc": true,
+        "gcDelay": "1s",
```
rchincha, line 6: As an example make this more realistic, maybe `1h`, same below

pkg/storage/storage_fs.go
```diff
@@ -1641,3 +1643,58 @@ func DirExists(d string) bool {
 
 	return true
 }
+
+func GetValidGCInterval(providedGCInterval time.Duration, log *zlog.Logger) time.Duration {
```
rchincha, line 1647: This should move to config validation, pls see `LoadConfiguration()` api call.

pkg/api/controller.go
```diff
@@ -258,6 +258,13 @@ func (c *Controller) InitImageStore(reloadCtx context.Context) error {
 				c.Config.Storage.Commit, c.Log, c.Metrics, store)
 		}
 
+		// Enable running garbage-collect periodically
```
rchincha, line 261: Can I have the `gcInterval` disabled at rootDir level but enabled at subPath level? Then how does this change?

Andreea-Lupu: The logic for gc periodic is the same as the one for gc inline. So basicaly this can be enabled/disabled per ImageStore which can be the DefaultStore or one of the SubStores. gcInterval can be disabled at rootDir level(if in config file "gc": false or "gc": true, but no value specified for "gcInterval") and in the same time can be enabled at subPath level if "gc":true and the value for "gcInterval" > 0 for one of the SubStores under "subPaths" field.

pkg/api/controller.go
```diff
@@ -258,6 +258,13 @@ func (c *Controller) InitImageStore(reloadCtx context.Context) error {
...
+		if c.Config.Storage.GC && c.Config.Storage.GCInterval != 0 {
+			c.Config.Storage.GCInterval = storage.GetValidGCInterval(c.Config.Storage.GCInterval, &c.Log)
+
+			defaultStore.RunGCPeriodically(c.Config.Storage.GCInterval)
```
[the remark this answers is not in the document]
Andreea-Lupu: Ok. I added a StartBackgroundTasks() method for these goroutines and it is called at the end of InitImageStore() method.

rchincha (dismissed):
lgtm

rchincha (approved):
lgtm

PR #448, opened 2022-03-07
renamed config variable

laurentiuNiculae:
Signed-off-by: Laurentiu Niculae [niculae.laurentiu1@gmail.com](mailto:niculae.laurentiu1@gmail.com)
**What type of PR is this?**

**Which issue does this PR fix**:
[Issue #441 ](https://github.com/project-zot/zot/issues/441)
**What does this PR do / Why do we need it**:

**If an issue # is not available please add repro steps and logs from IPAMD/CNI showing the issue**:

**Testing done on this change**:

**Automation added to e2e**:

**Will this break upgrades or downgrades. Has updating a running cluster been tested?**:

**Does this change require updates to the CNI daemonset config files to work?**:

**Does this PR introduce any user-facing change?**:

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #488, opened 2022-03-30
"make verify-config" target reports issues

### LINKED ISSUE #486: "make verify-config" target reports issues
**Describe the bug**
Run "make verify-config" and observe that examples/config-example.yaml is modified

{"level":"warn","error":"config: invalid config","time":"2022-03-29T00:01:16+03:00","message":"config dist-spec version: 0.1.0-dev differs from version actually used: 1.0.1, will be corrected automatically"}
...
{"level":"warn","error":"config: invalid config","time":"2022-03-29T00:01:17+03:00","message":"garbage-collect delay specified without enabling garbage-collect, will be ignored"}

**To Reproduce**
make verify-config

**Expected behavior**
All examples files should be reported as valid

**Screenshots**

**Additional context**

laurentiuNiculae:
**What type of PR is this?**

[Issue #486 ](https://github.com/project-zot/zot/issues/486)
Closes #486 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/cli/root.go
+++ b/pkg/cli/root.go
@@ -354,22 +354,21 @@ func applyDefaultValues(config *config.Config, viperInstance *viper.Viper) {
 		}
 	}
 
-	if config.DistSpecVersion != distspec.Version {
-		log.Warn().Err(errors.ErrBadConfig).
-			Msgf("config dist-spec version: %s differs from version actually used: %s, will be corrected automatically",
-				config.DistSpecVersion, distspec.Version)
-
-		// rewrite the config file
-		viperInstance.Set("distSpecVersion", distspec.Version)
-
-		err := viperInstance.WriteConfig()
-		if err != nil {
-			log.Warn().Err(errors.ErrBadConfig).
-				Msg("can't rewrite the config file")
-		}
-
-		config.DistSpecVersion = distspec.Version
+	if !config.Storage.GC && viperInstance.Get("storage::gcdelay") == nil {
+		config.Storage.GCDelay = 0
 	}
+}
+
+func updateDistSpecVersion(config *config.Config) {
+	if config.DistSpecVersion == distspec.Version {
+		return
+	}
+
+	log.Warn().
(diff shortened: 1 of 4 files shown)


laurentiuNiculae:
I have added make target that check for warnings when verifying configurations. This will force us to push only valid configurations.

rchincha:
Why is `make verify-config` even changing files???
It is supposed to be a read-only operation.

laurentiuNiculae:
When the distribution spec specified in the config file is wrong it will try to correct it by rewriting it. I made that change following [this](https://github.com/project-zot/zot/pull/448#discussion_r823058417) conversation.

pkg/cli/root.go
```diff
@@ -354,22 +354,21 @@ func applyDefaultValues(config *config.Config, viperInstance *viper.Viper) {
 		}
 	}
 
-	if config.DistSpecVersion != distspec.Version {
-		log.Warn().Err(errors.ErrBadConfig).
-			Msgf("config dist-spec version: %s differs from version actually used: %s, will be corrected automatically",
-				config.DistSpecVersion, distspec.Version)
+	if !config.Storage.GC && viperInstance.Get("storage::gcdelay") == nil {
```
rchincha, line 357: `storage::gcdelay`
^ interesting that I can scope like so.

pkg/cli/root.go
```diff
@@ -354,22 +354,21 @@ func applyDefaultValues(config *config.Config, viperInstance *viper.Viper) {
...
+}
 
-		// rewrite the config file
-		viperInstance.Set("distSpecVersion", distspec.Version)
+func updateDistSpecVersion(config *config.Config) {
+	if config.DistSpecVersion == distspec.Version {
+		return
+	}
 
-		err := viperInstance.WriteConfig()
```
rchincha, line 365: We should never have committed this. `verify` cannot modify config.

rchincha (dismissed):
lgtm

rchincha (approved):
lgtm

Issue #480, opened 2022-03-25
sync should use one temp .sync dir per repo, not per image


peusebiu:
**Describe the bug**

Right now sync will create a temporary oci layout (.sync dir) for each image it needs to sync, thus for a repo with many tags it will copy the same blobs multiple times wasting network.

**Expected behavior**

sync should create an oci layout per repo, so that all its tags will be copied into the same oci layout.

Issue #479, opened 2022-03-25
sync should check if blobs are already pushed before uploading


peusebiu:
**Describe the bug**
Currently sync pulls an image into a temporary oci layout then it takes the blobs and manifests and push them into the local storage.

**Expected behavior**

sync should check if blobs exists with storage.CheckBlob() before pushing them.

**Additional context**

It should behave like skopeo copy (first HEAD to check for blob existence then PUT if the blob isn't uploaded yet)

PR #493, opened 2022-04-05
Reduce downloads by grouping .sync dirs for all images in the same repo

chofnar:
**What type of PR is this?**

enhancement

#480 #479

This PR checks the existence of the blob before it being uploaded to the local imageStore, and only does so if it's not present.
Also, When syncing a repo with multiple images/tags, sync would create a local directory for each image, downloading the same blobs over and over again. This PR makes sync use the same folder per sync session, to massively reduce redownloading.

### DIFF
--- a/pkg/extensions/sync/sync.go
+++ b/pkg/extensions/sync/sync.go
@@ -190,10 +190,12 @@ func filterImagesBySemver(upstreamReferences *[]types.ImageReference, content Co
 // imagesToCopyFromRepos lists all images given a registry name and its repos.
 func imagesToCopyFromUpstream(ctx context.Context, registryName string, repos []string,
 	upstreamCtx *types.SystemContext, content Content, log log.Logger,
-) ([]types.ImageReference, error) {
-	var upstreamReferences []types.ImageReference
+) (map[string][]types.ImageReference, error) {
+	upstreamReferences := make(map[string][]types.ImageReference)
 
 	for _, repoName := range repos {
+		repoUpstreamReferences := make([]types.ImageReference, 0)
+
 		repoRef, err := parseRepositoryReference(fmt.Sprintf("%s/%s", registryName, repoName))
 		if err != nil {
 			log.Error().Err(err).Msgf("couldn't parse repository reference: %s", repoRef)
@@ -230,22 +232,26 @@ func imagesToCopyFromUpstream(ctx context.Context, registryName string, repos []
 				return nil, err
 			}
 
-			upstreamReferences = append(upstreamReferences, ref)
+			repoUpstreamReferences = append(repoUpstreamReferences, ref)
 		}
+
+		upstreamReferences[repoName] = repoUpstreamReferences
+
+		log.Debug().Msgf("repo: %s - upstream refs to be copied: %v", repoName, upstreamReferences)
+
+		err = filterImagesByTagRegex(&repoUpstreamReferences, content, log)
+		if err != nil {
+			return map[string][]types.ImageReference{}, err
+		}
+
+		log.Debug().Msgf("repo: %s - remaining upstream refs to be copied: %v", repoName, repoUpstreamReferences)
+
+		filterImagesBySemver(&repoUpstreamReferences, content, log)
+
+		log.Debug().Msgf("repo: %s - remaining upstream refs to be copied: %v", repoName, repoUpstreamReferences)
+
+		upstreamReferences[repoName] = repoUpstreamReferences
 	}
-
-	log.Debug().Msgf("upstream refs to be copied: %v", upstreamReferences)
-
-	err := filterImagesByTagRegex(&upstreamReferences, content, log)
-	if err != nil {
-		return []types.ImageReference{}, err
-	}
-
-	log.Debug().Msgf("remaining upstream refs to be copied: %v", upstreamReferences)
-
-	filterImagesBySemver(&upstreamReferences, content, log)
-
-	log.Debug().Msgf("remaining upstream refs to be copied: %v", upstreamReferences)
 
 	return upstreamReferences, nil
 }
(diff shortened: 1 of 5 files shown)


pkg/extensions/sync/sync.go
```diff
@@ -319,186 +326,227 @@ func syncRegistry(ctx context.Context, regCfg RegistryConfig, upstreamURL string
...
+				return err
+			}, retryOptions); err != nil {
+				log.Error().Err(err).Msgf("error while copying image %s to %s",
+					upstreamImageRef.DockerReference(), localCachePath)
 
-			return err
+				return err
+			}
+
+			upstreamImgRefsByTag[tag] = upstreamImageRef
```
peusebiu, line 492: Now that I think about it, let's pushSyncedLocalImage in this for, no need to do it later in another loop.

The blobs will still be cached in the same temp dir.

This way you don't need to create upstreamImgRefsByTag map, cacheImageStore  and tagsToSync.

pkg/extensions/sync/sync.go
```diff
@@ -319,186 +326,227 @@ func syncRegistry(ctx context.Context, regCfg RegistryConfig, upstreamURL string
...
-				upstreamImageRef.DockerReference(), localCachePath)
+		tagsToSync, err := cacheImageStore.GetImageTags(localRepo)
+		if len(tagsToSync) > 0 && err != nil {
+			log.Error().Err(err).Msgf("Error getting the cacheImageStore")
 
 			return err
 		}
+		// sync tag by tag
+		for _, tag := range tagsToSync {
+			err = pushSyncedLocalImage(localRepo, tag, localCachePath, imageStore, log)
```
peusebiu, line 512: move all the logic from this 'for' above

pkg/extensions/sync/sync.go
```diff
@@ -319,186 +326,227 @@ func syncRegistry(ctx context.Context, regCfg RegistryConfig, upstreamURL string
...
+	for _, imageList := range reposWithContentID {
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
 		default:
 			break
 		}
 
-		upstreamImageRef := image.ref
+		remoteRepo := getRepoFromRef(imageList[0].ref, upstreamAddr)
```
peusebiu, line 374: no need for this, you already have it in reposWithContentID

just write it like: for remoteRepo, imageList := range reposWithContentID  {}

pkg/extensions/sync/utils.go
```diff
@@ -363,18 +355,33 @@ func getImageRef(registryDomain, repo, tag string) (types.ImageReference, error)
...
+		}
+	}
+	// local cache for repo
+	fileList, err := os.ReadDir(localRepoPath)
+	if err != nil {
 		return nil, "", err
 	}
+	// using the existing one (which must be a dir and only one)
+	if !fileList[0].IsDir() || len(fileList) > 1 {
+		panic("Something wrong with the existing repo dir: " + localRepoPath)
```
peusebiu, line 381: this will fail in some scenarios, if a periodically sync is running on image A and a on demand runs on the same image A with another tag it will fail because there will be more than one sync subdir.

there should be a better way to treat this, maybe splitting this function in two?

one for getting a localCachePath given a repo
one for getting a localImageRef given a localCachePath and a tag

peusebiu:
check to see if you can add a test for this logic :D

pkg/extensions/sync/sync.go
chofnar: Moved push logic inside loop

pkg/extensions/sync/utils.go
chofnar: Split getLocalImageRef function by adding a getLocalCachePath that generates a unique folder so parallel execution will not be affected

pkg/extensions/sync/utils.go
```diff
@@ -362,28 +360,52 @@ func getImageRef(registryDomain, repo, tag string) (types.ImageReference, error)
...
+	// hard to reach test case, injected error, see pkg/test/dev.go
+	if err := test.Error(err); err != nil {
+		return "", err
+	}
+
+	localCachePath := path.Join(localRepoPath, uuid.String())
+
+	cachedRepoPath := path.Join(localCachePath, repo)
+	if err = os.MkdirAll(cachedRepoPath, storage.DefaultDirPerms); err != nil {
+		return "", err
```
peusebiu, line 405: you can cover this by injecting an error like on line 397.

chofnar: I wrote a test that didn't need to use injection: no write perimssion on already existing sync folder

pkg/extensions/sync/sync.go
```diff
@@ -319,186 +326,203 @@ func syncRegistry(ctx context.Context, regCfg RegistryConfig, upstreamURL string
...
-		remoteRepo := getRepoFromRef(upstreamImageRef, upstreamAddr)
-		localRepo := getRepoDestination(remoteRepo, image.content)
-
-		upstreamImageDigest, err := docker.GetDigest(ctx, upstreamCtx, upstreamImageRef)
+		localCachePath, err := getLocalCachePath(imageStore, remoteRepoCopy)
 		if err != nil {
-			log.Error().Err(err).Msgf("couldn't get upstream image %s manifest", upstreamImageRef.DockerReference())
-
-			return err
+			log.Error().Err(err).Msgf("couldn't get localCachePath for %s", remoteRepoCopy)
```
peusebiu, line 379: why don't you return error here? same with on_demand.go

peusebiu: if there is no reason then defer os.RemoveAll() on localCachePath.

chofnar: > why don't you return error here? same with on_demand.go

I must have missed it. Fixed it now, added a return err there

> if there is no reason then defer os.RemoveAll() on localCachePath.

There was a defer os.RemoveAll() on localCachePath, but it was at the end of the big loop that downloads the image, which in hindsight, was not the right place to be anyway. I moved it after the return err above

peusebiu (approved):

pkg/extensions/sync/utils.go
```diff
@@ -45,12 +45,14 @@ func getTagFromRef(ref types.ImageReference, log log.Logger) reference.Tagged {
 }
 
 // getRepoFromRef returns repo name from a registry ImageReference.
+/* Not used anymore
```
rchincha, line 48: No! Remove any unused dead code pls. How come lint is not catching this.

chofnar: Oops! Must have forgot that there. Removed it.

rchincha (approved):
lgtm

PR #495, opened 2022-04-07
redirect blob download to s3 server

### LINKED ISSUE #494: s3: add a 'redirect' flag in storageDriver params and based on it serve blobs or redirect to S3 URLs
**Is your feature request related to a problem? Please describe.**
When using s3 as backend storage zot first downloads blobs from s3 and then serves them thus wasting time.
It should be able to use redirects to s3 urls based on a 'redirect' flag in storageDriver config parameters.

**Describe the solution you'd like**

If redirect flag is true then:
1) use path := storage.BlobPath() to get path to blob
2) use url := storageDriver.URLFor(path)
3) in routes use http.Redirect(url) and serve the URLs instead of blobs.

we can add a new func to Storage interface URLForPath(path) that calls storageDriver.URLFor(path)
for filesystem storage return a err.NotImplemented.

laurentiuNiculae:
**What type of PR is this?**

[Issue ](https://github.com/project-zot/zot/issues/494)#494
Closes #494 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -26,6 +26,7 @@ import (
 	jsoniter "github.com/json-iterator/go"
 	notreg "github.com/notaryproject/notation/pkg/registry"
 	"github.com/opencontainers/distribution-spec/specs-go/v1/extensions"
+	godigest "github.com/opencontainers/go-digest"
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
 	artifactspec "github.com/oras-project/artifacts-spec/specs-go/v1"
 	httpSwagger "github.com/swaggo/http-swagger"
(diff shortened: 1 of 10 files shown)


pkg/api/routes.go
```diff
@@ -621,6 +622,26 @@ func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Requ
...
+			http.StatusBadRequest,
+			NewErrorList(NewError(DIGEST_INVALID, map[string]string{"digest": digest})))
+	}
+
+	if rh.c.Config.Storage.StorageDriver != nil {
+		redirect, ok := rh.c.Config.Storage.StorageDriver["redirect"].(bool)
+
+		if ok && redirect {
+			path := imgStore.BlobPath(name, digestObj)
+			if path, err := imgStore.URLForPath(path); err == nil {
```
peusebiu, line 637: please log the error in case there is one :D, it could be. and return a http.StatusInternalServerError

laurentiuNiculae: I've addressed this comment. I'll pay more attention to these details :D

peusebiu (dismissed):
LGTM

andaaron:
- Need to check if clients handle http redirects on a common basis. Need to check if OCI dist spec mentions redirects.
- After we validate the above, we need a rebase

rchincha:
@laurentiuNiculae is this PR still valid, else we should close it.

KuxaBeast:
Is this still relevant? I am looking forward to this feature... Is there any reason why haven't this PR been merged yet?

rchincha:
> Is this still relevant? I am looking forward to this feature... Is there any reason why haven't this PR been merged yet?

@KuxaBeast, we plan to merge this PR in a future milestone. Meanwhile, could you give us more details about your use case?

KuxaBeast:
@rchincha, it's just a home-lab setup for now. I was thinking about making use of the Backblaze storage and Cloudflare, utilizing the Bandwidth Alliance, making the download costs essentially zero. But then I found that it might also be a good idea to offload downloads directly to Cloudflare, instead of bombing the Zot instance on my K8s cluster.

PR #497, opened 2022-04-08
update metrics/Dockerfile to match current binary name format

Andreea-Lupu:

**What type of PR is this?**

After changing the name format of the binary(ex zot-$OS-$ARCH) trying to build images using metrics/Dockerfile doesn't work anymore because of name mismatch. This PR fixes this issue by updating the name of the binary to current format.

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/examples/metrics/Dockerfile
+++ b/examples/metrics/Dockerfile
@@ -30,11 +30,11 @@ RUN echo '{\n\
 # ---
 # Stage 2: Final image with nothing but certs, binary, and default config file
 # ---
-FROM scratch AS final
+FROM gcr.io/distroless/base AS final
 COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
-COPY --from=builder /go/src/github.com/project-zot/zot/bin/zot /zot
+COPY --from=builder /go/src/github.com/project-zot/zot/bin/zot-linux-amd64 /usr/bin/zot
 COPY --from=builder /go/src/github.com/project-zot/zot/config.json /etc/zot/config.json
-ENTRYPOINT ["/zot"]
+ENTRYPOINT ["/usr/bin/zot"]
 EXPOSE 5000
 VOLUME ["/var/lib/registry"]
 CMD ["serve", "/etc/zot/config.json"]
--- a/examples/metrics/Dockerfile-zxp
+++ b/examples/metrics/Dockerfile-zxp
@@ -23,9 +23,9 @@ RUN echo '{\n\
 # ---
 # Stage 2: Final image with nothing but binary and default config file
 # ---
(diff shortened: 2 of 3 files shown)


examples/metrics/Dockerfile
```diff
@@ -30,11 +32,13 @@ RUN echo '{\n\
 # ---
 # Stage 2: Final image with nothing but certs, binary, and default config file
 # ---
-FROM scratch AS final
+FROM gcr.io/distroless/base AS final
+ARG OS
+ARG ARCH
 COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
-COPY --from=builder /go/src/github.com/project-zot/zot/bin/zot /zot
```
adodon2go, line 35: I think what we need to do is to keep this file similar to Dockerfile-conformance from which this file was originally inspired from (maybe rename Dockerfile -> Dockerfile-conformance ?)

I'm thinking of changing this line to:
`COPY --from=builder /go/src/github.com/project-zot/zot/bin/zot-linux-amd64 /usr/bin/zot`
There ios no need for OS & ARCH args as the kind-setup.sh script was intended to be run on linux systems and was not tested on other platforms.

examples/metrics/Dockerfile-zxp
```diff
@@ -23,9 +25,11 @@ RUN echo '{\n\
 # ---
 # Stage 2: Final image with nothing but binary and default config file
 # ---
-FROM scratch AS final
-COPY --from=builder /go/src/github.com/project-zot/zot/bin/zxp /zxp
+FROM gcr.io/distroless/base AS final
+ARG OS
+ARG ARCH
+COPY --from=builder /go/src/github.com/project-zot/zot/bin/zxp-$OS-$ARCH /usr/bin/zxp
```
adodon2go, line 31: Better to use zxp-linux-amd64 directly and get rid of OS & ARCH args

examples/metrics/Makefile
```diff
@@ -1,8 +1,10 @@
 CONTAINER_RUNTIME := docker
+OS ?= linux
```
adodon2go, line 2: I don't think we really need to change this file.

examples/metrics/Dockerfile-zxp
rchincha: No why? I can reuse the same Dockerfile and it gets OS and ARCH from ci/cd

adodon2go: The idea was that kind-setup.sh script was never tested on other platforms to prove it works.
If the idea was to bring a kind cluster with all the zot pods installed (+ prometheus) than we need to validate that script works on different flavours of Mac processors too.

PR #500, opened 2022-04-13
Adding mocked tests for routes

### LINKED ISSUE #492: Increase code coverage of pkg/api/routes.go
Currently we have little unit tests in pkg/api/routes.go, it is mostly covered by integration tests. The integration tests may or may not hit some of the corner cases reliably, which results in different code coverage results when running the same tests on the same code.

We should implement more detailed tests to reproduce more of the errors which should be handled. Perhaps we could solve this by adding new tests which mock the storage, or other components we interact with in routes.go.

laurentiuNiculae:
**What type of PR is this?**

Issue #492
Closes #492
Raise cod coverage for routes.go

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -955,14 +955,6 @@ func (rh *RouteHandler) PatchBlobUpload(response http.ResponseWriter, request *h
 			return
 		}
 
-		contentRange := request.Header.Get("Content-Range")
-		if contentRange == "" {
-			rh.c.Log.Warn().Str("actual", request.Header.Get("Content-Range")).Msg("invalid content range")
-			response.WriteHeader(http.StatusRequestedRangeNotSatisfiable)
-
-			return
-		}
-
 		var from, to int64
 		if from, to, err = getContentRange(request); err != nil || (to-from)+1 != contentLength {
 			response.WriteHeader(http.StatusRequestedRangeNotSatisfiable)
--- a/pkg/api/routes_test.go
+++ b/pkg/api/routes_test.go
@@ -1,0 +1,1642 @@
+//go:build extended
+// +build extended
+
+package api_test
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/gorilla/mux"
+	"github.com/opencontainers/go-digest"
+	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+	artifactspec "github.com/oras-project/artifacts-spec/specs-go/v1"
+	. "github.com/smartystreets/goconvey/convey"
+	zerr "zotregistry.io/zot/errors"
+	"zotregistry.io/zot/pkg/api"
+	"zotregistry.io/zot/pkg/api/config"
+	"zotregistry.io/zot/pkg/api/constants"
+	"zotregistry.io/zot/pkg/storage"
+	"zotregistry.io/zot/pkg/test"
+)
+
+var ErrUnexpectedError = errors.New("error: unexpected error")
+
+type MockedImageStore struct {
+	dirExistsFn            func(d string) bool
+	rootDirFn              func() string
+	initRepoFn             func(name string) error
+	validateRepoFn         func(name string) (bool, error)
+	getRepositoriesFn      func() ([]string, error)
+	getImageTagsFn         func(repo string) ([]string, error)
+	getImageManifestFn     func(repo string, reference string) ([]byte, string, string, error)
+	putImageManifestFn     func(repo string, reference string, mediaType string, body []byte) (string, error)
+	deleteImageManifestFn  func(repo string, reference string) error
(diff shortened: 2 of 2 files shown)


pkg/api/routes_test.go
```diff
@@ -0,0 +1,841 @@
...
+	artifactspec "github.com/oras-project/artifacts-spec/specs-go/v1"
+	. "github.com/smartystreets/goconvey/convey"
+	zerr "zotregistry.io/zot/errors"
+	"zotregistry.io/zot/pkg/api"
+	"zotregistry.io/zot/pkg/api/config"
+	"zotregistry.io/zot/pkg/api/constants"
+	"zotregistry.io/zot/pkg/test"
+)
+
+type ImageStoreMock struct {
```
rchincha, line 27: call this "MockedImageStore" instead.

errors/errors.go
```diff
@@ -53,4 +53,5 @@ var (
 	ErrRegistryNoContent       = errors.New("sync: could not find a Content that matches localRepo")
 	ErrSyncSignatureNotFound   = errors.New("sync: couldn't find any upstream notary/cosign signatures")
 	ErrSyncSignature           = errors.New("sync: couldn't get upstream notary/cosign signatures")
+	ErrUnexpectedError         = errors.New("error: unexpected error")
```
rchincha, line 56: Where is this actually used?

pkg/api/routes.go
```diff
@@ -949,14 +949,6 @@ func (rh *RouteHandler) PatchBlobUpload(response http.ResponseWriter, request *h
 			return
 		}
 
-		contentRange := request.Header.Get("Content-Range")
```
rchincha, line 952: This part got deleted? why?

laurentiuNiculae: Because on line 937 we have this condition
`if request.Header.Get("Content-Length") == "" || request.Header.Get("Content-Range") == ""`

since the removed code is on the else branch, testing for Content-Range to be empty will always result false, so it's unreachable.

errors/errors.go
laurentiuNiculae: I use it just in tests to be more expressive on the meaning of the error I test for.

pkg/api/routes.go
peusebiu: maybe there should be an AND instead of OR?

laurentiuNiculae: I have tested this change and it breaks Zot.
`Summarizing 2 Failures:

[Fail] OCI Distribution Conformance Tests Push Blob Upload Streamed [It] PATCH request with blob in body should yield 202 response 
/go/src/github.com/opencontainers/distribution-spec/conformance/02_push_test.go:35

[Panic!] OCI Distribution Conformance Tests Push Blob Upload Streamed [It] PUT request to session URL with digest should yield 201 response 
/usr/local/go/src/runtime/panic.go:199
`

laurentiuNiculae: I think the initial logic is good if it's compliant with the OCI tests. They seem quite exhaustive. I propose to just remove the redundant `if`.

errors/errors.go
rchincha: Then move this error into the tests itself. It doesn't belong here.

pkg/api/routes.go
rchincha: https://github.com/project-zot/zot/pull/527

Refactored like so ^. This is what this check really means.

rchincha: https://github.com/project-zot/zot/runs/6375017852?check_suite_focus=true
and conformance passes ^

errors/errors.go
rchincha: maybe create a new errors/test_errors.go and move it there.

pkg/api/routes.go
rchincha: https://pkg.go.dev/net/http#Header.Set
^ Is it possible to set the key but empty value in this case?

laurentiuNiculae: I am sorry but I don't understand what you mean by the last comment. I have written a small (but exhaustive) equivalent example to show that the line is unreachable.

```
package main

import (
	"fmt"
)

func foo(contentIsPresent, rangeIsPresent bool) {
	if !contentIsPresent || !rangeIsPresent { // equivalent to line 943
		fmt.Println("PutBlobChunkStreamed")
	} else {
		if !rangeIsPresent {                  // equivalent to line 959
			fmt.Println("Unreachable line")   
		}
	}
}

func main() {
	truthValues := [2]bool{true, false}

	for _, i := range truthValues {
		for _, j := range truthValues {
			foo(i, j)
		}
	}
}
```

errors/errors.go
laurentiuNiculae: Ok! I have resolved this issue.

peusebiu (approved):
LGTM

rchincha (dismissed):
lgtm

rchincha (dismissed):
lgtm

rchincha (approved):
lgtm

PR #515, opened 2022-04-26
Make trivy external

laurentiuNiculae:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


laurentiuNiculae:
The plugin code for scanning is located [here](https://github.com/laurentiuNiculae/zot-scan-plugin).

examples/plugins/joinargs-plugin.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "JoinArgsPlugin",
+    "integrationPoints": [
+        {
+            "interface": "CLICommand",
```
rchincha, line 5: This I would consider as one option, where a rpc plugin standardizes the output around an arbitary locally invoked binary

rchincha:
https://www.howtogeek.com/devops/how-to-use-trivy-to-find-vulnerabilities-in-docker-containers/

Pls look for this section ...

```
Server Mode

Trivy can run in client-server mode. This moves the vulnerability database to the server, so clients don’t need to maintain their own versions.
```

rchincha:
Here is another service that can run remotely.

https://docs.clamav.net/manual/Usage/Services.html

laurentiuNiculae:
I have been inspired by [this](https://github.com/goharbor/community/blob/main/proposals/pluggable-image-vulnerability-scanning_proposal.md#proposal) proposal from goharbor which is exhaustive. There are also some very good ideas [here](https://docs.docker.com/engine/extend/plugin_api/#api-design) about plugin discovery. My current [zot-scan](https://github.com/laurentiuNiculae/zot-scan-plugin) proposal is very similar to the official [Trivy scanner](https://github.com/aquasecurity/harbor-scanner-trivy) plugin for goharbor which is short it uses the Zot API to scan remotely the images. I don't think the client-server model from Trivy is all that relevant in our case since half of the scanning process is done by the client and this is not possible in our general, plugin oriented design.

PR #519, opened 2022-05-04
Build extensions

alexstan12:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


alexstan12:
It looks like the init() function does indeed impact the binary size. I experimented as following, which proved the previous statement:
- included all extension functions in the source code with no init / empty init functon  -> same binary size for cli as the one resulted from main branch
- moved extension functions one by one in the init function -> binary size of cli increased with each one of them

It looks like init() is dragging some extra weight with it, I'll yet have to see if there is any workaround or if we should ditch init() and try some other mechanism for managing build extensions.

alexstan12:
When comparing main/reflect/init the following resulted:
- the Get Catalog test seems to be the most unstable test, such that even multiple runs of the same benchmark have very different results. Given that, I based my conclusion mainly on the other tests (which also vary in results but not as much).
main >= init >= reflect ( reflect and init being close to each other)
Also main is behaving worse in some tests than the other 2 ( ex main vs init , init behaves better in GetCatalog and Push Chunk Streamed 10MB etc.)

**Overall the results vary too much for me to choose an optimum solution.**

All benchmarks ran on my local machine.
Here are the results:

- https://pastebin.com/8yHpAEZs bench-local-init
- https://pastebin.com/Qr2kpBaD bench-local-init2 (second run)
- https://pastebin.com/kBejjNpi bench-local-main
- https://pastebin.com/B3Lkk4Df bench-local-main2 (second run)
- https://pastebin.com/9YJbgz5Z bench-local-reflect
- https://pastebin.com/nK7Z8Mp6 bench-local-reflect2 (second run)


- init = usage of init() function to implement extensions
- main = current state of main branch
- reflect = usage of reflect package in order to detect at runtime which extensions have been included in the binary

alexstan12:
Came up with a solution that uses only build tags, so it won't have performance decrease or unexpected binary size.
Referenced in new draft PR https://github.com/project-zot/zot/pull/532

Issue #507, opened 2022-04-15
label-based github workflow actions


rchincha:
**Is your feature request related to a problem? Please describe.**

Our ci/cd workflows are increasing in number, complexity and resource usage. We may need to be thrifty.

**Describe the solution you'd like**

One way is to use label-based actions.
https://github.com/marketplace/actions/label-actions
Be selective based on labels.

andaaron:
No longer relevant.

PR #526, opened 2022-05-10
Added sample for label-specific Github Actions

chofnar:
**What type of PR is this?**

enhancement

#507 

This PR contains a sample Github Actions workflow that only runs when a specific label is present

### DIFF
--- a/.github/workflows/final.yml
+++ b/.github/workflows/final.yml
@@ -1,30 +1,25 @@
 on:
-  push:
-    branches:
-      - main
   pull_request:
     branches:
       - main
+    types: [ labeled, synchronize ]
+  push:
+    branches:
+      - main
+    types: [ labeled, synchronize ]
   release:
     types:
       - published
-name: build-test
 
+name: complete-check
 
 permissions:
   contents: read
   packages: write
 
 jobs:
-  build-test:
-    name: Build and test ZOT
-    runs-on: ubuntu-latest
-    steps:
-      - name: Run build and test
-        timeout-minutes: 60
-        run: |
-            echo "job deprecated"
-  build-test-arch:
+  complete-check:
+    if: contains(github.event.pull_request.labels.*.name, 'final_stage')
     name: Build and test ZOT
     runs-on: ubuntu-latest
     strategy:
@@ -79,8 +74,7 @@ jobs:
           echo "Building for $OS:$ARCH"
           cd $GITHUB_WORKSPACE
           if [[ $OS == "linux" && $ARCH == "amd64" ]]; then
-            make OS=$OS ARCH=$ARCH
-            sudo env "PATH=$PATH" make privileged-test
+            make OS=$OS ARCH=$ARCH complete-check
           else
             make OS=$OS ARCH=$ARCH binary binary-minimal binary-debug cli bench exporter-minimal
           fi
@@ -90,8 +84,6 @@ jobs:
           AWS_SECRET_ACCESS_KEY: fake
           OS: ${{ matrix.os }}
           ARCH: ${{ matrix.arch }}
-      - name: Upload code coverage
-        uses: codecov/codecov-action@v3
       - if: github.event_name == 'release' && github.event.action == 'published'
         name: Publish artifacts on releases
         uses: svenstaro/upload-release-action@v2
--- a/.github/workflows/minimal.yml
+++ b/.github/workflows/minimal.yml
@@ -1,0 +1,64 @@
+on:
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
+name: minimal-check
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  minimal-check:
+    name: Build and test ZOT
+    runs-on: ubuntu-latest
+    services:
+      s3mock:
+        image: ghcr.io/project-zot/localstack/localstack:0.13.2
+        env:
+          SERVICES: s3
+        ports:
+          - 4563-4599:4563-4599
+          - 9090:8080
+    steps:
+      - name: Install go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.18.x
+      - name: Check out source code
+        uses: actions/checkout@v3
+      - name: Install dependencies
+        run: |
+          cd $GITHUB_WORKSPACE
+          go install github.com/swaggo/swag/cmd/swag@latest
+          go mod download
+          sudo apt-get update
+          sudo apt-get install rpm
+          sudo apt-get install snapd
+          sudo apt-get install libgpgme-dev libassuan-dev libbtrfs-dev libdevmapper-dev pkg-config
+          git clone https://github.com/containers/skopeo -b v1.6.1 $GITHUB_WORKSPACE/src/github.com/containers/skopeo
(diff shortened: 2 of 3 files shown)


chofnar:
It seems that label-specific workflows as I used them in this [commit ](https://github.com/project-zot/zot/commit/777d15092258aa33d7214a5323d1ec5a1e577d57) only run when the label is added. I will look into the possibility of adding label-specific execution paths to the existing workflows.

peusebiu (approved):

peusebiu (approved):

Makefile
```diff
@@ -20,6 +20,12 @@ BENCH_OUTPUT ?= stdout
 .PHONY: all
 all: modcheck swagger binary binary-minimal binary-debug cli bench exporter-minimal verify-config test covhtml check
 
+.PHONY: minimal
```
alexstan12, line 23: Can we give this another name, for it to be easily distinguishable from minimal binaries build ?

.github/workflows/final.yml
```diff
@@ -1,24 +1,12 @@
 on:
-  push:
-    branches:
-      - main
-  pull_request:
-    branches:
-      - main
-  release:
```
andaaron, line 8: How does the release fit into this new refactoring?

Makefile
alexstan12: Maybe minimal-conformance-check, since we check if it is meeting criteria in " check test covhtml " .

andaaron (commented):
These are good ideas, but we need to decided what gets into what job:

1. Where do we keep the release logic?
2. How do we trigger the logic current in 'final', I think adding a label would be better than waiting for someone to approved.
3. Most of the time is spent by running the tests themselves in a single make target. And this PR doesn't really solve that issue.

.github/workflows/enhancement.yml
```diff
@@ -0,0 +1,18 @@
...
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+    types: [ labeled ]
+jobs:
+  enhancement-specific-stuff:
+    if: ${{ github.event.label.name == 'enhancement' }}
```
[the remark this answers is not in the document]
chofnar: Yes, that is correct - they are incompatible. This particular workflow would only run the moment a PR is labeled.

Removed the "on push" trigger to better reflect the use case (as it was misleading anyway).

.github/workflows/minimal.yml
```diff
@@ -0,0 +1,61 @@
+on:
```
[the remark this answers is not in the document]
chofnar: Added

Makefile
chofnar: Renamed as suggested

Makefile
```diff
@@ -20,6 +20,12 @@ BENCH_OUTPUT ?= stdout
...
+minimal: check test covhtml
+
+.PHONY: final
```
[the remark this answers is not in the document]
chofnar: Renamed as suggested

.github/workflows/minimal.yml
```diff
@@ -0,0 +1,61 @@
...
+    branches:
+      - main
+
+name: minimal-test
+jobs:
+  minimal-test:
+    name: Build and test ZOT - minimal edition
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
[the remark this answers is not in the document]
chofnar: Removed the matrix and set the env vars to only linux and amd64

.github/workflows/final.yml
chofnar: Mistake on my part, I had removed the "on release" trigger but left the release part of the workflow in the file. I re-added the trigger

.github/workflows/final.yml
```diff
@@ -1,24 +1,15 @@
...
-      - main
+  pull_request_review:
+    types:
+      - submitted
   release:
     types:
       - published
-name: build-test
+
+name: complete-conformance-check
```
andaaron, line 9: I don't know if conformance is the right word, as that sounds too close to the dist spec conformance tests.

.github/workflows/final.yml
```diff
@@ -1,24 +1,15 @@
...
 jobs:
-  build-test:
```
andaaron, line 13: I don't remember why se still needed this after the refactoring, @rchincha ?

.github/workflows/final.yml
andaaron: And my previous comment still stands, the person who opened the reviw will have to wait for someone else to review before we run the final check. Can we do it using labels instead?

.github/workflows/enhancement.yml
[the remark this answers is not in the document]
chofnar: I just tested, and it works with `github.event.label...`
I reckon it's because `pull_request` is already a type of `event`.
A similar thing is in the official documentation, but for issues:
https://docs.github.com/en/actions/managing-issues-and-pull-requests/commenting-on-an-issue-when-a-label-is-added

.github/workflows/final.yml
```diff
@@ -1,24 +1,12 @@
...
-    branches:
-      - main
-  release:
+  pull_request_review:
     types:
-      - published
-name: build-test
+      - submitted
+
+name: reviewer-approved-checks
```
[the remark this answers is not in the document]
chofnar: Answered in another comment:

Changed the behavior - the complete-check workflow will run on any activity on the PR if it is labeled with the "final_stage" label

andaaron (commented):
Looks like code coverage dropped. Why is that?

andaaron:
Pending items:
- [ ] investigate missing coverage
- [ ] add mandatory check for - minimal-check / Build and test ZOT - minimal edition (pull_request)
- [ ] add mandatory check for - complete-check / Build and test ZOT (linux, amd64) (pull_request)
- [ ] remove mandatory check for - Build and test ZOT
@rchincha please confirm the check related changes.

.github/workflows/minimal-check.yml
```diff
@@ -0,0 +1,72 @@
...
+    branches:
+      - main
+name: minimal-check
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  build-test:
```
andaaron, line 15: This job is in 2 files?

Makefile
```diff
@@ -24,6 +24,12 @@ extended-name:=
 .PHONY: all
 all: modcheck swagger binary binary-minimal binary-debug cli bench exporter-minimal verify-config test covhtml check
 
+.PHONY: minimal-check
+minimal: check test covhtml
```
andaaron, line 28:
```suggestion
minimal-check: check test covhtml
```

Makefile
```diff
@@ -24,6 +24,12 @@ extended-name:=
...
+
+.PHONY: complete-check
+final: modcheck swagger binary binary-minimal binary-debug cli bench exporter-minimal verify-config
```
andaaron, line 31:
```suggestion
complete-check: modcheck swagger binary binary-minimal binary-debug cli bench exporter-minimal verify-config
```

andaaron (dismissed):
Looks good from my side.

@project-zot/contributors please review as this will change our PR workflows.

andaaron:
Let's close this for now.

PR #529, opened 2022-05-11
Fix periodic background tasks (gc and scrub)

Andreea-Lupu:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -20,6 +20,7 @@ import (
 	"zotregistry.io/zot/errors"
 	"zotregistry.io/zot/pkg/api/config"
 	ext "zotregistry.io/zot/pkg/extensions"
+	extconf "zotregistry.io/zot/pkg/extensions/config"
 	"zotregistry.io/zot/pkg/extensions/monitoring"
 	"zotregistry.io/zot/pkg/log"
 	"zotregistry.io/zot/pkg/storage"
@@ -350,23 +351,13 @@ func (c *Controller) Shutdown() {
 }
 
 func (c *Controller) StartBackgroundTasks(reloadCtx context.Context) {
-	// Enable running garbage-collect periodically for DefaultStore
-	if c.Config.Storage.GC && c.Config.Storage.GCInterval != 0 {
-		c.StoreController.DefaultStore.RunGCPeriodically(c.Config.Storage.GCInterval)
-	}
-
 	// Enable extensions if extension config is provided for DefaultStore
 	if c.Config != nil && c.Config.Extensions != nil {
 		ext.EnableExtensions(c.Config, c.Log, c.Config.Storage.RootDirectory)
 	}
 
 	if c.Config.Storage.SubPaths != nil {
-		for route, storageConfig := range c.Config.Storage.SubPaths {
-			// Enable running garbage-collect periodically for subImageStore
-			if storageConfig.GC && storageConfig.GCInterval != 0 {
-				c.StoreController.SubStore[route].RunGCPeriodically(storageConfig.GCInterval)
-			}
-
+		for _, storageConfig := range c.Config.Storage.SubPaths {
 			// Enable extensions if extension config is provided for subImageStore
 			if c.Config != nil && c.Config.Extensions != nil {
 				ext.EnableExtensions(c.Config, c.Log, storageConfig.RootDirectory)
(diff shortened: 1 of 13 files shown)


pkg/api/controller.go
```diff
@@ -373,6 +364,117 @@ func (c *Controller) StartBackgroundTasks(reloadCtx context.Context) {
 	}
 
 	if c.Config.Extensions != nil {
-		ext.EnableScrubExtension(c.Config, c.StoreController, c.Log)
+		ext.EnableScrubExtension(c.Config, c.Log, false, nil, "")
+	}
+
+	c.StartPeriodicTasks()
```
rchincha, line 370: There should only be one `go c.StartPeriodicTasks()` here

pkg/api/controller.go
```diff
@@ -373,6 +364,117 @@ func (c *Controller) StartBackgroundTasks(reloadCtx context.Context) {
...
+		gcInterval = configGCInterval
+		gc = true
+	}
+
+	if extensions != nil && extensions.Scrub != nil && extensions.Scrub.Interval != 0 {
+		scrubInterval = extensions.Scrub.Interval
+		scrub = true
+	}
+
+	switch {
```
rchincha, line 408: This looks should be a routine called func minPeriodicInterval() instead of this unrolled switch

pkg/api/controller.go
```diff
@@ -373,6 +364,117 @@ func (c *Controller) StartBackgroundTasks(reloadCtx context.Context) {
...
-		ext.EnableScrubExtension(c.Config, c.StoreController, c.Log)
+		ext.EnableScrubExtension(c.Config, c.Log, false, nil, "")
+	}
+
+	c.StartPeriodicTasks()
+}
+
+func (c *Controller) StartPeriodicTasks() {
+	// start periodic gc and/or scrub for DefaultStore
+	go StartPeriodicTasksForImageStore(c.StoreController.DefaultStore, c.Config.Storage.GC, c.Config.Storage.GCInterval,
```
rchincha, line 375: Remove this go() routine.

pkg/api/controller.go
```diff
@@ -373,6 +364,117 @@ func (c *Controller) StartBackgroundTasks(reloadCtx context.Context) {
...
+		c.Config.Extensions, c.Log)
+
+	if c.Config.Storage.SubPaths != nil {
+		for route, storageConfig := range c.Config.Storage.SubPaths {
+			// Enable running garbage-collect or/and scrub periodically for subImageStore
+			go StartPeriodicTasksForImageStore(c.StoreController.SubStore[route], storageConfig.GC, storageConfig.GCInterval,
```
rchincha, line 381: Same here

pkg/api/controller.go
```diff
@@ -373,6 +364,117 @@ func (c *Controller) StartBackgroundTasks(reloadCtx context.Context) {
...
+}
+
+func RunGCScrubPeriodically(imgStore storage.ImageStore, gc, scrub bool, log log.Logger) {
+	repos, err := imgStore.GetRepositories()
+	if err != nil {
+		log.Error().Err(err).Msg(fmt.Sprintf("error while running background task for %s", imgStore.RootDir()))
+	}
+
+	for _, repo := range repos {
+		if gc {
```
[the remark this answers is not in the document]
Andreea-Lupu: I already check it in for loop from StartPeriodicTasksForImageStore after each interval
`if !lastGC.IsZero() && time.Since(lastGC) >= gcInterval {
	gc = true
}`

pkg/api/controller.go
```diff
@@ -373,6 +364,117 @@ func (c *Controller) StartBackgroundTasks(reloadCtx context.Context) {
...
+	case scrub && gc && gcInterval < scrubInterval:
+		interval = gcInterval
+	default:
+		return
+	}
+
+	var lastGC, lastScrub time.Time
+
+	for {
+		log.Info().Msg(fmt.Sprintf("Starting periodic background tasks(gc and/or scrub) for %s", imageStore.RootDir()))
```
rchincha, line 424: no need to mention gc/scrub here

pkg/api/controller.go
```diff
@@ -373,6 +364,117 @@ func (c *Controller) StartBackgroundTasks(reloadCtx context.Context) {
...
+
+		// Enable running garbage-collect or/and scrub periodically for imageStore
+		RunGCScrubPeriodically(imageStore, gc, scrub, log)
+
+		log.Info().Msg(fmt.Sprintf("Finishing periodic background tasks(gc and/or scrub) for %s", imageStore.RootDir()))
```
rchincha, line 429: Same here!

pkg/api/controller.go
```diff
@@ -373,6 +364,117 @@ func (c *Controller) StartBackgroundTasks(reloadCtx context.Context) {
...
+			gc = true
+		}
+
+		if !lastScrub.IsZero() && time.Since(lastScrub) >= scrubInterval {
+			scrub = true
+		}
+	}
+}
+
+func RunGCScrubPeriodically(imgStore storage.ImageStore, gc, scrub bool, log log.Logger) {
```
rchincha, line 451: Call this RunBackgroundTasks()

peusebiu:
Wouldn't be more clear and easier to treat the two background tasks (scrub and gc) independently? 
Instead of RunBackgroundTasks RunGc and RunScrub and get rid of "if scrub", "if gc", "gc = true", "scrub = true", and each of them taking care of their time intervals and the need to run again?

Just asking, maybe I didn't fully grasp the logic :D

rchincha:
> Wouldn't be more clear and easier to treat the two background tasks (scrub and gc) independently? Instead of RunBackgroundTasks, have a RunGc and RunScrub and get rid of "if scrub", "if gc", "gc = true", "scrub = true", and each of them taking care of their time intervals and the need to run again?
> 
> Just asking, maybe I didn't fully grasp the logic :D

Not about clarity. At runtime, we cannot ... **CANNOT** ... become unavailable. So this gives us more control about how locks are held/released etc.

pkg/api/controller.go
```diff
@@ -382,6 +373,131 @@ func (c *Controller) StartBackgroundTasks(reloadCtx context.Context) {
...
+
+	if scrub {
+		return scrubInterval
+	}
+
+	if gc {
+		return gcInterval
+	}
+
+	return time.Duration(0)
```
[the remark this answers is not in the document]
Andreea-Lupu: Yes, if GC and/or scrub is enabled but intervals are set to 0 then this feature is disabled.

rchincha (approved):
lgtm

Issue #554, opened 2022-05-25
zli: cve scan doesn't print CRITICAL vulnerabilities for image


laurentiuNiculae:
**Describe the bug**
When we try to scan an image using `cve` command from the zli, CRITICAL severity CVEs are not printed to the console.

**To Reproduce**
1. Open Zot with `config-cve.json`
2. Upload wdhif/docker-eicar image to zot with skopeo:
`skopeo copy -f oci --dest-tls-verify=false docker://wdhif/docker-eicar docker://localhost:8080/eicar:latest`
3.  Scan the image:
`./bin/zli-linux-amd64 cve --url=http://localhost:8080 -I eicar:latest`

**Expected behavior**
This image should have 4 CVEs with CRITICAL severity but they are not printed to the console.

**Context**
The function `groupCVEsBySeverity` only checks for 3 types of the 5 possible severities. More cases are needed.

andaaron:
Fixed in https://github.com/project-zot/zot/pull/556

PR #556, opened 2022-05-25
zli: cve scan doesn't print CRITICAL vulnerabilities for image

laurentiuNiculae:
**What type of PR is this?**

#554

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/cli/service.go
+++ b/pkg/cli/service.go
@@ -367,12 +367,20 @@ func (service searchService) getCveByImage(ctx context.Context, config searchCon
 }
 
 func groupCVEsBySeverity(cveList []cve) []cve {
-	high := make([]cve, 0)
-	med := make([]cve, 0)
-	low := make([]cve, 0)
+	var (
+		unknown  = make([]cve, 0)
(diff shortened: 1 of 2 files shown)


rchincha:
@laurentiuNiculae lgtm, can you make this a proper PR?

pkg/cli/service.go
```diff
@@ -367,12 +367,23 @@ func (service searchService) getCveByImage(ctx context.Context, config searchCon
...
+		unknown  = make([]cve, 0)
+		none     = make([]cve, 0)
+		high     = make([]cve, 0)
+		med      = make([]cve, 0)
+		low      = make([]cve, 0)
+		critical = make([]cve, 0)
+	)
 
 	for _, cve := range cveList {
 		switch cve.Severity {
```
andaaron, line 380: Still thinking about the fact we found this bug pretty late.
Do you think it would make sense to add a default, in case they add more severities we don't handle today? At least we would see the issues in the cli output one way or another.

rchincha (approved):
lgtm

Issue #564, opened 2022-05-26
storage: root directory directly on cloud storage


rchincha:
**Is your feature request related to a problem? Please describe.**

If we want to support a stateless zot (helps for clustering) but also support dedupe, then we cannot have anything on instance-local private storage. Even the dedupe cache should be on shared cloud storage.

- [x] #667 
- [ ] #953

rchincha:
This need investigation/evaluation first.

rchincha:
@chofnar Pls. take @peusebiu's help since he is the feature owner of this area.

chofnar:
- [ ] #667 
- [ ] https://github.com/chofnar/zot/pull/18

andaaron:
In review.

PR #588, opened 2022-06-14
Add database interface for blob paths + DynamoDB integration + refactor BoltDB to use interface

chofnar:
**What type of PR is this?**

enhancement

#564 

This PR adds a database interface that allows the usage of vendor-specific database services by implementing their drivers.
The usage of BoltDB was refactored to use this new interface.
Added AWS DynamoDB driver.

If a client uses DynamoDB and decides to downgrade, they must make sure that the database is migrated to BoltDB to avoid data loss.

Adds the possibility to use AWS DynamoDB instead of BoltDB

```release-note

```

**Problems**
Right now, there are several problems with this PR:
- database_factory.go uses global variable driverFactories to keep track of the available drivers (gochecknoglobals)
- new entries to that variable are added with the Drivers' init() functions (gochecknoinits)
- need to specify protocol in front of dynamodb endpoint [(known issue)](https://github.com/aws/aws-sdk-go-v2/issues/1294)
- checking for db interface value to be nil perhaps less than ideal way: fmt.Sprintf("%v", is.cache) == fmt.Sprintf("%v", nil)

**Other comments**
Inspiration from https://github.com/distribution/distribution

**To do**
- check that the example config actually works
- add coverage tests
- rewrite some (if not all) cache tests for dynamodb


.github/workflows/ci-cd.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       s3mock:
         image: ghcr.io/project-zot/localstack/localstack:0.13.2
         env:
-          SERVICES: s3
+          SERVICES: s3,dynamodb
```
rchincha, line 32: This makes us a user of another aws service? :\

peusebiu: yes..I propose to keep backing up boltdb in s3, and locally we just download it to a temp dir, use it, and keep backing it up?

rchincha:
What happens when we interface with GCP instead? is there a dynamoDB equivalent for GCP?
Would prefer to have a IO driver per cloud provider instead.

docs/aws_databases.md
```diff
@@ -0,0 +1,38 @@
+# AWS Database Options
+
+We store blobs and identify them by the digest. That means that if we were to use a schema-based database, the schema would be pretty simple - but the blobs can get quite big.
```
peusebiu, line 3: blobs size is not an issue. we only store the blob digest as key and and its path as value, or I'm misunderstanding something.

peusebiu (approved):

peusebiu (changes requested):
approved by mistake

pkg/storage/cache.go
```diff
@@ -169,22 +164,22 @@ func (c *Cache) HasBlob(digest, blob string) bool {
 	return true
 }
 
-func (c *Cache) DeleteBlob(digest, path string) error {
+func (d *Driver) DeleteBlob(digest, path string) error {
```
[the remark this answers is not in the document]
chofnar: It was incomplete, I forgot to mention that. This is in my to-do list, and just changed the name to match the one I set in the dynamodb implementation, with the full implementation coming up

docs/aws_databases.md
chofnar: It was a misunderstanding on my part when I wrote the document. I will update it

pkg/storage/local.go
```diff
@@ -1032,7 +1041,7 @@ func (is *ImageStoreLocal) FinishBlobUpload(repo, uuid string, body io.Reader, d
 
 	dst := is.BlobPath(repo, dstDigest)
 
-	if is.dedupe && is.cache != nil {
+	if is.dedupe && fmt.Sprintf("%v", is.cache) != fmt.Sprintf("%v", nil) {
```
peusebiu, line 1044: what if we use a switch where we use type check for each implementation we know we have, this time the only implementation is dynamodatabase, if it's not that type assume cache is nil

later when adding more databases we add check for them.

Thoughts?

pkg/storage/s3/s3.go
```diff
@@ -81,15 +82,27 @@ func NewImageStore(rootDir string, cacheDir string, gc bool, gcDelay time.Durati
 		dedupe:           dedupe,
 	}
 
-	cachePath := path.Join(cacheDir, CacheDBName+storage.DBExtensionName)
+	databaseType := os.Getenv("ZOT_CACHEDB_TYPE")
```
peusebiu, line 85: who sets this env var?

peusebiu: I saw that you set it, ok.

pkg/storage/s3/s3.go
```diff
@@ -81,15 +82,27 @@ func NewImageStore(rootDir string, cacheDir string, gc bool, gcDelay time.Durati
...
-		imgStore.cache = storage.NewCache(cacheDir, CacheDBName, false, log)
-	} else {
-		// if dedupe was used in previous runs use it to serve blobs correctly
-		if _, err := os.Stat(cachePath); err == nil {
-			log.Info().Str("cache path", cachePath).Msg("found cache database")
-			imgStore.cache = storage.NewCache(cacheDir, CacheDBName, false, log)
+	switch databaseType {
+	case "dynamodb":
+		log.Debug().Msg("Using dynamodb")
+		imgStore.cache, _ = database.Create("dynamodb", dynamodatabase.DynamoDBDriverParameters{}, log)
```
peusebiu, line 90: I don't think there should be any difference between creating a dynamodb cache and a boltdb cache.

Issue #537, opened 2022-05-18
remove `allowReadAccess` and `ReadOnly` as they can now be controlled via access control policy


rchincha:
**Describe the bug**
`allowReadAccess` and `ReadOnly` were added until access control policy feature could be added, which we now have.
So should be deprecated.

andaaron:
> and then you want to hide a specific repo from unauthorized users (using authz config) it will not work, all the logic regarding longest matched pattern we use there will be bypased.

Can you elaborate how this works?
Sounds like a bug to be.

PR #594, opened 2022-06-23
deprecation: allowReadAccess and ReadOnly

nicoldr:

**What type of PR is this?**

resolves: #537 

Now that the access control policy feature was added, AllowReadAccess and ReadOnly can be marked as deprecated.
So should be deprecated.

**Automation added to e2e**:

yes

```release-note
ReadOnly and AllowReadAcces options are removed, AccessControl should be used instead
```

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -49,13 +49,6 @@ Additionally, TLS configuration can be specified with:
         },
 ```
 
-The registry can be deployed as a read-only service with:
-
-```
-        "ReadOnly": false
-    },
-```
-
 ## Storage
 
 Configure storage with:
@@ -209,7 +202,8 @@ create/update/delete can not be used without 'read' action, make sure read is al
           "actions": ["read", "create", "update"]
         }
       ],
-      "defaultPolicy": ["read", "create"]                      # default policy which is applied for all users => so all users can read/create repositories
+      "defaultPolicy": ["read", "create"],                     # default policy which is applied for authenticated users, other than "charlie"=> so these users can read/create repositories
+      "anonymousPolicy": ["read]                               # anonymous policy which is applied for unauthenticated users => so they can read repositories
     },
     "tmp/**": {                                                # matches all repos under tmp/ recursively
       "defaultPolicy": ["read", "create", "update"]            # so all users have read/create/update on all repos under tmp/ eg: tmp/infra/repo
--- a/examples/config-policy.json
+++ b/examples/config-policy.json
@@ -15,6 +15,7 @@
     },
     "accessControl": {
       "**": {
+        "anonymousPolicy": ["read"],
         "policies": [
           {
             "users": [
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -86,22 +86,6 @@ func noPasswdAuth(realm string, config *config.Config) mux.MiddlewareFunc {
 				return
 			}
 
-			if config.HTTP.AllowReadAccess &&
-				config.HTTP.TLS.CACert != "" &&
-				request.TLS.VerifiedChains == nil &&
-				request.Method != http.MethodGet && request.Method != http.MethodHead {
-				authFail(response, realm, 5) //nolint:gomnd
-
-				return
-			}
-
-			if (request.Method != http.MethodGet && request.Method != http.MethodHead) && config.HTTP.ReadOnly {
-				// Reject modification requests in read-only mode
-				response.WriteHeader(http.StatusMethodNotAllowed)
-
-				return
-			}
-
 			// Process request
 			next.ServeHTTP(response, request)
 		})
@@ -197,51 +181,29 @@ func basicAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 
 				return
 			}
-			if (request.Method == http.MethodGet || request.Method == http.MethodHead) && ctlr.Config.HTTP.AllowReadAccess {
+			if request.Header.Get("Authorization") == "" && anonymousPolicyExists(ctlr.Config.AccessControl) {
 				// Process request
 				next.ServeHTTP(response, request)
 
 				return
 			}
 
-			if (request.Method != http.MethodGet && request.Method != http.MethodHead) && ctlr.Config.HTTP.ReadOnly {
-				response.WriteHeader(http.StatusMethodNotAllowed)
-
-				return
-			}
-
-			basicAuth := request.Header.Get("Authorization")
-			if basicAuth == "" {
-				authFail(response, realm, delay)
-
-				return
-			}
-
-			splitStr := strings.SplitN(basicAuth, " ", 2) //nolint:gomnd
-
-			if len(splitStr) != 2 || strings.ToLower(splitStr[0]) != "basic" {
-				authFail(response, realm, delay)
-
-				return
-			}
-
-			decodedStr, err := base64.StdEncoding.DecodeString(splitStr[1])
+			username, passphrase, err := getUsernamePasswordBasicAuth(request)
 			if err != nil {
+				ctlr.Log.Error().Err(err).Msg("failed to parse authorization header")
 				authFail(response, realm, delay)
 
 				return
 			}
 
-			pair := strings.SplitN(string(decodedStr), ":", 2) //nolint:gomnd
-			// nolint:gomnd
-			if len(pair) != 2 {
-				authFail(response, realm, delay)
+			// some client tools might send Authorization: Basic Og== (decoded into ":")
+			// empty username and password
+			if username == "" && passphrase == "" && anonymousPolicyExists(ctlr.Config.AccessControl) {
+				// Process request
+				next.ServeHTTP(response, request)
 
 				return
 			}
-
-			username := pair[0]
-			passphrase := pair[1]
 
 			// first, HTTPPassword authN (which is local)
 			passphraseHash, ok := credMap[username]
@@ -296,4 +258,32 @@ func authFail(w http.ResponseWriter, realm string, delay int) {
 	w.Header().Set("WWW-Authenticate", realm)
 	w.Header().Set("Content-Type", "application/json")
 	WriteJSON(w, http.StatusUnauthorized, NewErrorList(NewError(UNAUTHORIZED)))
+}
+
+func getUsernamePasswordBasicAuth(request *http.Request) (string, string, error) {
+	basicAuth := request.Header.Get("Authorization")
+
+	if basicAuth == "" {
+		return "", "", errors.ErrParsingAuthHeader
+	}
+
+	splitStr := strings.SplitN(basicAuth, " ", 2) //nolint:gomnd
+	if len(splitStr) != 2 || strings.ToLower(splitStr[0]) != "basic" {
+		return "", "", errors.ErrParsingAuthHeader
+	}
+
+	decodedStr, err := base64.StdEncoding.DecodeString(splitStr[1])
+	if err != nil {
+		return "", "", err
+	}
+
+	pair := strings.SplitN(string(decodedStr), ":", 2) //nolint:gomnd
+	if len(pair) != 2 {                                //nolint:gomnd
+		return "", "", errors.ErrParsingAuthHeader
+	}
+
+	username := pair[0]
+	passphrase := pair[1]
+
+	return username, passphrase, nil
 }
--- a/pkg/api/authz.go
+++ b/pkg/api/authz.go
@@ -2,10 +2,8 @@ package api
 
 import (
 	"context"
-	"encoding/base64"
 	"fmt"
 	"net/http"
-	"strings"
 	"time"
 
 	glob "github.com/bmatcuk/doublestar/v4"
@@ -140,7 +138,14 @@ func isPermitted(username, action string, policyGroup config.PolicyGroup) bool {
 
 	// check defaultPolicy
 	if !result {
-		if common.Contains(policyGroup.DefaultPolicy, action) {
+		if common.Contains(policyGroup.DefaultPolicy, action) && username != "" {
+			result = true
+		}
+	}
+
+	// check anonymousPolicy
+	if !result {
+		if common.Contains(policyGroup.AnonymousPolicy, action) && username == "" {
 			result = true
 		}
 	}
@@ -184,10 +189,19 @@ func AuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
 			acCtrlr := NewAccessController(ctlr.Config)
 
 			// allow anonymous authz if no authn present and only default policies are present
-			username := ""
+			var username string
+			var err error
 
-			if isAuthnEnabled(ctlr.Config) {
-				username = getUsername(request)
+			/* To be implemented: verify client certs and get its username(subject DN)
+			if request.TLS.VerifiedChains != nil, then get subject DN
+			issue: https: //github.com/project-zot/zot/issues/614 */
+
+			if isAuthnEnabled(ctlr.Config) && request.Header.Get("Authorization") != "" {
+				username, _, err = getUsernamePasswordBasicAuth(request)
+
+				if err != nil {
+					authFail(response, ctlr.Config.HTTP.Realm, ctlr.Config.HTTP.Auth.FailDelay)
+				}
 			}
 
 			ctx := acCtrlr.getContext(username, request)
@@ -232,19 +246,23 @@ func AuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
 	}
 }
 
-func getUsername(r *http.Request) string {
-	// this should work because it was already parsed in authn middleware
-	basicAuth := r.Header.Get("Authorization")
-	s := strings.SplitN(basicAuth, " ", 2) //nolint:gomnd
-	b, _ := base64.StdEncoding.DecodeString(s[1])
-	pair := strings.SplitN(string(b), ":", 2) //nolint:gomnd
-
-	return pair[0]
-}
-
 func authzFail(w http.ResponseWriter, realm string, delay int) {
 	time.Sleep(time.Duration(delay) * time.Second)
 	w.Header().Set("WWW-Authenticate", realm)
 	w.Header().Set("Content-Type", "application/json")
 	WriteJSON(w, http.StatusForbidden, NewErrorList(NewError(DENIED)))
+}
+
+func anonymousPolicyExists(config *config.AccessControlConfig) bool {
+	if config == nil {
+		return false
+	}
+
+	for _, repository := range config.Repositories {
+		if len(repository.AnonymousPolicy) > 0 {
+			return true
+		}
+	}
+
+	return false
 }
--- a/pkg/api/config/config.go
+++ b/pkg/api/config/config.go
@@ -68,8 +68,6 @@ type HTTPConfig struct {
 	Auth             *AuthConfig
 	RawAccessControl map[string]interface{} `mapstructure:"accessControl,omitempty"`
 	Realm            string
-	AllowReadAccess  bool             `mapstructure:",omitempty"`
-	ReadOnly         bool             `mapstructure:",omitempty"`
 	Ratelimit        *RatelimitConfig `mapstructure:",omitempty"`
 }
 
@@ -112,8 +110,9 @@ type AccessControlConfig struct {
 type Repositories map[string]PolicyGroup
 
 type PolicyGroup struct {
-	Policies      []Policy
-	DefaultPolicy []string
+	Policies        []Policy
+	DefaultPolicy   []string
+	AnonymousPolicy []string
 }
 
 type Policy struct {
@@ -193,8 +192,11 @@ func (c *Config) LoadAccessControlConfig(viperInstance *viper.Viper) error {
 		}
 
 		defaultPolicy := viperInstance.GetStringSlice(fmt.Sprintf("http::accessControl::%s::defaultPolicy", policy))
-		policyGroup.Policies = policies
 		policyGroup.DefaultPolicy = defaultPolicy
+
+		anonymousPolicy := viperInstance.GetStringSlice(fmt.Sprintf("http::accessControl::%s::anonymousPolicy", policy))
+		policyGroup.Policies = policies
+		policyGroup.AnonymousPolicy = anonymousPolicy
 		c.AccessControl.Repositories[policy] = policyGroup
 	}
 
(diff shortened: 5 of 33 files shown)


pkg/api/authz.go
```diff
@@ -235,11 +235,15 @@ func AuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
 func getUsername(r *http.Request) string {
 	// this should work because it was already parsed in authn middleware
 	basicAuth := r.Header.Get("Authorization")
-	s := strings.SplitN(basicAuth, " ", 2) //nolint:gomnd
-	b, _ := base64.StdEncoding.DecodeString(s[1])
-	pair := strings.SplitN(string(b), ":", 2) //nolint:gomnd
+	if basicAuth != "" {
+		s := strings.SplitN(basicAuth, " ", 2) //nolint:gomnd
+		b, _ := base64.StdEncoding.DecodeString(s[1])
```
peusebiu, line 240: let's treat these errrors now like in authn.go

and also add a test where anonymous is enabled and authz have only default policies (one global ** and another one per repo) and make some requests with and without credentials


another one where anonymous is enabled and authz have users policies and make some requests with and without credentials

try to find corner cases

pkg/api/config/config.go
```diff
@@ -68,8 +68,6 @@ type HTTPConfig struct {
 	Auth             *AuthConfig
 	RawAccessControl map[string]interface{} `mapstructure:"accessControl,omitempty"`
 	Realm            string
-	AllowReadAccess  bool             `mapstructure:",omitempty"`
```
peusebiu, line 71: I think we shouldn't remove them for now, only throw an warning if they are not nil, telling clients that these options are deprecated and to use AccessControl instead, also we need to document the new field in examples/README.md

pkg/api/config/config.go
```diff
@@ -179,6 +178,11 @@ func (c *Config) LoadAccessControlConfig(viperInstance *viper.Viper) error {
 
 		var policyGroup PolicyGroup
 
+		if policy == "anonymousaccess" {
```
peusebiu, line 181: manually test that this works by running a zot with a anonymousaccess true, it should be case insensitive, but let's make sure.

peusebiu (changes requested):
Modify examples/config-authz.json to use this field with value false.

examples/README.md
```diff
@@ -202,6 +202,7 @@ create/update/delete can not be used without 'read' action, make sure read is al
 
 ```
 "accessControl": {
+    "anonymousAccess": true,                                   # when true, allowed operations are taken from access control policies, when false users can only read
```
rchincha, line 205: Just call this "anonymous" since you are already in "accessControl" context

pkg/api/config/config.go
```diff
@@ -179,6 +180,11 @@ func (c *Config) LoadAccessControlConfig(viperInstance *viper.Viper) error {
 
 		var policyGroup PolicyGroup
 
+		if policy == "anonymous" {
+			anonymousAccess := viperInstance.GetBool("http::accessControl::anonymous")
+			c.AccessControl.AnonymousAccess = anonymousAccess
```
andaaron, line 185: Do we need a `continue` instruction at the end this block?

pkg/cli/root.go
```diff
@@ -223,6 +223,14 @@ func validateConfiguration(config *config.Config) error {
 		}
 	}
 
+	if config.HTTP.ReadOnly {
+		log.Warn().Msgf("ReadOnly option is deprecated, use AccessControl instead")
```
andaaron, line 227: Deprecated means we are still supporting it, but we will not in the future.
In this PR we remove it, so I don't think it's the right wording.

Do we want to remove or deprecate it? 
Deprecate could be better for existing users of this setting if there are any, but would take some extra maintenance.

examples/config-policy.json
```diff
@@ -14,6 +14,7 @@
       "failDelay": 1
     },
     "accessControl": {
+      "anonymous": false,
       "**": {
```
rchincha, line 18: This "**" is the default right? Should "anonymous" move here instead?
And can "anonynous" be set per repo-policy?

peusebiu: Hei Ram, from what I see in the code we can do this in authz

basically we bypass authn if accessControl != nil and in authz decide if we throw authFail() based on anonymousAccess  flag combined with the condition that username is nil.

The problem arise in certificate authentication, this flag tls.RequireAndVerifyClientCert is used when we start the server.
What to do in this case? currently we look at the global flag and decide based on it, but per repo flags are used in the context of a request, not the whole server.

If we look at "**"'s anonymousAccess it will not be consistent, we can not authFail for a certain repo which has the anonymousAccess on false in case of certificates auth.

hope this is clear enough :D.

peusebiu: We spoke about this and we want to use one global bool flag in http.auth config, basically the same as allowReadAccess and based on it to decide if we let anonymous access with certs authentication. related issue: #614 

And use a per repo anonymous policy for request scoped authorization.

example:
```
      "**": {
        "anonymousAccess": ["read"],
        "policies": [
          {
            "users": [
              "charlie", "bob"
            ],
            "actions": [
              "read",
              "create",
              "update"
            ]
          }
        ],
        "defaultPolicy": [
          "read",
          "create"
        ]
      },
```

authenticated users charlie and bob can do read, cread, update (users policy)
authenticated users others than charlie and bob can do read, create (default policy)
unauthenticated users can do read (anonymous policy)

rchincha: Why not write a:

```
func allowReadAccess()  {
// iterate over all policies, and if there is any one with anonymous read access, then return true here
// and when we know the URL and method, we can enforce specifically
}

Comment (peusebiu) [reply examples/config-policy.json:18]: this will not work for TLS mutual auth.

either we make it optional or required at server startup, from what I understand, besides this, it would work...

Comment (peusebiu) [reply examples/config-policy.json:18]: ok, you are right :D, we make tls certs optional, but we verify them at request level, got it

File: pkg/cli/root.go [hunk 8]
```diff
@@ -219,7 +219,7 @@ func validateConfiguration(config *config.Config) error {
 	if config.HTTP.RawAccessControl != nil {
 		if config.HTTP.Auth == nil || (config.HTTP.Auth.HTPasswd.Path == "" && config.HTTP.Auth.LDAP == nil) {
 			// checking for default policy only authorization config: no users, no policies but default policy
-			checkForDefaultPolicyConfig(config)
+			checkForAnonymousPolicyConfig(config)
```
andaaron, line 222: I find this function name confusing, it's not clear what `check` means.
Why not call it `validateAuthzPolicies` and let the comments/log messages inside give more details? And return the error in case there is a a bad configuration.

pkg/cli/root.go
```diff
@@ -219,7 +219,7 @@ func validateConfiguration(config *config.Config) error {
 	if config.HTTP.RawAccessControl != nil {
 		if config.HTTP.Auth == nil || (config.HTTP.Auth.HTPasswd.Path == "" && config.HTTP.Auth.LDAP == nil) {
```
andaaron, line 220: I would move this `if` inside `checkForAnonymousPolicyConfig` / `validateAuthzPolicies ` (or whatever name it will finally have) as the relevant error message is there. Or move the error message from there over here, but I personally prefer the former.

pkg/cli/root.go
```diff
@@ -423,14 +423,21 @@ func isDefaultPolicyConfig(cfg *config.Config) bool {
...
+				log.Info().Interface("repository", repository).
+					Msg("repository with default policy detected, anonymous authorization disabled")
 
 				return false
 			}
 		}
 	}
 
-	log.Info().Msg("default authorization detected")
+	log.Info().Msg("anonymous authorization detected")
```
andaaron, line 440: So where do we check if anonymous policies are actually present in the config? So far we only check if other types of policies are missing.

This was probably not needed before as the read access flag was used in some other place.

pkg/cli/root.go
```diff
@@ -272,11 +272,11 @@ func validateConfiguration(config *config.Config) error {
...
 
-func checkForDefaultPolicyConfig(config *config.Config) {
-	if !isDefaultPolicyConfig(config) {
+func checkForAnonymousPolicyConfig(config *config.Config) {
+	if !isAnonymousPolicyConfig(config) {
 		log.Error().Err(errors.ErrBadConfig).
 			Msg("access control config requires httpasswd, ldap authentication " +
-				"or using only 'defaultPolicy' policies")
+				"or using only 'anonymousPolicy' policies")
 		panic(errors.ErrBadConfig)
```
andaaron, line 280: Why panic, and not return an error like in other bad configuration cases?

pkg/cli/root.go
```diff
@@ -408,13 +408,13 @@ func LoadConfiguration(config *config.Config, configPath string) error {
 	return nil
 }
 
-func isDefaultPolicyConfig(cfg *config.Config) bool {
+func isAnonymousPolicyConfig(cfg *config.Config) bool {
```
andaaron, line 411: I would rename this `doesAuthzContainOnlyAnonymousPolicies`

pkg/cli/root.go
```diff
@@ -408,13 +408,13 @@ func LoadConfiguration(config *config.Config, configPath string) error {
...
 	adminPolicy := cfg.AccessControl.AdminPolicy
 
-	log.Info().Msg("checking if default authorization is possible")
+	log.Info().Msg("checking if anonymous authorization is possible")
```
andaaron, line 414: `log.Info().Msg("checking if anonymous authorization is the only type of authorization policy configured")`

pkg/cli/root.go
```diff
@@ -408,13 +408,13 @@ func LoadConfiguration(config *config.Config, configPath string) error {
...
 
 	if len(adminPolicy.Actions)+len(adminPolicy.Users) > 0 {
-		log.Info().Msg("admin policy detected, default authorization disabled")
+		log.Info().Msg("admin policy detected, anonymous authorization disabled")
```
andaaron, line 417: `log.Info().Msg("admin policy detected, anonymous authorization is not the only authorization policy configured")`.

Similar updates for the other log messages.

andaaron (commented):
Overall looks good. I do have some comments though, mostly for the config validation code.

pkg/api/routes.go
```diff
@@ -140,7 +140,7 @@ func (rh *RouteHandler) CheckVersionSupport(response http.ResponseWriter, reques
 	response.Header().Set(constants.DistAPIVersion, "registry/2.0")
 	// NOTE: compatibility workaround - return this header in "allowed-read" mode to allow for clients to
 	// work correctly
-	if rh.c.Config.HTTP.AllowReadAccess {
```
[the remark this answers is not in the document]
peusebiu: we do return the auth header in the authn.go handler, always.

that was an extra check, because if allowReadAccess was true and the request was a HEAD or a GET then the request would pass authn.go, and went into routes.go, and in the case of /v2/ which is checked by all tools, we needed to return the WWW-Authenticate header.

in other words without this check we wouldn't return the WWW-Authenticate  header for GET and HEAD request when allowReadAccess was true.
 
hope this is clear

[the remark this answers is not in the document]
peusebiu: now, because of our change with anonymous policy, we may return a 403 with WWW-Authenticate headers, I wonder if tools know how to handle this.

pkg/api/routes.go
```diff
@@ -140,7 +140,7 @@ func (rh *RouteHandler) CheckVersionSupport(response http.ResponseWriter, reques
...
+	if checkAnonymousPolicyExists(rh.c.Config.AccessControl) {
```
peusebiu, line 143: I don't know if this makes sense though...
Maybe we should use the request context to mark a user as anonymous?

pkg/api/authz.go
```diff
@@ -184,10 +189,19 @@ func AuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
...
-				username = getUsername(request)
+			/* To do: verify client certs and get its username(subject DN)
+			if request.TLS.VerifiedChains != nil, then get subject DN
+			issue: https: //github.com/project-zot/zot/issues/614 */
+
+			if isAuthnEnabled(ctlr.Config) && request.Header.Get("Authorization") != "" {
+				username, _, err = getUsernamePassword(request, ctlr, response)
+
+				if err != nil {
+					authzFail(response, ctlr.Config.HTTP.Realm, ctlr.Config.HTTP.Auth.FailDelay)
```
peusebiu, line 203: I think this should be authFail()

pkg/cli/root.go
peusebiu: config needs to miss all default+admin+users policies right?
if config misses all of them it means we have at least one anonymous policy, otherwise config.HTTP.RawAccessControl would be nil and the check would not run.

maybe I'm missing something :D

examples/config-policy.json
andaaron: Let's resolve this

pkg/api/routes.go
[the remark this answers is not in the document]
andaaron: Actually it is not clear. I'm not saying we should remove the content of the if block. I'm saying we should always add the WWW-Authenticate header regardless of checkAnonymousPolicyExists. This just looks like an extra check which was redundant before and is not needed now either. Am I missing something?

pkg/cli/root.go
```diff
@@ -408,31 +412,37 @@ func LoadConfiguration(config *config.Config, configPath string) error {
...
 	adminPolicy := cfg.AccessControl.AdminPolicy
 
-	log.Info().Msg("checking if default authorization is possible")
+	log.Info().Msg("checking if anonymous authorization is the only type of authorization policy configured")
 
 	if len(adminPolicy.Actions)+len(adminPolicy.Users) > 0 {
-		log.Info().Msg("admin policy detected, default authorization disabled")
-
-		return false
+		log.Info().Msg("admin policy detected, anonymous authorization is not the only authorization policy configured")
```
andaaron, line 422: You still need to return false in case other policies besides anonymous policies are found.

pkg/cli/root.go
```diff
@@ -408,31 +412,37 @@ func LoadConfiguration(config *config.Config, configPath string) error {
...
 	}
 
 	for _, repository := range cfg.AccessControl.Repositories {
 		for _, policy := range repository.Policies {
 			if len(policy.Actions)+len(policy.Users) > 0 {
 				log.Info().Interface("repository", repository).
-					Msg("repository with non-empty policy detected, default authorization disabled")
+					Msg("repository with non-empty policy detected, " +
```
andaaron, line 429: We still need to return false after this log message

pkg/cli/root.go
```diff
@@ -408,31 +412,37 @@ func LoadConfiguration(config *config.Config, configPath string) error {
...
+						"anonymous authorization is not the only authorization policy configured")
+			}
 
-				return false
+			if len(repository.DefaultPolicy) > 0 {
+				log.Info().Interface("repository", repository).
```
andaaron, line 434: We still need to return false after this log message.

pkg/cli/root.go
```diff
@@ -272,13 +272,17 @@ func validateConfiguration(config *config.Config) error {
...
+func validateAuthzPolicies(config *config.Config) error {
+	if (config.HTTP.Auth == nil || (config.HTTP.Auth.HTPasswd.Path == "" && config.HTTP.Auth.LDAP == nil)) &&
+		!doesAuthzContainOnlyAnonymousPolicies(config) {
 		log.Error().Err(errors.ErrBadConfig).
 			Msg("access control config requires httpasswd, ldap authentication " +
-				"or using only 'defaultPolicy' policies")
-		panic(errors.ErrBadConfig)
+				"or using only 'anonymousPolicy' policies")
+
+		return errors.ErrParsingAuthHeader
```
andaaron, line 282: This error means something is wrong with a request header, not a config file.

andaaron (commented):
Please also check the failing test.

pkg/api/config/config.go
```diff
@@ -196,6 +195,11 @@ func (c *Config) LoadAccessControlConfig(viperInstance *viper.Viper) error {
 		policyGroup.Policies = policies
 		policyGroup.DefaultPolicy = defaultPolicy
 		c.AccessControl.Repositories[policy] = policyGroup
```
peusebiu, line 197: remove this, you have it below already line 202

pkg/api/config/config.go
```diff
@@ -196,6 +195,11 @@ func (c *Config) LoadAccessControlConfig(viperInstance *viper.Viper) error {
 		policyGroup.Policies = policies
```
peusebiu, line 195: remove this, you have it below already, line 200

andaaron (commented):
Code looks good from my point of view.
But before merging we need to:
- [ ] make sure the calls to get version are not getting error 403 before reaching CheckVersionSupport
- [ ] decided on the change for routes.go

examples/config-default-authz.json
```diff
@@ -9,25 +9,25 @@
     "realm": "zot",
     "accessControl": {
       "**": {
-        "defaultPolicy": [
```
peusebiu, line 12: rename this file to config-anonymous-authz.json

peusebiu:
there are some tests in root_test.go for default policy, you also need to modify those.
https://github.com/project-zot/zot/pull/594/files#diff-c3e95b9d0ad189ba68bf30694e60fd679eb406ac541ff635f3171f18a6d794a4R194

rchincha:
for your PR, a simple test is enable authN and anonymous, and you should be able to skopeo copy reads without passwords but writing should need a password

take sebi's help with this

can you also add a BATS test for this?

pkg/api/controller_test.go
```diff
@@ -2408,7 +2489,7 @@ func TestAuthorizationWithOnlyDefaultPolicy(t *testing.T) {
 		So(resp.StatusCode(), ShouldEqual, http.StatusForbidden)
 
 		if entry, ok := conf.AccessControl.Repositories[TestRepo]; ok {
-			entry.DefaultPolicy = []string{"create", "read"}
+			entry.AnonymousPolicy = []string{"create", "read"}
 			conf.AccessControl.Repositories[TestRepo] = entry
 		}
```
peusebiu, line 2495: there are some SetBasicAuth(username, passphrase). below this line, remove them as they are not needed for this test.

peusebiu:
We need a TestAnonymousPolicies() where we test anonymous policies used together with default and users policies.

we need to confirm that:
1. authenticated users permissions are taken from defaultPolicy and users Policies.
2. unauthenticated users permissions are taken from anonymousPolicy and not the other.

pkg/api/authn.go
```diff
@@ -197,52 +181,21 @@ func basicAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 
 				return
 			}
-			if (request.Method == http.MethodGet || request.Method == http.MethodHead) && ctlr.Config.HTTP.AllowReadAccess {
+			if request.Header.Get("Authorization") == "" && checkAnonymousPolicyExists(ctlr.Config.AccessControl) {
```
peusebiu, line 184: seems like we can't rely on this check because tools are sending this header when no user or password was provided:
Authorization: Basic Og==  (decoded value == ":")

The good news is that our function won't throw an error for this header value.

so we can add this check after line 191: username, passphrase, err := getUsernamePasswordBasicAuth(request)

```
			if username == "" && passphrase == "" && checkAnonymousPolicyExists(ctlr.Config.AccessControl) {
				// Process request
				next.ServeHTTP(response, request)

				return
			}
```

andaaron (approved):
LGTM, let's wait on approval from @rchincha

peusebiu (approved):

pkg/api/authz.go
```diff
@@ -232,19 +246,23 @@ func AuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
...
-}
-
 func authzFail(w http.ResponseWriter, realm string, delay int) {
 	time.Sleep(time.Duration(delay) * time.Second)
 	w.Header().Set("WWW-Authenticate", realm)
 	w.Header().Set("Content-Type", "application/json")
 	WriteJSON(w, http.StatusForbidden, NewErrorList(NewError(DENIED)))
 }
+
+func checkAnonymousPolicyExists(config *config.AccessControlConfig) bool {
```
rchincha, line 256: Just call this "anonymousPolicyExists()"

pkg/api/authn.go
```diff
@@ -86,22 +86,6 @@ func noPasswdAuth(realm string, config *config.Config) mux.MiddlewareFunc {
 				return
 			}
 
-			if config.HTTP.AllowReadAccess &&
```
rchincha, line 89: This check is saying if read-only access is enabled and mutual tls is enabled and methods are write, deny them. Is this getting subsumed by access control policy now?

peusebiu: handling mutual auth is not in this PR, will arrive in the Nicol's next one, maybe we should merge them because of this?

you are right, this is a bug.

peusebiu: besides tls mutual auth, the rest is handled in the authz.

pkg/api/authz.go
```diff
@@ -184,10 +189,19 @@ func AuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
 			acCtrlr := NewAccessController(ctlr.Config)
 
 			// allow anonymous authz if no authn present and only default policies are present
-			username := ""
+			var username string
+			var err error
 
-			if isAuthnEnabled(ctlr.Config) {
-				username = getUsername(request)
+			/* To do: verify client certs and get its username(subject DN)
```
rchincha, line 195: The keyword to use is either TODO or FIXME

rchincha: examples/metrics/kubernetes/prometheus/bundle.yaml:3762:                                      TODO: Add other useful fields. apiVersion, kind

like so ^

pkg/cli/root.go
```diff
@@ -408,31 +412,44 @@ func LoadConfiguration(config *config.Config, configPath string) error {
 	return nil
 }
 
-func isDefaultPolicyConfig(cfg *config.Config) bool {
+func doesAuthzContainOnlyAnonymousPolicies(cfg *config.Config) bool {
```
rchincha, line 415: Rename this as "authzContainsOnlyAnonymousPolicy()"

rchincha:
@nicoldr resolve the minor comments and re-push. This PR is ready to be merged.

pkg/api/authz.go
nicoldr: "TODO" would make lint fail when updating the PR. I replaced it with "To be implemented" if it's alright with you

rchincha (dismissed):
lgtm

rchincha (approved):
lgtm

Issue #534, opened 2022-05-16
report a nicer failure on a non-OCI compliant `docker push`


rchincha:
**Is your feature request related to a problem? Please describe.**

Given that `zot` is a container image registry, an expectation may be that `zot` supports docker protocol and manifests. However, `zot` is a OCI-only registry and it would be nice to return a more graceful error to indicate why for example `docker push` doesn't work.

Originally reported by @SteveLasker

**Describe the solution you'd like**
`
"[Steve Lasker]:
just a suggestion you might want to put that in the docs and samples. Maybe even throw an unsupported exception when you see the docker manifest on push. 
`

It would be nice to return a more graceful error to indicate why for example `docker push` doesn't work.

**Describe alternatives you've considered**

**Additional context**

rchincha:
Bring up zot and then try to do a `docker push` against it.

`docker push` assumes docker registry which is not oci dist-spec compliant for push. But this could be true for any non-oci compliant client. What we want to fix is that we return an error reporting the non-compliance.

andaaron:
This looks like a client side error? Can someone confirm this `expected integer` error is not coming from zot?

bogdanbiv:
indeed it is

Issue #586, opened 2022-06-09
add `regclient` to ecosystem tools github action


rchincha:
**Is your feature request related to a problem? Please describe.**

We currently verify against skopeo and oras. We should also do `regclient`.

**Describe the solution you'd like**

https://github.com/regclient/regclient

bogdanbiv:
makefile test regclient exists, install if missing

add to Zot blackbox push-pull tests
- regclient pull OCI image (if there is an image already in Zot)
- regclient push new OCI image

rchincha:
> makefile test regclient exists, install if missing
> 
> add to Zot blackbox push-pull tests (test/blackbox/pushpull.bats)
> 
>     * regclient pull OCI image (if there is an image already in Zot)
> 
>     * regclient push new OCI image

We can add the install step in the github workflow instead of in the Makefile, fwiw

andaaron:
Done in #668

PR #596, opened 2022-06-28
Make rejecting docker manifests more explicit

bogdanbiv:
Signed-off-by: Bogdan BIVOLARU <bbivolar@cisco.com>

**What type of PR is this?**
bug

In case we receive an invalid manifest (docker) we used to return 'Unsupported Media Type' HTTP error code 415, without any HTTP body. This adds a HTTP body which spells it out for the user.

Makes docker fail with a clear reason, since docker sees an invalid manifest error and retries with a valid OCI manifest format.

fixes: #534 

```bash
sudo docker push reg.mydomain2.com:8080/test:test
The push refers to repository [reg.mydomain2.com:8080/test]
65110baad718: Pushing [==================================================>]  1.947MB
0bc0768466af: Retrying in 4 seconds
bebfee1e62da: Pushing [==================================================>]  774.7kB
23f7bd114e4a: Layer already exists
expected integer   # it's unclear if it's successful or an error; it's not clear how to investigate the root cause
```

```skopeo --debug -v copy --dest-tls-verify=false docker://alpine:3.4 docker://reg.mydomain2.com:8080/test:test```

```bash
sudo dockerd --pidfile /var/run/docker.sock
sudo docker tag alpine:3.4 docker://reg.mydomain2.com/test:test
sudo docker push reg.mydomain2.com:8080/test:test
The push refers to repository [reg.mydomain2.com:8080/test]
65110baad718: Layer already exists
0bc0768466af: Layer already exists
bebfee1e62da: Layer already exists
23f7bd114e4a: Layer already exists
manifest invalid: manifest invalid
```

**Automation added to e2e**:

Yes

Our own compliance tests fail:
```
  * zotregistry.io/zot/pkg/compliance/v1_0_0/check_test.go 
  Line 32:
  Expected: 'bytes=0-23'
  Actual:   '0-23'
  (Should be equal)
  * zotregistry.io/zot/pkg/compliance/v1_0_0/check_test.go 
  Line 32:
  Expected: 'bytes=0-23'
  Actual:   '0-23'
  (Should be equal)
```

```release-note
When receiving a non-OCI compliant manifest, and the HTTP unsupported media code 415 is returned, send a response body with more detailed explanation
```

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -438,7 +438,9 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
 
 	mediaType := request.Header.Get("Content-Type")
 	if !storage.IsSupportedMediaType(mediaType) {
-		response.WriteHeader(http.StatusUnsupportedMediaType)
+		// response.WriteHeader(http.StatusUnsupportedMediaType)
+		WriteJSON(response, http.StatusUnsupportedMediaType,
+			NewErrorList(NewError(MANIFEST_INVALID, map[string]string{"mediaType": mediaType})))
 
 		return
 	}
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -725,7 +725,7 @@ func (rh *RouteHandler) DeleteBlob(response http.ResponseWriter, request *http.R
 // @Param   name				path    string     true        "repository name"
 // @Success 202 {string} string	"accepted"
 // @Header  202 {string} Location "/v2/{name}/blobs/uploads/{session_id}"
-// @Header  202 {string} Range "bytes=0-0"
+// @Header  202 {string} Range "0-0"
 // @Failure 404 {string} string "not found"
 // @Failure 500 {string} string "internal server error"
 // @Router /v2/{name}/blobs/uploads [post].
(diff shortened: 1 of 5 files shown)


andaaron:
Looks like they discussed this for the conformance tests: https://github.com/opencontainers/distribution-spec/pull/203

bogdanbiv:
@rchincha what's your take, can we move to using 'Content-Range' header as per OCI distro spec?

andaaron:
Let's circle back after we resolve https://github.com/project-zot/zot/issues/586

andaaron:
> @rchincha what's your take, can we move to using 'Content-Range' header as per OCI distro spec?

Per distro spec Content-Range is part of the headers in the request, not the response headers.
And we currently handle it in the code handling request headers.

This PR should be working well as is.

rchincha:
Add `regclient` to the ecosystem tools workflow.

andaaron:
```
Failures:

  * zotregistry.io/zot/pkg/compliance/v1_0_0/check_test.go 
  Line 32:
  Expected: 'bytes=0-23'
  Actual:   '0-23'
  (Should be equal)

  * zotregistry.io/zot/pkg/compliance/v1_0_0/check_test.go 
  Line 32:
  Expected: 'bytes=0-23'
  Actual:   '0-23'
  (Should be equal)


442 total assertions

--- FAIL: TestWorkflows (1.13s)
FAIL
coverage: 15.0% of statements in ./...
FAIL	zotregistry.io/zot/pkg/compliance/v1_0_0	2.251s
```

I think we may want to update the our tests. The dist-spec tests are passing.

andaaron (approved):

PR #601, opened 2022-07-01
Fuzzing for local storage

alexstan12:
**What type of PR is this?**
feature

1. It is meant to introduce fuzzing, using native go fuzzing engine and ~~https://github.com/AdaLogics/go-fuzz-headers~~ to generate data for fuzz tests

**UPDATE**: go-fuzz-headers fails to generate data as go fuzz engine mutates input. go-fuzz-headers requires a specific amount  of bytes, which can't be guaranteed when data is mutated.

2. Fuzz tests require an initial seed corpus, which in Zot's case is cloned right before fuzzing starts from a separate repo https://github.com/project-zot/test-data
3. A Make target was created for running all Fuzz tests sequentially, each for a default period of 10s ( which can be overriden)

**This PR contains fuzz tests for all implementations of storage interface in storage_fs.**
Other parts of code to be covered with fuzz in future-PRs

detect code vulnerabilities

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/local.go
+++ b/pkg/storage/local.go
@@ -7,13 +7,16 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
+	"unicode/utf8"
 
 	apexlog "github.com/apex/log"
 	guuid "github.com/gofrs/uuid"
@@ -186,6 +189,13 @@ func (is *ImageStoreLocal) Unlock(lockStart *time.Time) {
 
 func (is *ImageStoreLocal) initRepo(name string) error {
 	repoDir := path.Join(is.rootDir, name)
+
+	if !utf8.ValidString(name) {
+		is.log.Error().Msg("input is not valid UTF-8")
+
+		return zerr.ErrInvalidRepositoryName
+	}
+
 	// create "blobs" subdir
 	err := ensureDir(path.Join(repoDir, "blobs"), is.log)
 	if err != nil {
@@ -850,6 +860,7 @@ func (is *ImageStoreLocal) NewBlobUpload(repo string) (string, error) {
 	if err != nil {
 		return "", zerr.ErrRepoNotFound
 	}
+
 	defer file.Close()
 
 	return uid, nil
@@ -858,6 +869,12 @@ func (is *ImageStoreLocal) NewBlobUpload(repo string) (string, error) {
 // GetBlobUpload returns the current size of a blob upload.
 func (is *ImageStoreLocal) GetBlobUpload(repo, uuid string) (int64, error) {
 	blobUploadPath := is.BlobUploadPath(repo, uuid)
+
+	if !utf8.ValidString(blobUploadPath) {
+		is.log.Error().Msg("input is not valid UTF-8")
+
+		return -1, zerr.ErrInvalidRepositoryName
+	}
 
 	binfo, err := os.Stat(blobUploadPath)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/storage/local_test.go
```diff
@@ -213,6 +214,711 @@ func TestGetReferrers(t *testing.T) {
...
+			t.Error(err)
+		}
+	})
+}
+
+func FuzzPutBlobChunk(f *testing.F) {
+	f.Fuzz(func(t *testing.T, data string) {
+		dir := t.TempDir()
+		defer os.RemoveAll(dir)
+		t.Logf("Input argument is %s", data)
```
andaaron, line 241: Can you briefly explain how the go toolchain identifies the correct data in the separate repo for a specific test?
What is the repo used for? I was under the impression the data was supposed to be dynamically generated for each test execution.

alexstan12: The Go toolchain does not identify data from a separate repo. Go just expects a folder named 'testdata' to be present inside the package it tests, before fuzzing starts.
That's why we have these 2 lines in the newly added fuzz-all Make target:
```
git clone https://github.com/project-zot/test-data.git; \
mv test-data/storage pkg/storage/testdata; \
```
As fuzz tests are created for different functionalities/packages, the fuzz-all Make target needs completion with new "mv test-data/blabla pkg/blabla/testdata" commands.

The separate repo exists for fuzz data to be gathered, since it can get bigger over time.

See https://github.com/project-zot/zot/blob/f72e4dfdff2187e444781f217816cf79e19c0758/README_fuzz.md for a better view.

andaaron: Got it.
What about having the fuzz tests in the same file with the other tests? What happens when we run the usual go test command? Are they skipped because the data is missing? Do they count for coverage in any way?

alexstan12: Native Go fuzzing encourages having Fuzz tests in the same file with the other tests. The usual go test command won't run fuzz test without the "-fuzz" flag, followed by the name of the test to fuzz.
As I've seen , they won't count for coverage. Maybe using the -cover flag (as used with usual go test) will make Fuzz tests count.

andaaron: Looking at https://go.dev/doc/fuzz/

```
There are two modes of running your fuzz test: as a unit test (default go test), or with fuzzing (go test -fuzz=FuzzTestName).

Fuzz tests are run much like a unit test by default. Each [seed corpus entry](https://go.dev/doc/fuzz/#glos-seed-corpus) will be tested against the fuzz target, reporting any failures before exiting.
```
Since the seed corpus would not normally be available (since the test-data folders are not normally copied under the repo), are these not run at all?

andaaron: I think we should be fine.

alexstan12: Yes, you are right. Without "-fuzz", they will be run as unit tests. Without input data, they will finish instantly, as if skiped.

pkg/storage/local.go
```diff
@@ -183,6 +186,13 @@ func (is *ImageStoreLocal) Unlock(lockStart *time.Time) {
 
 func (is *ImageStoreLocal) initRepo(name string) error {
 	repoDir := path.Join(is.rootDir, name)
+
+	if !utf8.ValidString(name) {
+		is.log.Error().Msg("input is not valid UTF-8")
+
+		return zerr.ErrFuzz
```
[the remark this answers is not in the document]
alexstan12: I will think of a substitute for this and modify.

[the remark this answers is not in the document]
peusebiu: yes, just return the utf8 error and treat it in the test, same with other cases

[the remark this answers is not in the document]
peusebiu: ErrInvalidRepositoryName ? we have one already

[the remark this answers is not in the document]
alexstan12: I've used the one Sebi mentioned since it seems to fit our use case.

pkg/storage/local.go
```diff
@@ -1673,12 +1690,23 @@ func ifOlderThan(imgStore *ImageStoreLocal, repo string, delay time.Duration) ca
...
 func DirExists(d string) bool {
-	fi, err := os.Stat(d)
+	if !utf8.ValidString(d) {
+		return false
+	}
+
+	fileInfo, err := os.Stat(d)
+	if err != nil {
+		if e, ok := err.(*fs.PathError); ok && errors.Is(e.Err, syscall.ENAMETOOLONG) || //nolint: errorlint
+			errors.Is(e.Err, syscall.EINVAL) {
```
peusebiu, line 1700: you can cover this in storage/local_test.go?
by running DirExists() with a long name? :D

alexstan12: yes, I created a test with a long, invalid path. now waiting for checks to pass.

andaaron (approved):

PR #602, opened 2022-07-04
build: remove swagger install in stacker files

peusebiu:
it is currently installed in the Makefile

No need to install swagger in stacker files, that part is now in the Makefile as a dependency for binaries.


peusebiu:
testing it right here: https://github.com/peusebiu/zot/runs/7183719766?check_suite_focus=true

rchincha (approved):
lgtm

Issue #591, opened 2022-06-22
ui: add a freeform "search" api (with boltdb support)


rchincha:
**Is your feature request related to a problem? Please describe.**

Currently, we have graphQL APIs for very specific use cases. 
We need a "freeform" search API which does the following:
given a string, return a top-K list of descriptors and types (repo, image, layer-hash) containing that string.

**Describe the solution you'd like**

Create a new API that behaves like above.

rchincha:
Given a substring (not regex), return a list of results.

Dynamic search:
* Given a substring, return a list of top-K results of each type (repo, image:tag, layer hash)

Paginated search:
* Given a substring, return a list of results (repo, image:tag, layer hash) with paginated search (/query?offset=15&limit=15)

andaaron:
I propose we have a single API, and the results would be ordered by relevance.
For the dynamic search, we would only show the 1st page, which would contains the most relevant K results.
Laurentiu has a ideas on the criteria according to which the search results would be more relevant than others.

andaaron:
Let's keep track of this for the DB related work for servicing graphql requests.
The GQL side changes were already merged.

andaaron:
In review https://github.com/project-zot/zot/pull/979/commits

PR #618, opened 2022-07-12
Freeform search API

laurentiuNiculae:
sorting result by their match score

tests

Signed-off-by: Laurentiu Niculae <themelopeus@gmail.com>

**What type of PR is this?**

#591

**Automation added to e2e**:

```release-note

```


pkg/extensions/search/common/common_test.go
```diff
@@ -52,6 +57,50 @@ type ExpandedRepoInfoResp struct {
 	Errors           []ErrorGQL       `json:"errors"`
 }
 
+type GlobalSearchResultResp struct {
```
andaaron, line 60: Why do we need these definitions here instead of using the ones in models_gen.go?

pkg/extensions/search/common/oci_layout.go
```diff
@@ -238,15 +253,120 @@ func (olu OciLayoutUtils) checkCosignSignature(name, digest string) bool {
...
+func (olu BaseOciLayoutUtils) checkManifestSignature(name, digest, mediaType string) bool {
 	if !olu.checkCosignSignature(name, digest) {
 		return olu.checkNotarySignature(name, digest, mediaType)
 	}
 
 	return true
 }
 
-func (olu OciLayoutUtils) GetExpandedRepoInfo(name string) (RepoInfo, error) {
+func (olu BaseOciLayoutUtils) GetImageLastUpdated(repo string, manifestDigest godigest.Digest) time.Time {
```
shimish2, line 264: I think it is good idea to take imageInfo as an argument or define a new type because almost for all methods below we are first obtaining ImageInfo. 
```suggestion
func (olu OciLayoutUtils) GetImageLastUpdated(repo string, imageInfo ispec.Image) time.Time {
```
or 

``` 
type imageInfo ispec.ImageInfo 

func (imgInfo imageInfo) GetTimeStamp() time.Time {
        var timeStamp time.Time

	if len(imgInfo.History) != 0 {
		timeStamp = *imgInfo.History[0].Created
	} else {
		timeStamp = time.Time{}
	}

	return timeStamp
} 
```

Similar for all below methods as required.

pkg/extensions/search/schema.graphql
```diff
@@ -66,11 +66,56 @@ type LayerInfo {
...
+# We define an image as a pairing or a repo and a tag belonging to that repo
+type ImageSummary {
+     RepoName: String
+     Tag: String
+     LastUpdated: Time
+     IsSigned: Boolean
+     Size: String
+     Platform: OsArch
+     Vendor: String
+     Score: Int
```
rchincha, line 87: What is "Score" field used to indicate?

pkg/extensions/search/common/common_test.go
laurentiuNiculae: In tests, I need a structure to unmarshal the response into. The structures provided by models_gen.go have pointers as fields and the unmarshal library can't work with this kind of structure.

pkg/extensions/search/resolver.go
```diff
@@ -186,6 +188,187 @@ func (r *queryResolver) getImageListWithLatestTag(store storage.ImageStore) ([]*
...
+					layers = append(layers, &gql_generated.LayerSummary{
+						Digest: &layer.Digest,
+						Size:   &layer.Size,
+						Score:  &index,
+					})
+				}
+			}
+
+			imageSize := imageLayersSize + manifestSize + configSize
+			repoSize += manifestSize + configSize
```
andaaron, line 269: Per our discussion, please add tests for computing the image/repo/layer sizes.

andaaron (approved):
Looks good!

shimish2 (approved):
lgtm

PR #639, opened 2022-07-15
Update to latest gqlgen - 1.17.13

andaaron:
**What type of PR is this?**

Bug

We encountered some problems with using the existing folder structure,
but it looks like running the tooling with the latest versions works after
we regenerated the project using 'gql init' and refactoring to separate
the login previously in resolvers.go.

- the autogenerated code is now under the gql_generated folder
- the file resolvers.go now contains only the code which is not
rewritten by the gqlgen framework
- the file schema.resolvers.go is rewritten when gqlgen runs,
and we'll only keep there the actual resolvers matching query names
Changes we observed to schema.resolvers.go when gqlgen runs include
reordering methods, and renaming function parameters to match the
names used in schema.graphql
- we now have a gqlgen.yaml config file which governs the behavior of
gqlgen (can be tweaked to restructure the folder structure of the
generated code in the future)

Looks like the new graphql server has better validation
1 Returns 422 instead of 200 for missing query string - had to update tests
2 Correctly uncovered an error in a test for a bad `%` in query string.

As as result of 2, a `masked` bug was found in the way we check if images are
signed with Notary, the signatures were reasched for with the media type
of the image manifest itself instead of the media type for notation.
Fixed this bug, and improved error messages.
This bug would have also been reproducible with main branch if the bad `%`
in the test would have fixed.

Updated the linter to ignore some issues with the code which is
always rewritten when running:
`go run github.com/99designs/gqlgen@v0.17.13 generate`

Add a workflow to test gqlgen works and has no uncommitted changes

  Looks like the endpoint now returns 422 instead of 200 if no query is specified when making a GET request against the _search endpoint.


pkg/extensions/search/gql_generated/generated.go
```diff
@@ -754,73 +709,30 @@ func (ec *executionContext) field___Type_fields_args(ctx context.Context, rawArg
...
-		Object:     "CVE",
-		Field:      field,
-		Args:       nil,
-		IsMethod:   false,
-		IsResolver: false,
-	}
-
-	ctx = graphql.WithFieldContext(ctx, fc)
-	resTmp := ec._fieldMiddleware(ctx, obj, func(rctx context.Context) (interface{}, error) {
+	resTmp, err := ec.ResolverMiddleware(ctx, func(rctx context.Context) (interface{}, error) {
```
andaaron, line 728: I believe this is the actual fix, but with the previous folder structure, the old `_fieldMiddleware` function was still called.

rchincha: @shimish2 ^?

andaaron:
> @andaaron New schema looks good and clean, do you think is it possible to cover more lines of code by adding few more tests ?

I just reinitialized the project using `gqlgen init` using the existing schema and new yaml config file. No changes to the schema itself.

The only go changes were:
- move methods exposed as queries from resolver.go to schema.resolvers.go - which seems to be what the framework expects - it rewrites schema.resolvers.go when running `gql generate` reordering methods, and changing parameters names in signatures to match the names in the queries in schema.go.
- move generated code in a separate package, and update accordingly in places where it is used.
- fixed the tests now failing because of improved query validation, which in turn was also found a bug in the api

So with regards to coverage, there are no new functions, or new branches in existing functions - except in the code generated by gqlgen itself.

shimish2 (approved):
lgtm

PR #667, opened 2022-07-21
refactor(cache): Add database interface for blob paths + refactor BoltDB to use interface

chofnar:
**What type of PR is this?**

enhancement

Part of #564 

This PR adds a database interface that allows the usage of vendor-specific database services by implementing their drivers.
The usage of BoltDB was refactored to use this new interface.

**Other comments**:
This PR is a result of the split of #588, which now depends on this one. The dynamodb integration can be found [here](https://github.com/chofnar/zot/pull/66)


pkg/storage/database/database_factory_test.go
```diff
@@ -0,0 +1,33 @@
...
+		_, err := database.Create("", storage.BoltDBDriverParameters{}, log)
+		So(err, ShouldNotBeNil)
+
+		So(func() {
+			database.Register("nilfactory", nil)
+		}, ShouldPanic)
+
+		So(func() {
+			database.Register("boltdb", &mockDriverFactory{})
+		}, ShouldPanic)
```
andaaron, line 31: Why not add a database.Create() test for boltdb after this?

pkg/api/controller.go
```diff
@@ -215,9 +217,81 @@ func (c *Controller) Run(reloadCtx context.Context) error {
 	return server.Serve(listener)
 }
 
+// Will create a cache database using the configuration file.
+// Pass a non-nil first argument to override the config (for testing purposes, mostly).
+func (c *Controller) CreateCacheDatabaseDriver(configOverride interface{}, log log.Logger) (database.Driver, error) {
```
andaaron, line 222: Why use panics in this functions, if it potentially returns errors?

andaaron (commented):
Can you explain why we need this registering mechanism? Other drivers would be zot extensions with different build tags?

pkg/api/controller.go
chofnar: Changed logic to return errors instead

pkg/storage/database/database_factory_test.go
chofnar: Added

rchincha:
This PR is awaiting benchmark results. Both latency and throughput.

andaaron:
> This PR is awaiting benchmark results. Both latency and throughput.

This PR includes just adding an interface and refactoring boltdb. It doesn't include dynamodb related changes.

pkg/storage/database/database_driver.go
```diff
@@ -0,0 +1,23 @@
...
+
+	// Check if blob exists in database.
+	HasBlob(digest, path string) bool
+
+	// Delete a blob from the database.
+	DeleteBlob(digest, path string) error
+}
+
+type Blob struct {
+	Digest   string   `dynamodbav:"Digest,string"`
```
andaaron, line 21: Why use `dynamodbav` here? Isn't this generic?

pkg/storage/s3/s3.go
```diff
@@ -936,7 +945,7 @@ func (is *ObjectStorage) FinishBlobUpload(repo, uuid string, body io.Reader, dig
 	is.Lock(&lockLatency)
 	defer is.Unlock(&lockLatency)
 
-	if is.dedupe && is.cache != nil {
+	if is.dedupe && fmt.Sprintf("%v", is.cache) != fmt.Sprintf("%v", nil) {
```
rchincha, line 948: fmt.Sprintf() in the fast path?
also what is this check supposed to do? is.cache != nil check doesn't suffice?

pkg/storage/database/database_driver.go
chofnar: That is correct, I missed removing that.

pkg/storage/s3/s3.go
chofnar: > is.cache != nil check doesn't suffice?
We ran into this a while back, it's a bit of a pickle. is.cache is now an interface, and there's a catch there: when you check if an interface == nil, you check its type, not its value, and since it's not a nil interface, the check fails. You need to grab the underlying value, and you either do that with reflect, or with the hackish method you see there. 
More about this https://trstringer.com/go-nil-interface-and-interface-with-nil-concrete-value/

chofnar:
> Can you explain why we need this registering mechanism? Other drivers would be zot extensions with different build tags?

This is how [distribution does it](https://github.com/distribution/distribution/blob/main/registry/storage/driver/factory/factory.go). I reckon it's the easiest way to register the available factories, by calling the Register function from the Driver's init function, since the Create func uses the same map to return the Drivers where they're needed

andaaron (dismissed):
LGTM

pkg/storage/cache.go
```diff
@@ -17,20 +19,70 @@ const (
...
+	if parameters.Name != "" {
+		params.Name = parameters.Name
+	}
+	params.UseRelPaths = parameters.UseRelPaths
+
+	return NewCache(*params, log), nil
+}
+
+func NewCache(parameters BoltDBDriverParameters, log zlog.Logger) *Driver {
+	err := os.MkdirAll(parameters.RootDir, 0o700) //nolint:gomnd
```
rchincha, line 78: pkg/storage/local.go
43:     DefaultDirPerms  = 0o700

^ use this?

pkg/storage/cache.go
```diff
@@ -17,20 +19,70 @@ const (
...
+func (factory *boltDBDriverFactory) Create(parameters interface{}, log zlog.Logger) (cachedb.Driver, error) {
+	properParameters, ok := parameters.(BoltDBDriverParameters)
+	if !ok {
+		panic("Failed type assertion")
+	}
+
+	return FromParameters(properParameters, log)
+}
+
+// nolint:gochecknoinits
```
rchincha, line 53: We have managed to do without init() funcs so far. Let's not start now. Call an func inline from initialization path directly.

pkg/storage/cachedb/cachedb_factory.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	return driverFactory.Create(parameters, log)
+}
+
+// Register makes a database driver available by the provided name.
+// If Register is called twice with the same name or if driver factory is nil, it panics.
+// Additionally, it is not concurrency safe. Most Database Drivers call this function
+// in their init() functions.
+func Register(name string, factory DriverFactory) {
```
rchincha, line 34: We are not supporting any "dynamic" registration right? Just move "driverFactories" into here and that will also fix that nolint.

pkg/storage/cachedb/cachedb_factory.go
```diff
@@ -0,0 +1,45 @@
+package cachedb
+
+import (
+	"fmt"
+
+	"zotregistry.io/zot/errors"
+	zlog "zotregistry.io/zot/pkg/log"
+)
+
+// nolint:gochecknoglobals
```
rchincha, line 10: see comment below.

rchincha:
There is a "pkg/api/*db" file in the PR - do you intend to merge this file? or was this for a test? if so, create/generate it from the Makefile or some setup routine.

rchincha:
The code refactor is not very clear. I think we want to say:
1) zot has two major "states" - datapath blobs and metadata (which we call cache) for dedupe, gc, user-prefs etc
2) zot creates a cache interface like so: Create(name, driverName) -> cache
3) All "cache" instances implement the same api except over different backends.

For 2), put the interface under pkg/storage/cache.go, and move implementations out into their own files.
I think it is scattered currently around cachedb.go etc.

pkg/storage/cache.go
chofnar: Fixed

pkg/storage/cachedb/cachedb_factory.go
chofnar: Got rid of inits and moved the driver initialization to the factory file

pkg/storage/boltdb/boltdb.go
```diff
@@ -0,0 +1,265 @@
...
+	useRelPaths bool // weather or not to use relative paths, should be true for filesystem and false for s3
+}
+
+type DriverParameters struct {
+	RootDir, Name string
+	UseRelPaths   bool
+}
+
+const (
+	defaultRootDir     = "/tmp/zot/boltDB"
```
andaaron, line 29: I don't really like this path. Can we have one relative to the root dir? We had one in the old implementation.

andaaron (dismissed):
LGTM

pkg/storage/boltdb/boltdb.go
chofnar: Removed the default values as they aren't really ever used. From the controller func CreateCacheDatabaseDriver, if an incomplete config is used, an errors.ErrBadConfig with nil driver is returned. If no configuration is provided, it will default to the storage rootDir.

examples/config-boltdb.json
```diff
@@ -0,0 +1,19 @@
+{
+    "distSpecVersion": "1.0.1-dev",
+    "storage": {
+        "rootDirectory": "/tmp/zot",
+        "cacheDatabaseDriver": {
+            "name": "boltdb",
+            "cacheName": "cache1",
+            "rootDirectory": "/tmp/zot/cachedb",
+            "useRelPaths": "False"
```
peusebiu, line 9: shouldn't this be bool?

peusebiu (dismissed):
LGTM

examples/config-boltdb.json
chofnar: Rewrote config and CreateCacheDatabaseDriver to use interface instead, and used type assertion. Also updated the config

peusebiu (approved):
LGTM

chofnar:
Commit ID with the caching on subpaths functionality: 49e4464a8461b6ed64a001910bdadcde2970297f

pkg/api/controller.go
```diff
@@ -5,9 +5,9 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"io/ioutil"
```
andaaron: Let's not use `io/ioutil`

shimish2 (commented):
Please fix the lint

andaaron:
There are some changes to be made which didn't get into the PR yet (move config validation in root.go, make sure each image store has it's own blob cache to avoid hardlink attempts between blobs in different substores)

pkg/storage/s3/s3.go
```diff
@@ -86,15 +91,31 @@ func NewImageStore(rootDir string, cacheDir string, gc bool, gcDelay time.Durati
 		linter:           linter,
 	}
 
-	cachePath := path.Join(cacheDir, CacheDBName+storage.DBExtensionName)
+	cachePath := path.Join(cacheDir, CacheDBName+constants.DBExtensionName)
 
 	if dedupe {
-		imgStore.cache = storage.NewCache(cacheDir, CacheDBName, false, log)
+		if len(cacheDriver) != 0 {
```
andaaron, line 97: Let's not do this magic. Pass a single valid initialized cacheDriver in the function parameters.
There are too many fallbacks to boltdb in this PR.

pkg/storage/s3/s3.go
```diff
@@ -86,15 +91,31 @@ func NewImageStore(rootDir string, cacheDir string, gc bool, gcDelay time.Durati
...
+				RootDir:     cacheDir,
+				Name:        CacheDBName,
+				UseRelPaths: false,
+			}
+
+			imgStore.cache, _ = cachedbfactory.Create("boltdb", params, log)
+		}
 	} else {
 		// if dedupe was used in previous runs use it to serve blobs correctly
 		if _, err := os.Stat(cachePath); err == nil {
```
andaaron, line 110: I don't think this logic should be here? Can do this check when we validate the config, before initializing the controller and storage?

pkg/storage/boltdb/boltdb.go
```diff
@@ -0,0 +1,321 @@
...
+	params := &DriverParameters{
+		RootDir:     parameters.RootDir,
+		Name:        parameters.Name,
+		UseRelPaths: parameters.UseRelPaths,
+	}
+
+	return NewCache(*params, log), nil
+}
+
+func NewCache(parameters DriverParameters, log zlog.Logger) *Driver {
```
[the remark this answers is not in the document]
peusebiu: Yes, looks good

pkg/api/controller.go
```diff
@@ -418,6 +425,81 @@ func compareImageStore(root1, root2 string) bool {
...
+		RootDirectory:       global.RootDirectory,
+		StorageDriver:       global.StorageDriver,
+		CacheDatabaseDriver: global.CacheDatabaseDriver,
+	}
+}
+
+func CreateCacheDatabaseDriver(storageConfig interface{}, log log.Logger) cachedb.Driver {
+	absStorageConfig := abstractizeStorageConfig(storageConfig)
+	if absStorageConfig.Dedupe {
+		switch cacheType, _ := absStorageConfig.CacheDatabaseDriver["type"].(string); cacheType {
```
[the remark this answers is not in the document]
chofnar: Added to validation in root.go

pkg/cli/root.go
```diff
@@ -232,6 +232,22 @@ func validateStorageConfig(cfg *config.Config) error {
 	return nil
 }
 
+func validateCacheDBConfig(cfg *config.Config) error {
+	if cfg.Storage.Dedupe && cfg.Storage.CacheDatabaseDriver == nil {
+		log.Info().Msgf("Dedupe set to true for %s but no cachedb config provided. Will default to localstorage (boltdb)",
```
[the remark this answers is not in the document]
chofnar: Added

pkg/storage/local.go
```diff
@@ -134,7 +138,17 @@ func NewImageStore(rootDir string, gc bool, gcDelay time.Duration, dedupe, commi
 	}
 
 	if dedupe {
-		imgStore.cache = NewCache(rootDir, "cache", true, log)
+		if len(cacheDriver) != 0 {
```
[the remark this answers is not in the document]
chofnar: Added

pkg/storage/local.go
```diff
@@ -0,0 +1,2057 @@
+package storage
```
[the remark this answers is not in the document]
chofnar: Leftover file after rebase. Removed.

examples/config-boltdb.json
```diff
@@ -0,0 +1,20 @@
+{
+    "distSpecVersion": "1.0.1-dev",
+    "storage": {
+        "rootDirectory": "/tmp/zot",
+        "dedupe": true,
+        "cache": {
+            "type": "boltdb",
+            "cacheName": "cache1",
+            "rootDirectory": "/tmp/zot/cachedb",
```
rchincha, line 9: Why do we have another rootDirectory param here?

pkg/api/controller.go
```diff
@@ -420,6 +427,74 @@ func compareImageStore(root1, root2 string) bool {
 	return isSameFile
 }
 
+type abstractStorageDriver struct {
```
rchincha, line 430: Let's try and **not** use the word "abstract" here

examples/config-boltdb.json
chofnar: Removed

pkg/api/controller.go
chofnar: Got rid of the type entirely, used config.StorageConfig instead

pkg/api/config/config.go
```diff
@@ -19,13 +19,14 @@ var (
...
-	GCInterval    time.Duration
-	StorageDriver map[string]interface{} `mapstructure:",omitempty"`
+	RootDirectory       string
+	GC                  bool
+	Dedupe              bool
+	Commit              bool
+	GCDelay             time.Duration
+	GCInterval          time.Duration
+	StorageDriver       map[string]interface{} `mapstructure:",omitempty"`
+	CacheDatabaseDriver map[string]interface{} `mapstructure:"cache,omitempty"`
```
rchincha, line 29: Just call it "CacheDriver", that it is a db is probably implementation detail

pkg/storage/cachedb/cachedb_driver.go
```diff
@@ -0,0 +1,18 @@
+package cachedb
```
rchincha, line 1: As an interface, would have preferred this to be 
pkg/storage/cache.go

pkg/storage/boltdb/boltdb.go
```diff
@@ -0,0 +1,321 @@
...
+	zlog "zotregistry.io/zot/pkg/log"
+	"zotregistry.io/zot/pkg/storage/cachedb"
+	"zotregistry.io/zot/pkg/storage/constants"
+)
+
+type Driver struct {
+	rootDir     string
+	db          *bbolt.DB
+	log         zlog.Logger
+	useRelPaths bool // weather or not to use relative paths, should be true for filesystem and false for s3
```
rchincha, line 20: "whether"

rchincha: and this should be derived automatically from the type of storage

pkg/storage/cache.go
```diff
@@ -1,297 +0,0 @@
-package storage
```
rchincha, line 1: why delete this file here?

chofnar: The file contained the boltdb implementation, which was moved to its own folder

pkg/api/config/config.go
chofnar: Renamed to CacheDriver

pkg/storage/cachedb/cachedb_driver.go
chofnar: Moved to said location

pkg/storage/boltdb/boltdb.go
chofnar: Removed useRelPaths from config, it gets set from CreateDatabaseDriver according to storage type

pkg/api/config/config.go
```diff
@@ -21,8 +21,9 @@ var (
 
 type StorageConfig struct {
 	RootDirectory string
-	GC            bool
 	Dedupe        bool
+	CloudCache    bool
```
rchincha, line 25: No...call this LocalCache preferably (or RemoteCache - could be Cloud, NFS, anything else)

pkg/api/config/config.go
```diff
@@ -94,13 +95,7 @@ type LogConfig struct {
...
 
 type GlobalStorageConfig struct {
-	Dedupe        bool
-	GC            bool
-	Commit        bool
-	GCDelay       time.Duration
-	GCInterval    time.Duration
-	RootDirectory string
-	StorageDriver map[string]interface{} `mapstructure:",omitempty"`
+	StorageConfig `mapstructure:",squash"`
```
rchincha, line 98: 👍

pkg/storage/cache/factory/factory.go
```diff
@@ -0,0 +1,36 @@
+package factory
```
rchincha, line 1: Instead of "factory", maybe use "backend" instead

pkg/api/controller.go
```diff
@@ -25,6 +25,10 @@ import (
 	"zotregistry.io/zot/pkg/log"
 	"zotregistry.io/zot/pkg/scheduler"
 	"zotregistry.io/zot/pkg/storage"
+	"zotregistry.io/zot/pkg/storage/cache"
+	"zotregistry.io/zot/pkg/storage/cache/boltdb"
+	cachedbfactory "zotregistry.io/zot/pkg/storage/cache/factory"
```
rchincha, line 30: s/cachedb/cache/g or "cachebackend"
Let's not indicate that the cache backend is always only a db.

pkg/storage/cache/cache.go
```diff
@@ -0,0 +1,18 @@
+package cache
+
+type Cache interface {
```
rchincha, line 3: 👍

pkg/api/controller.go
andaaron: I would say this is not really a cache, it is a DB. We need to persist data there - for local this cache may be optional, but for S3 it is mandatory to persist as we can't recover the content otherwise.

chofnar: Moved factory to backend, renamed cachedbfactory to cacheBackend

pkg/storage/cache/factory/factory.go
chofnar: Moved factory to backend, renamed cachedbfactory to cacheBackend

pkg/api/config/config.go
chofnar: Renamed to RemoteCache

examples/config-boltdb.json
```diff
@@ -0,0 +1,15 @@
+{
+    "distSpecVersion": "1.0.1-dev",
+    "storage": {
+        "rootDirectory": "/tmp/zot",
+        "dedupe": true,
+        "cloudCache": false
```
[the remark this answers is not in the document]
chofnar: I thought that since false is the zero value in go, remoteCache would work better to default to false if none specified, so it defaults to local

examples/config-boltdb.json
```diff
@@ -0,0 +1,15 @@
+{
+    "distSpecVersion": "1.0.1-dev",
+    "storage": {
+        "rootDirectory": "/tmp/zot",
+        "dedupe": true,
+        "remoteCache": false
```
rchincha, line 6: this defaults to false if not specified i suppose?

pkg/api/controller_test.go
```diff
@@ -106,6 +107,34 @@ func TestNew(t *testing.T) {
...
+
+		err := os.Chmod(dir, 0o000)
+		if err != nil {
+			panic(err)
+		}
+
+		driver := api.CreateCacheDatabaseDriver(conf.Storage.StorageConfig, log)
+		So(driver, ShouldBeNil)
+
+		// remoteCache = true
```
rchincha, line 129: remove this comment

pkg/storage/constants/constants.go
```diff
@@ -0,0 +1,21 @@
+package constants
```
rchincha, line 1: does this need its own package?
why not under pkg/storage/constants.go itself under storage pkg?

pkg/api/controller_test.go
chofnar: Removed

examples/config-boltdb.json
chofnar: Yes

pkg/storage/constants/constants.go
chofnar: Moved into storage package

pkg/api/config/config.go
```diff
@@ -22,8 +22,9 @@ var (
 
 type StorageConfig struct {
 	RootDirectory string
-	GC            bool
 	Dedupe        bool
+	RemoteCache   bool
```
rchincha, line 26: If "RemoteCache" is **NOT** explicitly specified, then default behavior is as follows.
* local storage driver -> RemoteCache=false
* remote storage driver (s3) -> RemoteCache=true

ValidateConfig()
* local storage driver, but **explicit** RemoteCache=true, this is not valid, so error out
* maintain pairing (s3 storage, dynamodb), tomorrow, different pairing for (gcp storage, gcp db) etc

rchincha: (local storage, local db) -> dedupe
(remote storage, local db) -> dedupe but not stateful cluster
(remote storage, remote db) -> dedupe and stateful cluster

pkg/storage/constants/constants.go
chofnar: And I had to move them back due to the new changes.

pkg/storage/cache.go
```diff
@@ -1,301 +1,41 @@
...
-	if path == "" {
-		c.log.Error().Err(errors.ErrEmptyValue).Str("digest", digest.String()).Msg("empty path provided")
-
-		return errors.ErrEmptyValue
-	}
+func Create(dbtype string, parameters interface{}, log zlog.Logger) (Cache, error) {
+	switch dbtype {
+	case "boltdb":
+		{
+			boltDriverFactory := cache.BoltDBDriverFactory{}
```
rchincha, line 32: Just make this boltdb.Create() <- public method, remove all Factory references

pkg/storage/constants/constants.go
rchincha: pkg/storage/constants.go

so "import pkg/storage" is sufficient?

pkg/storage/cache.go
chofnar: Modified as requested

pkg/storage/constants/constants.go
chofnar: I also needed to move the cache driver interface to the cache package, as it would have created some more cyclic imports on the dynamodb implementation (in the cache package) if left in storage

rchincha (approved):
lgtm

PR #671, opened 2022-07-22
RepoSummary has a new attribute NewestTag of type ImageSummary

### LINKED ISSUE #666: GraphQL: RepoSummary to have an attribute NewestTag of type ImageSummary
**Is your feature request related to a problem? Please describe.**
ImageListWithLatestTag currently returns a list of ImageInfo objects. It needs to return consistent results with the API used for Global search as the same information will be used by the UI in the same type or cards.
So we need to update RepoSummary to include the data which right now is present in ImageInfo, but missing from RepoSummary (information on the latest tag in that specific repo).

**Describe the solution you'd like**
We will update the RepoSummary to include the missing data currently in  ImageInfo. Add a RepoSummary attribite NewestTag of type ImageSummary, This attribute would contain the newest tag/manifest pushed to the repo.

**Additional context**
Needed for zui.

andaaron:
**What type of PR is this?**
feature

#666

ImageListWithLatestTag currently returns a list of ImageInfo objects.
It needs to return consistent results with the API used for Global search as the same information will be used by the UI in the same type or cards.
So we need to update RepoSummary to include the data which right now is present in ImageInfo, but missing from RepoSummary (information on the latest tag in that specific repo).
Will update return type of ImageListWithLatestTag in a later PR (issue tracked in a separate GH issue)

Existing and updated automated tests.

No

Yes it contains extra information to be used by zot ui.

Closes #666

### DIFF
--- a/pkg/extensions/search/common/common_test.go
+++ b/pkg/extensions/search/common/common_test.go
@@ -87,15 +87,17 @@ type ImageSummary struct {
 	Platform    OsArch    `json:"platform"`
 	Vendor      string    `json:"vendor"`
 	Score       int       `json:"score"`
+	IsSigned    bool      `json:"isSigned"`
 }
 
 type RepoSummary struct {
-	Name        string    `json:"name"`
-	LastUpdated time.Time `json:"lastUpdated"`
-	Size        string    `json:"size"`
-	Platforms   []OsArch  `json:"platforms"`
-	Vendors     []string  `json:"vendors"`
-	Score       int       `json:"score"`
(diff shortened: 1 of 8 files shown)


rchincha (approved):
lgtm

shimish2 (commented):
looks good can we use LatestTag instead of NewestTag

andaaron:
> looks good can we use LatestTag instead of NewestTag

I think it would be confusing, since tag string is not necessarily "latest" if could be anything.

shimish2 (approved):

PR #680, opened 2022-07-26
Fixes/Improvements to pkg/cli/stress_test.go

andaaron:
**What type of PR is this?**
cleanup

pkg/cli/stress_test.go fails on all PRs

- Decrease RLIMIT_NOFILE and the number of goroutines used to reach this limit (from 512 to 100)
- Reset RLIMIT_NOFILE to the initial value before the test finishes
- Remove panic
- Use temporary dir managed by test framework
- Swith to using test logging in pkg/cli/stress_test.go
- Execute commands without `bash -c` in pkg/cli/stress_test.go

First item is needed as the GH runner seems to stop the test if stressed too much.
The lower number is still good enough to reproduce the test conditions

### DIFF
--- a/pkg/cli/stress_test.go
+++ b/pkg/cli/stress_test.go
@@ -22,7 +22,7 @@ import (
 )
 
 const (
-	MaxFileDescriptors = 512
+	MaxFileDescriptors = 100
 	WorkerRunningTime  = 60 * time.Second
 )
 
@@ -36,7 +36,7 @@ func TestSressTooManyOpenFiles(t *testing.T) {
 		// all the log files to figure out what happened in this test (zot log file, scrub output, storage rootFS tree)
 		SetDefaultFailureMode(FailureContinues)
 
-		err := setMaxOpenFilesLimit(MaxFileDescriptors)
+		initialLimit, err := setMaxOpenFilesLimit(MaxFileDescriptors)
 		So(err, ShouldBeNil)
 
 		port := test.GetFreePort()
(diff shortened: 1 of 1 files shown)


rchincha (approved):
lgtm

rchincha:
let @peusebiu also review this.

pkg/cli/stress_test.go
```diff
@@ -129,6 +125,9 @@ func TestSressTooManyOpenFiles(t *testing.T) {
 		}
 		wg.Wait()
 
+		_, err = setMaxOpenFilesLimit(initialLimit)
```
alexstan12, line 128: This is for my own curiosity - is it needed to reset the resource limit, since the previously set value will be available only for the current process ?

alexstan12 (approved):
Looks good

pkg/cli/stress_test.go
```diff
@@ -182,42 +186,45 @@ func worker(id int, zotPort, rootDir string) {
...
-	return nil
+	return initialLimit, nil
 }
 
 func startServer(c *api.Controller) {
 	// this blocks
 	ctx := context.Background()
 	if err := c.Run(ctx); err != nil {
-		return
+		fmt.Printf("\nerror on starting zot server: %s\n", err)
```
bogdanbiv, line 221: could you maybe move `\n` to the previous string/s

andaaron: What previous message? Can you guarantee a specific one?
printf because that was used for printing errors in the rest of the file.

peusebiu (approved):
LGTM

PR #683, opened 2022-07-27
Fix permissions for image sync and stale workflows

andaaron:
Looks like read|write is not a correct value:
https://github.com/project-zot/zot/actions/runs/2743961177
https://github.com/project-zot/zot/actions/runs/2743965531

Write should include both, so let's try to use that.


peusebiu (approved):
lgtm

peusebiu (approved):

peusebiu (dismissed):

PR #695, opened 2022-08-02
replace dependency of tagsInfo with list of manifests

alexstan12:
replace dependency of tagsInfo with list of manifests since only the manifest Digest is needed

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/search/resolver.go
+++ b/pkg/extensions/search/resolver.go
@@ -12,6 +12,7 @@ import (
 	godigest "github.com/opencontainers/go-digest"
 	"zotregistry.io/zot/pkg/log" // nolint: gci
 
+	ispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"zotregistry.io/zot/pkg/extensions/search/common"
 	cveinfo "zotregistry.io/zot/pkg/extensions/search/cve"
 	digestinfo "zotregistry.io/zot/pkg/extensions/search/digest"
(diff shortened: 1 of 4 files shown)


pkg/extensions/search/common/common_test.go
```diff
@@ -943,6 +943,20 @@ func TestBaseOciLayoutUtils(t *testing.T) {
 		So(size, ShouldBeZeroValue)
 	})
 
+	Convey("GetImageManifests", t, func() {
```
andaaron, line 946: In this case the test name is not representative of the actual behavior your are expecting and under what conditions.

pkg/extensions/search/resolver.go
```diff
@@ -216,39 +216,45 @@ func globalSearch(repoList []string, name, tag string, olu common.OciLayoutUtils
...
 
-		repoPlatforms := make([]*gql_generated.OsArch, 0, len(tagsInfo))
+		repoPlatforms := make([]*gql_generated.OsArch, 0, len(manifests))
 		repoVendors := make([]*string, 0, len(repoInfo.Manifests))
 
 		for i, manifest := range repoInfo.Manifests {
 			imageLayersSize := int64(0)
 
-			imageBlobManifest, err := olu.GetImageBlobManifest(repo, godigest.Digest(tagsInfo[i].Digest))
+			imageBlobManifest, err := olu.GetImageBlobManifest(repo, manifests[i].Digest)
```
andaaron, line 247: Why use `manifests[i].Digest` instead of using the digest from the `manifest` variable. I don't think there's a guarantee `repoInfo.Manifests` are in the same order as in `manifests`.
This was probably an issue before, but let's fix it now since we found it.

pkg/extensions/search/resolver.go
```diff
@@ -216,39 +216,45 @@ func globalSearch(repoList []string, name, tag string, olu common.OciLayoutUtils
...
+		repoInfo, err := olu.GetExpandedRepoInfo(repo)
 		if err != nil {
-			log.Error().Err(err).Msgf("can't get tags info for repo: %s", repo)
+			log.Error().Err(err).Msgf("can't get repo info for repo: %s", repo)
 
 			continue
 		}
 
-		repoInfo, err := olu.GetExpandedRepoInfo(repo)
+		manifests, err := olu.GetImageManifests(repo)
```
andaaron, line 226: Wondering why we need this, don't we have all the needed data in `repoInfo.Manifests`?

alexstan12: It is the same reason why we needed tagsInfo in the first place. repoInfo has a list of manifests with encoded digests, while olu.GetImageManifests(repo) gives us a list of manifests that have full digests ( i.e. digests that include the algorithm prefix).

pkg/extensions/search/common/common_test.go
alexstan12: Indeed, removed

pkg/extensions/search/resolver.go
alexstan12: Main logic is now changed, we use only a list of manifests.

andaaron (approved):

rchincha (approved):
lgtm

PR #696, opened 2022-08-02
Proposal for unified graphql API

andaaron:
GraphQL API proposal for 1st ZOT release. 

**Which issue does this PR fix**:
Several issues requesting graphql API changes need an `API` specification.
This is a proposal, not to be merged.

**Does this PR introduce any user-facing change?**:
Yes, it determines the data shown to the users in ZUI/ZLI.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #663, opened 2022-07-21
GraphQL: Replace ImageListWithLatestTag with RepoListWithNewestImage and return a list of RepoSummary objects


andaaron:
**Is your feature request related to a problem? Please describe.**
ImageListWithLatestTag currently returns a list of ImageInfo objects. It needs to return consistent results with the API used for Global search as the same information will be used by the UI in the same type or cards.

**Describe the solution you'd like**
We need to update ImageListWithLatestTag to return a list of RepoSummary objects.
But first we need to include missing data into RepoSummary objects: https://github.com/project-zot/zot/issues/666

Note point 1 will be necessary also for Global/Freeform search as the info is currently missing.

**Additional context**
Needed for zui.

PR #697, opened 2022-08-03
GraphQL: RepoListWithNewestImage to return list of RepoSummary

chofnar:
**What type of PR is this?**
ui

#663

From #663:

> ImageListWithLatestTag currently returns a list of ImageInfo objects. It needs to return consistent results with the API used for Global search as the same information will be used by the UI in the same type or cards.

Needed schema changes from #696 

### DIFF
--- a/pkg/extensions/search/common/common_test.go
+++ b/pkg/extensions/search/common/common_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/sigstore/cosign/cmd/cosign/cli/sign"
 	. "github.com/smartystreets/goconvey/convey"
 	"gopkg.in/resty.v1"
+	zerr "zotregistry.io/zot/errors"
 	"zotregistry.io/zot/pkg/api"
 	"zotregistry.io/zot/pkg/api/config"
 	"zotregistry.io/zot/pkg/api/constants"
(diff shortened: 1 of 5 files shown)


andaaron (changes requested):
Actually can you update the implementation to match what Alex did for https://github.com/project-zot/zot/pull/695?
See my comments there about iterating over manifests and tags using the same index, which may or may not be the same in practice.

andaaron (dismissed):

andaaron (commented):
In case there are errors, let's have the query return a partial list of "good" repos, and a list of all the errors encountered.
We don't want to break the entire query results for a few bad repos/images.

pkg/extensions/search/common/common_test.go
```diff
@@ -55,9 +56,9 @@ var (
 	subRootDir string
 )
 
-type ImgResponsWithLatestTag struct {
-	ImgListWithLatestTag ImgListWithLatestTag `json:"data"`
-	Errors               []ErrorGQL           `json:"errors"`
+type RepoResponseWithNewestImage struct {
```
rchincha, line 59: Call this `RepoWithNewestImageResponse` instead

chofnar: Renamed

andaaron (approved):

rchincha:
Pls fix the merge conflicts and resubmit.

andaaron (approved):

PR #700, opened 2022-08-03
Adding a task scheduler for background tasks

Andreea-Lupu:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


rchincha:
The general design is that any extension should be able to create a Task (which should be an interface) to the TaskScheduler.
The TaskScheduler shouldn't know what the task is, just that it handles it in the background and returns a status or calls a callback.

pkg/extensions/scheduler/task-scheduler.go
```diff
@@ -0,0 +1,94 @@
...
+				return
+			case task := <-scheduler.tasksQ:
+				scheduler.wg.Add(1)
+				status, callback := task.DoWork(scheduler.wg)
+
+				if callback != nil {
+					_ = callback(status)
+				}
+
+				if status == Running {
```
andaaron, line 65: Why is this needed? Under what condition is the status still Running at this line? Could it be any other value?

When is a task moved from running to a different status? I think that logic should be encapsulated in this package.

pkg/extensions/scheduler/task-scheduler.go
```diff
@@ -0,0 +1,94 @@
...
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				scheduler.StopScheduler()
+
+				return
+			case task := <-scheduler.tasksQ:
+				scheduler.wg.Add(1)
+				status, callback := task.DoWork(scheduler.wg)
```
andaaron, line 59: So the purpose of the scheduler is to execute tasks one by one?

pkg/extensions/extension_scrub.go
```diff
@@ -4,30 +4,107 @@
...
+				for {
+					if errIterator != nil && !errors.Is(errIterator, io.EOF) {
+						break
+					}
+
+					if repo == "" {
+						break
+					}
+
+					scrubTask := scrub.NewScrubTask(periodicScrubT.imgStore, repo, periodicScrubT.log)
```
andaaron, line 95: Can we have the task DoWork in a generic method, and the scrub logic be passed as a parameter of type function to it? Most of this method looks like code which would be duplicated in all DoWork implementations.

We would pass the actual "work"/scrub function as well as the callback function(s ?) for success or failure to the generic DoWork function. These work/context functions would be called using the same context as parameter (which would be instantiated just before calling DoWork for this specific task and passed on as a parameter), so if one is cancelled, all are cancelled.

pkg/extensions/extension_scrub.go
```diff
@@ -4,30 +4,107 @@
...
+	return periodicScrub
+}
+
+func (periodicScrubT *periodicScrubTask) DoWork(wgScheduler *goSync.WaitGroup) (scheduler.Status, scheduler.Cb) {
+	ticker := time.NewTicker(periodicScrubT.interval)
+
+	go func() {
+		for {
+			select {
+			case <-periodicScrubT.cancel:
```
andaaron, line 75: Why not use a context - also for handling timeouts and cancellation - see some examples here https://www.sohamkamani.com/golang/context-cancellation-and-values/

pkg/extensions/scrub/scrub.go
```diff
@@ -39,5 +44,47 @@ func RunScrubRepo(imgStore storage.ImageStore, repo string, log log.Logger) {
...
 	}
 
-	log.Info().Msg(fmt.Sprintf("scrub completed for %s", path.Join(imgStore.RootDir(), repo)))
+	log.Info().Msg(fmt.Sprintf("scrub successfully completed for %s", path.Join(imgStore.RootDir(), repo)))
+
+	return nil
+}
+
+type Task struct {
+	*scheduler.AbstractTask
```
andaaron, line 53: What do we need the abstract task for?

pkg/extensions/scheduler/task-scheduler.go
```diff
@@ -0,0 +1,88 @@
+package scheduler
```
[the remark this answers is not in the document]
andaaron: This will be under pkg/scheduler.

pkg/scheduler/scheduler.go
```diff
@@ -0,0 +1,244 @@
...
+		interval:      interval,
+		done:          false,
+		priority:      priority,
+		lastRepo:      "",
+		taskGenerator: taskGenerator,
+	}
+
+	scheduler.generatorsLock.Lock()
+	heap.Push(&scheduler.generators, newGenerator)
+	scheduler.generatorsLock.Unlock()
```
peusebiu, line 243: use defer :D

pkg/scheduler/scheduler.go
```diff
@@ -0,0 +1,244 @@
...
+				gen.done = false
+			} else {
+				return
+			}
+		} else {
+			return
+		}
+	}
+
+	task, repo, err := gen.taskGenerator.GenerateTask(gen.lastRepo)
```
peusebiu, line 217: should we call this item instead of repo? it can be anything right?

pkg/extensions/extension_scrub.go
```diff
@@ -9,25 +9,59 @@ import (
...
+	imgStore storage.ImageStore
+	log      log.Logger
+}
+
+func (gen taskGenerator) GenerateTask(lastRepo string) (scheduler.Task, string, error) {
+	var repo string
+
+	var err error
+
+	if lastRepo == "" {
```
peusebiu, line 60: lastRepo can be a taskGenerator attribute?

andaaron (commented):
Really nice work!
I left some comments on the doc.

pkg/storage/local.go
```diff
@@ -360,6 +361,106 @@ func (is *ImageStoreLocal) GetRepositories() ([]string, error) {
...
+	_, err := os.ReadDir(dir)
+	if err != nil {
+		is.log.Error().Err(err).Msg("failure walking storage root-dir")
+
+		return "", err
+	}
+
+	found := false
+	store := ""
+	err = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
```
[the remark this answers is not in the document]
andaaron: Let's stick to this for now. I'd rather we have this feature working instead of debating these details.

andaaron (approved):

rchincha (approved):
lgtm

Issue #660, opened 2022-07-20
New API to list dependencies for a specific Image


andaaron:
**Is your feature request related to a problem? Please describe.**
One of the views in zui will list image dependencies.
We need a query which given an image name or image digest would return a list of all images which share all their layers with the current image.

PR #457, opened 2022-03-11
[cloud usecase] zot cli should not download manifests when listing images

roxanaN:
Signed-off-by: Roxana Nemulescu <roxana.nemulescu@gmail.com>

**What type of PR is this?**

**Which issue does this PR fix**:

**What does this PR do / Why do we need it**:

**If an issue # is not available please add repro steps and logs from IPAMD/CNI showing the issue**:

**Testing done on this change**:

**Automation added to e2e**:

**Will this break upgrades or downgrades. Has updating a running cluster been tested?**:

**Does this change require updates to the CNI daemonset config files to work?**:

**Does this PR introduce any user-facing change?**:

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #713, opened 2022-08-08
list all images that have are derived from the given image

aokirisaki:
Signed-off-by: Lisca Ana-Roberta <ana.kagome@yahoo.com>

**What type of PR is this?**

#660 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/search/schema.graphql
+++ b/pkg/extensions/search/schema.graphql
@@ -117,5 +117,6 @@ type Query {
     ImageList(repo: String!): [ImageSummary!]
     ExpandedRepoInfo(repo: String!): RepoInfo!
     GlobalSearch(query: String!): GlobalSearchResult!
+    DerivedImageList(image: String!): [ImageSummary!]
     BaseImageList(image: String!): [ImageSummary!]
 }
--- a/pkg/extensions/search/schema.resolvers.go
+++ b/pkg/extensions/search/schema.resolvers.go
@@ -501,6 +501,74 @@ func (r *queryResolver) GlobalSearch(ctx context.Context, query string) (*gql_ge
 	}, nil
 }
 
+// DependencyListForImage is the resolver for the DependencyListForImage field.
+func (r *queryResolver) DerivedImageList(ctx context.Context, image string) ([]*gql_generated.ImageSummary, error) {
+	layoutUtils := common.NewBaseOciLayoutUtils(r.storeController, r.log)
+	imageList := make([]*gql_generated.ImageSummary, 0)
+
+	repoList, err := layoutUtils.GetRepositories()
+	if err != nil {
+		r.log.Error().Err(err).Msg("unable to get repositories list")
+
+		return nil, err
+	}
+
+	if len(repoList) == 0 {
+		r.log.Info().Msg("no repositories found")
+
+		return imageList, nil
+	}
+
+	imageDir, imageTag := common.GetImageDirAndTag(image)
+
+	imageManifest, err := layoutUtils.GetImageManifest(imageDir, imageTag)
+	if err != nil {
+		r.log.Info().Str("image", image).Msg("image not found")
+
+		return imageList, err
+	}
+
+	imageLayers := imageManifest.Layers
+
+	for _, repo := range repoList {
+		repoInfo, err := r.ExpandedRepoInfo(ctx, repo)
+		if err != nil {
+			r.log.Error().Err(err).Msg("unable to get image list")
+
+			return nil, err
+		}
+
+		imageSummaries := repoInfo.Images
+
+		// verify every image
+		for _, imageSummary := range imageSummaries {
+			if imageTag == *imageSummary.Tag && imageDir == repo {
+				continue
+			}
+
+			layers := imageSummary.Layers
+
+			sameLayer := 0
+
+			for _, l := range imageLayers {
+				for _, k := range layers {
+					if *k.Digest == l.Digest.Encoded() {
+						sameLayer++
+					}
+				}
+			}
+
+			// if all layers are the same
+			if sameLayer == len(imageLayers) {
+				// add to returned list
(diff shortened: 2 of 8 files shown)


peusebiu (commented):
remove things under hack/tools
remove regctl, zot-0.1.0.tgz and pushpull_output.txt?

hack/tools/lib/bats-core/common.bash
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env bash
```
andaaron, line 1: Remove the hack folder from the commit. See Sebi's comment.

pkg/extensions/search/schema.graphql
```diff
@@ -19,54 +19,11 @@ type PackageInfo {
      FixedVersion: String 
 }
 
-type ImgResultForCVE {
```
andaaron, line 22: Let's do this cleanup as part of https://github.com/project-zot/zot/pull/457.

In this PR, implement just https://github.com/project-zot/zot/pull/696/files#diff-84af1556304ab2f0edbb3541ee90f5e8eb88e4dde65c0a8069e517dea758c210R145, that's the new name of the query you implemented.

Clenan schema.graphql, add that query, and regenerate the code using gqlgen

pkg/extensions/search/schema.resolvers.go
```diff
@@ -419,7 +434,118 @@ func (r *queryResolver) GlobalSearch(ctx context.Context, query string) (*gql_ge
 	}, nil
 }
 
+// DependencyListForImage is the resolver for the DependencyListForImage field.
+func (r *queryResolver) DependencyListForImage(ctx context.Context, image string) ([]*gql_generated.ImageSummary, error) {
+	defaultStore := r.storeController.DefaultStore
+	imageList := make([]*gql_generated.ImageSummary, 0)
+
+	repoList, err := defaultStore.GetRepositories()
```
andaaron, line 442: We need to look into all stores. I think we have this problem in other queries as well.
We should have a new method in layoutUtils which returns all repos in all stores.

andaaron (commented):
Remove the files which are not supposed to be part of the review (zot-0.1.0.tgz, regctl, pushpull_output.txt)

Also I suggest a more detailed commit message.

Will circle back to this PR after the existing comments are addressed.

pkg/extensions/search/schema.resolvers.go
```diff
@@ -463,6 +463,74 @@ func (r *queryResolver) GlobalSearch(ctx context.Context, query string) (*gql_ge
...
+
+		// verify every image
+		for _, manifest := range manifests {
+			layers := manifest.Layers
+
+			sameLayer := 0
+
+			for _, l := range imageLayers {
+				for _, k := range layers {
+					if *k.Digest == string(l.Digest)[7:] {
```
andaaron, line 513: Let's use a clean way to do tics check.
layers are of type https://github.com/opencontainers/image-spec/blob/bc9c4bd9b2a4ec6f49f956e193f9c699d854507d/specs-go/v1/descriptor.go#L27
Which is of type Digest, and has various useful methods: https://github.com/opencontainers/go-digest/blob/master/digest.go#L37

Why would we do this comparison relative to the arbitrary characters starting from 7?

pkg/extensions/search/schema.resolvers.go
```diff
@@ -463,6 +463,74 @@ func (r *queryResolver) GlobalSearch(ctx context.Context, query string) (*gql_ge
 	}, nil
 }
 
+// DependencyListForImage is the resolver for the DependencyListForImage field.
+func (r *queryResolver) DerivedImageList(ctx context.Context, image string) ([]*gql_generated.ImageSummary, error) {
```
andaaron, line 467: We need to make sure all of the layers in the image sent as parameter are layers in the images returned by this method.
We don't what to know if the images have the exact same layers.
We want images with layers which are a superset of the set of layers sent in the image sent as parameter.

So if image `A` given as parameter has 3 layers `a b c`, and  image `B` has 4 layers `a b c d`, B meets the condition to be derived from A.
So if image `A` given as parameter has 3 layers `a b c`, and  image `C` has 2 layers `a b`, B meets the condition to be a base image for A (good for your other PR for BaseImageList), but doesn't meet the condition for this query.

pkg/extensions/search/schema.resolvers.go
```diff
@@ -463,6 +463,74 @@ func (r *queryResolver) GlobalSearch(ctx context.Context, query string) (*gql_ge
...
+					if *k.Digest == string(l.Digest)[7:] {
+						sameLayer++
+
+						break
+					}
+				}
+			}
+
+			// if all layers are the same
+			if sameLayer == len(imageLayers) {
```
andaaron, line 522: I think this should be `>=` and we should test for returning images which include all of the common layers, but also extra layers besides them.

andaaron (approved):

andaaron (changes requested):
See my previous review.

pkg/extensions/search/schema.resolvers.go
aokirisaki: in order to compare a digest with "sha256:" in front with one which doesn't have it (this is the way it is provided by the function I'm calling to get the digest)

pkg/extensions/search/schema.resolvers.go
aokirisaki: I added tests to verify all possible cases, equal layers, less layers than A and more layers than A.

andaaron (approved):

peusebiu:
LGTM

andaaron (dismissed):
Looks good.

rchincha:
@aokirisaki can you pls rebase and push this again. Too many merge conflicts.

andaaron (approved):

PR #718, opened 2022-08-10
storage: different subpaths can point to same root directory

shimish2:
currently different subpaths can only point to same root directory only
when one of storage config does not enable dedupe.

different subpath should be able to point to same root directory and in
that case their storage config should be same i.e GC,Dedupe, GC delay
and GC interval

Signed-off-by: Shivam Mishra <shimish2@cisco.com>

### DIFF
--- a/pkg/api/config/config.go
+++ b/pkg/api/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/getlantern/deepcopy"
@@ -142,6 +143,27 @@ func New() *Config {
 		HTTP:            HTTPConfig{Address: "127.0.0.1", Port: "8080", Auth: &AuthConfig{FailDelay: 0}},
 		Log:             &LogConfig{Level: "debug"},
 	}
+}
+
+func (expConfig StorageConfig) ParamsEqual(actConfig StorageConfig) bool {
+	return expConfig.GC == actConfig.GC && expConfig.Dedupe == actConfig.Dedupe &&
+		expConfig.GCDelay == actConfig.GCDelay && expConfig.GCInterval == actConfig.GCInterval
+}
+
+// SameFile compare two files.
+// This method will first do the stat of two file and compare using os.SameFile method.
+func SameFile(str1, str2 string) (bool, error) {
+	sFile, err := os.Stat(str1)
+	if err != nil {
+		return false, err
+	}
+
+	tFile, err := os.Stat(str2)
+	if err != nil {
+		return false, err
+	}
+
+	return os.SameFile(sFile, tFile), nil
 }
 
 // Sanitize makes a sanitized copy of the config removing any secrets.
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -21,6 +21,7 @@ import (
 	"zotregistry.io/zot/pkg/api/config"
 	ext "zotregistry.io/zot/pkg/extensions"
 	extconf "zotregistry.io/zot/pkg/extensions/config"
+	"zotregistry.io/zot/pkg/extensions/lint"
 	"zotregistry.io/zot/pkg/extensions/monitoring"
 	"zotregistry.io/zot/pkg/log"
 	"zotregistry.io/zot/pkg/storage"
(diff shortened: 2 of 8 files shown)


pkg/api/controller.go
```diff
@@ -221,6 +221,8 @@ func (c *Controller) InitImageStore(reloadCtx context.Context) error {
 
 	linter := ext.GetLinter(c.Config, c.Log)
 
+	imageStoreMap := make(map[string]storage.ImageStore, 0)
```
rchincha, line 224: imgStoreMap

pkg/api/controller.go
```diff
@@ -299,8 +303,14 @@ func (c *Controller) InitImageStore(reloadCtx context.Context) error {
 				if storageConfig.StorageDriver == nil {
 					// false positive lint - linter does not implement Lint method
 					// nolint: typecheck
-					subImageStore[route] = storage.NewImageStore(storageConfig.RootDirectory,
-						storageConfig.GC, storageConfig.GCDelay, storageConfig.Dedupe, storageConfig.Commit, c.Log, c.Metrics, linter)
+					imageStore, ok := imageStoreMap[storageConfig.RootDirectory]
```
rchincha, line 306: imgStore

pkg/cli/root.go
```diff
@@ -202,6 +203,34 @@ func NewCliRootCmd() *cobra.Command {
...
+			log.Error().Err(errors.ErrBadConfig).Msg("storage subpaths can not have default storage root directory")
+
+			return errors.ErrBadConfig
+		}
+
+		expConfig, ok := expConfigMap[storageConfig.RootDirectory]
+		if ok {
+			equal := expConfig.Equals(storageConfig)
+			if !equal {
+				log.Error().Err(errors.ErrBadConfig).Msg("storage config with same root directory should have same settings")
```
rchincha, line 222: "...same parameters"

pkg/cli/root.go
```diff
@@ -202,6 +203,34 @@ func NewCliRootCmd() *cobra.Command {
...
 }
 
+func validateStorageConfig(cfg *config.Config) error {
+	expConfigMap := make(map[string]config.StorageConfig, 0)
+
+	defaultRootDir := cfg.Storage.RootDirectory
+
+	for _, storageConfig := range cfg.Storage.SubPaths {
+		if strings.EqualFold(defaultRootDir, storageConfig.RootDirectory) {
+			log.Error().Err(errors.ErrBadConfig).Msg("storage subpaths can not have default storage root directory")
```
rchincha, line 213: "...cannot use the default ..."

pkg/api/controller.go
```diff
@@ -299,8 +303,11 @@ func (c *Controller) InitImageStore(reloadCtx context.Context) error {
 				if storageConfig.StorageDriver == nil {
 					// false positive lint - linter does not implement Lint method
 					// nolint: typecheck
-					subImageStore[route] = storage.NewImageStore(storageConfig.RootDirectory,
-						storageConfig.GC, storageConfig.GCDelay, storageConfig.Dedupe, storageConfig.Commit, c.Log, c.Metrics, linter)
+					_, ok := imgStoreMap[storageConfig.RootDirectory]
```
rchincha, line 306: We need filesystem equality check and not string equality here.

SameFile (fi1, fi2 FileInfo)
^ use this.

andaaron (approved):
LGTM

pkg/api/config/config.go
```diff
@@ -144,6 +145,27 @@ func New() *Config {
 	}
 }
 
+func (expConfig StorageConfig) Equals(actConfig StorageConfig) bool {
```
rchincha, line 148: Let's call this ParamsEqual()

pkg/api/config/config.go
```diff
@@ -144,6 +145,27 @@ func New() *Config {
...
+	return expConfig.GC == actConfig.GC && expConfig.Dedupe == actConfig.Dedupe &&
+		expConfig.GCDelay == actConfig.GCDelay && expConfig.GCInterval == actConfig.GCInterval
+}
+
+// CompareFiles compare two files.
+// This method will first do the stat of two file.
+func CompareFiles(str1, str2 string) (bool, error) {
```
rchincha, line 155: call this SameFile() itself?

shimish2: Yes

pkg/api/config/config.go
shimish2: Ok

pkg/api/config/config.go
```diff
@@ -144,6 +145,27 @@ func New() *Config {
 	}
 }
 
+func (expConfig StorageConfig) ParamsEqual(actConfig StorageConfig) bool {
+	return expConfig.GC == actConfig.GC && expConfig.Dedupe == actConfig.Dedupe &&
+		expConfig.GCDelay == actConfig.GCDelay && expConfig.GCInterval == actConfig.GCInterval
+}
+
+// CompareFiles compare two files.
```
rchincha, line 153: Fix comment

PR #728, opened 2022-08-16
graphql: Apply authorization on /_search endpoint

### LINKED ISSUE #615: graphql: Apply authorization on /_search endpoint
**Describe the bug**
Apply authorization for _search endpoint (graphql)

Currently authorization is applied only for dist-spec routes and not for graphql.

On a request context we store the repositories on which a user has read access, and later used them when the user GETs the catalog for example, to filter out the repos on which he doesn't have access.

Do the same for _search endpoint.

alexstan12:
- AccessControlContext now resides in a separate package from where it can be imported,
along with the contextKey that will be used to set and retrieve this context value.

- AccessControlContext has a new field called Username, that will be of use for future
implementations in graphQL resolvers.

- GlobalSearch resolver now uses this context to filter repos available to the logged user.

Closes #615

**What type of PR is this?**
feature

using authz context for use on GraphQl resolvers

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/authz.go
+++ b/pkg/api/authz.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	glob "github.com/bmatcuk/doublestar/v4"
@@ -12,9 +13,8 @@ import (
 	"zotregistry.io/zot/pkg/api/constants"
 	"zotregistry.io/zot/pkg/common"
 	"zotregistry.io/zot/pkg/log"
+	localCtx "zotregistry.io/zot/pkg/requestcontext"
 )
-
-type contextKey int
 
 const (
 	// actions.
@@ -22,21 +22,12 @@ const (
 	READ   = "read"
 	UPDATE = "update"
 	DELETE = "delete"
-
-	// request-local context key.
-	authzCtxKey contextKey = 0
 )
 
 // AccessController authorizes users to act on resources.
 type AccessController struct {
 	Config *config.AccessControlConfig
 	Log    log.Logger
-}
-
-// AccessControlContext context passed down to http.Handlers.
-type AccessControlContext struct {
-	globPatterns map[string]bool
-	isAdmin      bool
 }
 
 func NewAccessController(config *config.Config) *AccessController {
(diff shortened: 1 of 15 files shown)


pkg/context/context.go
```diff
@@ -0,0 +1,18 @@
+package context
```
peusebiu, line 1: I don't like the naming of this, it gets confusing.

peusebiu:
lgtm

pkg/context/context.go
alexstan12: what about userContext? if this file will gather over time all types of contexts useful to zot, maybe just zotContext ?

peusebiu: I would say requestContext or reqContext.

alexstan12: package renamed to requestcontext .

pkg/api/controller_test.go
```diff
@@ -5147,6 +5146,138 @@ func TestPeriodicTasks(t *testing.T) {
...
+			ctlr.Config.Storage.RootDirectory = tempDir
+
+			go startServer(ctlr)
+			defer stopServer(ctlr)
+			test.WaitTillServerReady(baseURL)
+
+			query := `
+			{
+				GlobalSearch(query:""){
+					Repos {
```
andaaron, line 5236: I think we should have a test according to which we verify that one repo is returned because the user has permissions to view it, while another is not returned as the user doesn't have permissions to view the second one.

andaaron (commented):
Looks good overall.

andaaron (dismissed):

andaaron (approved):

andaaron (approved):

peusebiu (approved):
LGTM

pkg/api/authz.go
```diff
@@ -213,6 +208,12 @@ func AuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
 				return
 			}
 
+			if strings.Contains(request.RequestURI, constants.ExtSearchPrefix) {
+				next.ServeHTTP(response, request.WithContext(ctx))
```
rchincha, line 212: Why are we not falling back to the normal path below?

rchincha: And what is the "next" handler?

rchincha (commented):
Just some clarifications, the PR looks ok

pkg/api/authz.go
alexstan12: We are not falling back on the normal path because on line [239](https://github.com/project-zot/zot/pull/728/files/7517df3258d350ac53c0d84be6d91f02f6575aed#diff-fa881d434bc95a6195440f2122784dacca2bcbc19b93baf9e575feed1265340aL239) the "resource" variable will be empty, since this in a request variable for REST API calls. Thus authz will fail - line 241.

alexstan12: > And what is the "next" handler?

idk what do you mean by this, nothing has changed in this approach. 
```func(next http.Handler) http.Handler``` is the signature of a MiddlewareFunc, and in our case the "next" handler will be the router inside the Controller.
We serve back the incoming http requests with modified context ```next.ServeHTTP(response, request.WithContext(ctx))```

andaaron (approved):

PR #729, opened 2022-08-16
Added support for changing config via API

### LINKED ISSUE #268: support configuration change using API calls
Currently the zot sync configuration can be changes only by modifying the zot configuration file.
We should implement an API to update this sync configuration, so other service can modify it while zot is running.

At a minimum we should be able to specify what images and tags we need to sync for a given registry.

### LINKED ISSUE #698: Zot should expose the current configuration using an API call
Needed for ZUI in order to determine if local authentication is enabled.
If local authentication is enabled, the user should be allowed to change his password.

peusebiu:
closes #268, closes #698

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


pkg/api/controller.go
```diff
@@ -125,15 +137,14 @@ func (c *Controller) Run(reloadCtx context.Context) error {
 
 	engine.Use(
 		c.CORSHeaders(),
-		SessionLogger(c),
```
peusebiu, line 128: need to restore these

pkg/api/config/config.go
```diff
@@ -62,14 +60,16 @@ type RatelimitConfig struct {
...
-	Realm            string
-	Ratelimit        *RatelimitConfig `mapstructure:",omitempty"`
+	Address         string
+	Port            string
+	AllowOrigin     string // comma separated
+	TLS             *TLSConfig
+	Auth            *AuthConfig
+	AccessControl   *AccessControlConfig
+	Realm           string
+	AllowReadAccess bool             `mapstructure:",omitempty"`
```
rchincha, line 70: AllowReadAccess was deprecated?

peusebiu: oops

pkg/extensions/extensions-config.go
```diff
@@ -0,0 +1,90 @@
...
+) {
+	if config.Extensions.SysConfig != nil && *config.Extensions.SysConfig.Enable {
+		log := log.Logger{Logger: l.With().Caller().Timestamp().Logger()}
+		log.Info().Msg("setting up extensions routes")
+
+		configPath := viper.GetViper().ConfigFileUsed()
+
+		handler := NewConfigHandler(configPath, log)
+
+		if config.Extensions.SysConfig != nil && *config.Extensions.SysConfig.Enable {
```
peusebiu, line 86: remove this line :D

PR #731, opened 2022-08-19
Fix file handlers not being closed after calls to ImageStore.GetBlob

andaaron:
**What type of PR is this?**
bug

This indirectly fixes hitting the FD limit when reading blobs from the disk when using the graphql API

Makes sure file handlers are timely closed after reading blobs.

Tested changeset in https://github.com/project-zot/zot/pull/457.
The environment had roughly 100 images (4 repos with multiple tags in each, 1042 individual blob files) in storage.
Before the fix, `bin/zli-linux-amd64 images local` used to cause FD limit to be reached when the file handlers for blobs remained open.

Repeated the test mentioned above with the patchset cherry-picked on top of https://github.com/project-zot/zot/pull/457, with this change the opened blob files are closed immediately.

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -655,6 +655,7 @@ func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Requ
 
 		return
 	}
+	defer repo.Close()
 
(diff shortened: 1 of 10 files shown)


peusebiu (approved):
LGTM

Issue #661, opened 2022-07-20
New API to list dependants for a specific Image 


andaaron:
**Is your feature request related to a problem? Please describe.**
One of the views in zui will list image dependants.
We need a query which given an image name or image digest would return a list of all images which include all the layers of the given image.

andaaron:
Done in https://github.com/project-zot/zot/pull/813

PR #732, opened 2022-08-19
list all images that have all layers of the base image included 

aokirisaki:
**What type of PR is this?**

#661 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/search/common/common_test.go
+++ b/pkg/extensions/search/common/common_test.go
@@ -14,6 +14,7 @@ import (
 	"os/exec"
 	"path"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
@@ -708,7 +709,7 @@ func TestExpandedRepoInfo(t *testing.T) {
 		So(err, ShouldBeNil)
 		So(responseStruct.ExpandedRepoInfo.RepoInfo.Summary, ShouldNotBeEmpty)
 		So(responseStruct.ExpandedRepoInfo.RepoInfo.Summary.Name, ShouldEqual, "test1")
-		So(len(responseStruct.ExpandedRepoInfo.RepoInfo.Images), ShouldEqual, 2)
+		So(len(responseStruct.ExpandedRepoInfo.RepoInfo.ImageSummaries), ShouldEqual, 2)
 	})
 
 	Convey("Test expanded repo info", t, func() {
@@ -791,10 +792,10 @@ func TestExpandedRepoInfo(t *testing.T) {
 
 		err = json.Unmarshal(resp.Body(), responseStruct)
 		So(err, ShouldBeNil)
-		So(len(responseStruct.ExpandedRepoInfo.RepoInfo.Images), ShouldNotEqual, 0)
-		So(len(responseStruct.ExpandedRepoInfo.RepoInfo.Images[0].Layers), ShouldNotEqual, 0)
+		So(len(responseStruct.ExpandedRepoInfo.RepoInfo.ImageSummaries), ShouldNotEqual, 0)
+		So(len(responseStruct.ExpandedRepoInfo.RepoInfo.ImageSummaries[0].Layers), ShouldNotEqual, 0)
 		found := false
-		for _, m := range responseStruct.ExpandedRepoInfo.RepoInfo.Images {
+		for _, m := range responseStruct.ExpandedRepoInfo.RepoInfo.ImageSummaries {
 			if m.Digest == "63a795ca90aa6e7cca60941e826810a4cd0a2e73ea02bf458241df2a5c973e29" {
 				found = true
 				So(m.IsSigned, ShouldEqual, false)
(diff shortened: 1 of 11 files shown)


andaaron (changes requested):
Same comments as for https://github.com/project-zot/zot/pull/713

rchincha:
Why are there so many BATS file changes? Do you want to split this PR into two instead?

rchincha:
@aokirisaki don't see the test case which actually builds a chain of images.
Basically, create a manifest with layers and include on of these layers in another manifest.

andaaron:
> @aokirisaki don't see the test case which actually builds a chain of images. Basically, create a manifest with layers and include one or more of these layers in another manifest. Add a negative case as well.

Isn't this what you are looking for https://github.com/project-zot/zot/pull/732/files#diff-b86e11fa5a3102b336caebec3b30a9d35e26af554dd8658f124dba2404b7d24aR1065?

rchincha:
> > @aokirisaki don't see the test case which actually builds a chain of images. Basically, create a manifest with layers and include one or more of these layers in another manifest. Add a negative case as well.
> 
> Isn't this what you are looking for https://github.com/project-zot/zot/pull/732/files#diff-b86e11fa5a3102b336caebec3b30a9d35e26af554dd8658f124dba2404b7d24aR1065?

Is the link right?

Also @aokirisaki can you post the zli output for this.

andaaron:
Look at pkg/extensions/search/common/common_test.go line ~1065. The link doesn't work right because the file is large.

pkg/extensions/search/common/common_test.go
```diff
@@ -982,6 +984,505 @@ func TestUtilsMethod(t *testing.T) {
 	})
 }
 
+func TestGetImageManifest(t *testing.T) {
+	Convey("Test inexistent image", t, func() {
```
rchincha, line 988: "non-existent"

pkg/extensions/search/common/common_test.go
```diff
@@ -982,6 +984,505 @@ func TestUtilsMethod(t *testing.T) {
...
+	// shut down server
+
+	defer func() {
+		ctx := context.Background()
+		_ = ctlr.Server.Shutdown(ctx)
+	}()
+
+	Convey("Test base image list for image working", t, func() {
+		// create test images
+		config := ispec.Image{
```
rchincha, line 1067: cblob, cdigest := test.GetRandomImageConfig()
^ why not use this instead?

pkg/extensions/search/common/common_test.go
```diff
@@ -982,6 +984,505 @@ func TestUtilsMethod(t *testing.T) {
...
+				Config:   config,
+				Layers:   layers,
+				Tag:      "latest",
+			},
+			baseURL,
+			repoName,
+		)
+		So(err, ShouldBeNil)
+
+		// create image with the same layers
```
rchincha, line 1136: Also create an image with NO shared layers. And it should show up in the results.

andaaron: Would this be the same image you make the query for, if I understood correctly?
If yes, then why would the UI should you a tile with the same image you are already viewing the "dependants" or "dependency" tabs for? As a user I would want to see other images this one is related to, not the current image, it would have to be filtered out on front-end?

pkg/extensions/search/common/common_test.go
aokirisaki: I made the modifications

pkg/extensions/search/common/common_test.go
aokirisaki: Added tests for no shared layers and added a verification so that the given image won't appear in the search

aokirisaki:
> > > @aokirisaki don't see the test case which actually builds a chain of images. Basically, create a manifest with layers and include one or more of these layers in another manifest. Add a negative case as well.
> > 
> > 
> > Isn't this what you are looking for https://github.com/project-zot/zot/pull/732/files#diff-b86e11fa5a3102b336caebec3b30a9d35e26af554dd8658f124dba2404b7d24aR1065?
> 
> Is the link right?
> 
> Also @aokirisaki can you post the zli output for this.

This is the output:
ana@ALISCA-NB:~/aokirisaki/zot$ bin/zli-linux-amd64 images config-test --base-images busybox
IMAGE NAME                        TAG                       DIGEST        SIZE      
busybox                           latest                    b9590ac9      773kB     
busybox                           new                       b9590ac9      773kB     
busybox                           1.0.1                     2ac88360      773kB

pkg/extensions/search/schema.resolvers.go
```diff
@@ -494,6 +495,87 @@ func (r *queryResolver) GlobalSearch(ctx context.Context, query string) (*gql_ge
...
+							foundLayer = true
+							layerSize, _ := strconv.Atoi(*l.Size)
+							imageSize += layerSize
+
+							break
+						}
+					}
+
+					if !foundLayer {
+						addImageToList = false
```
andaaron, line 559: We should set this to false also if the condition at line 544 is not met.

pkg/extensions/search/schema.resolvers.go
```diff
@@ -494,6 +495,87 @@ func (r *queryResolver) GlobalSearch(ctx context.Context, query string) (*gql_ge
...
+
+						break
+					}
+				}
+			}
+
+			if addImageToList {
+				name := repo
+				manifest.RepoName = &name
```
andaaron, line 568: This field should already be populated from ExpendedRepoInfo.
Same goes for computing the image size at line 570.

pkg/extensions/search/schema.resolvers.go
```diff
@@ -494,6 +495,87 @@ func (r *queryResolver) GlobalSearch(ctx context.Context, query string) (*gql_ge
...
+
+	for _, repo := range repoList {
+		repoInfo, err := r.ExpandedRepoInfo(ctx, repo)
+		if err != nil {
+			r.log.Error().Err(err).Msg("unable to get image list")
+
+			return nil, err
+		}
+
+		manifests := repoInfo.Images
```
andaaron, line 535: Rename this to imageSummaries or something relevant for the type of object.

pkg/extensions/search/schema.resolvers.go
```diff
@@ -494,6 +495,87 @@ func (r *queryResolver) GlobalSearch(ctx context.Context, query string) (*gql_ge
...
+	imageManifest, err := layoutUtils.GetImageManifest(imageDir, imageTag)
+	if err != nil {
+		r.log.Info().Str("image", image).Msg("image not found")
+
+		return imageList, err
+	}
+
+	imageLayers := imageManifest.Layers
+
+	for _, repo := range repoList {
```
andaaron, line 527: Add a comment on line 526 about concerns about future performance when the system scales.

aokirisaki: Added the comment

pkg/extensions/search/schema.resolvers.go
aokirisaki: I made the modification and now it works

andaaron:
Merged as part of https://github.com/project-zot/zot/pull/813

PR #733, opened 2022-08-19
Add the hack folder to .gitignore

andaaron:
To avoid committing it in the future


rchincha (approved):
lgtm

Issue #704, opened 2022-08-03
GraphQL API: ignore manifests which have no annotation for tags.


andaaron:
At the moment our search logic assumes that if a manifest does not have a tag annotation, the tag should be `latest`: https://github.com/project-zot/zot/blob/main/pkg/extensions/search/common/oci_layout.go#L358.

We should fix this by simply skipping that manifest.

andaaron:
https://github.com/project-zot/zot/pull/695

PR #754, opened 2022-08-26
Skip manifest which have no annotation for tags

alexstan12:
**What type of PR is this?**
feature

Closes https://github.com/project-zot/zot/issues/704

It skips manifests of a repo that don't have an annotation tag.

tests add a manifest that doesn't have a tag as a manifest.Annotations field and expect this manifest will be skipped
on calls such as ExpandedRepoInfo
**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/search/common/oci_layout.go
+++ b/pkg/extensions/search/common/oci_layout.go
@@ -384,7 +384,9 @@ func (olu BaseOciLayoutUtils) GetExpandedRepoInfo(name string) (RepoInfo, error)
 
 		tag, ok := man.Annotations[ispec.AnnotationRefName]
 		if !ok {
-			tag = "latest"
+			olu.Log.Info().Msgf("skipping manifest with digest %s because it doesn't have a tag", string(man.Digest))
+
+			continue
 		}
 
 		manifestInfo.Tag = tag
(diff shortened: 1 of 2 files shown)


pkg/extensions/search/common/oci_layout.go
```diff
@@ -384,7 +384,7 @@ func (olu BaseOciLayoutUtils) GetExpandedRepoInfo(name string) (RepoInfo, error)
 
 		tag, ok := man.Annotations[ispec.AnnotationRefName]
 		if !ok {
-			tag = "latest"
+			continue
```
peusebiu, line 387: add a log here where you specify that you are skipping manifest with digest X because it doesn't have a tag

alexstan12: done.

andaaron (dismissed):

shimish2 (approved):
lgtm

rchincha:
`skopeo --insecure-policy copy --format=oci --dest-tls-verify=false --multi-arch=all \
        docker://public.ecr.aws/docker/library/busybox:latest \
        docker://127.0.0.1:8080/busybox:latest
`

can you also do this and make sure the image is returned. 
This path is different because '--multi-arch' creates a Manifest Index (instead of just a Manifest) that in turn refers to other manifests, and let's make sure this is returned.

Issue #762, opened 2022-08-30
http Range Support for Image Pull


mwopfner:
**Is your feature request related to a problem? Please describe.**
I try to pull an artifact (binary file, not an image layer) from the repository. This perfectly works if I use `wget` but if I do it with the desired tool I get an http response that states: "range requests not supported by server".

**Describe the solution you'd like**
`zot` should support http `Range` as described in the spec use case "Resumeable Pull".
https://github.com/opencontainers/distribution-spec/blob/main/spec.md#resumable-pull

Unfortunately there is no detailed description for "Pulling a blob in chunks" as it is there for Pushing but I think the use case clearly states that this is a feature that should exist.

**Describe alternatives you've considered**
I verified that pulling the complete artifact works by using `wget` but the tool that must actually pull the artifact uses http `Range` because it does it in a streaming manner.

**Additional context**
The logs I get from `zot` are:
```
{"level":"info","module":"http","clientIP":"192.168.111.93:50614","method":"GET","path":"/v2/linux-os/blobs/sha256:34d8f27da12f02a3a57653c19b44815fe53cc9fde16a03ec242dc2fac673a8ac","statusCode":200,"latency":"690.051µs","bodySize":720896,"headers":{"Accept":["*/*"],"Range":["bytes=0-3"],"User-Agent":["rauc/1.7"]},"goroutine":29,"caller":"zotregistry.io/zot@v1.4.1/pkg/api/session.go:132","time":"2022-08-30T08:08:13.978160735+02:00","message":"HTTP API"}
{"level":"error","error":"write tcp 192.168.111.93:8080->192.168.111.93:50616: write: broken pipe","goroutine":45,"caller":"zotregistry.io/zot@v1.4.1/pkg/api/routes.go:1334","time":"2022-08-30T08:08:14.987854242+02:00","message":"copying data into http response"}
```
I looked at the specific code lines and as I see it `zot` actually does not support http `Range` requests. Is this correct?

rchincha:
> zot should support http Range as described in the spec use case "Resumeable Pull".
> https://github.com/opencontainers/distribution-spec/blob/main/spec.md#resumable-pull

@mwopfner Thanks for trying out zot and reporting this issue. We will fix this.

Just curious, is this an embedded device use case? None of standard tooling that we are aware of has this requirement.

mwopfner:
@rchincha Yes, your are correct, it is an embedded device which uses http `Range` to avoid a complete download of the artifact to the filesystem. 
[...]

PR #765, opened 2022-08-30
routes: support resumable pull

rchincha:

**What type of PR is this?**

feature

issue #762 

As per dist-spec, resumable pulls should be supported. Common tools in the container ecosystem don't use this. But per issue #762, some tools do require it.

No.

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -18,6 +18,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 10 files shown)


pkg/storage/local.go
```diff
@@ -1510,6 +1510,84 @@ func (is *ImageStoreLocal) copyBlob(repo, blobPath, dstRecord string) (int64, er
...
+
+	blobReadCloser, err := NewBlobStream(blobPath, from, to)
+	if err != nil {
+		is.log.Error().Err(err).Str("blob", blobPath).Msg("failed to open blob")
+
+		return nil, -1, true, err
+	}
+
+	partial := true
+	if from == 0 || (to-from)+1 >= binfo.Size() {
```
mwopfner, line 1583: The `||` must be a `&&` otherwise range request like `0-100` are not partial which is wrong.
```suggestion
	if from == 0 && (to-from)+1 >= binfo.Size() {
```

pkg/storage/local.go
```diff
@@ -1510,6 +1510,84 @@ func (is *ImageStoreLocal) copyBlob(repo, blobPath, dstRecord string) (int64, er
...
+		partial = false
+	}
+
+	// The caller function is responsible for calling Close()
+	return blobReadCloser, to - from + 1, partial, nil
```
mwopfner, line 1588: We should also return the complete blob size because it is needed in routes.go.
```suggestion
	return blobReadCloser, to - from + 1, binfo.Size(), partial, nil
```

pkg/api/routes.go
```diff
@@ -659,9 +742,17 @@ func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Requ
 	defer repo.Close()
 
 	response.Header().Set("Content-Length", fmt.Sprintf("%d", blen))
-	response.Header().Set(constants.DistContentDigestKey, digest)
+
+	status := http.StatusOK
+
+	if partial {
+		status = http.StatusPartialContent
```
mwopfner, line 749: The `Content-Range` header must be set (see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests#single_part_ranges)

`bsize`is the total blob size returned as an extra return value from `imgStore.GetBlobPartial`
`(from+blen-1)` is used so that we limit the `to` value to the maximum blob size

```suggestion
		status = http.StatusPartialContent
		response.Header().Set("Content-Range", fmt.Sprintf("bytes %d-%d/%d", from, (from+blen-1), bsize))
```

mwopfner (changes requested):
I tested the patch with my setup but I had to change some things got get it working.

pkg/storage/local.go
rchincha: fixed, thx

rchincha:
https://www.rfc-editor.org/rfc/rfc9110.html#name-range-requests
^ as reference

pkg/storage/local.go
```diff
@@ -1510,6 +1510,80 @@ func (is *ImageStoreLocal) copyBlob(repo, blobPath, dstRecord string) (int64, er
 	return -1, zerr.ErrBlobNotFound
 }
 
+// blobStream is using to serve blob range requests.
+type blobStream struct {
+	reader io.Reader
+	closer io.Closer
+}
+
+func NewBlobStream(blobPath string, from, to int64) (io.ReadCloser, error) {
```
andaaron, line 1519: Just a thought, if we're not going to add this as part of the interface (and I see signature is different between s3 and local), maybe we should not export it?

andaaron (approved):

mwopfner:
I have tested the newest PR version an now it works as expected.

pkg/storage/local.go
rchincha: Yes, this method is internal to the implementations, we can un-import. Fixed.

andaaron (approved):

shimish2 (approved):
lgtm

PR #774, opened 2022-09-08
Refactor s3 dedupe

peusebiu:
Under digest bucket create 2 buckets, one for storing origin blob and one for storing deduped blobs.

PutBlob() - puts an origin blob in both buckets
          - puts a deduped blob in deduped bucket
GetBlob() - returns blobs only from origin bucket
DeleteBlob() - deletes an origin blob from both buckets
             and moves one deduped blob into origin bucket
             - deletes a deduped blob from deduped bucket

[storage] When deleting an origin blob, next time we GetBlob() we get a deduped blob with no content and we will move the content from the deleted origin blob to it.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/cache.go
+++ b/pkg/storage/cache.go
@@ -12,11 +12,15 @@ import (
 )
 
 const (
-	BlobsCache              = "blobs"
+	// global bucket.
+	BlobsCache = "blobs"
+	// bucket where we store all blobs from storage(deduped blobs + original blob).
+	DedupedBucket = "deduped"
+	/* bucket where we store only the original/source blob (used by s3 to know which is the blob with content)
+	it should contain only one blob, this is the only place from which we'll get blobs. */
+	OriginBucket            = "origin"
 	DBExtensionName         = ".db"
 	dbCacheLockCheckTimeout = 10 * time.Second
-	// always mark the first key inserted in the BlobsCache with a value.
-	firstKeyValue = "first"
 )
 
 type Cache struct {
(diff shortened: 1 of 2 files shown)


pkg/storage/cache.go
```diff
@@ -211,28 +243,46 @@ func (c *Cache) DeleteBlob(digest, path string) error {
...
-		key, _ := cur.First()
-		if key == nil {
-			c.log.Debug().Str("digest", digest).Str("path", path).Msg("deleting empty bucket")
-			if err := root.DeleteBucket([]byte(digest)); err != nil {
-				c.log.Error().Err(err).Str("digest", digest).Str("path", path).Msg("unable to delete")
+					return err
+				}
 
-				return err
+				// move next candidate to origin bucket, next GetKey will return this one and storage will move the content here
```
andaaron, line 265: I understand this is for updating the boltdb content, but is there a similar logic for storage?
I mean if I have multiple `duplicate` blobs in S3 and they are empty, if the user makes a call to delete the `original` blob, do we have code copying the blob content from the `original` blob to a previously empty candidate?

peusebiu: yes, we do, look here https://github.com/project-zot/zot/blob/main/pkg/storage/s3/s3.go#L1549

1. client wants to delete blob X
2. we get a blob from cache based on the digest from client
3. we delete the cache entry for 1. (here we are moving from deduped bucket to origin bucket)
4. we compare the cache entry from 2 with the blob path we received from client, if they are same then it means that is the blob with content
5. because we did 3, and 4 is true then next time we get a blobpath from cache it will be one without content. (we already moved from deduped bucket to origin bucket)
6. move content

rchincha (approved):
lgtm

Issue #710, opened 2022-08-05
authN: support GitLab authentication


rchincha:
**Is your feature request related to a problem? Please describe.**

GitLab can be used as an authenticator. It is useful when developers are on GitLab and is desirable to avoid separately managing authentication for zot.

**Describe the solution you'd like**

https://github.com/xanzy/go-gitlab
https://github.com/xanzy/go-gitlab/blob/master/gitlab.go

rchincha:
This is related to issue #709

andaaron:
Fixed in #1381

Issue #709, opened 2022-08-05
authN: support GitHub authentication


rchincha:
**Is your feature request related to a problem? Please describe.**

GitHub can be used as an authenticator. It is useful when developers are on GitHub and is desirable to avoid separately managing authentication for zot.

**Describe the solution you'd like**

https://github.com/google/go-github
https://github.com/google/go-github/blob/master/example/basicauth/main.go

andaaron:
This library can also be used with OAuth2 (golang.org/x/oauth2).

See an example at:
https://blog.kowalczyk.info/article/f/accessing-github-api-from-go.html

andaaron:
Fixed in #1381

PR #775, opened 2022-09-08
Proposal for Social Login implementation

andaaron:
Credt for investigation into possible alternatives goes to Raul Kele and Alex Stan. Diabrams by Alex Stan.

**What type of PR is this?**
feature
documentation

Does not fix but it proposes a fix for https://github.com/project-zot/zot/issues/709 and https://github.com/project-zot/zot/issues/710

It proposes design alternatives for 'social login' using GitHub/GitLab/Google.

No, but if we implement the proposal it will.

### DIFF
--- a/docs/SocialLogin.md
+++ b/docs/SocialLogin.md
@@ -1,0 +1,380 @@
+# Social login flow
+
+This document describes our implementation options for implementing social login in zot
+By social login we understand login with a Github/Gitlab/Google account.
+The social login feature leverages OAuth2 infrastructure implemented by the social login provider.
+Note not all social login providers implement all OAuth2 flows, and there are differences in implementation.
+If we are to support these 3 providers we need to take this into consideration.
+
+## OAuth2 flows for UI
+
+There are several OAuth2 flows we looked at:
+
+* Authorization Code Flow
+
+    * https://oauth.net/2/grant-types/authorization-code/
+    * https://datatracker.ietf.org/doc/html/rfc6749
+    * This is for server-side apps making calls to social login providers,
+    most of the logic would be in zot code itself
+    * The zot code would need a `client_id` and `client_secret` in order to access
+    the social login provider and request a `code`. Based on this `code` the
+    zot server would make another request to get the `token`, which it would
+    later pass to zui as a query parameter. The workflow itself is more complex
+    but the basic idea is that we need a `client_secret` for athentication, which is not
+    safe to send to zui, save in zui code, or store in the browser.
+    * We would need to implement specific endpoints for login for each social login provider
+    or zui to send us the social provider name/id to a single login endpoint which we implement
+    specifically for oauth
+    * We would also need to implement a callback endpoint in zot, which is where the social login
+    provider would redirect the browser after the user authorized our app to access his data
+
+* Implicit Flow
+
+    * https://datatracker.ietf.org/doc/html/rfc6749
+    * https://oauth.net/2/grant-types/implicit/
+    * This flow was supposed to be for browser-based single-page apps, but is now legacy for security reasons
+    * This could theoretically be implemented without needing a `client_secret` browser-side
+
+* PKCE Flow
+
+    - https://datatracker.ietf.org/doc/html/rfc7636
+    - https://oauth.net/2/pkce/
+    - This could theoretically be implemented mostly in the browser without needing a `client_secret`
+    accessible there
+    - This flow is not supported by all social login providers we take into consideration - it's supported
+    on GitLab, but not on GitHub
+
+Given the limittations of the flows above, we are limitted with regards to options if we want all 3 to work.
+Login with GitHub requires an implementation of the Authorization Code Flow, which implies more zot server side code. 
+
+## Data-Path login
+
+The above is mostly applicable for zot UI side.
+For Data-Path, we can use personal token from GitHub. We could investigate further to see how we could obtain
+GitLab or Google tokens.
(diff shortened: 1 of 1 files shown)


rchamarthy:
The authentication flow is understood, but i would like to see details on how authorization also works once we have the user authenticated.

andaaron:
> The authentication flow is understood, but i would like to see details on how authorization also works once we have the user authenticated.

The authorization flow is the same as the one we use now. The github tokens do not contain information on user permissions (and we don't have a way to decode them even if they did). 
The proposal is to assign specific policies to the user email in the zot config.json as we support now - assuming we don't have a large number of users for which we need to manually configure the policies.

docs/SocialLogin.md
```diff
@@ -0,0 +1,368 @@
...
+all configured providers.
+
+Someone suggested custom headers on requests from frontend to specify token type (github/google/gitlab), but
+this would only work for zot UI, and not for the data path.
+
+In order to validate a token there is no straight forward way to make a standard call, it depends on the provider.
+For example GitHub provides a non-rate-limitted API for validating oauth2 `gho_*` tokens,
+but there is no such API for personal `ghp_*` tokens.
+
+This means we have to call another rate-limitted API to check if login with said token works on GitHun, or
```
raulkele, line 73: Typo, GitHub not GitHun

docs/SocialLogin.md
```diff
@@ -0,0 +1,368 @@
...
+the social login provider in general.
+We need to find a way around this.
+
+One option is to keep the tokens in memory or in a database (just the valid one, or also the invalid ones?).
```
raulkele, line 77: To add to this possible solution to handle rate limited checks, if we go with this implementation. we can store the refresh token and the token generation timestamp in the DB too, then if the active token expires, perform the token refresh flow. It would still reduce the need to query the possibly rate-limited social provider while keeping the token valid(if we care about this).

alexstan12: +1 on that. For example, the "Backend with session cookies" sequence diagram contains the "[access_token valid]" and "[access_token expired]" steps which do exactly this : 

1. extract period of validity from the in memory map / DB
2. check validity and decide if refresh_token will be used to generate another access_token

docs/SocialLogin.md
```diff
@@ -0,0 +1,368 @@
...
+This way we won't hit the rate limitting, but we also do not know if the token has expired in the meantime.
+A possible solution would be to validate the tokens once, and then use a different token which we emit ourselves or cookie - at
```
alexstan12, line 79: There won't be a need to validate against a Social Login provider endpoint if the token is generated on Zot server side.
As explained in the above comment,  period of validity is available in the in-memory map / DB so no extra query for that. We'll just use the refresh_token to generate access_token, if that's the case.

docs/SocialLogin.md
```diff
@@ -0,0 +1,368 @@
...
+        Zot Server ->> Social Server: REQ - https://social-provider.com/login/oauth/token (gets token using code from prev request)
+        Social Server ->> Zot Server: RESP - returns access_token, refresh_token, period of validity using previously returned code as a query param
+        Zot Server -->> Zot Server : creates session token from access_token and updates sessions map
+        Note right of Zot Server: cookie containing user data extracted from token
+        Zot Server ->> Client app: RESP - /auth/gh/login - set session token as cookie
+        Client app->>Zot Server: send cookie containing session token, asking for resource
+        Zot Server -->> Zot Server : verifies session token exists in sessions map
+        alt session token  exists
+            alt access_token valid
+            Note right of Zot Server: we may or we may not implement this, as we already got the user data before generating the token
```
alexstan12, line 319: Does this mean it doesn't matter if the access_token is expired? do we only care about the fact that the first access_token we get is enough to get user:email and other useful data ?

docs/SocialLogin.md
```diff
@@ -0,0 +1,368 @@
...
+        end
+```
+
+### 4: Backend with session cookies
+
+#### Short overview of the workflow
+
+    - Similar to flow#3
+    - Instead of generating JWT tokens use sessions to store information
+
```
alexstan12, line 289: I think it is worth mentioning one of the great advantages of this workflow - in case of any Social provider that implements  Oauth login (so Github, Gitlab, Google) it doesn't need to verify **token validity and origin** by calling every social provider API. That is because:

1.  we query for the token on Zot server side -> we know the origin of the token by knowing what Social provider was queried
2. we get the **period of validity** as a response on the https://social-provider.com/login/oauth/token call

In case of clients that don't implement Oauth but use PATs (eg Skopeo), we have the 5th workflow.

alexstan12 (changes requested):

docs/SocialLogin.md
andaaron: Agreed. We'd do this for token we have the expiry/refresh information on. For PATs we could use a default expiry in out code based on when the token was last used/validated.

docs/SocialLogin.md
andaaron: Not all tokens are generated on Zot server side. There's also personal access tokens.

docs/SocialLogin.md
andaaron: Yes, that's what I was thinking.
As long as our session token is not expired we don't really care about expiry of access_token.
Assuming the session token is designed to expire in a reasonable timeframe.

andaaron:
Let's close this as social login support was merged in https://github.com/project-zot/zot/pull/1381
The UI side needs to get some updates, and should be ready soon.

PR #787, opened 2022-09-13
graphql: Populate ImageSummary missing fields:

peusebiu:
Description, Labels, Licenses, Title, Documentation, Source

closes #786

Signed-off-by: Petu Eusebiu <peusebiu@cisco.com>

**What type of PR is this?**

**Which issue does this PR fix**:

**What does this PR do / Why do we need it**:

**If an issue # is not available please add repro steps and logs showing the issue**:

**Testing done on this change**:

**Automation added to e2e**:

**Will this break upgrades or downgrades?**

**Does this PR introduce any user-facing change?**:

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #790, opened 2022-09-15
Validate Annotations present in image manifest and fallback to annota…

### LINKED ISSUE #788: Zot annotation validation - check labels are on the config, and fallback to annotations on the manifest if not available
Currently we only validate Annotations present in image manifest. Most images contain the same metadata in config labels.
We need to check both.

Most likely the manifest annotations should be checked first.

nicoldr:
…tions in the config if not available

**What type of PR is this?**

Closes #788 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/lint/lint.go
+++ b/pkg/extensions/lint/lint.go
@@ -53,17 +53,54 @@ func (linter *Linter) CheckMandatoryAnnotations(repo string, manifestDigest godi
 		return false, err
 	}
 
-	annotations := manifest.Annotations
+	mandatoryAnnotationsMap := make(map[string]bool)
+	for _, annotation := range mandatoryAnnotationsList {
+		mandatoryAnnotationsMap[annotation] = false
+	}
 
-	for _, annot := range mandatoryAnnotationsList {
-		_, found := annotations[annot]
-
-		if !found {
-			// if annotations are not found, return false but it's not an error
-			linter.log.Error().Msgf("linter: missing %s annotations", annot)
-
-			return false, nil
+	manifestAnnotations := manifest.Annotations
+	for annotation := range manifestAnnotations {
+		if _, ok := mandatoryAnnotationsMap[annotation]; ok {
+			mandatoryAnnotationsMap[annotation] = true
 		}
+	}
+
+	missingAnnotations := getMissingAnnotations(mandatoryAnnotationsMap)
+	if len(missingAnnotations) == 0 {
+		return true, nil
+	}
+
+	// if there are mandatory annotations missing in the manifest, get config and check these annotations too
+	configDigest := manifest.Config.Digest
+
+	content, err = imgStore.GetBlobContent(repo, string(configDigest))
+	if err != nil {
+		linter.log.Error().Err(err).Msg("linter: couldn't get config JSON " + string(configDigest))
+
+		return false, err
+	}
+
+	var imageConfig ispec.Image
+	if err := json.Unmarshal(content, &imageConfig); err != nil {
+		linter.log.Error().Err(err).Msg("linter: couldn't unmarshal config JSON " + string(configDigest))
+
+		return false, err
+	}
+
+	configAnnotations := imageConfig.Config.Labels
+
+	for annotation := range configAnnotations {
+		if _, ok := mandatoryAnnotationsMap[annotation]; ok {
+			mandatoryAnnotationsMap[annotation] = true
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/extensions/lint/lint.go
```diff
@@ -53,17 +53,57 @@ func (linter *Linter) CheckMandatoryAnnotations(repo string, manifestDigest godi
...
-			// if annotations are not found, return false but it's not an error
-			linter.log.Error().Msgf("linter: missing %s annotations", annot)
+	configAnnotations := manifest.Config.Annotations
+	for annotation := range configAnnotations {
+		if _, ok := mandatoryAnnotationsMap[annotation]; ok {
+			mandatoryAnnotationsMap[annotation] = true
+		}
+	}
+
+	if !checkAllAnnotationsFound(mandatoryAnnotationsMap) {
```
peusebiu, line 95: What if we have a single function getMissingAnnotations(annotations, mandatoryAnnotations map[string]bool)
```
		for annotation, flag := range mandatoryAnnotationsMap {
			if !flag {
				missingAnnotations = append(missingAnnotations, annotation)
			}
		}
```

and the check after manifest would be if len(missingAnnotations) == 0 { return true, nil}
check after config would be 
```
              if len(missingAnnotations) > 0 {
                   		linter.log.Error().Msgf("linter: missing %s annotations:", missingAnnotations)
                                return false, nil
               } 
```

nicoldr: I implemented a new method as you suggested, replacing those checks in one method getMissingAnnotations()

pkg/extensions/lint/lint.go
```diff
@@ -53,19 +53,57 @@ func (linter *Linter) CheckMandatoryAnnotations(repo string, manifestDigest godi
...
-			return false, nil
+	configAnnotations := manifest.Config.Annotations
+	for annotation := range configAnnotations {
+		if _, ok := mandatoryAnnotationsMap[annotation]; ok {
+			mandatoryAnnotationsMap[annotation] = true
 		}
 	}
 
+	missingAnnotations = getMissingAnnotations(mandatoryAnnotationsMap)
+	if len(missingAnnotations) == 0 {
```
peusebiu, line 97: remove this, no need for it

pkg/extensions/lint/lint.go
```diff
@@ -53,19 +53,53 @@ func (linter *Linter) CheckMandatoryAnnotations(repo string, manifestDigest godi
...
-		if !found {
-			// if annotations are not found, return false but it's not an error
-			linter.log.Error().Msgf("linter: missing %s annotations", annot)
+	if err != nil {
+		linter.log.Error().Err(err).Msg("linter: couldn't get config JSON" + string(configDigest))
+
+		return false, err
+	}
+
+	if err := json.Unmarshal(content, &manifest); err != nil {
```
peusebiu, line 83: this should be config struct not manifest

pkg/extensions/lint/lint.go
```diff
@@ -53,19 +53,53 @@ func (linter *Linter) CheckMandatoryAnnotations(repo string, manifestDigest godi
...
+		linter.log.Error().Err(err).Msg("linter: couldn't unmarshal config JSON")
 
-			return false, nil
+		return false, err
+	}
+
+	configAnnotations := manifest.Config.Annotations
```
peusebiu, line 89: here you are looking on config descriptor annotations, which is found in the manifest, we need to look on config labels from config struct (not its descriptor)

look at my pr: https://github.com/project-zot/zot/pull/787

so you need to unmarshal the config blob in a ispec.ImageConfig var and look on its labels.

Right @andaaron ?

pkg/extensions/lint/lint_test.go
```diff
@@ -241,6 +242,60 @@ func TestVerifyMandatoryAnnotations(t *testing.T) {
...
+
+		manifestBlob := resp.Body()
+		var manifest ispec.Manifest
+		err = json.Unmarshal(manifestBlob, &manifest)
+		So(err, ShouldBeNil)
+
+		manifest.Annotations = make(map[string]string)
+
+		manifest.Annotations["annotation1"] = "testFail1"
+		manifest.Config.Annotations = make(map[string]string)
```
andaaron, line 286: Needs to be updated according to Sebi's comment, to verify an object of type  ispec.ImageConfig

peusebiu (approved):
LGTM

andaaron (approved):
Looks good.

Issue #606, opened 2022-07-07
Add graphql playground to zot


bogdanbiv:
**Is your feature request related to a problem? Please describe.**
Nice to have: documenting APIs in an user interactive way (like Swagger/OpenAPI UIs)

**Describe the solution you'd like**
Graphql playground allows self documenting APIs. Possible implementation (candidates): https://github.com/graphql-go/playground

**Describe alternatives you've considered**
 * Postman
 * cURL
 * graphqlbin https://www.graphqlbin.com/v2/6RQ6TM

**Additional context**
<img width="674" alt="GraphQL API Playground for SWAPI - The Star Wars API Database Hub" src="https://user-images.githubusercontent.com/104334/177834345-a8c1b531-7246-42c4-aca5-21b1a482dac9.png">

Demo workhorse 
 + https://swapi-graphql.netlify.app/.netlify/functions/index, 
 + https://swapi-graphql.netlify.app

andaaron:
Also interesting to look at: https://github.com/graphql/graphiql

PR #810, opened 2022-09-22
ci(workflows): changed workflow to generate introspection json when gql schema changed

chofnar:
**What type of PR is this?**

documentation

Part of #606 

This PR adds a few new steps that only run on linux-amd64. They start a zot instance and do a GraphQL introspection query, saves the result to a json and uploads it as an artifact.

I added a small comment in the GraphQL schema for testing purposes. For speed, I modified the workflow to only build the binary and not run tests. This will be reverted when the PR is ready to merge.

Example result, at the bottom of the page: https://github.com/project-zot/zot/actions/runs/3106672044


.github/introspection-query.json
```diff
@@ -0,0 +1 @@
+{"query": "{__schema{queryType{name}mutationType{name}subscriptionType{name}types{...FullType}directives{name description locations args{...InputValue}}}}fragment FullType on __Type{kind name description fields(includeDeprecated:true){name description args{...InputValue}type{...TypeRef}isDeprecated deprecationReason}inputFields{...InputValue}interfaces{...TypeRef}enumValues(includeDeprecated:true){name description isDeprecated deprecationReason}possibleTypes{...TypeRef}}fragment InputValue on __InputValue{name description type{...TypeRef}defaultValue}fragment TypeRef on __Type{kind name ofType{kind name ofType{kind name ofType{kind name ofType{kind name ofType{kind name ofType{kind name ofType{kind name}}}}}}}}"}
```
rchincha, line 1: can we format this better?

shimish2 (commented):
@andaaron Is it the same PR you mentioned during our discussion that will update documents as per schema ?

andaaron:
> @andaaron Is it the same PR you mentioned during our discussion that will update documents as per schema

This is the one I meant.

.github/introspection-query.json
chofnar: Unfortunately there's not much I can do without breaking other stuff. It's a pretty standard introspection query, it's not something specific to our schema

rchincha, line 1: No, move this file under pkg/debug/

.github/workflows/ci-cd.yml
```diff
@@ -105,6 +105,31 @@ jobs:
...
+          mkdir introspection
+          curl -X POST -H "Content-Type: application/json" -d @.github/introspection-query.json http://localhost:5000/v2/_zot/ext/search | jq > introspection/introspection-result.json
+          pkill zot
+
+      - name: Upload GraphQL Introspection JSON
+        uses: actions/upload-artifact@v3
+        if: steps.changed-files-specific.outputs.any_changed == 'true' && matrix.os == 'linux' && matrix.arch == 'amd64'
+        with:
+          name: introspection-query.json
+          path: introspection/introspection-result.json
```
rchincha, line 131: ok, this makes sense.

chofnar: Renamed the artifact name to introspection-result to avoid confusion

.github/introspection-query.json
chofnar: Moved

rchincha (approved):
lgtm

PR #822, opened 2022-09-27
use Created field to get the timestamp of image

### LINKED ISSUE #821: Fix "latest" timestamp calculation for the latest tag in a repo
Current implementation takes:
https://github.com/project-zot/zot/blob/main/pkg/extensions/search/common/oci_layout.go#L282

This is not correct, as that is the oldest layer in the image history. We should take the newest. Or check the value of the `Created` timestamp of the image itself.

Proposal is to:
1. Try to get: https://github.com/opencontainers/image-spec/blob/main/specs-go/v1/config.go#L84
2. If the data in #1 is not available, use `imageInfo.History[len(imageInfo.History)-1].Created`
3. If that is not available either, use the current time per current behavior.

shimish2:
if Created field is not set use last entry of history field present in Image object. Last entry of history list is latest layer entry present in image.

if History field is not set use empty time.Time object.

Closes #821

Signed-off-by: Shivam Mishra <shimish2@cisco.com>

**What type of PR is this?**

bug fix

#821 

### DIFF
--- a/pkg/extensions/search/common/oci_layout.go
+++ b/pkg/extensions/search/common/oci_layout.go
@@ -276,12 +276,12 @@ func (olu BaseOciLayoutUtils) GetImageTagsWithTimestamp(repo string) ([]TagInfo,
 				return tagsInfo, err
 			}
 
-			var timeStamp time.Time
+			timeStamp := *imageInfo.Created
 
-			if len(imageInfo.History) != 0 {
-				timeStamp = *imageInfo.History[0].Created
-			} else {
-				timeStamp = time.Time{}
+			defaultTime := time.Time{}
+
+			if timeStamp == defaultTime && len(imageInfo.History) != 0 {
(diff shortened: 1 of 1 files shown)


pkg/extensions/search/common/oci_layout.go
```diff
@@ -338,12 +338,12 @@ func (olu BaseOciLayoutUtils) CheckManifestSignature(name string, digest godiges
 }
 
 func (olu BaseOciLayoutUtils) GetImageLastUpdated(imageInfo ispec.Image) time.Time {
```
andaaron, line 340: I think this should be a function instead of a method of `olu`. It doesn't need anything from this object, doesn't read from the disk.
We we could call it in `GetImageTagsWithTimestamp` instead of duplicating the code. Or at least implement a new function in common.go and call it in both these existing places if we don't want to be disruptive..

shimish2:
Closing this PR, changes are done in this PR #863

PR #825, opened 2022-09-27
Remove forking logger

laurentiuNiculae:
- no longer needed, the race conditions were fixed

**What type of PR is this?**
cleanup

We are cleaning the forks made to the logger because they are no longer needed. Issue with race condition fixed in #817 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/extension_metrics.go
+++ b/pkg/extensions/extension_metrics.go
@@ -26,10 +26,8 @@ func EnableMetricsExtension(config *config.Config, log log.Logger, rootDir strin
 }
 
 func SetupMetricsRoutes(config *config.Config, router *mux.Router, storeController storage.StoreController,
-	l log.Logger,
+	log log.Logger,
 ) {
-	// fork a new zerolog child to avoid data race
-	log := log.Logger{Logger: l.With().Caller().Timestamp().Logger()}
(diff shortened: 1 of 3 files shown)


peusebiu (approved):
lgtm

pkg/extensions/extension_metrics.go
```diff
@@ -26,10 +26,8 @@ func EnableMetricsExtension(config *config.Config, log log.Logger, rootDir strin
 }
 
 func SetupMetricsRoutes(config *config.Config, router *mux.Router, storeController storage.StoreController,
-	l log.Logger,
+	log log.Logger,
 ) {
-	// fork a new zerolog child to avoid data race
-	log := log.Logger{Logger: l.With().Caller().Timestamp().Logger()}
```
rchincha, line 32: Just make sure the original logger has all these options set up.

laurentiuNiculae: The original logger has those options. All loggers created with NewLogger are created like this:
```
Logger{Logger: log.Hook(goroutineHook{}).With().Caller().Timestamp().Logger()}
```

rchincha (approved):
lgtm

PR #827, opened 2022-09-27
Add enable/disable option for scrub extension

Andreea-Lupu:

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/config/config.go
+++ b/pkg/extensions/config/config.go
@@ -39,5 +39,6 @@ type PrometheusConfig struct {
 }
 
 type ScrubConfig struct {
+	Enable   *bool
 	Interval time.Duration
 }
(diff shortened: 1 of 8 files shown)


pkg/extensions/config/config.go
```diff
@@ -39,5 +39,6 @@ type PrometheusConfig struct {
 }
 
 type ScrubConfig struct {
+	Enable   *bool
```
rchincha, line 42: Yes, we missed this in the last review.

pkg/extensions/extension_scrub.go
```diff
@@ -20,7 +20,7 @@ func EnableScrubExtension(config *config.Config, log log.Logger, storeController
 	sch *scheduler.Scheduler,
 ) {
 	if config.Extensions.Scrub != nil &&
-		config.Extensions.Scrub.Interval != 0 {
+		*config.Extensions.Scrub.Enable {
```
rchincha, line 23: Also add the Interval check, what if it was set to zero by mistake?

Andreea-Lupu: We have a minimum value for this interval which is 2h, so if this interval is set to zero in config then it will be automatically replaced by this minimum value.

rchincha:
Mark this ready for review if ready.

rchincha (approved):
lgtm

Issue #806, opened 2022-09-21
Support storing image logos to be later shown in ZUI for specific repos


andaaron:
Harbor seems to have defined it's own oci artifacts with specific labels for this purpose: https://goharbor.io/docs/2.3.0/administration/user-defined-oci-artifact/

andaaron:
This is in the UI integration branch.

PR #833, opened 2022-09-28
Support storing image logos as manifest annotations

alexstan12:
This commit allows storing base64 encoded logos under image manifest annotations, using com.zot.logo key. Logos have type limitations (only jpeg, png and gif allowed) and a maximum allowed size (200x200 px).

**What type of PR is this?**
feature

closes https://github.com/project-zot/zot/issues/806

**Automation added to e2e**:

### DIFF
--- a/pkg/extensions/lint/lint.go
+++ b/pkg/extensions/lint/lint.go
@@ -4,13 +4,23 @@
 package lint
 
 import (
+	"bytes"
+	"encoding/base64"
 	"encoding/json"
+	"image"
+	_ "image/gif"  // nolint:gci  // imported for the registration of it's decoder func
+	_ "image/jpeg" // nolint:gci // imported for the registration of it's decoder func
+	_ "image/png"  // nolint:gci  // imported for the registration of it's decoder func
 
 	godigest "github.com/opencontainers/go-digest"
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"zotregistry.io/zot/pkg/extensions/config"
 	"zotregistry.io/zot/pkg/log"
 	"zotregistry.io/zot/pkg/storage"
+)
+
+const (
+	logoKey string = "com.zot.logo"
 )
 
 type Linter struct {
@@ -61,6 +71,13 @@ func (linter *Linter) CheckMandatoryAnnotations(repo string, manifestDigest godi
 	manifestAnnotations := manifest.Annotations
 	for annotation := range manifestAnnotations {
 		if _, ok := mandatoryAnnotationsMap[annotation]; ok {
+			if annotation == logoKey {
+				if ok := linter.checkIfLogoAnnotationValid(manifestAnnotations[annotation]); ok {
+					mandatoryAnnotationsMap[annotation] = true
+				}
+
+				continue
+			}
 			mandatoryAnnotationsMap[annotation] = true
 		}
 	}
@@ -88,9 +105,15 @@ func (linter *Linter) CheckMandatoryAnnotations(repo string, manifestDigest godi
 	}
 
 	configAnnotations := imageConfig.Config.Labels
-
 	for annotation := range configAnnotations {
 		if _, ok := mandatoryAnnotationsMap[annotation]; ok {
+			if annotation == logoKey {
+				if ok := linter.checkIfLogoAnnotationValid(configAnnotations[annotation]); ok {
+					mandatoryAnnotationsMap[annotation] = true
+				}
+
+				continue
+			}
 			mandatoryAnnotationsMap[annotation] = true
 		}
 	}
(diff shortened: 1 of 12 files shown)


andaaron (changes requested):
Please move the logo validation logic to the linter.

pkg/extensions/search/common/oci_layout.go
```diff
@@ -404,6 +405,11 @@ func (olu BaseOciLayoutUtils) GetExpandedRepoInfo(name string) (RepoInfo, error)
 			continue
 		}
 
+		logo, ok := man.Annotations["com.zot.logo"]
```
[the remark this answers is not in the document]
alexstan12: logic changed, now gathering logo info from the result of common.GetAnnotations().

pkg/extensions/lint/lint.go
```diff
@@ -61,6 +71,13 @@ func (linter *Linter) CheckMandatoryAnnotations(repo string, manifestDigest godi
 	manifestAnnotations := manifest.Annotations
 	for annotation := range manifestAnnotations {
```
andaaron, line 72: Why not pick the annotation value from here and pass it to linter.checkIfLogoAnnotationValid() as the parameter?

Also can we add the same logic for lines 109-113?

pkg/extensions/lint/lint.go
```diff
@@ -123,3 +140,38 @@ func getMissingAnnotations(mandatoryAnnotationsMap map[string]bool) []string {
 
 	return missingAnnotations
 }
+
+func (linter *Linter) checkIfLogoAnnotationValid(annotation string, mandatoryAnnotations map[string]string) bool {
+	if val, ok := mandatoryAnnotations[logoKey]; ok {
```
andaaron, line 145: If we pass the annotation value as parameter we should not need this if. It should be cleaner. This function should be responsible for validating annotations, and it shouldn't need to be aware that some of them are mandatory and some ore not.

pkg/extensions/lint/lint.go
```diff
@@ -123,3 +140,38 @@ func getMissingAnnotations(mandatoryAnnotationsMap map[string]bool) []string {
...
+		if err != nil {
+			linter.log.Error().Err(err).Msg("unable to decode value")
+
+			return false
+		}
+		imageVal := bytes.NewBuffer(decodedVal)
+
+		logoConf, format, err := image.DecodeConfig(imageVal)
+		if err != nil {
+			linter.log.Error().Err(err).Msg("unable to decode image")
```
andaaron, line 156: Maybe we should include repo name and digest in error messages.

pkg/extensions/lint/lint.go
```diff
@@ -123,3 +140,38 @@ func getMissingAnnotations(mandatoryAnnotationsMap map[string]bool) []string {
...
+
+			return false
+		}
+
+		if format != "jpeg" && format != "gif" && format != "png" {
+			linter.log.Error().Err(err).Msg("encoded logo is of incorrect format, allowed formats are jpeg/png/gif")
```
andaaron, line 162:
```suggestion
			linter.log.Error().Msg("encoded logo is of incorrect format, allowed formats are jpeg/png/gif")
```

pkg/extensions/lint/lint.go
```diff
@@ -123,3 +140,38 @@ func getMissingAnnotations(mandatoryAnnotationsMap map[string]bool) []string {
...
+
+			return false
+		}
+
+		if logoConf.Height > 200 || logoConf.Width > 200 {
+			linter.log.Error().Err(err).Msg("encoded logo is of incorrect size")
```
andaaron, line 168:
```suggestion
			linter.log.Error().Msg("encoded logo is of incorrect size")
```

pkg/extensions/search/common/common.go
```diff
@@ -215,6 +223,11 @@ func GetAnnotations(annotations, labels map[string]string) ImageAnnotations {
 		vendor = GetVendor(labels)
 	}
 
+	logo := GetLogo(annotations)
+	if logo == "" {
+		logo = GetLogo(annotations)
```
andaaron, line 228:
```suggestion
		logo = GetLogo(labels)
```

If we want to support labels. I think we can. Otherwise the if is not needed.

pkg/storage/common.go
```diff
@@ -0,0 +1,98 @@
+package storage
```
andaaron, line 1: Remove this change from the PR if it is unrelated to the issue we are resolving.

pkg/extensions/search/common/common.go
alexstan12: Done, changed.

pkg/storage/common.go
alexstan12: removed.

pkg/extensions/lint/lint.go
alexstan12: Changed the signature and it's usages.

pkg/extensions/lint/lint.go
alexstan12: Did the changes. Since the logo annotation is not present in config annotations, I don't see why changes at 109-113 are needed.

andaaron: Because images built using Docker can't set annotations, only labels. So if we want to support images built using Docker we should add support for labels.

pkg/extensions/search/resolver.go
```diff
@@ -575,6 +577,8 @@ func BuildImageInfo(repo string, tag string, manifestDigest godigest.Digest,
 		lastUpdated = imageConfig.History[len(imageConfig.History)-1].Created
 	}
 
+	manifestAnnotations := common.GetAnnotations(manifest.Annotations, manifest.Config.Annotations)
```
andaaron, line 580: Reuse the annotations object defined above. manifest,Config will not contain annotations anyway.

andaaron (commented):
See my reply to reading the labels in the config file.

andaaron (dismissed):

rchincha:
We should really be using #936 for this.

andaaron:
> We should really be using #936 for this.

I agree.
The graphql API would be the same, and return the base64 encoded image?
I think the image blob itself should not be compressed, right? That would all an overhead?

andaaron:
Will redesign this and implement as part of: https://github.com/project-zot/zot/pull/1018

PR #838, opened 2022-10-03
build(swagger): removed swagger requirement from binary-minimal and binary

chofnar:
Also added the "debug" build tag in the binary-debug Make target

**What type of PR is this?**
cleanup

swagger is not needed for the minimal and regular binary builds

Also moved binary-minimal to be the first in the Makefile


rchincha (approved):
lgtm

PR #840, opened 2022-10-03
fix(sync): revert code which removed image destination feature

### LINKED ISSUE #793: Changing image URIs during zot sync feature got dropped in a commit
**Describe the bug**
This feature https://github.com/project-zot/zot/issues/433 got dropped in a new commit by mistake: https://github.com/project-zot/zot/commit/20a60cbad465f6147b0814f1096b4f8483b693e4

peusebiu:
Added an end to end test for this feature, closes #793

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/cli/root.go
+++ b/pkg/cli/root.go
@@ -598,9 +598,17 @@ func validateSync(config *config.Config) error {
 				for _, content := range regCfg.Content {
 					ok := glob.ValidatePattern(content.Prefix)
 					if !ok {
-						log.Error().Err(glob.ErrBadPattern).Str("pattern", content.Prefix).Msg("sync pattern could not be compiled")
+						log.Error().Err(glob.ErrBadPattern).Str("prefix", content.Prefix).Msg("sync prefix could not be compiled")
 
 						return glob.ErrBadPattern
+					}
+
+					if content.StripPrefix && !strings.Contains(content.Prefix, "/*") && content.Destination == "/" {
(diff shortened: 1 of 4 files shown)


pkg/cli/root.go
```diff
@@ -598,10 +598,18 @@ func validateSync(config *config.Config) error {
...
-						log.Error().Err(glob.ErrBadPattern).Str("pattern", content.Prefix).Msg("sync pattern could not be compiled")
+						log.Error().Err(glob.ErrBadPattern).Str("prefix", content.Prefix).Msg("sync prefix could not be compiled")
 
 						return glob.ErrBadPattern
 					}
+
+					if content.StripPrefix && !strings.Contains(content.Prefix, "/*") && content.Destination == "/" {
+						log.Error().Err(errors.ErrBadConfig).
+							Interface("sync content", content).
+							Msg("sync config: can not use stripPrefix true and destination '/' without using glob patterns in prefix")
```
peusebiu, line 609: given a image named /a/b/c
and sync content config:
```
			{
				content:  sync.Content{Prefix: "/a/b/c", Destination: "/", StripPrefix: true},
			}
```

It will put /a/b/c directly into zot-storage.RootDir(), because repoName would be "/", and not under zot-storage.RootDir()/c, so it will break things in zot.

StripPrefix should be use with a valid destination or when using globaPatterns like /a/b/* and it will only strip /a/b/ without c

PR #844, opened 2022-10-03
Add support for Metadata in specific nested buckets

bogdanbiv:
Each user gets one bucket identified by userid

Signed-off-by: Bogdan BIVOLARU <104334+bogdanbiv@users.noreply.github.com>

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -11,7 +11,6 @@ import (
 
 const (
 	S3StorageDriverName = "s3"
-	DefaultGCDelay      = 1 * time.Hour
 )
 
 type ImageStore interface {
--- a/pkg/storage/storage_controller.go
+++ b/pkg/storage/storage_controller.go
@@ -3,11 +3,14 @@ package storage
 import (
 	"fmt"
 	"strings"
+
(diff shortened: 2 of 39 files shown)


pkg/storage/storage_controller.go
```diff
@@ -3,11 +3,14 @@ package storage
...
+
+	"zotregistry.io/zot/pkg/metadata"
 )
 
 type StoreController struct {
-	DefaultStore ImageStore
-	SubStore     map[string]ImageStore
+	DefaultStore  ImageStore
+	SubStore      map[string]ImageStore
+	MetadataStore metadata.Store
```
andaaron, line 13: The MetadataStore is unrelated to the StoreController. Let's keep it separate.

@rchincha ^^

pkg/storage/storage.go
```diff
@@ -11,7 +11,6 @@ import (
 
 const (
 	S3StorageDriverName = "s3"
-	DefaultGCDelay      = 1 * time.Hour
```
andaaron, line 14: Can we get away without updating these imports? If the metadata store is not imported to be used in the storecontroller we should be fine right?

Issue #749, opened 2022-08-25
Implement API to return supported authentication methods for this specific ZOT instance


andaaron:
This API should be accessible without authentication and include all authentication options configured on the server:

- local username/password
- ldap username/password
- certificate authentication
- oauth with github
- oauth with gitlab
- oauth with google

And so on.

andaaron:
In review.

andaaron:
Implemented in https://github.com/project-zot/zot/pull/1198

PR #853, opened 2022-10-06
feat: Implement API to return supported authentication methods, close…

peusebiu:
…s #749

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


rchincha:
ideally, this URI should only accept `localhost` connections

rchincha:
rebase this pls

peusebiu:
done

pkg/extensions/config/config.go
```diff
@@ -12,11 +12,16 @@ type BaseConfig struct {
...
-	Sync    *sync.Config
-	Metrics *MetricsConfig
-	Scrub   *ScrubConfig
-	Lint    *LintConfig
+	Search    *SearchConfig
+	Sync      *sync.Config
+	Metrics   *MetricsConfig
+	Scrub     *ScrubConfig
+	Lint      *LintConfig
+	AuthQuery *AuthQueryConfig
```
rchincha, line 20: Let's call this Mgmt instead, so that we can return/manage config from here

pkg/api/constants/extensions.go
```diff
@@ -5,6 +5,8 @@ const (
 	ExtCatalogPrefix     = "/_catalog"
 	ExtOciDiscoverPrefix = "/_oci/ext/discover"
 	// zot specific extensions.
-	ExtSearchPrefix  = "/_zot/ext/search"
-	FullSearchPrefix = RoutePrefix + ExtSearchPrefix
+	ExtSearchPrefix     = "/_zot/ext/search"
+	ExtAuthQueryPrefix  = "/_zot/ext/authquery"
```
rchincha, line 9: This should be "mgmt"

pkg/extensions/extension_auth_query.go
```diff
@@ -0,0 +1,113 @@
...
+func (authHandler *authHandler) getInfo(response http.ResponseWriter, request *http.Request) {
+	remoteAddr, err := getIP(request)
+	if err != nil {
+		authHandler.log.Error().Err(err).Msg("auth query: couldn't obtain request ip")
+		response.WriteHeader(http.StatusInternalServerError)
+
+		return
+	}
+
+	if !strings.Contains(remoteAddr, "127.0.0.1") {
```
rchincha, line 83: This part is not correct (my bad actually)

server (remote-ip) <-> client (client-ip)

Client has to reach this server URL without authN

pkg/api/constants/extensions.go
rchincha: mgmt/config, mgmt/<something-else>

https://github.com/opencontainers/distribution-spec/tree/main/extensions
^ think we can do that

peusebiu:
ok, will fix

pkg/extensions/config/config.go
peusebiu: but in this case we return only the authentication type
should we call it mgmt/auth ?

rchincha: no, mgmt/config and we return only parts based on access policy

andaaron:
Superseded by https://github.com/project-zot/zot/pull/1198

PR #857, opened 2022-10-06
build(tags): remove redundant build tag ui_base

andaaron:
It was not used for UI, it had become a CLI dependency with the same functionality as search

**What type of PR is this?**
build
cleanup
ci

We had a redundant build tag `ui_base` which was basically the same as `search`.
It was used for cli, not ui.

Cleanup in preparation of adding a `ui` tag

The `ui_base` build tag/label will have no effect.


rchincha (dismissed):
lgtm

peusebiu (approved):
lgtm

PR #863, opened 2022-10-10
fix(lastUpdated): fix image lastUpdated timestamp logic

andaaron:
The lastUpdated field was picked from the first entry in image history Now it is the created time of the image, or the last entry in image history, if created time is unavailable

**What type of PR is this?**
bug

lastUpdated was computed based on wrong fields

### DIFF
--- a/pkg/extensions/search/common/common.go
+++ b/pkg/extensions/search/common/common.go
@@ -75,6 +75,27 @@ func GetImageDirAndTag(imageName string) (string, string) {
 	return imageDir, imageTag
 }
 
+// GetImageLastUpdated This method will return last updated timestamp.
+// The Created timestamp is used, but if it is missing, look at the
+// history field and, if provided, return the timestamp of last entry in history.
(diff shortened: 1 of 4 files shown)


pkg/extensions/search/common/common.go
```diff
@@ -75,6 +75,24 @@ func GetImageDirAndTag(imageName string) (string, string) {
 	return imageDir, imageTag
 }
 
+func GetImageLastUpdated(imageInfo ispec.Image) time.Time {
```
shimish2, line 78: Can we add a comment for this method 

e.g. 
`// GetImageLastUpdated This method will return last updated timestamp. If history field is provided it will return timestamp of last entry in history`.

Feel free to update the comment accordingly.

shimish2 (commented):
Change looks good, added one request to add a comment for GetImageLastUpdated.

shimish2 (approved):
lgtm

PR #871, opened 2022-10-11
refactor(tests): remove hardcoded digests

chofnar:
**What type of PR is this?**
cleanup

Removes hardcoded digests where possible

### DIFF
--- a/pkg/test/common.go
+++ b/pkg/test/common.go
@@ -29,6 +29,29 @@ const (
 	SleepTime     = 100 * time.Millisecond
 )
 
+// which: manifest, config, layer
+func GetTestBlobDigest(image, which string) godigest.Digest {
+	prePath := "../test/data"
+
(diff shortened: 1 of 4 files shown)


pkg/test/common.go
```diff
@@ -29,6 +29,20 @@ const (
 	SleepTime     = 100 * time.Millisecond
 )
 
+//nolint:gochecknoglobals
+var TestBlobDigests = map[string]map[int]string{
+	"zot-cve-test": {
+		0: "7a0437f04f83f084b7ed68ad9c4a4947e12fc4e1b006b38129bac89114ec3621",
```
andaaron, line 35: Can we avoid using the digest values?
The issue is different versions of skopeo copy the image with different hashes.
So different developers may have different values on their environments.

rchincha (approved):
lgtm

PR #872, opened 2022-10-11
test: Build images with annotations

nicoldr:

**What type of PR is this?**

Solves #850 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/.github/workflows/ecosystem-tools.yaml
+++ b/.github/workflows/ecosystem-tools.yaml
@@ -35,19 +35,22 @@ jobs:
           skopeo -v
       - name: Run push-pull tests
         run: |
-            make push-pull
+            make test-push-pull
       - name: Run metrics tests
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/ecosystem-tools.yaml
```diff
@@ -51,3 +51,6 @@ jobs:
       - name: Run anonymous-push-pull tests
         run: |
             make anonymous-push-pull
```
rchincha, line 53: we should really rename these Makefile targets with a test-* prefix
make test-annotations test-anon-push-pull etc
@nicoldr can we fix this also, your PR otherwise looks ok

nicoldr: Added "test-" before every target of that kind

rchincha (approved):
lgtm

PR #883, opened 2022-10-17
fix(s3): remove tracking multipart uploads

peusebiu:
Remove sticky sessions from clustering

**What type of PR is this?**
cleanup

Removes the need of sticky sessions in zot cluster load balancer (haproxy)

no

### DIFF
--- a/pkg/storage/s3/s3.go
+++ b/pkg/storage/s3/s3.go
@@ -42,14 +42,10 @@ type ObjectStorage struct {
 	lock        *sync.RWMutex
 	blobUploads map[string]storage.BlobUpload
 	log         zerolog.Logger
-	// We must keep track of multi part uploads to s3, because the lib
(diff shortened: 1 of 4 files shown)


andaaron (dismissed):
LGTM

pkg/storage/s3/s3.go
```diff
@@ -45,11 +45,10 @@ type ObjectStorage struct {
 	// We must keep track of multi part uploads to s3, because the lib
```
rchincha, line 45: remove this comment now?

rchincha: so we don't need to track the multipart then?

peusebiu: yes, we don't anymore :D , I'll remove it

rchincha (approved):
lgtm

Issue #891, opened 2022-10-19
Fix code coverage issues with scheduler


andaaron:
![Image](https://user-images.githubusercontent.com/25718736/196672794-c556ca0e-7d7e-4473-b014-5da3490dbbfd.png)

PR #893, opened 2022-10-19
fix: flaky scheduler coverage

chofnar:
**What type of PR is this?**
bug

#891

Coverage sometimes failed for a function within the scheduler code. There was a range of a slice that was iterated upon and then modified, leading to unexpected behavior. 

Also adds a test to check for the waiting task generator (tested with 150 runs) + fixes a periodic GC test that sometimes failed because it didn't have enough time to execute its async task (ran it 500 times on my machine to ensure it passes consistently)


rchincha:
@Andreea-Lupu can you pls review this also?

Andreea-Lupu:
LGTM

andaaron (approved):

Issue #566, opened 2022-05-30
signatures: call notation-go libs instead of invoking notation directly in UTs


rchincha:
**Is your feature request related to a problem? Please describe.**

`zot` invokes "notation" directly in UTs.
https://github.com/project-zot/zot/blob/main/pkg/api/controller_test.go#L3457

We can test/assert individual steps.

**Describe the solution you'd like**

Instead, can we link to this?
https://github.com/notaryproject/notation-go

**Additional context**

https://github.com/notaryproject/notation/pull/141

andaaron:
In progress.

andaaron:
Fixed in #1104

PR #896, opened 2022-10-19
Notation verify implementation

roxanaN:

https://github.com/project-zot/zot/issues/566

call notation-go libs instead of invoking "notation" directly in UTs.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module zotregistry.io/zot
 
-go 1.19
+go 1.18
 
 require (
 	github.com/99designs/gqlgen v0.17.22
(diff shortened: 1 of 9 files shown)


go.mod
```diff
@@ -53,10 +53,14 @@ require (
 require (
 	github.com/aquasecurity/trivy v0.0.0-00010101000000-000000000000
 	github.com/containers/image/v5 v5.22.0
-	github.com/notaryproject/notation-go v0.10.0-alpha.3
+	github.com/notaryproject/notation v0.11.0-alpha.4
```
dtzar, line 56: You shouldn't need the `notation` since it's only for CLI exection. 
If you need `notation-core-go` - the notation team would like to know why as notation-go should be all you need to do whatever you need to do.

rchincha: @roxanaN FYI a little context wrt @dtzar 's suggestion.
 https://pkg.go.dev/github.com/notaryproject/notation-go@main

andaaron:
Work continues as part of https://github.com/project-zot/zot/pull/1104

Issue #869, opened 2022-10-11
Use consistent Digest types throughout zot


andaaron:
Within the zot code we don't have a consistent practice for handling digests. Sometimes we:
- Use data type from: github.com/opencontainers/go-digest
- Use data type from: github.com/google/go-containerregistry/pkg/v1
- Use string data type with algorithm included
- Use string data type without algorithm

Ideally we would use something like github.com/opencontainers/go-digest and convert it to something else if a 3rd party library expects something else (thinking about CVE related code).

andaaron:
Done

PR #898, opened 2022-10-20
refactor(digests): standardise representation of digests to digest.Digest

andaaron:
- Digests were represented by different ways
  - We needed a uniform way to represent the digests and enforce a format
  - also replace usage of github.com/google/go-containerregistry/pkg/v1 with github.com/opencontainers/image-spec/specs-go/v1

Signed-off-by: Laurentiu Niculae <niculae.laurentiu1@gmail.com>
(cherry picked from commit 96b2f29d6d57070a913ce419149cd481c0723815) 
(cherry picked from commit 3d41b583daea654c98378ce3dcb78937d71538e8)

**What type of PR is this?**
refactoring

https://github.com/project-zot/zot/issues/869

In some cases it's unclear if the digest strings are supposed to contain just the hex or also the algorithm.
API now returns full digests with the algorithm name, and the cli now removes it for easy of view purpose (in any case if we don't remove it the algorithm name uses half the characters allowed by the digest column width).
Using the go-digest.Digest struct has some additional benefits such as having validation as methods.
Also we were using 2 different structs for manifest used in some places, resulting in incompatibility between these types when we wanted to compare objects, and duplicating methods returning basically the same information in different struct types. 

### DIFF
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -3294,12 +3294,12 @@ func TestCrossRepoMount(t *testing.T) {
 			baseURL, constants.RoutePrefix, constants.Blobs, godigest.SHA256, blob))
 
 		// Check os.SameFile here
-		cachePath := path.Join(ctlr.Config.Storage.RootDirectory, "zot-d-test", "blobs/sha256", dgst.Hex())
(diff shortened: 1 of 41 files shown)


rchincha (approved):
lgtm

PR #922, opened 2022-10-25
fix(sync): PreserveDigests flag prevents docker mediatype manifests t…

peusebiu:
…o be converted to oci mediatype

**What type of PR is this?**
bug

Because of preserveDigests flag, sync on demand doesn't work with docker.io (dockerhub) images
because it can not converts them from docker mediatype to oci mediatype

Fix sync on demand with dockerhub

no

no

```release-note

```


andaaron (dismissed):

peusebiu:
fixed in https://github.com/project-zot/zot/pull/932

PR #932, opened 2022-10-27
fix(sync): also sync on demand digests, not only tags, closes #902

### LINKED ISSUE #902: Fix sync on demand: On demand works only with tags but not with digests
**Describe the bug**
For example
```skopeo copy docker://localhost:8080/alpine@sha256:528b2606ec71fa73fad684f2aa7a1bcce795e4d1569e26c4ce0340273a65ab98```

it will query based on digest and sync on demand will get an error because it expects a tag.

**Expected behavior**
sync on demand should work both with tags and digests.

### PR DESCRIPTION
**Pull request by peusebiu**

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


rchincha (approved):
lgtm

PR #948, opened 2022-11-01
feat: gc untagged manifests and signatures without references

peusebiu:

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/local/local.go
+++ b/pkg/storage/local/local.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
@@ -18,6 +19,7 @@ import (
 	apexlog "github.com/apex/log"
 	guuid "github.com/gofrs/uuid"
 	"github.com/minio/sha256-simd"
+	notreg "github.com/notaryproject/notation-go/registry"
 	godigest "github.com/opencontainers/go-digest"
 	imeta "github.com/opencontainers/image-spec/specs-go"
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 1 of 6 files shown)


pkg/storage/local/local.go
```diff
@@ -1477,12 +1491,131 @@ func ensureDir(dir string, log zerolog.Logger) error {
...
+	foundSubject = false
+
+	for _, notationDesc := range notationDescriptors {
+		// check if we can find the manifest which the signature points to
+		artManifest, err := getArtifactManifestByDesc(dir, notationDesc, is.log)
+		if err != nil {
+			return err
+		}
+
+		if artManifest.ArtifactType != "application/vnd.cncf.notary.v2.signature" {
```
andaaron, line 1597: Isn't there a constant defined in the notation library for this?

pkg/storage/storage_test.go
```diff
@@ -508,16 +508,16 @@ func TestStorageAPIs(t *testing.T) {
 							manifestBuf, err = json.Marshal(manifest)
 							So(err, ShouldBeNil)
 							digest := godigest.FromBytes(manifestBuf)
-							_, err = imgStore.PutImageManifest("test", digest.String(),
+							_, err = imgStore.PutImageManifest("test", "0.0.1",
```
andaaron, line 511: Why do we need these changes?

pkg/storage/local/local.go
peusebiu: fixed, thanks

pkg/storage/storage_test.go
peusebiu: this was before adding isBlobOlderThan(gc.delay) check, thanks for the heads up, I have reverted these changes.

rchincha (approved):
lgtm

PR #951, opened 2022-11-02
fix(storage): deleting manifests with identical digests

peusebiu:
Suppose we push two identical manifests (sharing same digest) but with different tags, then deleting by digest should throw an error otherwise we end up deleting all image tags (with gc) or dangling references (without gc)

This behaviour is controlled via Authorization, added a new policy action named detectManifestsCollision which enables this behaviour

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -175,8 +175,11 @@ Should authentication fail, to prevent automated attacks, a delayed response can
 ## Identity-based Authorization
 
 Allowing actions on one or more repository paths can be tied to user
-identities. An additional per-repository default policy can be specified for
-identities not in the whitelist. Furthermore, a global admin policy can also be
+identities. Two additional per-repository policies can be specified for identities not in the whitelist:
+- anonymousPolicy - applied for unathenticated users.
+- defaultPolicy - applied for authenticated users.
+
+Furthermore, a global admin policy can also be
 specified which can override per-repository policies.
 
 Glob patterns can also be used as repository paths.
@@ -191,7 +194,15 @@ because it will be longer. So that's why we have the option to specify an adminP
 
 Basically '**' means repositories not matched by any other per-repository policy.
 
-create/update/delete can not be used without 'read' action, make sure read is always included in policies!
+Method-based action list:
+- "read" - list/pull images
+- "create" - push images (needs "read")
+- "update" - overwrite tags (needs "read" and "create")
+- "delete" - delete images (needs "read")
+
+Behaviour-based action list
+- "detectManifestCollision" - delete manifest by digest will throw an error if multiple manifests have the same digest (needs "read" and "delete")
+
 
 ```
 "accessControl": {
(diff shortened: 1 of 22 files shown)


pkg/api/routes.go
```diff
@@ -475,7 +475,7 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
 			// could be syscall.EMFILE (Err:0x18 too many opened files), etc
 			rh.c.Log.Error().Err(err).Msg("unexpected error: performing cleanup")
 
-			if err = imgStore.DeleteImageManifest(name, reference); err != nil {
+			if err = imgStore.DeleteImageManifest(name, reference, false); err != nil {
```
rchincha, line 478: can we pass a "policy" argument here, so as to become future proof?

pkg/api/routes.go
```diff
@@ -521,14 +521,30 @@ func (rh *RouteHandler) DeleteManifest(response http.ResponseWriter, request *ht
...
+	acCtx, err := localCtx.GetAccessControlContext(request.Context())
+	if err != nil {
+		response.WriteHeader(http.StatusInternalServerError)
+
+		return
+	}
+
+	var detectCollision bool
+	if acCtx != nil {
+		detectCollision = acCtx.CanDetectManifestsCollision(name)
```
rchincha, line 534: perfect

pkg/api/routes.go
rchincha: Let's do this later.

pkg/extensions/search/resolver.go
```diff
@@ -722,40 +721,21 @@ func BuildImageInfo(repo string, tag string, manifestDigest godigest.Digest,
 	return imageInfo
 }
 
-// returns either a user has or not rights on 'repository'.
-func matchesRepo(globPatterns map[string]bool, repository string) bool {
```
rchincha, line 726: Why is this part removed?

peusebiu: I moved it here: https://github.com/project-zot/zot/pull/951/files#diff-78f3d35f6fab9bf5af09ca967e0e55ec0ccab9cd7ab67fe5dfce6f51ad59f7a6R46
to be more explicit.

pkg/api/authz.go
```diff
@@ -18,11 +18,13 @@ import (
...
 
 const (
-	// actions.
+	// method actions.
 	CREATE = "create"
 	READ   = "read"
 	UPDATE = "update"
 	DELETE = "delete"
+	// behaviour actions.
+	DETECT_MANIFESTS_COLLISION = "detectManifestsCollision" //nolint: stylecheck,revive
```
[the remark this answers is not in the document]
peusebiu: ok, will change

pkg/api/authz.go
```diff
@@ -18,11 +18,13 @@ import (
 )
 
 const (
-	// actions.
+	// method actions.
 	CREATE = "create"
```
rchincha, line 22: @peusebiu change these to golang camelcase. These are not special, so no reason to do it this way.

pkg/requestcontext/context.go
```diff
@@ -12,7 +19,54 @@ func GetContextKey() *Key {
...
+	if authCtx := ctx.Value(authzCtxKey); authCtx != nil {
+		acCtx, ok := authCtx.(AccessControlContext)
+		if !ok {
+			return nil, errors.ErrBadType
+		}
+
+		return &acCtx, nil
+	}
+
+	//nolint: nilnil
```
rchincha, line 41: put this coment on the same line

pkg/api/authz.go
rchincha: also remove the nolint

pkg/api/authz.go
```diff
@@ -18,11 +18,13 @@ import (
...
-	READ   = "read"
-	UPDATE = "update"
-	DELETE = "delete"
+	// method actions.
+	Create = "create"
+	Read   = "read"
+	Update = "update"
+	Delete = "delete"
+	// behaviour actions.
+	DetectManifestCollission = "detectManifestCollision"
```
rchincha, line 27: "Collision"

examples/README.md
```diff
@@ -191,7 +194,13 @@ because it will be longer. So that's why we have the option to specify an adminP
 
 Basically '**' means repositories not matched by any other per-repository policy.
 
-create/update/delete can not be used without 'read' action, make sure read is always included in policies!
+Actions list:
```
rchincha, line 197: ## Method-based
...

## Behavior-based
...

rchincha (commented):
Add a bats test case so we know how external tools will behave when they hit this policy.

"skopeo delete should fail"
"regctl"

rchincha (approved):
lgtm

PR #953, opened 2022-11-03
feat(cache): dynamodb implementation

chofnar:
**What type of PR is this?**

enhancement

Part of #564 

Added AWS DynamoDB driver.

If a client uses DynamoDB and decides to downgrade, they must make sure that the database is migrated to BoltDB to avoid data loss.

New validations that might break stuff:
local storage driver, but explicit RemoteCache=true, this is not valid, so error out
remote storage driver, but no "cacheDriver" config section, error out

Defaults:
local storage driver -> RemoteCache=false
remote storage driver (s3) -> RemoteCache=true

Adds the possibility to use AWS DynamoDB instead of BoltDB

**Problems**
- need to specify protocol in front of dynamodb endpoint [(known issue)](https://github.com/aws/aws-sdk-go-v2/issues/1294)

**To do**
- [x] fix lint problems
- [x] remove comment from controller.go line 461
- [x] fix misleading example config (local storage with remote cache, is invalid)
- [x] check that the example config actually works
- [x] add negative tests
- [x] previous tests pass
- [x] add coverage tests


peusebiu:
let's add a new testCase in TestS3Dedupe https://github.com/project-zot/zot/blob/main/pkg/storage/s3/s3_test.go#L935

Currently TestS3Dedupe it's using boltdb, let's use both boltdb and DynamoDB, this should be a small change.

rchincha:
Thanks for tracking your PR with task items, makes it easier to follow.

examples/config-dynamodb.json
```diff
@@ -0,0 +1,29 @@
+{
+    "distSpecVersion": "1.0.1-dev",
+    "storage": {
```
rchincha, line 3: trivy's CVE db is the only one that should take up local disk. If that is disabled in config, can we make sure that nothing is present on local disk except config?

.github/workflows/ci-cd.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       s3mock:
         image: ghcr.io/project-zot/localstack/localstack:1.2.0
         env:
-          SERVICES: s3
+          SERVICES: s3,dynamodb
```
rchincha, line 41: `localstack` is used to mock s3. What is being used for dynamodb?

chofnar: It is used to mock dynamodb for tests

examples/config-dynamodb.json
chofnar: Do you mean that nothing should be written on the disk by zot if all the config is cloud-based, and a mechanism to check for that should be implemented? I'm not sure I understand

Issue #930, opened 2022-10-26
ZLI: signatures are always "false" if graphql API is not available


andaaron:
If the zot server doesn't have GraphQL available and zli is using only REST calls, the images are always shown as not signed.
We could make additional calls to get signature information.

To reproduce add this test:

```
diff --git a/pkg/cli/image_cmd_test.go b/pkg/cli/image_cmd_test.go
index 2ec1e310..a3671d9b 100644
--- a/pkg/cli/image_cmd_test.go
+++ b/pkg/cli/image_cmd_test.go
@@ -362,6 +362,8 @@ func TestSignature(t *testing.T) {
                So(err, ShouldBeNil)

                t.Logf("%s", ctlr.Config.Storage.RootDirectory)
+
+               t.Log("Test getting all images using a using graphql query")
                args := []string{"imagetest"}
                configPath := makeConfigFile(fmt.Sprintf(`{"configs":[{"_name":"imagetest","url":"%s","showspinner":false}]}`, url))
                defer os.Remove(configPath)
@@ -376,7 +378,20 @@ func TestSignature(t *testing.T) {
                str := space.ReplaceAllString(buff.String(), " ")
                actual := strings.TrimSpace(str)
                So(actual, ShouldContainSubstring, "IMAGE NAME TAG DIGEST SIGNED SIZE")
-               So(actual, ShouldContainSubstring, "repo7 test:1.0 883fc0c5 true")
+               So(actual, ShouldContainSubstring, "repo7 test:1.0 883fc0c5 true 15B")
+
+               t.Log("Test getting all images using rest calls to get catalog and individual manifests")
+               cmd = MockNewImageCommand(new(searchService))
+               buff = &bytes.Buffer{}
+               cmd.SetOut(buff)
+               cmd.SetErr(buff)
+               cmd.SetArgs(args)
+               err = cmd.Execute()
+               So(err, ShouldBeNil)
+               str = space.ReplaceAllString(buff.String(), " ")
+               actual = strings.TrimSpace(str)
+               So(actual, ShouldContainSubstring, "IMAGE NAME TAG DIGEST SIGNED SIZE")
+               So(actual, ShouldContainSubstring, "repo7 test:1.0 883fc0c5 true 15B")

                err = os.Chdir(currentWorkingDir)
                So(err, ShouldBeNil)
```

PR #954, opened 2022-11-03
fix: signatures now show correctly if signed or not

aokirisaki:
Signed-off-by: Ana-Roberta Lisca <ana.kagome@yahoo.com>

**What type of PR is this?**

#930 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/cli/client.go
+++ b/pkg/cli/client.go
@@ -17,10 +17,15 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
+	notreg "github.com/notaryproject/notation-go/registry"
+	"github.com/sigstore/cosign/pkg/oci/remote"
+
 	zotErrors "zotregistry.io/zot/errors"
+	"zotregistry.io/zot/pkg/api"
 	"zotregistry.io/zot/pkg/storage/local"
 )
 
@@ -297,6 +302,33 @@ func (p *requestsPool) doJob(ctx context.Context, job *manifestJob) {
 		p.outputCh <- stringResult{"", err}
 	}
 
+	isSigned := false
+	cosignTag := strings.Replace(digestStr, ":", "-", 1) + "." + remote.SignatureTagSuffix
+
+	_, err = makeGETRequest(ctx, *job.config.servURL+"/v2/"+job.imageName+
+		"/manifests/"+cosignTag, job.username, job.password,
+		*job.config.verifyTLS, *job.config.debug, &job.manifestResp, job.config.resultWriter)
+	if err == nil {
+		isSigned = true
+	}
+
+	var referrers api.ReferenceList
+
+	if !isSigned {
+		_, err = makeGETRequest(ctx, fmt.Sprintf("%s/oras/artifacts/v1/%s/manifests/%s/referrers?artifactType=%s",
+			*job.config.servURL, job.imageName, digestStr, notreg.ArtifactTypeNotation), job.username, job.password,
+			*job.config.verifyTLS, *job.config.debug, &referrers, job.config.resultWriter)
+		if err == nil {
+			for _, reference := range referrers.References {
+				if reference.ArtifactType == notreg.ArtifactTypeNotation {
+					isSigned = true
+
+					break
+				}
+			}
+		}
+	}
+
 	size += uint64(manifestSize)
 
 	image := &imageStruct{}
@@ -307,6 +339,7 @@ func (p *requestsPool) doJob(ctx context.Context, job *manifestJob) {
 	image.Size = strconv.Itoa(int(size))
 	image.ConfigDigest = configDigest
 	image.Layers = layers
+	image.IsSigned = isSigned
 
 	str, err := image.string(*job.config.outputFormat, len(job.imageName), len(job.tagName))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/cli/client.go
```diff
@@ -290,6 +294,22 @@ func (p *requestsPool) doJob(ctx context.Context, job *manifestJob) {
 		)
 	}
 
+	isSigned := false
+	cosignTag := strings.Replace(digestStr, ":", "-", 1) + "." + remote.SignatureTagSuffix
+	resp, _ := resty.R().Head(*job.config.servURL + "/v2/" + job.imageName + "/manifests/" + cosignTag)
```
peusebiu, line 299: check for error and handle it. same for the notation signature

aokirisaki: I added the error check

peusebiu: I don' t see it yet :D

don't ignore resty errors.

aokirisaki: I forgot to push the update, I'm sorry! It should be fine now.

pkg/cli/client.go
```diff
@@ -17,9 +17,13 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
+	"github.com/sigstore/cosign/pkg/oci/remote"
+	"gopkg.in/resty.v1"
```
rchincha, line 25: Try to avoid using resty or other http libs in core zot code. We want to minimize this.

pkg/cli/image_cmd_test.go
```diff
@@ -376,7 +376,20 @@ func TestSignature(t *testing.T) {
 		str := space.ReplaceAllString(buff.String(), " ")
 		actual := strings.TrimSpace(str)
 		So(actual, ShouldContainSubstring, "IMAGE NAME TAG DIGEST SIGNED SIZE")
-		So(actual, ShouldContainSubstring, "repo7 test:1.0 883fc0c5 true")
+		So(actual, ShouldContainSubstring, "repo7 test:1.0 883fc0c5 true 15B")
```
peusebiu, line 379: why the size was not shown before?

you didn't modify anything on this part

pkg/cli/client.go
```diff
@@ -290,6 +294,30 @@ func (p *requestsPool) doJob(ctx context.Context, job *manifestJob) {
 		)
 	}
 
+	isSigned := false
+	cosignTag := strings.Replace(digestStr, ":", "-", 1) + "." + remote.SignatureTagSuffix
+
+	resp, err := resty.R().Head(*job.config.servURL + "/v2/" + job.imageName + "/manifests/" + cosignTag)
```
andaaron, line 300: Do not use resty in production code please. Use the makeGETRequest method.

pkg/cli/image_cmd_test.go
andaaron: It was shown, but not verified.

pkg/cli/image_cmd_test.go
```diff
@@ -376,7 +376,20 @@ func TestSignature(t *testing.T) {
...
+
+		t.Log("Test getting all images using rest calls to get catalog and individual manifests")
```
andaaron, line 381: What is the difference between this test and the previous one?

pkg/cli/service.go
```diff
@@ -327,8 +327,6 @@ func getImage(ctx context.Context, config searchConfig, username, password, imag
 		hasTagPrefix := strings.HasPrefix(tag, "sha256-")
 		hasTagSuffix := strings.HasSuffix(tag, ".sig")
 
-		// check if it's an image or a signature
```
andaaron, line 330: I would keep the comment

pkg/cli/client.go
```diff
@@ -297,6 +300,25 @@ func (p *requestsPool) doJob(ctx context.Context, job *manifestJob) {
...
+		"/manifests/"+cosignTag, job.username, job.password,
+		*job.config.verifyTLS, *job.config.debug, &job.manifestResp, job.config.resultWriter)
+	if err == nil {
+		isSigned = true
+	}
+
+	if !isSigned {
+		_, err = makeGETRequest(ctx, fmt.Sprintf("%s/oras/artifacts/v1/%s/manifests/%s/referrers",
+			*job.config.servURL, job.imageName, digestStr), job.username, job.password,
+			*job.config.verifyTLS, *job.config.debug, &job.manifestResp, job.config.resultWriter)
```
andaaron, line 316: I think we should check also the type of artefact is referring to this image. To know it's an actual signature.

pkg/cli/image_cmd_test.go
aokirisaki: I modified it as mentioned in the task description

andaaron: Nevermind. I did not notice the `cmd` value change.

aokirisaki:
example output:
```

ana@CSCO-W-PF32FHV9:~/aokirisaki/zot$ bin/zli-linux-amd64 images config-multiple-cve5000 
IMAGE NAME                        TAG                       DIGEST      SIGNED      SIZE      
busybox  latest  3e2f9a5a    false       773kB     
python   latest  b0cda079    true        355MB 

```

rchincha (approved):
lgtm

PR #1010, opened 2022-11-22
fix(oras): relax artifactType check to allow other types

rchincha:

**What type of PR is this?**

oras cli now allows for different artifactTypes. Originally, zot was interested in the notation use case only. No reason to constrain.

**Automation added to e2e**:

```release-note

```


peusebiu (approved):
LGTM

PR #1017, opened 2022-11-24
fix(config): warn if cve is used with remote storage driver

chofnar:
**What type of PR is this?**

config hardening

Checks if remote storage driver is used while CVE is enabled at config verification time, warns that they are incompatible and exits with BadConfig. Also adjusts tests to reflect this.

If running zot with cve enabled and an s3 storage driver, the cve extension will treat the s3 root directory as a local directory leading to unexpected behavior

### DIFF
--- a/examples/config-search.json
+++ b/examples/config-search.json
@@ -12,10 +12,7 @@
     },
     "extensions": {
         "search": {
-          "enable": true,
-          "cve": {
-            "updateInterval": "24h"
-          }
+          "enable": true
         }
     }
 }
--- a/pkg/cli/root.go
+++ b/pkg/cli/root.go
@@ -306,6 +306,26 @@ func validateCacheConfig(cfg *config.Config) error {
 	return nil
 }
 
+func validateExtensionsConfig(cfg *config.Config) error {
+	//nolint:lll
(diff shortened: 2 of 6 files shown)


rchincha:
But what happens if I actually make a CVE query request?

chofnar:
> But what happens if I actually make a CVE query request?

I tried with CVEListForImage in the GQL Playground after I added some images with skopeo and it seems to be working

pkg/cli/root.go
```diff
@@ -306,6 +306,24 @@ func validateCacheConfig(cfg *config.Config) error {
 	return nil
 }
 
+func validateExtensionsConfig(cfg *config.Config) {
+	//nolint:lll
+	if cfg.Storage.StorageDriver != nil && cfg.Extensions != nil && cfg.Extensions.Search != nil && cfg.Extensions.Search.CVE != nil {
```
peusebiu, line 311: do not modify the internal config :D, will create issues with the configloader.

just return an error, look at sync+s3 combination which also panics

chofnar: Added

rchincha (dismissed):
lgtm

examples/config-search.json
```diff
@@ -12,10 +12,7 @@
     },
     "extensions": {
         "search": {
-          "enable": true,
```
rchincha, line 15: This is not correct. Existing configurations should work as is. Add a new example if need be called config-remote-only.json

Issue #908, opened 2022-10-22
Queries (rest and graphql) containing only repo name, without a reference (tag or digest) should return "bad request" or equivalent


andaaron:
Return "bad request" or equivalent.

PR #1040, opened 2022-12-02
fix: queries with images without a reference should return an error

alexstan12:
Currently there is no push-back on queries that should contain image names but have only the repo name. This commit adds a check that will return an error for images w/o a reference(tag or digest).

**What type of PR is this?**
feature

closes https://github.com/project-zot/zot/issues/908

**Automation added to e2e**:

```release-note

```


andaaron (approved):
Looks good.

rchincha (approved):
lgtm

PR #1041, opened 2022-12-02
refactor: Cleanup/simplify test cases

nicoldr:

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -196,36 +196,14 @@ func TestCreateCacheDatabaseDriver(t *testing.T) {
 func TestRunAlreadyRunningServer(t *testing.T) {
 	Convey("Run server on unavailable port", t, func() {
 		port := test.GetFreePort()
-		baseURL := test.GetBaseURL(port)
 		conf := config.New()
 		conf.HTTP.Port = port
 
-		ctlr := api.NewController(conf)
+		ctlr := makeController(conf, t.TempDir(), "")
+		cm := test.NewControllerManager(ctlr)
 
-		globalDir := t.TempDir()
-
-		ctlr.Config.Storage.RootDirectory = globalDir
-
-		go func() {
-			if err := ctlr.Run(context.Background()); err != nil {
-				return
-			}
-		}()
-
-		// wait till ready
-		for {
-			_, err := resty.R().Get(baseURL)
-			if err == nil {
-				break
-			}
-
-			time.Sleep(100 * time.Millisecond)
-		}
-
-		defer func() {
-			ctx := context.Background()
-			_ = ctlr.Server.Shutdown(ctx)
-		}()
+		cm.StartAndWait(port)
+		defer cm.StopServer()
 
 		err := ctlr.Run(context.Background())
 		So(err, ShouldNotBeNil)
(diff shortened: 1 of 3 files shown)


pkg/api/controller_test.go
```diff
@@ -200,32 +200,11 @@ func TestRunAlreadyRunningServer(t *testing.T) {
...
-		// wait till ready
-		for {
-			_, err := resty.R().Get(baseURL)
-			if err == nil {
-				break
-			}
-
-			time.Sleep(100 * time.Millisecond)
-		}
+		ctlr := makeController(conf, t.TempDir(), "")
```
rchincha, line 203: test.NewController(...)

In golang world, we New()... everything

pkg/api/controller_test.go
```diff
@@ -200,32 +200,11 @@ func TestRunAlreadyRunningServer(t *testing.T) {
...
 
-		defer func() {
-			ctx := context.Background()
-			_ = ctlr.Server.Shutdown(ctx)
-		}()
+		go startServer(ctlr)
```
rchincha, line 205:
```
ctlr := test.NewController(...)
go ctlr.Start(wait=true/false)
defer ctlr.Stop()
```

pkg/api/controller_test.go
nicoldr: Thank you for the review, Ram! My first thought for implementing this new controller function was to create it in the test/common.go and use it, but for that, we would need to import conf and api in common.go, and that would cause a cyclic import error since test is imported in api and so on. That's why I implemented it here, to save some lines in controller_test.go. Or are you just suggesting renaming it?

pkg/api/controller_test.go
nicoldr: Made these changes, but now I see the part where we specify the wait=true/false. Will also add this

nicoldr: Now when wait would be true, we call go ctlr.Start(baseUrl), or ctlr.Start("") when wait would be false, since url is needed

rchincha: ctlr.Start(wait=true/false)?

pkg/test/common.go
```diff
@@ -167,6 +167,43 @@ func CopyFiles(sourceDir, destDir string) error {
 	return nil
 }
 
+type Controller interface {
+	Run(ctx context.Context) error
+	Shutdown()
+	GetPort() int
+}
+
+type ControllerStruct struct {
```
andaaron, line 176:
```suggestion
type TestController struct {
```

pkg/test/common.go
```diff
@@ -167,6 +167,43 @@ func CopyFiles(sourceDir, destDir string) error {
...
+	c Controller
+}
+
+func (c *ControllerStruct) StartServer() {
+	// this blocks
```
andaaron, line 181: This comment is no longer relevant?

pkg/test/common.go
```diff
@@ -167,6 +167,43 @@ func CopyFiles(sourceDir, destDir string) error {
...
+
+func NewControllerTest(c Controller) ControllerStruct {
+	ctrl := ControllerStruct{
+		c: c,
+	}
+
+	return ctrl
+}
+
+func (c *ControllerStruct) GetPort() int {
```
andaaron, line 203: Add a method for waiting instead of this one. That method would call c.c.GetPort() and pass the result to test.GetBaseURL(port). The url would be later passed to test.WaitTillServerReady(baseURL) in the same method.

rchincha: Make sure you use https://github.com/phayes/freeport for getting free ports, else port collisions and failed tests. Just fyi.

pkg/test/common.go
nicoldr: I added the comment back in controller_test, just before cm.StartServer()

pkg/api/routes_test.go
```diff
@@ -46,9 +46,10 @@ func TestRoutes(t *testing.T) {
 			panic(err)
 		}
 
-		go startServer(ctlr)
-		defer stopServer(ctlr)
-		test.WaitTillServerReady(baseURL)
+		cm := test.NewControllerManager(ctlr)
+		cm.StartServer()
```
rchincha, line 50: StartAndWait(port) <- one call? this is usually the call everyone makes

rchincha: Add this as another API btw, and call that instead here, we may need a separate StartServer() call elsewhere

rchincha (approved):
lgtm

PR #1065, opened 2022-12-14
fix(sync): fix sync on demand with docker library

peusebiu:

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/sync/utils.go
+++ b/pkg/extensions/sync/utils.go
@@ -18,6 +18,7 @@ import (
 	glob "github.com/bmatcuk/doublestar/v4"
 	"github.com/containers/image/v5/docker"
 	"github.com/containers/image/v5/docker/reference"
+	"github.com/containers/image/v5/manifest"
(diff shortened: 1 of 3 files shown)


pkg/extensions/sync/utils.go
```diff
@@ -678,7 +679,8 @@ func descriptorEqual(desc1, desc2 ispec.Descriptor) bool {
 
 func isSupportedMediaType(mediaType string) bool {
 	return mediaType == ispec.MediaTypeImageIndex ||
-		mediaType == ispec.MediaTypeImageManifest
+		mediaType == ispec.MediaTypeImageManifest ||
+		mediaType == manifest.DockerV2ListMediaType
```
rchincha, line 683: wait, what? docker.io supports OCI on the wire right?

peusebiu: yes, but here is the upstream mediatype and in the case of docker.io is not OCI

rchincha (approved):
lgtm

PR #1068, opened 2022-12-15
chore(trivy): update trivy version and enforce OCI compliant repo names in local image storage

andaaron:
Originally 4 commits, squashed so the can be signed after rebasing.

1. chore(trivy): update trivy library version

The trivy team switched github.com/urfave/cli for viper so
there are some other code changes as well.

Since we don't use github.com/urfave/cli directly in our software
we needed to add a tools.go in order for "go mod tidy" to not delete it.
See this pattern explained in:
- https://github.com/99designs/gqlgen#quick-start
- https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
- https://github.com/go-modules-by-example/index/blob/master/010_tools/README.md#walk-through

The jobs using "go get -u" have been updated to use "go install", since go get
modifies the go.mod by upgrading some of the packages, but downgrading trivy to an older
version with broken dependencies

2. fix(storage) Update local storage to ignore folder names not compliant with dist spec
Also updated trivy to download the DB and cache results under the rootDir/_trivy folder

3. fix(s3): one of the s3 tests was missing the skipIt call
This caused a failure when running locally without s3 being available

4. make sure the offline scanning is enabled, and zot only downloads the trivy DB
on the regular schedule, and doesn't download the DB on every image scan


shimish2 (commented):
@rchincha 

With this change we are not using our forks (trivy, trivy-db and fanal) and now we can configure our cache dir accordingly.

rchincha:
@shimish2 can you review this pls?

pkg/storage/local/local.go
```diff
@@ -154,6 +155,12 @@ func (is *ImageStoreLocal) initRepo(name string) error {
 		return zerr.ErrInvalidRepositoryName
 	}
 
+	if !zreg.FullNameRegexp.MatchString(name) {
+		is.log.Error().Msg("repository name is not compliant with the spec")
```
rchincha, line 159: Add "name" to the Str("repo", name)

Msg("Invalid repo name")

pkg/storage/local/local_test.go
```diff
@@ -2483,10 +2489,49 @@ func TestValidateRepo(t *testing.T) {
...
+		metrics := monitoring.NewMetricsServer(false, log)
+		cacheDriver, _ := storage.Create("boltdb", cache.BoltDBDriverParameters{
+			RootDir:     dir,
+			Name:        "cache",
+			UseRelPaths: true,
+		}, log)
+		imgStore := local.NewImageStore(dir, true, storage.DefaultGCDelay,
+			true, true, log, metrics, nil, cacheDriver)
+
+		_, err := imgStore.ValidateRepo(".")
```
rchincha, line 2506: Also add a ".." case here and below.

pkg/storage/s3/s3_test.go
```diff
@@ -446,6 +446,8 @@ func TestStorageDriverStatFunction(t *testing.T) {
 }
 
 func TestGetOrasAndOCIReferrers(t *testing.T) {
```
rchincha, line 448: Add tests under s3 same as local, just to be sure?

rchincha (approved):
lgtm

PR #1101, opened 2023-01-10
feat(cve): cache trivy results for an image:tag

andaaron:
**What type of PR is this?**
feature

Right now the trivy scanner logic is triggered every time a call is made to scan an image.

By caching the results, we can avoid some of the repetitive calls made for the same image (specifically in the UI use case).
We can also add pagination on the results if we already have them cached (which is a separate commit which will be ported to main branch after this one is merged).

### DIFF
--- a/pkg/extensions/search/cve/trivy/cache.go
+++ b/pkg/extensions/search/cve/trivy/cache.go
@@ -1,0 +1,41 @@
+package trivy
+
+import (
+	lru "github.com/hashicorp/golang-lru"
+
+	cvemodel "zotregistry.io/zot/pkg/extensions/search/cve/model"
+	"zotregistry.io/zot/pkg/log"
+)
+
+type CveCache struct {
+	cache *lru.Cache
+	log   log.Logger
+}
+
+func NewCveCache(size int, log log.Logger) *CveCache {
+	cache, _ := lru.New(size)
+
+	return &CveCache{cache: cache, log: log}
+}
+
+func (cveCache *CveCache) Add(image string, cveMap map[string]cvemodel.CVE) {
+	cveCache.cache.Add(image, cveMap)
+}
+
+func (cveCache *CveCache) Get(image string) map[string]cvemodel.CVE {
+	value, ok := cveCache.cache.Get(image)
+	if !ok {
+		return nil
+	}
+
+	cveMap, ok := value.(map[string]cvemodel.CVE)
+	if !ok {
+		return nil
+	}
+
+	return cveMap
+}
+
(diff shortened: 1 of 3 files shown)


rchincha:
@shimish2 doesn't trivy already maintain its own cache for this use case?

shimish2:
> @shimish2 doesn't trivy already maintain its own cache for this use case?

Yes it does, now with this change we are configuring where we can store this cache i.e. (`root-dir+_trivy`).

andaaron:
@shimish2 You are probably thinking of the other PR open as we speak.

@rchincha Theoretically trivy does this but
1. It is slow, and it's causing long delays for the UI.
2. If we cache the results, then we can actually to pagination of the CVE in that cache entry (which will be a separate PR, since it depends on some other commits on the integration branch I haven't ported yet to main).

shimish2:
> @shimish2 You are probably thinking of the other PR open as we speak.

Yes, I just edited my reply.

andaaron:
@shimish2 the UI made a call to request all CVEs for a given image, and after that for each CVE run the scan to get the tags where it was fixed. It was basically hundreds of image scans in a single UI screen.

pkg/extensions/search/cve/trivy/cache.go
```diff
@@ -0,0 +1,41 @@
+package trivy
+
+import (
+	lru "github.com/hashicorp/golang-lru"
+
+	cvemodel "zotregistry.io/zot/pkg/extensions/search/cve/model"
+	"zotregistry.io/zot/pkg/log"
+)
+
+type CveCache struct {
```
rchincha, line 10: Does this need locks?

pkg/extensions/search/cve/trivy/cache.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func NewCveCache(size int, log log.Logger) *CveCache {
+	cache, _ := lru.New(size)
+
+	return &CveCache{cache: cache, log: log}
+}
+
+func (cveCache *CveCache) Add(image string, cveMap map[string]cvemodel.CVE) {
+	cveCache.cache.Add(image, cveMap)
```
rchincha, line 22: Writelock here

pkg/extensions/search/cve/trivy/cache.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func (cveCache *CveCache) Get(image string) map[string]cvemodel.CVE {
```
rchincha, line 25: Readlocks here

rchincha (approved):
lgtm

PR #1110, opened 2023-01-13
fix(ci): do not build zot container image for the annotation tests

andaaron:

**What type of PR is this?**
cleanup
ci

Building the zot container image just to use as a test image to copy over takes too long, uses too many resources, and is not necessary.

Build a simpler image for the purpose of testing annotations instead.


rchincha (approved):
lgtm

PR #1118, opened 2023-01-18
feat(cve): the cve related calls to use repodb and add pagination on image results

andaaron:


pkg/extensions/search/cve/trivy/scanner.go
```diff
@@ -221,7 +221,7 @@ func (scanner Scanner) IsImageFormatScannable(image string) (bool, error) {
 		}
 	}
 
-	return false, nil
+	return true, nil
```
rchincha, line 224: Behavior in this func is changing, in what way?

rchincha (approved):

pkg/extensions/search/cve/trivy/scanner.go
andaaron: Because in the previous version only the first layer was checked for compatibility. Both cases were returning from the function on the first iteration of the for loop.

andaaron: Previous version of line 224 wasn't even reached, and I noticed while checking coverage.

Issue #983, opened 2022-11-09
[Feat]: Add a group mechanism for authZ


voor:
### Is your feature request related to a problem? Please describe.

Providing unique permissions to individual users works at a small scale when you're managing dozens of users, but once you're dealing with larger amounts of users you need a group mechanism that can automatically apply users to different groups and assign permissions to groups rather than individuals.  Additionally, this eliminates the need for shared accounts like an `admin` account or similar, which is typically a security violation.

### Describe the solution you'd like

As an authenticated user I am already applied to the `Default` group, which is the failover for when the individual does not have a user based policy.  Add another option for `groups` inside the policies type that can apply policies to entire groups.

```json
"http": {
...
  "accessControl": {
    "**": {
      "policies": [{
        "users": ["charlie"],
        "groups": ["admins","developers"],
        "actions": ["read", "create", "update"]
      }],
      "defaultPolicy": ["read", "create"]
    },
    "tmp/**": {
      "defaultPolicy": ["read", "create", "update"]
    },
    "infra/*": {
      "policies": [{
          "users": ["alice", "bob"],
          "groups": ["maintainers","platformteam"],
          "actions": ["create", "read", "update", "delete"]
        },
        {
          "users": ["mallory"],
          "actions": ["create", "read"]
        }
      ],
      "defaultPolicy": ["read"]
    },
    "repos2/repo": {
      "policies": [{
          "users": ["bob"],
          "groups": ["sparkle_team","repo2_team"],
          "actions": ["read", "create"]
        },
        {
          "users": ["mallory"],
          "actions": ["create", "read"]
        }
      ],
      "defaultPolicy": ["read"]
    },
    "adminPolicy": {
      "users": ["admin"],
      "groups": ["admins","developers"],
      "actions": ["read", "create", "update", "delete"]
    }
  }
```

### Additional context

Coordinating this with https://github.com/project-zot/zot/issues/978 would provide an easy mechanism to discovery user-group pairings without needing to create an internal configuration to match a user to a group.

alexstan12:
Hi @voor ! In the solution you've described above, where would the **group** value come from? Is it from the ID token retrieved from OIDC flow?
If so, then 'group' isn't really a default claim on this token and would require a modification on the provider side ( e.g. Github, Gitlab, Google ) which I don't think it's possible, since we're not implementing the provider side of the OIDC flow

voor:
[...]

PR #1123, opened 2023-01-19
feat(authz): added groups mechanism

aokirisaki:
**What type of PR is this?**

#983 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/examples/config-anonymous-authz.json
+++ b/examples/config-anonymous-authz.json
@@ -8,29 +8,31 @@
     "port": "8080",
     "realm": "zot",
     "accessControl": {
-      "**": {
-        "anonymousPolicy": [
-          "read",
-          "create"
-        ]
-      },
-      "tmp/**": {
-        "anonymousPolicy": [
-          "read",
-          "create",
-          "update"
-        ]
-      },
-      "infra/**": {
-        "anonymousPolicy": [
-          "read"
-        ]
-      },
-      "repos2/repo": {
-        "anonymousPolicy": [
-          "read"
-        ]
-      }
+      "repositories": {
+        "**": {
+          "anonymousPolicy": [
+            "read",
+            "create"
+          ]
+        },
+        "tmp/**": {
+          "anonymousPolicy": [
+            "read",
+            "create",
+            "update"
+          ]
+        },
+        "infra/**": {
+          "anonymousPolicy": [
+            "read"
+          ]
+        },
+        "repos2/repo": {
+          "anonymousPolicy": [
+            "read"
+          ]
+        }
+      } 
     }
   },
   "log": {
--- a/examples/config-ldap.json
+++ b/examples/config-ldap.json
@@ -1,0 +1,73 @@
+{
+    "distSpecVersion": "1.1.0-dev",
+    "storage": {
+      "rootDirectory": "/tmp/zot"
+    },
+    "http": {
+      "address": "127.0.0.1",
+      "port": "8080",
+      "realm": "zot",
+      "tls": {
+        "cert": "test/data/server.cert",
+        "key": "test/data/server.key"
+      },
+      "auth": {
+        "ldap": {
+          "address": "ldap.example.org",
+          "port": 389,
+          "startTLS": false,
+          "baseDN":"ou=Users,dc=example,dc=org",
+          "userAttribute": "uid",
+          "userGroupAttribute": "memberOf",
+          "bindDN":"cn=ldap-searcher,ou=Users,dc=example,dc=org",
+          "bindPassword":"ldap-searcher-password",
+          "skipVerify": true,
+          "subtreeSearch": true
+        },
+        "failDelay": 5
+      },
+      "accessControl": {
+        "repositories": {
+          "**": {
+            "policies": [{
+              "users": ["charlie"],
+              "groups": ["admins", "developers", "cn=ldap-group,ou=Groups,dc=example,dc=org"],
+              "actions": ["read", "create", "update"]
+            },
+            {
+              "users": ["mary"],
+              "groups": ["group2"],
+              "actions": ["read", "create", "update", "delete"]
+            }],
+            "defaultPolicy": []
+          },
+          "tmp/**": {
+            "defaultPolicy": ["read", "create", "update"]
+          },
+          "repos2/repo": {
+            "policies": [{
+                "users": ["bob"],
+                "groups": ["sparkle_team","repo2_team"],
+                "actions": ["read", "create"]
+              },
+              {
+                "users": ["mallory"],
+                "actions": ["create", "read"]
+              }
+            ],
+            "defaultPolicy": ["read"]
+          }  
+        },
+        "adminPolicy": {
+          "users": ["admin"],
+          "groups": ["admins","developers"],
+          "actions": ["read", "create", "update", "delete"]
+        }
+      }
+    },
+    "log": {
+      "level": "debug",
+      "output": "/tmp/zot.log",
+      "audit": "/tmp/zot-audit.log"
+    }
+}
--- a/examples/config-policy.json
+++ b/examples/config-policy.json
@@ -4,107 +4,68 @@
     "rootDirectory": "/tmp/zot"
   },
   "http": {
-    "address": "127.0.0.1",
-    "port": "8080",
-    "realm": "zot",
     "auth": {
       "htpasswd": {
         "path": "test/data/htpasswd"
-      },
-      "failDelay": 1
+      }
     },
     "accessControl": {
-      "**": {
-        "anonymousPolicy": ["read"],
-        "policies": [
-          {
-            "users": [
-              "charlie"
-            ],
-            "actions": [
-              "read",
-              "create",
-              "update"
-            ]
-          }
-        ],
-        "defaultPolicy": [
-          "read",
-          "create",
-          "delete",
-          "detectManifestCollision"
-        ]
+      "groups": {
+        "group1": {
+          "users": ["jack", "john", "jane", "ana"]
+        },
+        "group2": {
+          "users": ["alice", "mike", "jim"]
+        }
       },
-      "tmp/**": {
-        "defaultPolicy": [
-          "read",
-          "create",
-          "update"
-        ]
-      },
-      "infra/**": {
-        "policies": [
-          {
-            "users": [
-              "alice",
-              "bob"
-            ],
-            "actions": [
-              "create",
-              "read",
-              "update",
-              "delete"
-            ]
+      "repositories": {
+        "**": {
+          "policies": [{
+            "users": ["charlie"],
+            "groups": ["admins", "developers", "group1"],
+            "actions": ["read", "create", "update"]
           },
           {
-            "users": [
-              "mallory"
-            ],
-            "actions": [
-              "create",
-              "read"
-            ]
-          }
-        ],
-        "defaultPolicy": [
-          "read"
-        ]
-      },
-      "repos2/repo": {
-        "policies": [
-          {
-            "users": [
-              "charlie"
-            ],
-            "actions": [
-              "read",
-              "create"
-            ]
-          },
-          {
-            "users": [
-              "mallory"
-            ],
-            "actions": [
-              "create",
-              "read"
-            ]
-          }
-        ],
-        "defaultPolicy": [
-          "read"
-        ]
+            "users": ["mary"],
+            "groups": ["group2"],
+            "actions": ["read", "create", "update", "delete"]
+          }],
+          "defaultPolicy": ["read", "create"]
+        },
+        "tmp/**": {
+          "defaultPolicy": ["read", "create", "update"]
+        },
+        "infra/*": {
+          "policies": [{
+              "users": ["alice", "bob"],
+              "groups": ["maintainers","platformteam"],
+              "actions": ["create", "read", "update", "delete"]
+            },
+            {
+              "users": ["mallory"],
+              "actions": ["create", "read"]
+            }
+          ],
+          "defaultPolicy": ["read"]
+        },
+        "repos2/repo": {
+          "policies": [{
+              "users": ["bob"],
+              "groups": ["sparkle_team","repo2_team"],
+              "actions": ["read", "create"]
+            },
+            {
+              "users": ["mallory"],
+              "actions": ["create", "read"]
+            }
+          ],
+          "defaultPolicy": ["read"]
+        }  
       },
       "adminPolicy": {
-        "users": [
-          "admin"
-        ],
-        "actions": [
-          "read",
-          "create",
-          "update",
-          "delete"
-        ]
+        "users": ["admin"],
+        "groups": ["admins","developers"],
+        "actions": ["read", "create", "update", "delete"]
       }
     }
   },
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"bufio"
+	"context"
 	"crypto/x509"
 	"encoding/base64"
 	"fmt"
@@ -17,6 +18,7 @@ import (
 
 	"zotregistry.io/zot/errors"
 	"zotregistry.io/zot/pkg/api/config"
+	localCtx "zotregistry.io/zot/pkg/requestcontext"
 )
 
 const (
@@ -87,7 +89,8 @@ func noPasswdAuth(realm string, config *config.Config) mux.MiddlewareFunc {
 			}
 
 			// Process request
-			next.ServeHTTP(response, request)
+			ctx := getReqContextWithAuthorization("", []string{}, request)
+			next.ServeHTTP(response, request.WithContext(ctx)) //nolint:contextcheck
 		})
 	}
 }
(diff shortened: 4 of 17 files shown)


peusebiu:
to summarize: group policies have priority over user policies which have priority over default policy.

can you add a test please in controller_test?

examples/config-policy.json
```diff
@@ -4,107 +4,63 @@
...
-          }
-        ],
-        "defaultPolicy": [
-          "read"
-        ]
+      "repos": {
+        "**": {
+          "policies": [{
+            "users": ["charlie"],
+            "groups": ["admins","developers"],
```
andaaron, line 25: The groups used in the `policies` are not defined under `groups`. Let's give an example of a valid configuration.
Is there a specific reason for them not to match?

pkg/api/authz.go
```diff
@@ -133,9 +143,37 @@ func (ac *AccessController) getContext(username string, request *http.Request) c
...
+			}
+		}
+	}
+
+	return ""
+}
+
 // isPermitted returns true if username can do action on a repository policy.
-func isPermitted(username, action string, policyGroup config.PolicyGroup) bool {
+func isPermitted(ac *AccessController, username, action string, policyGroup config.PolicyGroup) bool {
```
andaaron, line 160: Any specific reason why this function is not a member of the AccessController struct? Since it needs an AccessController parameter anyway.
Same goes for `getUserGroup`

pkg/api/authz.go
```diff
@@ -246,6 +284,7 @@ func AuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
 				action = Delete
 			}
 
+			// acCtrlr.Config.Groups
```
andaaron, line 287: What is this comment for?

pkg/api/authz.go
```diff
@@ -133,9 +143,37 @@ func (ac *AccessController) getContext(username string, request *http.Request) c
 	return ctx
 }
 
+func getUserGroup(ac *AccessController, username string) string {
```
andaaron, line 146: I think there's also the case where is user is part of multiple groups, and the permissions cumulate.

pkg/api/controller_test.go
```diff
@@ -6704,6 +6704,238 @@ func TestSearchRoutes(t *testing.T) {
 			So(string(resp.Body()), ShouldNotContainSubstring, repoName)
 			So(string(resp.Body()), ShouldNotContainSubstring, inaccessibleRepo)
 		})
+
+		Convey("Testing group permissions", func(c C) {
```
andaaron, line 6708: There should also be tests for the user being part of multiple groups

examples/config-policy.json
Skiepp: @aokirisaki are the group membershipt taken from ldap?
Or is it hardcoded in the config file?

Skiepp:
@aokirisaki please prove me wrong but right now in your config **groups are hardcoded** into the config file and not taken from LDAP
Those kind of approch **should be avoided** like the plague, as it's not scalable neither it's correct as LDAP should be the only source of thruth for group membership

This is not solving the issue #983 which clearly states that group should be read by the userinfo endpoint.
Downvoting this

andaaron:
Some notes:
1. We need a mechanism to be able to use groups also for local users, not just ldap, so even if the PR does not resolve #983, some of the group related code is the same.
2. For LDAP there will be extra settings needed, but I'm not sure how the config would look (for example attribute of the LDAP group to match against policies, basically the group name).

rchincha:
wrt LDAP (served by Windows AD backend at least),
"sAMAccountName" and "memberOf" return the two pieces of info relevant here.
Pls. enhance LDAP config to also include GroupAttribute to allow for @Skiepp's ask.

Skiepp:
Thank you @rchincha, but as @andaaron suggested the attribute should be part of config.

Your proposal is perfect for us (we are using AD), but many other people use other services like OpenLDAP and samba, which uses different attributes.

Restrict to those filelds only would be limitative.
Thanks to all for the interest

rchincha:
@Skiepp understood, these attributes are not hard-coded, and you are allowed to configure what the attributes are called and zot would just query of those attributes. If there is sample backend config that you can share, we will be happy to include it in our CI/CD tests.

pkg/requestcontext/context.go
```diff
@@ -63,10 +64,23 @@ func (acCtx *AccessControlContext) matchesRepo(globPatterns map[string]bool, rep
...
 func (acCtx *AccessControlContext) CanDetectManifestCollision(repository string) bool {
-	return acCtx.matchesRepo(acCtx.DmcGlobPatterns, repository)
+	if acCtx.DmcGlobPatterns != nil {
+		return acCtx.matchesRepo(acCtx.DmcGlobPatterns, repository)
+	}
+
+	return false
+}
+
+//nolint:unused
```
andaaron, line 83: Why is this unused? Shall we delete it in this case?

pkg/api/authn.go
```diff
@@ -123,6 +126,7 @@ func basicAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 				SkipTLS:            !ldapConfig.StartTLS,
 				Base:               ldapConfig.BaseDN,
 				BindDN:             ldapConfig.BindDN,
+				GroupAttribute:     "memberOf", // from config
```
andaaron, line 129: This should be configurable.

pkg/api/authn.go
```diff
@@ -210,18 +217,20 @@ func basicAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 			if ok {
 				if err := bcrypt.CompareHashAndPassword([]byte(passphraseHash), []byte(passphrase)); err == nil {
 					// Process request
-					next.ServeHTTP(response, request)
+					ctx := getReqContextWithAuthorization(username, []string{}, request)
```
andaaron, line 220: We should also set the local groups (for HTTPPassword users) from the config file here, right?

pkg/api/authz.go
```diff
@@ -110,32 +130,78 @@ func (ac *AccessController) isAdmin(username string) bool {
...
+	userGroups := ac.getUserGroups(groups, username)
+
+	if userGroups != nil {
+		for _, p := range policyGroup.Policies {
+			if common.Contains(p.Actions, action) {
+				for _, group := range p.Groups {
+					if common.Contains(userGroups, group) {
+						result = true
+
+						return result
```
andaaron, line 198: I think we should first check the user permissions, and if those are missing, fallback to the group permissions, and if those are missing, fallback to the default, and anonymous at the end.

And we should make sure we test this order.

pkg/api/authz.go
```diff
@@ -89,14 +98,25 @@ func (ac *AccessController) can(username, action, repository string) bool {
...
 	// check matched repo based policy
 	pg, ok := ac.Config.Repositories[longestMatchedPattern]
 	if ok {
-		can = isPermitted(username, action, pg)
+		can = ac.isPermitted(groups, username, action, pg)
 	}
 
 	// check admins based policy
 	if !can {
+		if ac.isAnyGroupInAdminPolicy(groups, username) && common.Contains(ac.Config.AdminPolicy.Actions, action) {
```
andaaron, line 116: First check username, and afterwards the group.

pkg/api/authz.go
```diff
@@ -110,32 +130,78 @@ func (ac *AccessController) isAdmin(username string) bool {
...
+}
+
 // isPermitted returns true if username can do action on a repository policy.
-func isPermitted(username, action string, policyGroup config.PolicyGroup) bool {
+func (ac *AccessController) isPermitted(groups []string, username, action string,
+	policyGroup config.PolicyGroup,
+) bool {
 	var result bool
+
+	userGroups := ac.getUserGroups(groups, username)
```
andaaron, line 189: This shouldn't be here, the local groups should be set on the context by the the authentication handler as well.

pkg/api/authz.go
```diff
@@ -110,32 +130,78 @@ func (ac *AccessController) isAdmin(username string) bool {
...
 	}
 
 	authzCtxKey := localCtx.GetContextKey()
-	ctx := context.WithValue(request.Context(), authzCtxKey, acCtx)
+	ctx := context.WithValue(request.Context(), authzCtxKey, *acCtx)
 
 	return ctx
 }
 
+func (ac *AccessController) getUserGroups(groupNames []string, username string) []string {
```
andaaron, line 170: This adds additional groups after the ones already passed as parameter. If we're going to do it this way, we should determine this full list in the authn middleware, and pass it to the context.

pkg/api/ldap.go
```diff
@@ -183,26 +186,28 @@ func (lc *LDAPClient) Authenticate(username, password string) (bool, map[string]
...
 		lc.Log.Error().Err(err).Str("bindDN", lc.BindDN).Str("username", username).
 			Str("baseDN", lc.Base).Msg("too many entries")
 
-		return false, nil, err
+		return false, nil, nil, err
 	}
 
 	userDN := search.Entries[0].DN
+	userAttributes := search.Entries[0].Attributes[0]
+	userGroups := userAttributes.Values
```
andaaron, line 210: There will be the FQDNs of the groups, right? Is there a way to map this to the user groups in the zot config? Or the user needs to use the FQDN value in all the authz policies, without needing additional mapping to a more friendly name in the config?

pkg/api/routes.go
```diff
@@ -1531,7 +1531,7 @@ func (rh *RouteHandler) ListRepositories(response http.ResponseWriter, request *
 		return
 	}
 
-	if acCtx != nil {
+	if acCtx != nil && len(combineRepoList) != 0 {
```
andaaron, line 1534: Why do we need this change? Theoretically we should be able to iterate over an empty slice.

peusebiu: because a sync test was failing, but I guess we can modify the test.

the test expected []string{} but now it gets []string(nil) because now acCtx will not be nil, being set in authn.

let's make sure we don't return {"repositories": [nil]} in /v2/_catalog response when there is no repos.

andaaron: Why would we return `[nil]`? I guess that's the bug.

pkg/api/authn.go
aokirisaki: Only the ldap groups are kept on the context, the local groups are still kept as before. The getUserGroups function, which is in authz.go, does take both into consideration.

andaaron: Why not add all groups in the context?

aokirisaki: I have added them

examples/config-policy.json
```diff
@@ -4,107 +4,68 @@
...
-        ],
-        "defaultPolicy": [
-          "read",
-          "create",
-          "delete",
-          "detectManifestCollision"
-        ]
+      "groups": {
+        "group1": {
+          "users": ["jack", "john", "jane", "ana"]
```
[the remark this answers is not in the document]
aokirisaki: I added an example config

pkg/api/authn.go
aokirisaki: Now it is configurable

pkg/api/authz.go
aokirisaki: I changed the order of the checks so that the username's permissions are checked first, then for the groups

pkg/api/authz.go
aokirisaki: Now all the groups are passed to the context in authn.go

pkg/api/routes.go
aokirisaki: Before setting acCtx in authn.go, it entered the else branch, where the repo list was set on nil, now because acCtx isn't nil anymore here, it would enter the if branch and remain a list

pkg/api/authz.go
aokirisaki: I moved the logic in authn.go

pkg/api/authz.go
aokirisaki: I modified the order and added tests which reflect the checks order

pkg/requestcontext/context.go
aokirisaki: I removed the unused code

pkg/api/routes.go
andaaron: We need to clarify this in a call.

examples/config-ldap.json
```diff
@@ -0,0 +1,75 @@
...
+          "subtreeSearch": true
+        },
+        "failDelay": 5
+      },
+      "accessControl": {
+        "repos": {
+          "**": {
+            "policies": [{
+              "users": ["charlie"],
+              "groups": ["admins", "developers", "ldapgroupexample"],
```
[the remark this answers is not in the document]
aokirisaki: I have made the changes

pkg/api/authn.go
```diff
@@ -232,6 +257,33 @@ func basicAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 	}
 }
 
+func (ac *AccessController) getUserGroups(username string) []string {
```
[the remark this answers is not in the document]
aokirisaki: I have moved it to authz.go

andaaron (dismissed):
Looks good.

peusebiu (approved):
lgtm
just a note that groups feature will not work for clients who authenticates through tls certificates.
we can fix it in a new PR.

examples/config-anonymous-authz.json
```diff
@@ -8,29 +8,31 @@
...
-        "anonymousPolicy": [
-          "read"
-        ]
-      },
-      "repos2/repo": {
-        "anonymousPolicy": [
-          "read"
-        ]
-      }
+      "repos": {
```
rchincha, line 11: This is a breaking change actually!

rchincha: In the commit description/message, follow this approach

BREAKING CHANGE: a commit that has the text BREAKING CHANGE: at the beginning of its optional body or footer section introduces a breaking API change (correlating with [MAJOR](http://semver.org/#summary) in semantic versioning). A breaking change can be part of commits of any type. e.g., a fix:, feat: & chore: types would all be valid, in addition to any other type.

examples/config-ldap.json
```diff
@@ -0,0 +1,75 @@
...
+        "key": "test/data/server.key"
+      },
+      "auth": {
+        "ldap": {
+          "address": "ldap.example.org",
+          "port": 389,
+          "startTLS": false,
+          "baseDN":"ou=Users,dc=example,dc=org",
+          "userAttribute": "uid",
+          "groupMembershipAttribute": "memberOf",
```
andaaron, line 21: Please update the example to use the new key.

rchincha:
In the commit msg, also describe the authn/authz behavior including ldap. This will likely be a long commit description.

andaaron (approved):

examples/config-ldap.json
```diff
@@ -0,0 +1,75 @@
...
+          "userGroupAttribute": "memberOf",
+          "bindDN":"cn=ldap-searcher,ou=Users,dc=example,dc=org",
+          "bindPassword":"ldap-searcher-password",
+          "skipVerify": true,
+          "subtreeSearch": true
+        },
+        "failDelay": 5
+      },
+      "accessControl": {
+        "repos": {
```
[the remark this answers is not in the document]
aokirisaki: I have a question about repoPaths, what would its structure be? Would it have a key(the name of the repo) and value(the path)? And how would we identify then which groups/users/policies are for each repo? And what would be the advantages of  this change?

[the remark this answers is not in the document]
rchincha: RepoPaths []map[string]RepoPath

RepoPath {
...
}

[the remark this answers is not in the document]
andaaron: Rename `repos` to `repositories`

examples/config-anonymous-authz.json
rchincha: feat(groups): ...
<NEWLINE>
BREAKING CHANGE: <only related to breaking change>
<NEWLINE>
Explanation of the commit itself.

andaaron (dismissed):

examples/config-anonymous-authz.json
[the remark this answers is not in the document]
andaaron: Renamed to "repositories"

andaaron (approved):

examples/config-anonymous-authz.json
aokirisaki: I changed the commit message

andaaron (dismissed):
Looks good.

rchincha (commented):
the commit msg has not been fixed yet?

rchincha:
"BREAKING CHANGE: A groups mechanism has been added, with policies for groups"

^ this is NOT the breaking change.

andaaron (approved):

rchincha:
```
feat(groups)!: added "groups" mechanism for authZ

BREAKING CHANGE: repository paths are now specified under a new config key called "repositories" under "accessControl" section in order to handle "groups" feature. Previously the repository paths were specified directly under "accessControl".

This PR adds the ability to create groups of users which can be used for authZ policies, instead of just users.

{
"http": {
   "accessControl": {
       "groups": {
 
Just like the users, groups can be part of repository policies/default policies/admin policies. The 'groups' field in accessControl can be missing if there are no groups. The permissions priority is user>group>default>admin policy, verified in this order (in authz.go), and permissions are cumulative. It works with LDAP too, and the group attribute name is configurable. The DN of the group is used as the group name and the functionality is the same. All groups for the given user are added to the context in authn.go. Repository paths are now specified under a new keyword called "repositories" under "accessControl" section in order to handle "groups" feature.
```

andaaron (approved):

rchincha (approved):
lgtm

PR #1141, opened 2023-01-26
chore: update the version of go-lru we use to the latest available

andaaron:
We are now using v2.0.1 in the cve cache logic.
Unfortunately we are also using v0.5.4 indirectly, as it is required for gqlgen, see: https://github.com/99designs/gqlgen/blob/e6114a2c6af22bcdc92180660a58e6125e7946ad/go.mod#L7


rchincha (approved):
lgtm

PR #1146, opened 2023-01-27
feat(scheduler): use an worker pool for scheduler

peusebiu:

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -85,26 +85,55 @@ func NewScheduler(logC log.Logger) *Scheduler {
 	}
 }
 
-const rateLimit = 5 * time.Second
+const (
+	rateLimit  = 5 * time.Second
+	numWorkers = 3
+)
+
+func (scheduler *Scheduler) poolWorker(numWorkers int, tasks chan Task) {
+	for i := 0; i < numWorkers; i++ {
+		go func(workerID int) {
+			for task := range tasks {
+				scheduler.log.Debug().Int("worker", workerID).Msg("scheduler: starting task")
+
+				if err := task.DoWork(); err != nil {
+					scheduler.log.Error().Int("worker", workerID).Err(err).Msg("scheduler: error while executing task")
+				}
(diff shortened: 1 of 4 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -85,24 +85,42 @@ func NewScheduler(logC log.Logger) *Scheduler {
...
 }
 
-const rateLimit = 5 * time.Second
+const (
+	rateLimit  = 5 * time.Second
+	numWorkers = 5
+)
+
+func (scheduler *Scheduler) taskWorker(jobs chan Task) {
+	for j := range jobs {
```
laurentiuNiculae, line 94: This for loop will run indefinitely until the channel jobs is closed?

peusebiu: yes

pkg/scheduler/scheduler.go
```diff
@@ -85,26 +85,55 @@ func NewScheduler(logC log.Logger) *Scheduler {
...
-				task := scheduler.getTask()
-				if task != nil {
-					if err := task.DoWork(); err != nil {
-						scheduler.log.Error().Err(err).Msg("error while executing task")
+				i := 0
+				for i < numWorkers {
+					task := scheduler.getTask()
+					if task != nil {
+						// push tasks into worker pool
+						scheduler.log.Debug().Msg("scheduler: pushing task into worker pool")
```
rchincha, line 130: This has the problem that we no longer get FIFO semantics.

andaaron: Please elaborate on this, what are you referring to?

peusebiu: is "enqueue task" better? or this is about implementation and not logging messaging?

rchincha (approved):
Note that when tasks are dequeued and run, we no longer get FIFO semantics.
If we need stronger guarantees, we will revisit.

Issue #1145, opened 2023-01-27
[Bug]: Not helpful error message when `linter` extension rejects upload


smoser:
### zot version

unknown (zothub.io on 2023-01-27)

### Describe the bug

I tried to upload to zot with stacker and received the following error message:

```
[...]

PR #1150, opened 2023-01-30
feat(routes): better error message in case of missing annotations

peusebiu:
putting this info into error detail would be ideal, but skopeo doesn't print them, so overwrite the error message.

**What type of PR is this?**

#1145

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/errors.go
+++ b/pkg/api/errors.go
@@ -11,6 +11,12 @@ type Error struct {
 	Detail      interface{} `json:"detail,omitempty"`
 }
 
+func (e Error) WithMessage(msg string) Error {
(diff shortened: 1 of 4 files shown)


andaaron (approved):

rchincha:
Let's look at the dist spec and make sure we are not doing something way off.

rchamarthy (approved):

rchincha (approved):
lgtm

rchincha:
FYI, only the codes matter, rest can be bespoke.
https://github.com/opencontainers/distribution-spec/blob/main/spec.md#error-codes

Issue #1167, opened 2023-02-03
Investigate increase in time to run the `Build and test ZOT` workflow


andaaron:
Recently we've had to increase maximum execution timeout from 60 minutes to over 70 minutes in the `Run build and test` step which executes the make target for building and testing zot.
Shortly afterwards another increase to 80 minutes was needed, and a week later (around February 1st) the timeout seems still not enough for certain runs.

andaaron:
Attaching the test execution time as recorded from https://github.com/project-zot/zot/actions/runs/4084763446/jobs/7041855696
See upload below.

[test_times.txt](https://github.com/project-zot/zot/files/10580208/test_times.txt)

I also noticed The build command starts around 2023-02-03T13:54:27.4478122Z and ends around 2023-02-03T14:02:21.1875185Z, around 8 minutes. Note this is not the first build on this PR, so there is some caching, the first time I looked it was over 10 minutes.
See my other upload below.

[build_time.txt](https://github.com/project-zot/zot/files/10580238/build_time.txt)

rchincha:
Should we preserve the go mod cache?
https://github.com/magnetikonline/action-golang-cache

andaaron:
We are already caching. The new plugin is just syntactic sugar for the logic we already have - https://github.com/magnetikonline/action-golang-cache/blob/main/action.yaml.

But I did find a bug in caching, like this plugin, we're not taking count of runner architecture, just runner OS.
Current cache is for Linux ARM, that's why it's not working. We'll need to open a PR to fix this.

andaaron:
PRs for reducing test times:
- https://github.com/project-zot/zot/pull/1169
- https://github.com/project-zot/zot/pull/1170
- https://github.com/project-zot/zot/pull/1172
- https://github.com/project-zot/zot/pull/1173

andaaron:
Will also working on mocking Trivy as part of https://github.com/project-zot/zot/pull/1178 or a follow up PR.

PR #1169, opened 2023-02-03
ci(cache): split go build cache from go modules cache

andaaron:
The cache should not be the same if we build for different OSs/architectures

See #1167


rchincha (approved):
lgtm

PR #1170, opened 2023-02-03
test: refactor CVE tests in CLI package

andaaron:
cli tests: remove redundant conveys in CVE tests
cli tests: reduce duration of CVE tests by reducing the number of DB downloads

### DIFF
--- a/pkg/cli/cve_cmd_test.go
+++ b/pkg/cli/cve_cmd_test.go
@@ -5,7 +5,6 @@ package cli //nolint:testpackage
 
 import (
 	"bytes"
-	"context"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


rchincha:
Why are these tests redundant?

andaaron:
> Why are these tests redundant?

Changes are:

A.
There were nested conveys following the pattern:
- outer convey was a positive test
- inner conveys were negative tests
The positive and negative tests did not depend on one another, but the positive test was repeated for each negative test.

So for example if there was 1 positive test and 2 negative tests what go would do is:
1. execute the outer convey with the positive test
2. execute the 1st inner convey with the 1st negative test
3. execute the outer convey with the positive test
4. execute the 2nd inner convey with the 2nd negative test

Since there were no dependencies between these tests they can be moved on the same level.

B.
The test function bringin up the controller for `TestServerCVEResponseGQL` was the same for `TestServerCVEResponse`, so I removed `TestServerCVEResponseGQL` and moved the Conveys under `TestServerCVEResponse`. This will save time on the initial TrivyDB download, since it's done only once when the controller starts.

C. The `Convey("without flags", func() {` was actually the same as the test under which it was nested: `Convey("Test CVE invalid url port", t, func() {` no code differences so I removed it.

As can be seen in the coverage report there are no changes, as this PR just moves the code to make it more readable (also in test execution output) and efficient, it does not change the test logic.

andaaron:
Rebased and update commit message with info above.

rchincha (approved):
lgtm

PR #1172, opened 2023-02-04
test: avoid running trivy tests if search build label is missing

andaaron:
The 'test' makefile target runs the tests for both 'minimal' and equivalent of the former 'extended' build. The trivy package tests were run twice, even if the trivy logic is unreachable if search is disabled. With this update we should see a cut of about 150s of test time.


rchincha (approved):
lgtm

PR #1178, opened 2023-02-06
refactor(cve): remove the global of type cveinfo.CveInfo from the ext…

andaaron:
…ensions package

Replace it with an attribute on controller level


andaaron:
Codecov for this PR seems to be broken.
Replace by https://github.com/project-zot/zot/pull/1183.

PR #1193, opened 2023-02-13
feat(http): add support for listening on a unix socket

smoser:
This adds support for listening on a local unix socket by specifying in config http.address of "unix:/path/to/name.socket".

You can then interact with the api over that socket with:

curl --unix-socket /tmp/zot.socket http://ignored.example.com/v2/name/tags/list

**What type of PR is this?**
feature - enable listening on a unix socket.

N/A

I'm looking at a use case where I populate zot storage by running a zot and loading content to it.  In this scenario there is no need for networking or exposing the zot to even localhost network.

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -166,12 +166,23 @@ func (c *Controller) Run(reloadCtx context.Context) error {
 	c.Server = server
 
 	// Create the listener
-	listener, err := net.Listen("tcp", addr)
-	if err != nil {
-		return err
-	}
+	var listener net.Listener
+	var err error
 
-	if c.Config.HTTP.Port == "0" || c.Config.HTTP.Port == "" {
+	if strings.HasPrefix(c.Config.HTTP.Address, "unix:") {
+		// unix:/path/to/socket allows listening on unix socket
+		c.Server.Addr = c.Config.HTTP.Address
+		socketPath := c.Config.HTTP.Address[len("unix:"):]
+		listener, err = net.Listen("unix", socketPath)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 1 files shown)


pkg/api/controller.go
```diff
@@ -166,12 +166,23 @@ func (c *Controller) Run(reloadCtx context.Context) error {
...
-	listener, err := net.Listen("tcp", addr)
-	if err != nil {
-		return err
-	}
-
-	if c.Config.HTTP.Port == "0" || c.Config.HTTP.Port == "" {
+	var listener net.Listener
+	var err error
+
+	if strings.HasPrefix(c.Config.HTTP.Address, "unix:") {
```
andaaron, line 172: There should be a test to cover this type of configuration.

rchincha: Wait, why not just listen on localhost?

smoser: > Wait, why not just listen on localhost?

Any process on the system can reach a 'localhost' endpoint while unix sockets are protected by filesystem permissions. I agree that this isn't a *huge* win, as I could protect access to the zot via httpasswd auth and a randomly generated username/key, but that comes for free with a unix socket.

I'll accept that unix socket support isn't needed at the moment, but rejecting the use of unix sockets due to the availability of 'localhost' isn't an idea that would stand on its own very long. I really did this as a proof of concept.

hallyn: And in addition to protecting unix socket with file perms, this also opens the door for potential zot configs saying "only uid 100000-165536 can connect", using SO_PEERCRED.

smoser:
I think this would be useful, but its not something I'm going to follow up on now.
Thanks everyone for work on zot.

PR #1198, opened 2023-02-14
feat(mgmt): added mgmt extension which returns current zot configuration

peusebiu:

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


andaaron (commented):
This would work, I calling this API "mgmt" doesn't seem right to me. I don't see how this implementation fits into a holistic "mgmt" approach where in the future:
- users see part of the config based on their permissions
- users would also be able to POST a config based only on the fields they have permissions to set.

"mgmt" implies that you manage zot, meaning you would also be able to configure it in the future, while this implementation which skips all authentication on purpose (even for good reason) falls short with regards to a management API. 
If you don't have a way to specify which configuration you want based on query parameters/url, and you also can't figure it out based on the user permission, this "mgmt" API seems to be a dead end.

Sure. we can call it something else, which means expectations with regards to how this will evolve in the future would be different.

peusebiu:
> This would work, I calling this API "mgmt" doesn't seem right to me. I don't see how this implementation fits into a holistic "mgmt" approach where in the future:
> 
> * users see part of the config based on their permissions
> * users would also be able to POST a config based only on the fields they have permissions to set.
> 
> "mgmt" implies that you manage zot, meaning you would also be able to configure it in the future, while this implementation which skips all authentication on purpose (even for good reason) falls short with regards to a management API. If you don't have a way to specify which configuration you want based on query parameters/url, and you also can't figure it out based on the user permission, this "mgmt" API seems to be a dead end.
> 
> Sure. we can call it something else, which means expectations with regards to how this will evolve in the future would be different.

auth is skipped only if you don't provide credentials, if you give it credentials then authn and authz will take care of the request.

peusebiu:
modified the API response as follows:

no auth:
```
{
  "binaryType": "",
  "http": {
    "auth": {}
  }
}
```
htpasswd auth:
```
{
  "binaryType": "",
  "http": {
    "auth": {
      "htpasswd": {}
    }
  }
}
```

httpasswd + ldap:
```
{
  "binaryType": "",
  "http": {
    "auth": {
      "htpasswd": {},
      "ldap": {
        "address": "ldap.example.org"
      }
    }
  }
}
```

ldap auth:
```
{
  "binaryType": "",
  "http": {
    "auth": {
      "ldap": {
        "address": "ldap.example.org"
      }
    }
  }
}
```

bearer auth:
```
{
  "binaryType": "-sync-search-scrub-metrics-lint-ui-mgmt",
  "http": {
    "auth": {
      "bearer": {
        "Realm": "https://auth.myreg.io/auth/token",
        "Service": "myauth"
      }
    }
  }
}
```

ldap + htpasswd + bearer auth:
```
{
  "binaryType": "-sync-search-scrub-metrics-lint-ui-mgmt",
  "http": {
    "auth": {
      "htpasswd": {},
      "bearer": {
        "Realm": "https://auth.myreg.io/auth/token",
        "Service": "myauth"
      },
      "ldap": {
        "address": "ldap.example.org"
      }
    }
  }
}
```
@raulkele what do you think about this?

raulkele:
The API response looks good to me

andaaron (dismissed):

andaaron:
> > This would work, I calling this API "mgmt" doesn't seem right to me. I don't see how this implementation fits into a holistic "mgmt" approach where in the future:
> > 
> > * users see part of the config based on their permissions
> > * users would also be able to POST a config based only on the fields they have permissions to set.
> > 
> > "mgmt" implies that you manage zot, meaning you would also be able to configure it in the future, while this implementation which skips all authentication on purpose (even for good reason) falls short with regards to a management API. If you don't have a way to specify which configuration you want based on query parameters/url, and you also can't figure it out based on the user permission, this "mgmt" API seems to be a dead end.
> > Sure. we can call it something else, which means expectations with regards to how this will evolve in the future would be different.
> 
> auth is skipped only if you don't provide credentials, if you give it credentials then authn and authz will take care of the request.

Thanks, it's clear now.

pkg/extensions/mgmt.md
```diff
@@ -0,0 +1,46 @@
...
+  "ReleaseTag": "v1.1.15-rc1",
+  "BinaryType": "-sync-search-scrub-metrics-lint-mgmt",
+  "HTTP": {
+    "Auth": {
+      "FailDelay": 0,
+      "HTPasswd": {
+        "Path": "test/data/htpasswd"
+      },
+      "LDAP": {
+        "Address": "ldap.example.org"
```
andaaron, line 39: This behavior was changed in the last amends, right?

peusebiu: oh, yes, will fix

andaaron (approved):

rchincha (approved):

PR #1202, opened 2023-02-16
build(ui): the ui is now included in the zot binary by default

andaaron:
Update the default value of the EXTENSIONS variable in the makefile.
Also cleanup binary-ui and other make targets assuming the UI was not included by default.
Enable the ui by default in the zot container image
Switch back to using the distroless images, as c3 only has linux amd64 images.
Fix updating security events in github (permission issue)


pkg/extensions/extension_ui_test.go
```diff
@@ -0,0 +1,106 @@
...
+	. "github.com/smartystreets/goconvey/convey"
+	"gopkg.in/resty.v1"
+
+	"zotregistry.io/zot/pkg/api"
+	"zotregistry.io/zot/pkg/api/config"
+	extconf "zotregistry.io/zot/pkg/extensions/config"
+	"zotregistry.io/zot/pkg/test"
+)
+
+func TestUiExtension(t *testing.T) {
```
rchincha, line 22: TestUIExtension

rchincha: Also, use commit-id in the Makefile for zui instead of a release tag.

rchincha (approved):
lgtm

PR #1204, opened 2023-02-17
fix: trivydb update now uses task scheduler

aokirisaki:
**What type of PR is this?**

#1127

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/extension_search.go
+++ b/pkg/extensions/extension_search.go
@@ -5,6 +5,7 @@ package extensions
 
 import (
 	"net/http"
+	"sync"
 	"time"
 
 	gqlHandler "github.com/99designs/gqlgen/graphql/handler"
@@ -18,10 +19,20 @@ import (
 	"zotregistry.io/zot/pkg/extensions/search/gql_generated"
 	"zotregistry.io/zot/pkg/log"
 	"zotregistry.io/zot/pkg/meta/repodb"
+	"zotregistry.io/zot/pkg/scheduler"
 	"zotregistry.io/zot/pkg/storage"
 )
 
-type CveInfo cveinfo.CveInfo
+type (
+	CveInfo cveinfo.CveInfo
+	state   int
+)
+
+const (
+	pending state = iota
+	running
+	done
+)
 
 func GetCVEInfo(config *config.Config, storeController storage.StoreController,
 	repoDB repodb.RepoDB, log log.Logger,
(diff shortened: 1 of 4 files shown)


pkg/extensions/extension_search.go
```diff
@@ -51,30 +59,94 @@ func EnableSearchExtension(config *config.Config, storeController storage.StoreC
...
+func newTrivyTask(interval time.Duration, cveInfo cveinfo.CveInfo,
+	generator *trivyTaskGenerator, log log.Logger,
+) *trivyTask {
+	return &trivyTask{interval, cveInfo, generator, time.Second, log}
+}
+
+func (trivyT *trivyTask) DoWork() error {
+	trivyT.generator.SetTrivyGeneratorStatus(statusRunning)
+	trivyT.log.Info().Msg("updating the CVE database")
+	time.Sleep(trivyT.waitTime)
```
andaaron, line 126: This is actually not OK. we need to wait before generating the task. If we wait here, we block one of the goroutines of the scheduler. Not sure where's the correct place to put this. @Andreea-Lupu any thoughts?

pkg/extensions/extension_search.go
```diff
@@ -51,30 +62,105 @@ func EnableSearchExtension(config *config.Config, storeController storage.StoreC
...
+	gen.waitTime = 0
+	gen.lock.Unlock()
+}
 
-		time.Sleep(updateInterval)
+type trivyTask struct {
+	interval  time.Duration
+	cveInfo   cveinfo.CveInfo
+	generator *trivyTaskGenerator
+	waitTime  time.Duration
```
andaaron, line 133: This should no longer be needed here.

pkg/extensions/extension_search.go
```diff
@@ -51,30 +62,105 @@ func EnableSearchExtension(config *config.Config, storeController storage.StoreC
...
+	interval     time.Duration
+	cveInfo      CveInfo
+	log          log.Logger
+	status       state
+	waitTime     time.Duration
+	lastTaskTime time.Time
+	lock         *sync.Mutex
+}
+
+func (gen *trivyTaskGenerator) SetTrivyGeneratorStatus(status state) {
```
andaaron, line 89: Not really sure we need the setter/getter anymore, since we don't lock here anymore.

andaaron (approved):
Looks good!

rchincha (approved):
lgtm

PR #1212, opened 2023-02-17
chore(codecov): use a token to authenticate to codecov

andaaron:
This should not be necessary for public repos,
but should help with rate limiting


rchincha (approved):
lgtm

PR #1223, opened 2023-02-23
fix: add additional blob path validation

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -1441,6 +1441,7 @@ github.com/rs/zerolog v1.29.1 h1:cO+d60CHkknCbvzEWxP0S9K6KqyTjrCNUy1LdQLCGPc=
 github.com/rs/zerolog v1.29.1/go.mod h1:Le6ESbR7hc+DP6Lt1THiV8CQSdkkNrd3R0XbEgp3ZBU=
 github.com/rubenv/sql-migrate v1.2.0 h1:fOXMPLMd41sK7Tg75SXDec15k3zg5WNV6SjuDRiNfcU=
 github.com/rubenv/sql-migrate v1.2.0/go.mod h1:Z5uVnq7vrIrPmHbVFfR4YLHRZquxeHpckCnRq0P/K9Y=
+github.com/russross/blackfriday v1.6.0 h1:KqfZb0pUVN2lYqZUYRddxF4OR8ZMURnJIG5Y3VRLtww=
 github.com/russross/blackfriday/v2 v2.0.1/go.mod h1:+Rmxgy9KzJVeS9/2gXHxylqXiyQDYRxCVz55jmeOWTM=
 github.com/russross/blackfriday/v2 v2.1.0 h1:JIOH55/0cWyOuilr9/qlrm0BSXldqnqwMsf35Ld67mk=
 github.com/russross/blackfriday/v2 v2.1.0/go.mod h1:+Rmxgy9KzJVeS9/2gXHxylqXiyQDYRxCVz55jmeOWTM=
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -4374,7 +4374,81 @@ func TestArtifactReferences(t *testing.T) {
 					Get(baseURL + fmt.Sprintf("/v2/%s/referrers/%s", repoName, digest.String()))
 				So(err, ShouldBeNil)
 				So(resp.StatusCode(), ShouldEqual, http.StatusOK)
+
+				// create a bad manifest (constructed manually)
+				content := `{"schemaVersion":2,"mediaType":"application/vnd.oci.image.manifest.v1+json","subject":{"mediaType":"application/vnd.oci.image.manifest.v1+json","digest":"sha256:71dbae9d7e6445fb5e0b11328e941b8e8937fdd52465079f536ce44bb78796ed","size":406}}` //nolint: lll
+				resp, err = resty.R().SetHeader("Content-Type", ispec.MediaTypeImageManifest).
+					SetBody(content).Put(baseURL + fmt.Sprintf("/v2/%s/manifests/1.0", repoName))
+				So(err, ShouldBeNil)
+				So(resp.StatusCode(), ShouldEqual, http.StatusBadRequest)
+
+				// missing layers
+				mcontent := []byte("this is a missing blob")
+				digest = godigest.FromBytes(mcontent)
+				So(digest, ShouldNotBeNil)
+
+				manifest.Layers = append(manifest.Layers, ispec.Descriptor{
+					MediaType: "application/vnd.oci.image.layer.v1.tar",
+					Digest:    digest,
+					Size:      int64(len(mcontent)),
+				})
+
+				mcontent, err = json.Marshal(manifest)
+				So(err, ShouldBeNil)
+				resp, err = resty.R().SetHeader("Content-Type", ispec.MediaTypeImageManifest).
+					SetBody(mcontent).Put(baseURL + fmt.Sprintf("/v2/%s/manifests/1.0", repoName))
+				So(err, ShouldBeNil)
+				So(resp.StatusCode(), ShouldEqual, http.StatusBadRequest)
+
+				// invalid schema version
+				manifest.SchemaVersion = 1
+
+				mcontent, err = json.Marshal(manifest)
+				So(err, ShouldBeNil)
+				resp, err = resty.R().SetHeader("Content-Type", ispec.MediaTypeImageManifest).
+					SetBody(mcontent).Put(baseURL + fmt.Sprintf("/v2/%s/manifests/1.0", repoName))
+				So(err, ShouldBeNil)
+				So(resp.StatusCode(), ShouldEqual, http.StatusBadRequest)
+
+				// upload image config blob
+				resp, err = resty.R().Post(baseURL + fmt.Sprintf("/v2/%s/blobs/uploads/", repoName))
+				So(err, ShouldBeNil)
+				So(resp.StatusCode(), ShouldEqual, http.StatusAccepted)
+				loc := test.Location(baseURL, resp)
+				cblob = []byte("{}")
+				cdigest = godigest.FromBytes(cblob)
+				So(cdigest, ShouldNotBeNil)
+
+				resp, err = resty.R().
+					SetContentLength(true).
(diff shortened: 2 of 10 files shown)


pkg/storage/common.go
```diff
@@ -91,6 +92,16 @@ func ValidateManifest(imgStore ImageStore, repo, reference, mediaType string, bo
 				return "", zerr.ErrBadManifest
 			}
 		}
+
+		// a lightweight check if the blob is present
+		for _, layer := range manifest.Layers {
+			blobPath := imgStore.BlobPath(repo, layer.Digest)
+			if _, err := os.Stat(blobPath); err != nil {
```
andaaron, line 99: This is common code for both local and s3. This does not work. You need to call an image store API.

pkg/storage/common.go
```diff
@@ -91,6 +91,16 @@ func ValidateManifest(imgStore ImageStore, repo, reference, mediaType string, bo
 				return "", zerr.ErrBadManifest
 			}
 		}
+
+		// a lightweight check if the blob is present
+		for _, layer := range manifest.Layers {
+			_, _, err := imgStore.CheckBlob(repo, layer.Digest)
```
peusebiu, line 97: CheckBlob has a lock inside, and ValidateManifest is already locked from outside.

use imgStore.GetBlobContent() instead. it will throw ErrBlobNotFound.

pkg/storage/common.go
```diff
@@ -68,6 +71,36 @@ func ValidateManifest(imgStore ImageStore, repo, reference, mediaType string, bo
...
+		schema, err := gojsonschema.NewSchema(loader)
+		if err != nil {
+			log.Error().Err(err).Msg("unable to unmarshal JSON")
+
+			return "", zerr.ErrBadManifest
+		}
+
+		result, err := schema.Validate(documentLoader)
+		if err != nil {
+			log.Error().Err(err).Msg("unable to unmarshal JSON")
```
andaaron, line 87:
```suggestion
			log.Error().Err(err).Msg("manifest JSON not compliant with the OCI spec")
```

pkg/storage/common.go
```diff
@@ -68,6 +71,36 @@ func ValidateManifest(imgStore ImageStore, repo, reference, mediaType string, bo
...
+		loader := gojsonschema.NewStringLoader(string(b))
+		documentLoader := gojsonschema.NewStringLoader(string(body))
+		schema, err := gojsonschema.NewSchema(loader)
+		if err != nil {
+			log.Error().Err(err).Msg("unable to unmarshal JSON")
+
+			return "", zerr.ErrBadManifest
+		}
+
+		result, err := schema.Validate(documentLoader)
```
andaaron, line 85: Is it possible to initialize the `schema` beforehand, when the server starts? Instead of creating these objects every time.

pkg/storage/common.go
```diff
@@ -68,6 +71,36 @@ func ValidateManifest(imgStore ImageStore, repo, reference, mediaType string, bo
...
+		if err != nil {
+			log.Error().Err(err).Msg("unable to unmarshal JSON")
+
+			return "", zerr.ErrBadManifest
+		}
+
+		if !result.Valid() {
+			fmt.Printf("The document is not valid. see errors :\n")
```
andaaron, line 93: This should be a log message.

pkg/storage/common.go
rchincha: Will have to become a global var likely :(

pkg/storage/common.go
```diff
@@ -56,6 +58,51 @@ func GetManifestDescByReference(index ispec.Index, reference string) (ispec.Desc
...
+		for _, err := range result.Errors() {
+			field := err.Details()["field"]
+			property := err.Details()["property"]
+
+			if field == "(root)" &&
+				(property != "layers" &&
+					property != "subject" &&
+					property != "annotations") {
+				err := zerr.ErrBadManifest
+				log.Error().Err(err).Interface("property", property).Msg("missing field in OCIv1 image manifest")
```
andaaron, line 96:
```suggestion
				log.Error().Err(err).Interface("property", property).Msg("property not present in OCIv1 image manifest spec")
```

pkg/storage/common.go
```diff
@@ -99,6 +153,22 @@ func ValidateManifest(imgStore ImageStore, repo, reference, mediaType string, bo
 				return "", zerr.ErrBadManifest
 			}
 		}
+
+		// a lightweight check if the blob is present
+		for _, layer := range manifest.Layers {
```
andaaron, line 158: This duplicates code from validateOCIManifest.

pkg/storage/common.go
```diff
@@ -76,21 +123,32 @@ func ValidateManifest(imgStore ImageStore, repo, reference, mediaType string, bo
...
 
-				return digest, err
+				return "", err
 			}
 		}
 
+		// validate "subject"
 		if manifest.Subject != nil {
 			var m ispec.Descriptor
 			if err := json.Unmarshal(body, &m); err != nil {
```
andaaron, line 154: What exactly is this check doing? 
Manifest Subject is already type descriptor after the unmarshal?
This check verifies the manifest itself is a descriptior?
Do we need this?

andaaron:
> https://github.com/project-zot/zot/pull/1223/files#diff-f0da7068358f30587e2628f244293247918e02939f1df8ccafc9bd89d25146cbR132
> 
> We already validate manifest layers in validateOCIManifest
> 
> We do it with GetBlobContent which is bad because it will 'download' content and we don't need it.
> 
> We can have a VerifyBlob() function with CheckBlob logic but without locks, and use locks in
> 
> ```
> func CheckBlob() {
>   lock()
>   defer unlock()
>   return VerifyBlob()
> }
> ```

Ok I see Sebi has the same concern as me.

rchincha:
```
warning: config sha256:205eed24cbec29ad9cb4593a73168ef1803402370a82f7d51ce25646fc2f943a has an unknown media type: com.artifact.test                                      warning: layer sha256:b8ef38a79a15865d2a635e90fead3da1d8adff96c56365397da07852fc3cbc99 has an unknown media type: application/octet-stream                                {"level":"error","error":"[config.mediaType: Does not match pattern '^[A-Za-z0-9][A-Za-z0-9!#$&-^_.+]{0,126}/[A-Za-z0-9][A-Za-z0-9!#$&-^_.+]{0,126}$']","goroutine":104,"caller":"/local/rchincha/go/src/github.com/project-zot/zot/pkg/storage/common.go:90","time":"2023-05-10T21:03:50.299060476Z","caller":"/local/rchincha/go/src/github.com/project-zot/zot/pkg/storage/common.go:90","message":"OCIv1 image manifest schema validation failed"}                                                                        {"level":"info","module":"http","clientIP":"127.0.0.1:32856","method":"PUT","path":"/v2/artifact-ref/manifests/sha256:b87cd2df7da8c4dbf32d5d8af838bc4b9513a1ede8727ba82e4376c0a12f0732","statusCode":400,"latency":"0s","bodySize":423,"headers":{"Accept":["application/vnd.oci.image.manifest.v1+json"],"Accept-Encoding":["gzip"],"Content-Length":["561"],"Content-Type":["application/vnd.oci.image.manifest.v1+json"],"User-Agent":["go-resty/1.12.0 (https://github.com/go-resty/resty)"]},"goroutine":104,"caller":"/local/rchincha/go/src/github.com/project-zot/zot/pkg/api/session.go:131","time":"2023-05-10T21:03:50.299152421Z","message":"HTTP API"}                                     ✘

                                                                                                                                                                          Failures:
                                                                                                                                                                            * /local/rchincha/go/src/github.com/project-zot/zot/pkg/extensions/search/search_test.go                                                                                  Line 1140:                                                                                                                                                                Expected: nil
  Actual:   'can't put blob'
  ```

**Comment by rchincha** [comment]
```
go test -v -cover -run ^TestGetReferrersGQL -tags sync,search,scrub,metrics,lint,ui,mgmt,userprefs,containers_image_openpgp ./pkg/extensions/search/...
```

**Comment by rchincha** [comment]
```
{"level":"error","error":"[layers.0.data: Invalid type. Expected: string, given: null]","goroutine":97,"caller":"zotregistry.io/zot/pkg/storage/common/common.go:79","time":"2023-06-12T21:54:52.824219066Z","caller":"zotregistry.io/zot/pkg/storage/common/common.go:79","message":"OCIv1 image manifest schema validation failed"}               {"level":"info","module":"http","clientIP":"127.0.0.1:54348","method":"PUT","path":"/v2/oci-conformance/distribution-test/manifests/tagtest0","statusCode":400,"latency":"0s","bodySize":360,"headers":{"Content-Length":["671"],"Content-Type":["application/vnd.oci.image.manifest.v1+json"],"User-Agent":["distribution-spec-conformance-tests"]},"goroutine":97,"caller":"zotregistry.io/zot/pkg/api/session.go:131","time":"2023-06-12T21:54:52.824392519Z","message":"HTTP API"}    
```

peusebiu:
> Run a latest conformance test against zot with this PR, and understand why this is.
> 
> ```
> {"level":"error","error":"[layers.0.data: Invalid type. Expected: string, given: null]","goroutine":97,"caller":"zotregistry.io/zot/pkg/storage/common/common.go:79","time":"2023-06-12T21:54:52.824219066Z","caller":"zotregistry.io/zot/pkg/storage/common/common.go:79","message":"OCIv1 image manifest schema validation failed"}               {"level":"info","module":"http","clientIP":"127.0.0.1:54348","method":"PUT","path":"/v2/oci-conformance/distribution-test/manifests/tagtest0","statusCode":400,"latency":"0s","bodySize":360,"headers":{"Content-Length":["671"],"Content-Type":["application/vnd.oci.image.manifest.v1+json"],"User-Agent":["distribution-spec-conformance-tests"]},"goroutine":97,"caller":"zotregistry.io/zot/pkg/api/session.go:131","time":"2023-06-12T21:54:52.824392519Z","message":"HTTP API"}    
> ```
> 
> Secondly, layers can be empty but upstream image-spec requires at least one layer, so maybe we cannot use this validation.

I can't fully grasp the validation code... I think it's because this config descriptor:
https://github.com/opencontainers/distribution-spec/blob/main/conformance/setup.go#L289
has 'Data' field and the layers don't have it which is null, and the library expects a string.

andaaron:
Superseded by https://github.com/project-zot/zot/pull/1609

PR #1232, opened 2023-02-28
refactor(test): update cve tests to stop duplicating test/data if not…

nicoldr:
… needed

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/search/cve/cve_test.go
+++ b/pkg/extensions/search/cve/cve_test.go
@@ -8,7 +8,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"net/http"
 	"os"
 	"path"
 	"strings"
@@ -407,8 +406,8 @@ func TestCVESearchDisabled(t *testing.T) {
 			},
 		}
 
-		dbDir, err := testSetup(t)
-		So(err, ShouldBeNil)
+		dbDir := t.TempDir()
+
 		conf.Storage.RootDirectory = dbDir
 		defaultVal := true
 		searchConfig := &extconf.SearchConfig{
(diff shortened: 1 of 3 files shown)


pkg/extensions/search/cve/cve_test.go
```diff
@@ -727,8 +727,6 @@ func TestCVEConfig(t *testing.T) {
 
 		secondDir := t.TempDir()
 
-		CopyTestFiles("../../../../test/data", path.Join(secondDir, "a"))
-
 		ctlr.Config.Storage.RootDirectory = firstDir
```
andaaron, line 730: It seems to me like the purpose of this test was to enable the search, and obtain coverage on:
https://github.com/project-zot/zot/commit/b0ed625a2e0f7c5724a6d3c613c066ac874dbc87

However the CVE extensions/configuration was removed and the test was repurposed to test multiple storage subpaths in:
https://github.com/project-zot/zot/commit/28974e81dcd449e78d4735f08b960b88f5ea068a#diff-5e37dc387d048f301d803af010fae652fc9fb27118226fd3145ed4372fb22900L707

I suggest we:
- rename `TestCVEConfig` to `TestCVEMultipleSubstores`
- we should either add images with CVEs (keep CopyTestFiles and scan these images), or mock the scanner.
- add some calls at the end of the tests to verify the CVEs present in the images using the graphql API.

nicoldr: Commented this test and I will check how this affects coverage and if it can be removed or has to be modified

andaaron (approved):

rchincha (approved):
lgtm

PR #1240, opened 2023-03-03
fix(go.mod): replace opencontainers/umoci dependency with project-sta…

Andreea-Lupu:
…cker/umoci

**What type of PR is this?**

Currently, github.com/opencontainers/umoci doesn't offer support for OCI Artifacts. This is the reason for what notation signature blobs are not marked as reachable when GC is performed, so these blobs are removed even if the manifests which contain them still exist. This problem was fixed in github.com/project-stacker/umoci/tree/stacker, so this PR updates the go.mod with this replacement(github.com/opencontainers/umoci => github.com/project-stacker/umoci) in order to solve the presented issue.

**Automation added to e2e**:

```release-note

```


rchincha (approved):
lgtm

PR #1019, opened 2022-11-24
fix(routes): set Content-Length on GetManifest

peusebiu:
notation sign image index exits with "missing Content-Length" error

Signed-off-by: Petu Eusebiu <peusebiu@cisco.com>

**What type of PR is this?**

**Which issue does this PR fix**:

**What does this PR do / Why do we need it**:

**If an issue # is not available please add repro steps and logs showing the issue**:

**Testing done on this change**:

**Automation added to e2e**:

**Will this break upgrades or downgrades?**

**Does this PR introduce any user-facing change?**:

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1256, opened 2023-03-10
bug(zli): fix compatibility with servers not returning Content-Length for the /manifests API

andaaron:
This is the case of older zot servers running without graphql support, and which do not have the fix: https://github.com/project-zot/zot/pull/1019


peusebiu (approved):
lgtm

Issue #909, opened 2022-10-24
Global search: for multiple nested folders inside the repo name search only matches the first and last item in the path


andaaron:
If I want to find `c3/debian/base-amd64`, I can only search by `c3` or `base`.
If the input for the search query is `debian`, the result list does not include `c3/debian/base-amd64`.

andaaron:
Done in #1257

PR #1257, opened 2023-03-10
feat(search): update search pattern matching rules

laurentiuNiculae:
**What type of PR is this?**

#909 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/meta/repodb/boltdb-wrapper/boltdb_wrapper.go
+++ b/pkg/meta/repodb/boltdb-wrapper/boltdb_wrapper.go
@@ -897,7 +897,7 @@ func (bdw *DBWrapper) SearchRepos(ctx context.Context, searchText string, filter
 				return err
 			}
 
-			if score := common.ScoreRepoName(searchText, string(repoName)); score != -1 {
+			if rank := common.RankRepoName(searchText, string(repoName)); rank != -1 {
 				var (
 					// specific values used for sorting that need to be calculated based on all manifests from the repo
 					repoDownloads     = 0
@@ -1006,7 +1006,7 @@ func (bdw *DBWrapper) SearchRepos(ctx context.Context, searchText string, filter
 
 				pageFinder.Add(repodb.DetailedRepoMeta{
 					RepoMeta:   repoMeta,
-					Score:      score,
+					Rank:       rank,
 					Downloads:  repoDownloads,
 					UpdateTime: repoLastUpdated,
 				})
--- a/pkg/meta/repodb/common/common.go
+++ b/pkg/meta/repodb/common/common.go
@@ -62,7 +62,21 @@ func ValidateRepoReferenceInput(repo, reference string, manifestDigest godigest.
 	return nil
 }
 
-func ScoreRepoName(searchText string, repoName string) int {
+// These constants are meant used to describe how high or low in rank a match is.
+// Note that the "higher rank" relates to a lower number so ranks are sorted in a
+// ascending order.
+const (
+	lowPriority          = 100
+	mediumPriority       = 10
+	highPriority         = 1
+	perfectMatchPriority = 0
+)
+
+// RankRepoName associates a rank to a given repoName given a searchText.
+// The imporance of the value grows inversly proportional to the int value it has.
+// For example: rank(1) > rank(10) > rank(100)...
+func RankRepoName(searchText string, repoName string) int {
+	searchText = strings.Trim(searchText, "/")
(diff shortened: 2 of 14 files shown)


pkg/meta/repodb/common/common.go
```diff
@@ -71,34 +77,47 @@ func ScoreRepoName(searchText string, repoName string) int {
...
 	for i := 0; i < len(searchTextSlice)-1; i++ {
 		if searchTextSlice[i] != repoNameSlice[i] {
-			return -1
+			begginingMatches = false
+
+			break
+		}
+	}
+
+	if begginingMatches {
```
rchincha, line 110: typo

beginning

pkg/meta/repodb/common/common.go
```diff
@@ -62,6 +62,12 @@ func ValidateRepoReferenceInput(repo, reference string, manifestDigest godigest.
 	return nil
 }
 
+const (
```
rchincha, line 65: what is this for?

pkg/meta/repodb/common/common.go
```diff
@@ -71,34 +77,47 @@ func ScoreRepoName(searchText string, repoName string) int {
 	}
```
andaaron, line 77: Make sure and exact match always has lowest score.

pkg/meta/repodb/common/common.go
```diff
@@ -71,34 +77,47 @@ func ScoreRepoName(searchText string, repoName string) int {
...
 
 	if len(searchTextSlice) == 1 {
-		// check if it maches first or last name in path
+		// check if it maches first or last directory name in path, we'll give priority for these names
```
andaaron, line 80:
```suggestion
		// check if it matches last directory name in path, we'll give priority for these names
```

andaaron (approved):

pkg/meta/repodb/boltdb-wrapper/boltdb_wrapper.go
```diff
@@ -999,7 +999,7 @@ func (bdw *DBWrapper) SearchRepos(ctx context.Context, searchText string, filter
 				return err
 			}
 
-			if score := common.ScoreRepoName(searchText, string(repoName)); score != -1 {
+			if score := common.RankRepoName(searchText, string(repoName)); score != -1 {
```
rchincha, line 1002: call "score" rank also?

rchincha:
So what is the overall logic for ranking? image:tag first, repo names, and any wildcards? We should eventually also allow searching within annotations. This use case is coming up often.

andaaron:
> So what is the overall logic for ranking? image:tag first, repo names, and any wildcards? We should eventually also allow searching within annotations. This use case is coming up often.

- If a tag is specified (`:`), the search string for the repo needs to be an exact match. That is a tag search, not a repo search. The logic for that was not changed in this PR.
- If a `:` is not specified, the repo string is compared to the search string according to the new comments.

We did not decide on any wildcard, or annotation searches to my knowledge.
At the moment the presence of `:` signals if this is a repo or a tag search.

What would be a good example of a string the user sends as an annotation search? How would the servers know that is an annotation search, not a repo or a tag search?

Actually I am not convinced searches by annotations are a good idea: which annotations?, on the manifest itself?, on an index?, on descriptors inside a manifest or index?, by annotation key? by annotation value?, by an annotation value on a specific key?

Why would I want to search on an annotation anyway? What is the use case?

andaaron (dismissed):
It's good to see we removed `score` from the graphql api, as we had no benefit in exposing it.

pkg/meta/repodb/common/common.go
```diff
@@ -62,43 +62,89 @@ func ValidateRepoReferenceInput(repo, reference string, manifestDigest godigest.
...
+
+// RankRepoName associates a rank to a given repoName given a searchText.
+// The imporance of the value grows inversly proportional to the int value it has.
+// For example: rank(1) > rank(10) > rank(100)...
+func RankRepoName(searchText string, repoName string) int {
+	searchText = strings.Trim(searchText, "/")
 	searchTextSlice := strings.Split(searchText, "/")
 	repoNameSlice := strings.Split(repoName, "/")
 
+	// ex: zot/test/alpine, "alpine"
```
rchincha, line 83: "repo/repo1/abc" use other names
or remove the comment, if there are examples in test cases

andaaron (dismissed):

rchincha (approved):
lgtm

Issue #1242, opened 2023-03-06
GraphQL: Add ability to filter on the CVE list for an image


andaaron:
If a UI user needs to find a specific CVE he now has to scroll through the entire CVE list for that image (which could contain hundreds of CVEs). It would be an improvement to be able to search for a specific CVE.
Note the pagination logic on the backend needs to take into consideration this search string.

andaaron:
Done in https://github.com/project-zot/zot/pull/1264

PR #1264, opened 2023-03-14
fix(cve): Search by CVE title/id (full or partial) when listing an im…

aokirisaki:
…age's CVEs

**What type of PR is this?**

#1242 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/cli/cve_cmd.go
+++ b/pkg/cli/cve_cmd.go
@@ -117,6 +117,7 @@ func NewCveCommand(searchService SearchService) *cobra.Command {
 func setupCveFlags(cveCmd *cobra.Command, variables cveFlagVariables) {
 	variables.searchCveParams["imageName"] = cveCmd.Flags().StringP("image", "I", "", "List CVEs by IMAGENAME[:TAG]")
 	variables.searchCveParams["cveID"] = cveCmd.Flags().StringP("cve-id", "i", "", "List images affected by a CVE")
+	variables.searchCveParams["searchedCVE"] = cveCmd.Flags().StringP("search", "s", "", "Search specific CVEs by name/id")
 
 	cveCmd.Flags().StringVar(variables.servURL, "url", "", "Specify zot server URL if config-name is not mentioned")
 	cveCmd.Flags().StringVarP(variables.user, "user", "u", "", `User Credentials of `+
--- a/pkg/extensions/search/cve/cve.go
+++ b/pkg/extensions/search/cve/cve.go
@@ -3,6 +3,7 @@ package cveinfo
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	godigest "github.com/opencontainers/go-digest"
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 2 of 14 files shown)


pkg/extensions/search/schema.graphql
```diff
@@ -550,6 +550,8 @@ type Query {
         image: String!,
         "Sets the parameters of the requested page"
         requestedPage: PageInput
+        "Search term for specific CVE by title/id"
+        searchedCVE: String!
```
andaaron, line 554: searchedCVE should not be mandatory. And if it is not present, all CVEs should be returned, just like current behavior on main.
```suggestion
        searchedCVE: String
```

pkg/extensions/search/cve/cve.go
```diff
@@ -215,7 +216,18 @@ func (cveinfo BaseCveInfo) GetImageListWithCVEFixed(repo, cveID string) ([]commo
 	return fixedTags, nil
 }
 
-func (cveinfo BaseCveInfo) GetCVEListForImage(repo, tag string, pageInput PageInput) (
+func filterCVEList(cveMap map[string]cvemodel.CVE, searchedCVE string, pageFinder *CvePageFinder) {
+	searchedCVE = strings.ToUpper(searchedCVE)
+
```
andaaron, line 221: So what happens if searchedCVE is empty string?

pkg/cli/cve_cmd.go
```diff
@@ -123,6 +126,7 @@ func setupCveFlags(cveCmd *cobra.Command, variables cveFlagVariables) {
 		`zot server in USERNAME:PASSWORD format`)
 	cveCmd.Flags().StringVarP(variables.outputFormat, "output", "o", "", "Specify output format [text/json/yaml]."+
 		" JSON and YAML format return all info for CVEs")
+	cveCmd.Flags().StringVarP(variables.searchedCVE, "search", "s", "", "Search specific CVEs by name/id")
```
andaaron, line 129: I think this should be saved together with the image name and cveID in the map searchCveParams. Is that possible?
The code would be cleaner.

```suggestion
	variables.searchCveParams["searchedCVE"] = cveCmd.Flags().StringP("search", "s", "", Search specific CVEs by name/id")
```

pkg/extensions/search/cve/cve.go
aokirisaki: It will match all the CVEs, because all CVE titles/ids will contain the "" string

pkg/extensions/search/cve/cve.go
```diff
@@ -19,7 +20,7 @@ import (
 type CveInfo interface {
 	GetImageListForCVE(repo, cveID string) ([]common.TagInfo, error)
 	GetImageListWithCVEFixed(repo, cveID string) ([]common.TagInfo, error)
-	GetCVEListForImage(repo, tag string, pageinput PageInput) ([]cvemodel.CVE, PageInfo, error)
+	GetCVEListForImage(repo, tag string, pageinput PageInput, searchedCVE *string) ([]cvemodel.CVE, PageInfo, error)
```
andaaron, line 23: I think we should convert it from pointer to string before calling this function. For consistency with the parameters `repo` and `tag`. Also reorder the parameters for `pageinput` to be last.

pkg/cli/cve_cmd.go
```diff
@@ -83,9 +83,10 @@ func NewCveCommand(searchService SearchService) *cobra.Command {
 				fixedFlag:     &fixedFlag,
 				verifyTLS:     &verifyTLS,
 				verbose:       &verbose,
-				debug:         &debug,
-				resultWriter:  cmd.OutOrStdout(),
-				spinner:       spinnerState{spin, isSpinner},
+
```
andaaron, line 86: Revert this change, is it not needed.

andaaron (approved):
Looks good.

rchincha (approved):
lgtm

Issue #1263, opened 2023-03-14
GraphQL: Having OCI artifacts which are tagged breaks calls returning repo lists


andaaron:
Specifically if the user pushes artifacts with tags to zot, the calls succeed, but subsequent GraphQL calls invoking `zotregistry.io/zot/pkg/extensions/search/convert.RepoMeta2RepoSummary` are failing.

```bash
remote_dest_image_ref=${REGISTRY}/${image}:${tag}

# Upload image logo as artifact media type
regctl artifact put --media-type "application/vnd.oci.artifact.manifest.v1+json" --annotation artifact.type=com.zot.logo.artifact --annotation format=oci \
    --artifact-type "application/vnd.zot.logo.v1" --subject ${remote_dest_image_ref} ${remote_dest_image_ref}-logo-artifact << EOF
${logo}
EOF

# Upload image logo as image media type
regctl artifact put --annotation artifact.type=com.zot.logo.image --annotation format=oci \
    --artifact-type "application/vnd.zot.logo.v1" --subject ${remote_dest_image_ref} ${remote_dest_image_ref}-logo-image << EOF
${logo}
EOF
```

The error as shown in the zot logs:
```
runtime/debug.Stack()
        runtime/debug/stack.go:24 +0x65
runtime/debug.PrintStack()
        runtime/debug/stack.go:16 +0x19
github.com/99designs/gqlgen/graphql.DefaultRecover({0x55ce2d83e9ff?, 0x0?}, {0x55ce33b72ac0?, 0x55ce36c7bf70?})
        github.com/99designs/gqlgen@v0.17.25/graphql/recovery.go:17 +0x71
github.com/99designs/gqlgen/graphql.(*OperationContext).Recover(0xc00029a400?, {0x55ce34191540, 0xc00272d1d0}, {0x55ce33b72ac0?, 0x55ce36c7bf70?})
        github.com/99designs/gqlgen@v0.17.25/graphql/context_operation.go:124 +0x3e
zotregistry.io/zot/pkg/extensions/search/gql_generated.(*executionContext)._Query_RepoListWithNewestImage.func1()
        zotregistry.io/zot/pkg/extensions/search/gql_generated/generated.go:5121 +0x6a
panic({0x55ce33b72ac0, 0x55ce36c7bf70})
        runtime/panic.go:884 +0x212
[...]

PR #1265, opened 2023-03-14
fix(repodb): GQL request for ExpandedRepoInfo errors when artifacts w…

laurentiuNiculae:
…ith tags are present

If we push an artifact and give it a tag, repodb would crash because of the null pointer dereferencing

Now when iterating over the tags of a repo and stumbling upon a unsupported media type, it's being ignored

**What type of PR is this?**

#1263 

**Automation added to e2e**:

```release-note

```


rchincha (approved):
lgtm

PR #1267, opened 2023-03-15
fix(mgmt): skip bearer authn for mgmt route

peusebiu:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -55,7 +55,18 @@ func bearerAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 			}
 			vars := mux.Vars(request)
 			name := vars["name"]
+
+			// we want to bypass auth for mgmt route
+			isMgmtRequested := request.RequestURI == constants.FullMgmtPrefix
+
 			header := request.Header.Get("Authorization")
+
+			if (header == "" || header == "Basic Og==") && isMgmtRequested {
+				next.ServeHTTP(response, request)
+
+				return
+			}
+
 			action := auth.PullAction
 			if m := request.Method; m != http.MethodGet && m != http.MethodHead {
 				action = auth.PushAction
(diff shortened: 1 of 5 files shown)


pkg/api/authn.go
```diff
@@ -55,7 +55,18 @@ func bearerAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 			}
 			vars := mux.Vars(request)
 			name := vars["name"]
+
```
andaaron, line 58: Why do we need this?
The request reaches this code if the client already has a bearer token, and if he already knows it needs to use a bearer token, it doesn't need the mgmt api?

rchincha: anonymous case?

peusebiu: hmm I think you have a point, although this code is reached without token first
then it will authFail(response, permissions.WWWAuthenticateHeader, 0) with bearer headers
then client will get the token using bearer headers
then it will try the same request but with token this time.

But yeah, if UI makes a /v2/ GET request it can tell by the WWWAuthenticateHeader that a bearer auth is in place.

andaaron: Do we have a use case with bearer auth for the UI?

andaaron: Ok, let's fix this.

pkg/common/bearer.go
andaaron, line 1: Move this to pkg/test/bearer.go.
My understanding is this file is only relevant for testing.

pkg/api/authn.go
```diff
@@ -55,7 +55,18 @@ func bearerAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
...
+			// we want to bypass auth for mgmt route
```
rchincha, line 59: This will not be entirely correct when accessControl is enforced over mgmt config push and data.

rchincha (approved):
lgtm

PR #1273, opened 2023-03-17
feat: cleanup error msgs

nicoldr:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/local/local.go
+++ b/pkg/storage/local/local.go
@@ -986,7 +986,7 @@ retry:
 
 		dstRecordFi, err := os.Stat(dstRecord)
 		if err != nil {
-			is.log.Error().Err(err).Str("blobPath", dstRecord).Msg("dedupe: unable to stat")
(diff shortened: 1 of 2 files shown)


pkg/storage/local/local.go
```diff
@@ -1112,13 +1110,21 @@ func (is *ImageStoreLocal) checkCacheBlob(digest godigest.Digest) (string, error
 
 	dstRecord, err := is.cache.GetBlob(digest)
 	if err != nil {
+		if errors.Is(err, zerr.ErrCacheMiss) {
```
rchincha, line 1113: Just make this a if-else on the error type and single return?

andaaron:
Can you also check the DynamoDB implementation of cache for the same cases?

nicoldr:
> Can you also check the DynamoDB implementation of cache for the same cases?

I also checked DynamoDB and changer log level in some cases and added log messages

andaaron (approved):

rchincha (approved):
lgtm

PR #1289, opened 2023-03-21
fix(loadrepodb): statistics are now prezerved after reloading zot

### LINKED ISSUE #1274: RepoDB: repodb.RepoMetadata should be reset on zot restart, and only the statistics be kept
Tags/Referrers/Signatures should be reset when server restarts (to make sure the data in the DB matches the latest struct updates, if not for other reasons), but the Statistics should remain (number of downloads should not be lost). 

In current implementation, the statistics are deleted together with the rest of RepoMeta on zot start.

laurentiuNiculae:
- before, the download count for a manifest and repo star count were lost after reload

- now we are keeping these values when we reset the repo-meta structure

**What type of PR is this?**

Closes #1274

**Automation added to e2e**:

```release-note

```


andaaron (dismissed):
Looks good.

rchincha:
"prezerved"?

andaaron (approved):

andaaron (approved):

Issue #1269, opened 2023-03-16
Cleanup: remove unused package pkg/extensions/search/digest


andaaron:
Package pkg/extensions/search/digest is no longer used for the functionality it provides.
There are some tests there which exercise the GraphQL API, meaning they use the new implementation in a different package. Those should be kept and moved to a different package.

andaaron:
Solved in https://github.com/project-zot/zot/pull/1298

PR #1298, opened 2023-03-23
chore(deps): remove unused package pkg/extensions/search/digest

nicoldr:
**What type of PR is this?**

Solves #1269 

**Automation added to e2e**:

```release-note

```


pkg/extensions/search/digest_test.go
```diff
@@ -1,28 +1,22 @@
-//go:build search
```
andaaron, line 1: You probably need to keep the build labels, as there tests would probably not work if the rest of the search code is not included in the build.

andaaron (approved):

Issue #1322, opened 2023-03-30
[Bug]: listening port chosen by kernel is printed to the console instead of being logged


andaaron:
### zot version

v2.0.0-rc3

### Describe the bug

The port selected automatically by the kernel in issue https://github.com/project-zot/zot/issues/717 is printed to the console instead of the logs. See https://github.com/project-zot/zot/pull/1260.

### To reproduce

See implementation of https://github.com/project-zot/zot/issues/717 in https://github.com/project-zot/zot/pull/770.

### Expected behavior

Instead of printing to the console, this information should be shown in the zot logs.
Preferably as part of the config json shown when zot starts. If that is not possible, it should be shown in a separate log message.

thesayyn:
It’s also printed to the console if you configured your logger properly.

andaaron:
Fixed in https://github.com/project-zot/zot/pull/1339

PR #770, opened 2022-09-06
report listening port when chosen by kernel

rchincha:
Based off of the PR by @thesayyn
https://github.com/project-zot/zot/pull/720

Signed-off-by: Ramkumar Chinchani <rchincha@cisco.com>

**What type of PR is this?**

feature

**Which issue does this PR fix**:

issue #717 

**What does this PR do / Why do we need it**:

Supports kernel-chosen random ports

**Will this break upgrades or downgrades?**

No

**Does this PR introduce any user-facing change?**:

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #720, opened 2022-08-11
feat: print the port chosen by kernel if port is unspecified

thesayyn:
**What type of PR is this?**

feature

**Which issue does this PR fix**:

#717

**What does this PR do / Why do we need it**:

**If an issue # is not available please add repro steps and logs showing the issue**:

**Testing done on this change**:
```bash 
{"level":"info","params":{"distSpecVersion":"1.0.1-dev","GoVersion":"","Commit":"","BinaryType":"","AccessControl":null,"Storage":{"Dedupe":true,"GC":true,"Commit":false,"GCDelay":3600000000000,"GCInterval":0,"RootDirectory":"/tmp/zot","StorageDriver":null,"SubPaths":null},"HTTP":{"Address":"127.0.0.1","Port":"0","AllowOrigin":"","TLS":null,"Auth":{"FailDelay":0,"HTPasswd":{"Path":""},"LDAP":null,"Bearer":null},"RawAccessControl":null,"Realm":"","Ratelimit":null},"Log":{"Level":"debug","Output":"","Audit":""},"Extensions":null},"goroutine":1,"caller":"/Users/thesayyn/Documents/zot/pkg/api/controller.go:107","time":"2022-08-11T18:06:37.75729+03:00","message":"configuration settings"}
{"level":"info","cpus":8,"max. open files":10240,"goroutine":1,"caller":"/Users/thesayyn/Documents/zot/pkg/api/controller.go:102","time":"2022-08-11T18:06:37.757374+03:00","message":"runtime params"}
{"level":"warn","goroutine":1,"caller":"/Users/thesayyn/Documents/zot/pkg/extensions/extensions_lint_disabled.go:13","time":"2022-08-11T18:06:37.757525+03:00","message":"lint extension is disabled because given zot binary doesn'tinclude this feature please build a binary that does so"}
{"level":"info","port":64680,"address":"127.0.0.1","goroutine":1,"caller":"/Users/thesayyn/Documents/zot/pkg/api/controller.go:175","time":"2022-08-11T18:06:37.80581+03:00","message":"Port is unspecified. zot will listen on an available port chosen by the kernel."}

```
**Automation added to e2e**:

**Will this break upgrades or downgrades?**
No

**Does this PR introduce any user-facing change?**:

```release-note
Print the port chosen by the kernel when port is unspecified
```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1324, opened 2023-04-03
bug: removed printing port and now the port is chosen before validation

aokirisaki:
**What type of PR is this?**

#1322 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -9,7 +9,6 @@ import (
 	"net/http"
 	"os"
 	"runtime"
-	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -48,8 +47,6 @@ type Controller struct {
 	Server          *http.Server
 	Metrics         monitoring.MetricServer
 	CveInfo         ext.CveInfo
-	// runtime params
-	chosenPort int // kernel-chosen port
 }
 
 func NewController(config *config.Config) *Controller {
@@ -111,10 +108,6 @@ func DumpRuntimeParams(log log.Logger) {
 	evt.Msg("runtime params")
 }
 
-func (c *Controller) GetPort() int {
-	return c.chosenPort
-}
-
 func (c *Controller) Run(reloadCtx context.Context) error {
 	c.StartBackgroundTasks(reloadCtx)
 
@@ -164,25 +157,6 @@ func (c *Controller) Run(reloadCtx context.Context) error {
 	listener, err := net.Listen("tcp", addr)
 	if err != nil {
 		return err
-	}
-
-	if c.Config.HTTP.Port == "0" || c.Config.HTTP.Port == "" {
-		chosenAddr, ok := listener.Addr().(*net.TCPAddr)
-		if !ok {
-			c.Log.Error().Str("port", c.Config.HTTP.Port).Msg("invalid addr type")
-
-			return errors.ErrBadType
-		}
-
-		c.chosenPort = chosenAddr.Port
-
-		c.Log.Info().Int("port", chosenAddr.Port).IPAddr("address", chosenAddr.IP).Msg(
-			"port is unspecified, listening on kernel chosen port",
-		)
-	} else {
-		chosenPort, _ := strconv.ParseInt(c.Config.HTTP.Port, 10, 64)
-
-		c.chosenPort = int(chosenPort)
 	}
 
 	if c.Config.HTTP.TLS != nil && c.Config.HTTP.TLS.Key != "" && c.Config.HTTP.TLS.Cert != "" {
--- a/pkg/cli/root.go
+++ b/pkg/cli/root.go
@@ -14,6 +14,7 @@ import (
 	glob "github.com/bmatcuk/doublestar/v4"
 	"github.com/mitchellh/mapstructure"
 	distspec "github.com/opencontainers/distribution-spec/specs-go"
+	"github.com/phayes/freeport"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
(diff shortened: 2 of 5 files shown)


pkg/cli/root.go
```diff
@@ -619,6 +620,18 @@ func LoadConfiguration(config *config.Config, configPath string) error {
 	// defaults
 	applyDefaultValues(config, viperInstance)
 
+	// set port if 0
+	if config.HTTP.Port == "" || config.HTTP.Port == "0" {
+		newPort, err := freeport.GetFreePort()
+		if err != nil {
+			log.Error().Str("port", config.HTTP.Port).Msg("invalid port")
```
andaaron, line 627: It would be better to include the original error in the log message.

pkg/cli/root.go
```diff
@@ -619,6 +620,18 @@ func LoadConfiguration(config *config.Config, configPath string) error {
 	// defaults
 	applyDefaultValues(config, viperInstance)
```
andaaron, line 621: Maybe the new logic should be inside `applyDefaultValues`.

andaaron (approved):

andaaron (dismissed):

pkg/cli/root_test.go
```diff
@@ -1018,11 +1018,56 @@ func TestVerify(t *testing.T) {
...
+		_, err = cfgfile.Write([]byte(content))
+		So(err, ShouldBeNil)
+
+		os.Args = []string{"cli_test", "serve", cfgfile.Name()}
+		go func() {
+			err = cli.NewServerRootCmd().Execute()
+			So(err, ShouldBeNil)
+		}()
+
+		_, err = ReadLogFileAndSearchString(logPath, "CVE config not provided, skipping CVE update", 90*time.Second)
```
andaaron, line 1045: Why wait for the CVE related string?
Wait for the port related log message (`HTTP port `). That is what you are testing after all.

pkg/cli/root.go
```diff
@@ -432,6 +433,18 @@ func validateAuthzPolicies(config *config.Config) error {
 	return nil
 }
 
+func setPort(config *config.Config) {
+	// set port if 0
+	if config.HTTP.Port == "" || config.HTTP.Port == "0" {
+		newPort, err := freeport.GetFreePort()
+		if err != nil {
+			log.Error().Err(err).Str("port", config.HTTP.Port).Msg("invalid port")
```
andaaron, line 441:
```suggestion
			log.Error().Err(err).Str("port", config.HTTP.Port).Msg("failed to assign HTTP port automatically")
```

pkg/cli/root_test.go
```diff
@@ -1018,11 +1018,56 @@ func TestVerify(t *testing.T) {
...
+			panic(err)
+		}
+
+		data, err := os.ReadFile(logFile.Name())
+		So(err, ShouldBeNil)
+		So(string(data), ShouldContainSubstring, "\"Port\":")
+	})
+}
+
 func TestLoadConfig(t *testing.T) {
```
andaaron, line 1056: Is there a port-related reason this test needed to change?

pkg/cli/root_test.go
aokirisaki: I made the changes

pkg/cli/root.go
aokirisaki: I changed the message

pkg/cli/root_test.go
aokirisaki: I changed it because now the port is set when the config is verified (the port is now set in roots.go), and in the old version of this test the port on the controller would remain 0 because no config was loaded, and the verification wouldn't take place, thus leaving the port with value 0.

andaaron (dismissed):

pkg/cli/root_test.go
```diff
@@ -1018,11 +1018,56 @@ func TestVerify(t *testing.T) {
...
+		_, err = cfgfile.Write([]byte(content))
+		So(err, ShouldBeNil)
+
+		os.Args = []string{"cli_test", "serve", cfgfile.Name()}
+		go func() {
+			err = cli.NewServerRootCmd().Execute()
+			So(err, ShouldBeNil)
+		}()
+
+		_, err = ReadLogFileAndSearchString(logPath, "HTTP port", 90*time.Second)
```
peusebiu, line 1045: this function return err only if it can't read the log file, the other return param is whether or not the string was found.
please check that the other param is true.

pkg/api/controller.go
```diff
@@ -166,24 +162,8 @@ func (c *Controller) Run(reloadCtx context.Context) error {
...
-		c.Log.Info().Int("port", chosenAddr.Port).IPAddr("address", chosenAddr.IP).Msg(
-			"port is unspecified, listening on kernel chosen port",
-		)
-	} else {
-		chosenPort, _ := strconv.ParseInt(c.Config.HTTP.Port, 10, 64)
-
-		c.chosenPort = int(chosenPort)
-	}
+	chosenPort, _ := strconv.ParseInt(c.Config.HTTP.Port, 10, 64)
+	c.chosenPort = int(chosenPort)
```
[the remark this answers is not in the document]
aokirisaki: I have removed it now, as it is no longer needed because I removed GetPort

andaaron (approved):

pkg/api/controller.go
```diff
@@ -111,10 +108,6 @@ func DumpRuntimeParams(log log.Logger) {
 	evt.Msg("runtime params")
 }
 
-func (c *Controller) GetPort() int {
```
rchincha, line 114: This was an external contribution, so that one could get at the auto-assigned port.

rchincha: commit f3faae0e09264dfa13dc7a61d38ad45ffb83bdbd
Author: Ramkumar Chinchani <45800463+rchincha@users.noreply.github.com>
Date:   Thu Sep 8 22:41:13 2022 -0700

    report listening port when chosen by kernel (#770)

    Based off of the PR by @thesayyn
    https://github.com/project-zot/zot/pull/720

rchincha: cc: @thesayyn

thesayyn: Yeap. I don’t think this change is acceptable.

thesayyn: https://github.com/google/go-containerregistry/pull/1586#discussion_r1128651260

thesayyn: Also: https://github.com/google/go-containerregistry/pull/1586#discussion_r1128646547

andaaron: @rchincha we update the config, and by the time the controller is started the port is already chosen.
@thesayyn ok, I see there's a chance for some other process to pick the same port in between the call to GetFreePort() and the controller tries using it.

Let's close this PR?

rchincha: But the debug log still needs to be removed?

andaaron: I merged that in a separate PR.

rchincha: Then this PR can be closed. Thx @thesayyn for the review.

PR #1328, opened 2023-04-04
fix(zap): add Content-Security-Policy to the UI routes

andaaron:
The zap scanner started to check the csp header, which is causing a warning.

We also need to ignore the rule, as both settings are read by the scanner.

Per https://w3c.github.io/webappsec-csp/#example-7bb4ce67 we can have multiple
Content-Security-Policy headers, and the most restrictive policies apply.
This rule doesn't seem to be applied by zap.

* URL: http://localhost:8080/
  * Method: `GET`
  * Parameter: `content-security-policy`
  * Attack: ``
  * Evidence: ` default-src 'none'; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com/; font-src 'self' https://fonts.googleapis.com/ https://fonts.gstatic.com/; connect-src *; img-src 'self'; manifest-src 'self'; base-uri 'self'`
* URL: http://localhost:8080/
  * Method: `GET`
  * Parameter: `Content-Security-Policy`
  * Attack: ``
  * Evidence: `default-src 'none'; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com/; font-src 'self' https://fonts.googleapis.com/ https://fonts.gstatic.com/; connect-src 'self'; img-src 'self'; manifest-src 'self'; base-uri 'self'`

### DIFF
--- a/.zap/rules.tsv
+++ b/.zap/rules.tsv
@@ -37,7 +37,7 @@
 10050	WARN	(Retrieved from Cache)
 10052	WARN	(X-ChromeLogger-Data (XCOLD) Header Information Leak)
 10054	WARN	(Cookie without SameSite Attribute)
-10055	WARN	(CSP)
+10055	IGNORE	(CSP)  We set CSP in both UI (with *) and backend (no *), unfortunately the UI needs to use * in case it is deployed separately
 10056	WARN	(X-Debug-Token Information Leak)
 10057	WARN	(Username Hash Found)
 10061	WARN	(X-AspNet-Version Response Header)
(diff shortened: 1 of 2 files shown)


.zap/rules.tsv
```diff
@@ -37,7 +37,7 @@
 10050	WARN	(Retrieved from Cache)
 10052	WARN	(X-ChromeLogger-Data (XCOLD) Header Information Leak)
 10054	WARN	(Cookie without SameSite Attribute)
-10055	WARN	(CSP)
+10055	IGNORE	(CSP)  We set CSP in both UI (with *) and backend (no *), unfortunately the UI needs to use * in case it is deployed seprately
```
rchincha, line 40: "separately", typo

andaaron:
Closed in favor of #1372

PR #1341, opened 2023-04-06
fix(routes): get blob upload session should include 'bytes=' in Range…

peusebiu:
… header

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -1001,7 +1001,7 @@ func (rh *RouteHandler) DeleteBlob(response http.ResponseWriter, request *http.R
 // @Param   name				path    string     true        "repository name"
 // @Success 202 {string} string	"accepted"
 // @Header  202 {string} Location "/v2/{name}/blobs/uploads/{session_id}"
-// @Header  202 {string} Range "0-0"
+// @Header  202 {string} Range "bytes=0-0"
 // @Failure 404 {string} string "not found"
 // @Failure 500 {string} string "internal server error"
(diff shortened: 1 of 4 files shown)


pkg/api/routes.go
```diff
@@ -1189,7 +1189,7 @@ func (rh *RouteHandler) GetBlobUpload(response http.ResponseWriter, request *htt
 	}
 
 	response.Header().Set("Location", getBlobUploadSessionLocation(request.URL, sessionID))
-	response.Header().Set("Range", fmt.Sprintf("0-%d", size-1))
+	response.Header().Set("Range", fmt.Sprintf("bytes=0-%d", size-1))
```
andaaron, line 1192: Not sure about this. We made some changes in https://github.com/project-zot/zot/pull/596/files
See that discussions.

rchincha:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range
<unit> doesn't look optional. So would just make this PR official.
Can we also look at all Range responses and fix them pls.

jdolitsky (approved):

andaaron:
The range responses were modified as part of https://github.com/project-zot/zot/pull/596/files
And there was a previous discussion on the dist spec in https://github.com/opencontainers/distribution-spec/pull/203

jdolitsky:
i think this can be closed, we just merged https://github.com/opencontainers/distribution-spec/pull/401 with missing bytes prefix

Issue #1342, opened 2023-04-06
Remove pkg/extensions/search/common and move the code to the appropriate packages


andaaron:
Further cleanup to `pkg/extensions/search/common/`:
1. Move the model (from `model.go` and some of the model in `common_test.go`) to the file `pkg/common/model.go` -> this code could be imported in the `pkg/cli` (eliminating duplicate structures), in `pkg/extensions/search` and in `pkg/test`
2. Move the `common_test.go` file into the `pkg/extensions/search` package, maybe rename it to `search_test.go`.
3. The content of `common.go` could be split, some of it could go under `pkg/common`, some of it could be removed;
- GetRootDir can be removed as it is used only it it's own tests.
- GetRepo can be removed as it is used only it it's own tests.
- GetImageDirAndTag can be moved to `pkg/common/utils.go`
- GetImageDirAndDigest can be moved to `pkg/common/utils.go`
- GetImageDirAndReference can be moved to `pkg/common/utils.go`
- GetImageLastUpdated can be moved to `pkg/common/utils.go`
- GetFixedTags can be moved to `pkg/extensions/search/cve`
- GetLatestTag can be moved to `pkg/test`
- GetRoutePrefix can be removed, and we'd make the similar function under `pkg/storage` exportable, and use that one instead.
- GetAnnotations and related annotation code could be moved under `pkg/extensions/search/convert/annotations.go`

andaaron:
Done in #1358

PR #1357, opened 2023-04-11
refactor: remove pkg/extensions/search/common and move the code to th…

nicoldr:
…e appropriate packages

**What type of PR is this?**

#1342 

**Automation added to e2e**:

```release-note

```


pkg/test/common.go
```diff
@@ -184,6 +186,14 @@ func MakeHtpasswdFile() string {
 	return MakeHtpasswdFileFromString(content)
 }
 
+func GetLatestTag(allTags []cvemodel.TagInfo) cvemodel.TagInfo {
```
andaaron, line 189: Why not move this to `pkg/test/oci-layout/oci_layout.go`? Is it used somewhere else?

pkg/storage/common.go
```diff
@@ -56,6 +57,19 @@ func GetManifestDescByReference(index ispec.Index, reference string) (ispec.Desc
 	return manifestDesc, false
 }
 
+func GetRoutePrefix(name string) string {
```
andaaron, line 60: there is also a `getRoutePrefix` function in `pkg/storage/storage_controller.go`, can you modify that one instead of adding this new function?

pkg/extensions/search/search_test.go
```diff
@@ -1798,110 +1800,87 @@ func TestExpandedRepoInfo(t *testing.T) {
 
 func TestUtilsMethod(t *testing.T) {
```
andaaron, line 1801: This should be split into multiple tests, and those be moved to the appropriate packages. For example the annotation tests should be in the `convert` package.

pkg/extensions/search/search_test.go
```diff
@@ -1798,110 +1800,87 @@ func TestExpandedRepoInfo(t *testing.T) {
...
-		license = common.GetLicenses(labels)
+		license = convert.GetLicenses(labels)
 		So(license, ShouldEqual, "zot-label-license")
 
-		vendor = common.GetVendor(labels)
+		vendor = convert.GetVendor(labels)
 		So(vendor, ShouldEqual, "zot-vendor")
 
-		routePrefix := common.GetRoutePrefix("test:latest")
+		routePrefix := storage.GetRoutePrefix("test:latest")
```
andaaron, line 1851: These checks should be in the storage package.

pkg/extensions/search/search_test.go
```diff
@@ -1798,110 +1800,87 @@ func TestExpandedRepoInfo(t *testing.T) {
...
 		So(routePrefix, ShouldEqual, "/a")
 
-		routePrefix = common.GetRoutePrefix("a/b/test:latest")
+		routePrefix = storage.GetRoutePrefix("a/b/test:latest")
 		So(routePrefix, ShouldEqual, "/a")
 
 		allTags, vulnerableTags := getTags()
 
-		latestTag := common.GetLatestTag(allTags)
+		latestTag := GetLatestTag(allTags)
```
andaaron, line 1862: This is a test to be moved in the test package.

pkg/extensions/search/search_test.go
```diff
@@ -1798,110 +1800,87 @@ func TestExpandedRepoInfo(t *testing.T) {
...
 		So(latestTag.Name, ShouldEqual, "1.0.3")
 
-		fixedTags := common.GetFixedTags(allTags, vulnerableTags)
+		fixedTags := cveinfo.GetFixedTags(allTags, vulnerableTags)
```
andaaron, line 1865: Can we move this to the folder where the cve package tests are found?

pkg/extensions/search/search_test.go
```diff
@@ -1910,16 +1889,6 @@ func TestUtilsMethod(t *testing.T) {
...
-
-		dir := common.GetRootDir("a/zot-cve-test", storeController)
-
-		So(dir, ShouldEqual, rootDir)
-
-		dir = common.GetRootDir("b/zot-cve-test", storeController)
-
-		So(dir, ShouldEqual, subRootDir)
-
 		repo, digest := common.GetImageDirAndDigest("image")
```
andaaron, line 1892: This should be a test under `pkg/common`

pkg/extensions/search/search_test.go
andaaron, line 1: There may be some structs defined in this package which could move to pkg/common/model.go, and be reused in the zli code. Please look into that as well.

pkg/test/common.go
nicoldr: It is also used in pkg/extesions/search

Issue #1295, opened 2023-03-22
Enhance the way we handle building/enabling prerequisite extensions for the UI


andaaron:
Make sure the dependencies between the UI extension and other extensions are easily managed.
1. `ui` depends on `search` for reading the data
2. `ui` depends on `mgmt` for authentication
3. `ui` may depend on others by the time it is relased

We need a way to handle:
a) build time misconfiguration (the ui is not built with all the needed extensions)
b) run time misconfiguration (the ui is enabled, but other extensions are not)

To solve a) the proposal is to update the extensions list to include missing extensions using the makefile.
To solve b) there are several options:
- document the dependencies, so the user can configure them correctly.
- throw an error if zot is misconfigured
- enable the missing extensions implicitly at runtime (may be tricky

andaaron:
Done in #1369

PR #1369, opened 2023-04-13
feat: verifying and enabling necessary extensions for ui

aokirisaki:
**What type of PR is this?**

#1295 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,9 +32,17 @@ OS ?= linux
 ARCH ?= amd64
 BENCH_OUTPUT ?= stdout
 EXTENSIONS ?= sync,search,scrub,metrics,lint,ui,mgmt,userprefs
+UI_DEPENDENCIES := search,mgmt,userprefs
 comma:= ,
+space := $(null) #
 hyphen:= -
 extended-name:=
+
+define add-extensions =
+	$(shell echo $(shell echo $(subst $(space),$(comma), $(strip $(shell echo $(subst $(comma),$(space), $(if $(findstring ui,$(EXTENSIONS)), "$(EXTENSIONS)$(comma)$(UI_DEPENDENCIES)", $(EXTENSIONS))) | tr ' ' '\n' | sort -u | tr '\n' ' '))) | cut -c2-) | tr -d " \t\n\r")
+endef
+
+BUILD_LABELS = $(strip $(call add-extensions))
 
 .PHONY: all
 all: modcheck swagger binary binary-minimal binary-debug cli bench exporter-minimal verify-config test covhtml check check-gh-actions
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -29,10 +29,18 @@ OS ?= linux
 ARCH ?= amd64
 BENCH_OUTPUT ?= stdout
 EXTENSIONS ?= sync,search,scrub,metrics,lint,ui,mgmt
+UI_DEPENDENCIES := search,mgmt
 comma:= ,
+space := $(null) #
```
andaaron, line 34:
```suggestion
space := $(null)
```

pkg/cli/root.go
```diff
@@ -312,6 +312,20 @@ func validateCacheConfig(cfg *config.Config) error {
 }
 
 func validateExtensionsConfig(cfg *config.Config) error {
+	if cfg.Extensions != nil && cfg.Extensions.UI != nil && cfg.Extensions.UI.Enable != nil && *cfg.Extensions.UI.Enable {
+		if cfg.Extensions.Mgmt == nil || !*cfg.Extensions.Mgmt.Enable {
+			log.Warn().Err(errors.ErrBadConfig).Msg("Mgmt extension is disabled")
```
andaaron, line 317: You should also mention why mgmt being disabled is not acceptable in this case.
Same for search.

Makefile
```diff
@@ -31,10 +31,18 @@ OS ?= linux
...
 hyphen:= -
 extended-name:=
 
+define unique =
```
andaaron, line 40: it seems to me like half the logic is in the `BUILD_LABELS` assignment statement and half in this function, and the reasoning for splitting it this way is not clear. `unique` seems to do a very specific thing and cannot be called for another purpose somewhere else. Maybe a name change for it should be done as well, after the logic is refactored.

Makefile
```diff
@@ -31,11 +31,19 @@ TESTDATA := $(TOP_LEVEL)/test/data
 OS ?= linux
 ARCH ?= amd64
 BENCH_OUTPUT ?= stdout
-EXTENSIONS ?= sync,search,scrub,metrics,lint,ui,mgmt,userprefs
+EXTENSIONS ?= sync,search,scrub,metrics,lint,ui,mgmt
+UI_DEPENDENCIES := search,mgmt
```
andaaron, line 35:
```suggestion
UI_DEPENDENCIES := search,mgmt,userprefs
```

pkg/cli/root.go
```diff
@@ -314,6 +314,20 @@ func validateCacheConfig(cfg *config.Config) error {
...
 
 func validateExtensionsConfig(cfg *config.Config) error {
+	if cfg.Extensions != nil && cfg.Extensions.UI != nil && cfg.Extensions.UI.Enable != nil && *cfg.Extensions.UI.Enable {
+		if cfg.Extensions.Mgmt == nil || !*cfg.Extensions.Mgmt.Enable {
+			log.Warn().Err(errors.ErrBadConfig).Msg("UI functionality can't be used without mgmt extension.")
+
+			return errors.ErrBadConfig
+		}
+
+		if cfg.Extensions.Search == nil || !*cfg.Extensions.Search.Enable {
```
andaaron, line 324: Will need a similar checks for the `userprefs` extension.

examples/config-ui.json
```diff
@@ -18,6 +18,9 @@
     },
     "ui": {
       "enable": true
+    },
+    "mgmt": {
+      "enable": true
```
andaaron, line 23: Add userprefs as well.

rchincha (approved):
lgtm

PR #1415, opened 2023-05-03
Update referrers response

### LINKED ISSUE #1410: [Feat]: referrers API MUST respond with the response header `OCI-Subject: true`
https://github.com/opencontainers/distribution-spec/commit/16cc054de70442ea08e1aa269cee1c6cda5e7c96

```
When processing a request for an image or artifact manifiest with the `subject` field, a registry implementation that supports the [referrers API](#listing-referrers) MUST respond with the response header `OCI-Subject-Processed: true` to indicate to the client that the registry understands the request's `subject`.
```

laurentiuNiculae:
**What type of PR is this?**

Closes #1410

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -606,7 +606,7 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
 		return
 	}
 
-	digest, err := imgStore.PutImageManifest(name, reference, mediaType, body)
+	digest, subjectDigest, err := imgStore.PutImageManifest(name, reference, mediaType, body)
 	if err != nil {
 		if errors.Is(err, zerr.ErrRepoNotFound) { //nolint:gocritic // errorslint conflicts with gocritic:IfElseChain
 			WriteJSON(response, http.StatusNotFound,
@@ -654,6 +654,10 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
 
 			return
 		}
+	}
+
+	if subjectDigest.String() != "" {
+		response.Header().Set(constants.SubjectDigestKey, subjectDigest.String())
 	}
 
 	response.Header().Set("Location", fmt.Sprintf("/v2/%s/manifests/%s", name, digest))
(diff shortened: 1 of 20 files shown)


pkg/storage/local/local.go
```diff
@@ -499,11 +499,11 @@ func (is *ImageStoreLocal) PutImageManifest(repo, reference, mediaType string, /
 
 	updateIndex, oldDgst, err := storage.CheckIfIndexNeedsUpdate(&index, &desc, is.log)
 	if err != nil {
-		return "", err
+		return "", "", err
 	}
 
 	if !updateIndex {
-		return desc.Digest, nil
+		return desc.Digest, "", nil
```
andaaron, line 506: Shouldn't the subject digest be returned also in this case?

pkg/storage/s3/s3.go
```diff
@@ -404,11 +404,11 @@ func (is *ObjectStorage) PutImageManifest(repo, reference, mediaType string, //n
 
 	updateIndex, oldDgst, err := storage.CheckIfIndexNeedsUpdate(&index, &desc, is.log)
 	if err != nil {
-		return "", err
+		return "", "", err
 	}
 
 	if !updateIndex {
-		return desc.Digest, nil
+		return desc.Digest, "", nil
```
andaaron, line 411: Shouldn't this also return the subject digest?

andaaron (approved):

pkg/api/routes.go
```diff
@@ -656,6 +656,10 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
 		}
 	}
 
+	if subjectDigest.String() != "" {
+		response.Header().Set(constants.SubjectDigestKey, subjectDigest.String())
```
rchincha, line 660: Where is the "OCI-Subject-Processed" header set?

laurentiuNiculae: In the latest version of the [distribution spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pushing-manifests-with-subject) there is no "OCI-Subject-Processed" header. It was changed to "OCI-Subject".

rchincha (approved):
lgtm

PR #1439, opened 2023-05-11
fix(sync): fixed way of updating repodb when syncing a signature

laurentiuNiculae:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


pkg/meta/repodb/storage_parsing.go
```diff
@@ -118,15 +109,18 @@ func ParseRepo(repo string, repoDB RepoDB, storeController storage.StoreControll
...
-			// We'll ignore signatures now because the order in which the signed image and signature are added into
-			// the DB matters. First we add the normal images then the signatures
-			signaturesFound = append(signaturesFound, foundSignatureData{
-				repo:                 repo,
-				tag:                  tag,
-				signatureType:        signatureType,
-				signedManifestDigest: signedManifestDigest.String(),
-				signatureDigest:      digest.String(),
-			})
+			err := repoDB.AddManifestSignature(repo, digest,
```
andaaron, line 112: Does it matter if the signature is added before or after the image it is signing? The old code added the signatures after.

pkg/meta/repodb/boltdb-wrapper/boltdb_wrapper.go
```diff
@@ -747,8 +733,33 @@ func (bdw *DBWrapper) AddManifestSignature(repo string, signedManifestDigest god
 		buck := tx.Bucket([]byte(bolt.RepoMetadataBucket))
 
 		repoMetaBlob := buck.Get([]byte(repo))
-		if repoMetaBlob == nil {
-			return zerr.ErrManifestMetaNotFound
+
+		if len(repoMetaBlob) == 0 {
+			var err error
+			// create a new object
+			repoMeta := repodb.RepoMetadata{
```
andaaron, line 740: So it doesn't matter if the image or the signature is added first? Does this change take care of my concern in the other comment?

pkg/meta/repodb/storage_parsing.go
laurentiuNiculae: With this PR the order doesn't matter anymore.

pkg/meta/repodb/boltdb-wrapper/boltdb_wrapper.go
laurentiuNiculae: Yes, this change takes care of the case when the signature is added before the signe manifest

rchincha (approved):
lgtm

rchincha:
There are conflicts

andaaron (approved):

PR #1469, opened 2023-05-23
 feat(mgmt): update mgmt handler to return info about anonymous policy

alexstan12:
**What type of PR is this?**

featyre

adds more info on the mgmt api call response regarding accessControl, specifically on anonymous policy

**Automation added to e2e**:

```release-note

```


peusebiu:
let me test it manually before merging

andaaron:
We decided not to implement this, just use the `/v2/` endpoint to determine if the user has access to at least one repo.

PR #1493, opened 2023-05-31
chore: fix security alerts

rchincha:
https://github.com/project-zot/zot/security/code-scanning/4293

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -10,10 +10,7 @@ on:
       - published
 name: build-test
 
-
-permissions:
-  contents: read
-  packages: write
+permissions: read-all
 
 jobs:
   build-test:
@@ -355,6 +352,9 @@ jobs:
     if: github.event_name == 'release' && github.event.action== 'published'
     needs: push-image
     name: Update Helm Chart
+    permissions:
+      contents: write
+      packages: write
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/commit-msg.yaml
+++ b/.github/workflows/commit-msg.yaml
@@ -10,6 +10,8 @@ on:
     branches:
       - main
 
+permissions: read-all
+
 jobs:
   check-commit-message-style:
     name: Check commit message style
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -23,7 +23,7 @@ jobs:
       - name: Install dependencies
         run: |
           cd $GITHUB_WORKSPACE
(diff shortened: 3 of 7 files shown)


go.mod
```diff
@@ -461,3 +461,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/aws/aws-sdk-go => github.com/aws/aws-sdk-go v1.44.271
```
andaaron, line 465: Why is this needed?

rchincha: https://github.com/project-zot/zot/security/code-scanning/4293

That is not enough because go.sum still has an older version via an indirect dependency. This forces to move everywhere.

rchincha: IINM, `require` specifies a direct dependency but does not guarantee that all indirect ones respect that also. You can try it yourself. Vuln scanners look at go.sum also btw.

andaaron: Actually this go.mod is already up to date on main. Why would we be vulnerable?

rchincha: https://github.com/project-zot/zot/blob/main/go.sum#L400

andaaron: That is a list of possible versions to use, not the version to be included in the binary at build time. v1.44.271 is there, since we have another higher version - that one should be used https://github.com/project-zot/zot/blob/main/go.sum#L403.

rchincha: $ go mod graph  | grep "github.com/aws/aws-sdk-go@v1.20.6"
github.com/apex/log@v1.9.0 github.com/aws/aws-sdk-go@v1.20.6

$ go mod graph | grep "github.com/apex/log@v1.9.0"
github.com/opencontainers/umoci@v0.4.8-0.20210922062158-e60a0cc726e6 github.com/apex/log@v1.9.0

^ here is the actual dep graph

This is the whole point of go.sum

andaaron: Yes. but only a single version of the library is used at compile time, the maximum of all the minim versions specified as dependencies.

andaaron: From the link above
```
MVS starts at the main modules (special vertices in the graph that have no version) and traverses the graph, tracking the highest required version of each module. At the end of the traversal, the highest required versions comprise the build list: they are the minimum versions that satisfy all requirements.
```

rchincha: https://go.dev/ref/mod#minimal-version-selection

yes, but this or we find a way around the vuln. scanner

andaaron (approved):

PR #1501, opened 2023-06-08
chore: fix dependabot alerts

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


rchincha:
https://github.com/opencontainers/image-spec/commit/25fc5534d6a7a206f993ba03d63199eea22d7136#diff-f644ae378dd9d27890b86e85b2427c53f10853e498ee3aa38c794b28b879c9d8

^ has been merged

go.mod
```diff
@@ -473,3 +483,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/google/go-containerregistry => github.com/google/go-containerregistry v0.14.1-0.20230409045903-ed5c185df419
```
andaaron, line 487: Why upgrade this package and then use a replace here for an older version?

rchincha: because trivy and cosign deps are for different versions and the newer version breaks.

andaaron: Maybe it's better to use top of main for cosign?

andaaron (approved):

go.mod
andaaron: Actually I think we're ok if everything compiles successfully and the tests pass

PR #1518, opened 2023-06-14
ci: fix go build caching

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/.github/workflows/cluster.yaml
+++ b/.github/workflows/cluster.yaml
@@ -56,15 +56,19 @@ jobs:
                 minio/minio:edge-cicd server /data
       - name: Install py minio
         run: pip3 install minio
-        
       - name: Wait for minio to come up
         run: |
-          curl --connect-timeout 5 \
-            --max-time 10 \
-            --retry 12 \
-            --retry-max-time 120 \
-            'http://localhost:9000/minio/health/live'
-          
+          while true; do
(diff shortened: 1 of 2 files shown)


andaaron:
Maybe the caches are too large because we have too many older 3rd party versions we no longer need. Maybe we should clear the cache entries instead? Removing it entirely would increase the build time.

.github/workflows/cluster.yaml
```diff
@@ -56,15 +56,19 @@ jobs:
...
-        
       - name: Wait for minio to come up
         run: |
-          curl --connect-timeout 5 \
-            --max-time 10 \
-            --retry 12 \
-            --retry-max-time 120 \
-            'http://localhost:9000/minio/health/live'
-          
+          while true; do
```
andaaron, line 61: Why is this needed?

andaaron:
The go build amd-64 cache size dropped from ~3 GB back to ~1.5 GB after the previous cache entries were deleted (an older version of the cache is used as starting point if an exact match is not found).
Something may have changed in between Go versions in order to cause this.

Let's close this PR.

.github/workflows/cluster.yaml
andaaron: Sebi fixed this on top of main earlier today.

PR #1528, opened 2023-06-16
fix(ui): fix login screen when anonymous and auth are both enabled

andaaron:
Update to the latest zui version


rchincha (approved):
lgtm

PR #1530, opened 2023-06-19
added swagger doc generation for mgmt and userprefs

aokirisaki:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/extension_mgmt.go
+++ b/pkg/extensions/extension_mgmt.go
@@ -70,6 +70,14 @@ type mgmt struct {
 	log    log.Logger
 }
 
+// mgmtHandler godoc
+// @Summary Get current server configuration
+// @Description Get current server configuration
+// @Router 	/v2/_zot/ext/mgmt [get]
+// @Accept  json
+// @Produce json
+// @Success 200 {object} 	extensions.StrippedConfig
+// @Failure 500 {string} 	string 				"internal server error".
 func (mgmt *mgmt) handler() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		sanitizedConfig := mgmt.config.Sanitize()
--- a/pkg/extensions/extension_userprefs.go
+++ b/pkg/extensions/extension_userprefs.go
@@ -43,6 +43,19 @@ func SetupUserPreferencesRoutes(config *config.Config, router *mux.Router, store
 	}
 }
 
+// ListTags godoc
+// @Summary Add bookmarks/stars info
+// @Description Add bookmarks/stars info
+// @Router 	/v2/_zot/ext/userprefs [put]
+// @Accept  json
+// @Produce json
+// @Param 	action	 	 query 	 string 		true	"specify action" Enums("toggleBookmark", "toggleStar")
+// @Param   repo     	 query    string			true	"repository name"
+// @Success 200 {string}	string				"ok"
+// @Failure 404 {string} 	string 				"not found"
+// @Failure 403 {string} 	string 				"forbidden"
+// @Failure 500 {string} 	string 				"internal server error"
+// @Failure 400 {string} 	string 				"bad request".
 func HandleUserPrefs(repoDB repodb.RepoDB, log log.Logger) func(w http.ResponseWriter, r *http.Request) {
 	return func(rsp http.ResponseWriter, req *http.Request) {
 		if !queryHasParams(req.URL.Query(), []string{"action"}) {
--- a/swagger/docs.go
+++ b/swagger/docs.go
@@ -141,6 +141,96 @@ const docTemplate = `{
                 }
             }
         },
+        "/v2/_zot/ext/mgmt": {
+            "get": {
+                "description": "Get current server configuration",
+                "consumes": [
+                    "application/json"
+                ],
+                "produces": [
+                    "application/json"
+                ],
+                "summary": "Get current server configuration",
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -209,7 +209,7 @@ check: ./golangcilint.yaml $(GOLINTER)
 
 swagger/docs.go: 
 	swag -v || go install github.com/swaggo/swag/cmd/swag@$(SWAGGER_VERSION)
-	swag init --parseDependency -o swagger -g pkg/api/routes.go -q
+	swag init --parseDependency -o swagger -g pkg/api/routes.go -g pkg/extensions/extension_mgmt.go -q
```
andaaron, line 212: Does this work without adding pkg/extensions/extension_userprefs.go here?

pkg/extensions/extension_userprefs.go
```diff
@@ -43,6 +43,18 @@ func SetupUserPreferencesRoutes(config *config.Config, router *mux.Router, store
 	}
 }
 
+// ListTags godoc
+// @Summary Add bookmarks/ stars info
+// @Description Add bookmarks/ stars info
+// @Router 	/v2/_zot/ext/userprefs [put]
+// @Accept  json
+// @Produce json
+// @Param 	action	 	 query 	 string 		true	"specify action"
```
andaaron, line 52: Is there a way we can mention the supported values for `action`?
Also there is a `repo` Parameter for the existing `action`s, we should also mention that

swagger/docs.go
```diff
@@ -1030,12 +1168,12 @@ const docTemplate = `{
 
 // SwaggerInfo holds exported Swagger Info so clients can modify it
 var SwaggerInfo = &swag.Spec{
-	Version:          "v1.1.0-dev",
```
andaaron, line 1033: These should not be changed. They are read from routes.go from what I can tell. Why are they gone now?

Makefile
aokirisaki: I tested and it works without it so I removed it for mgmt too

pkg/extensions/extension_userprefs.go
aokirisaki: I made the changes

swagger/docs.go
aokirisaki: I regenerated the docs and now they appear, I think I accidentally deleted the line

pkg/extensions/extension_userprefs.go
```diff
@@ -43,6 +43,19 @@ func SetupUserPreferencesRoutes(config *config.Config, router *mux.Router, store
...
 }
 
+// ListTags godoc
+// @Summary Add bookmarks/ stars info
+// @Description Add bookmarks/ stars info
+// @Router 	/v2/_zot/ext/userprefs [put]
+// @Accept  json
+// @Produce json
+// @Param 	action	 	 query 	 string 		true	"specify action" Enums(ToggleRepoBookmarkAction, ToggleRepoStarAction)
+// @Param   repo     	 path    string			true	"repository name"
```
andaaron, line 53:
```suggestion
// @Param   repo     	 query    string			true	"repository name"
```

Makefile
andaaron: Thanks

swagger/docs.go
andaaron: I think it has something to do with sending multiple files as parameters to the swagger command in the makefile. But since it's working now, let's mark this as resolved.

pkg/extensions/extension_mgmt.go
```diff
@@ -70,6 +70,15 @@ type mgmt struct {
...
 }
 
+// mgmtHandler godoc
+// @Summary Get current server configuration
+// @Description Get current server configuration
+// @Router 	/v2/_zot/ext/mgmt [get]
+// @Accept  json
+// @Produce json
+// @Success 200 {object} 	extensions.StrippedConfig
+// @Failure 404 {string} 	string 				"not found"
```
peusebiu, line 80: I think we should remove this 404 and 400 responses, mgmt shouldn't respond with them in any scenario.

also add 500 internal server error

200 and 500 should be the only responses

aokirisaki: I made the changes

andaaron (approved):

swagger/docs.go
```diff
@@ -141,6 +141,96 @@ const docTemplate = `{
...
+                        "schema": {
+                            "type": "string"
+                        }
+                    }
+                }
+            }
+        },
+        "/v2/_zot/ext/userprefs": {
+            "put": {
+                "description": "Add bookmarks/ stars info",
```
laurentiuNiculae, line 172: We can write it like this I think:
```suggestion
                "description": "Add bookmarks/stars info",
```

peusebiu (dismissed):
lgtm

swagger/docs.go
```diff
@@ -141,6 +141,96 @@ const docTemplate = `{
...
+                    "application/json"
+                ],
+                "produces": [
+                    "application/json"
+                ],
+                "summary": "Add bookmarks/ stars info",
+                "parameters": [
+                    {
+                        "enum": [
+                            "ToggleRepoBookmarkAction",
```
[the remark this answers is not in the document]
aokirisaki: I made the changes

andaaron (approved):

laurentiuNiculae (approved):

PR #1541, opened 2023-06-21
fix(authz): assign identity to authz context in tls mutual authentica…

peusebiu:
…tion

this causes a bug in extensions by not having the identity for the authenticated user and couldn't apply his permissions, just the default ones.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


alexstan12 (approved):
lgtm

andaaron (approved):

PR #1548, opened 2023-06-22
fix(CVE): attempt to scan now returns early with an error if trivyDB metadata json is missing

andaaron:
Without the changes to check if the CVE DB is already downloaded, relying on the trivy library itself for error handling had some issues, as sometimes the files it opens (the cache DB, different from the CVE DB - called fanal.db) were not closed.
This was causing issues on subsequent calls, even if the CVE DB was downloaded in the meantime.

zli will now make repeated attempts to search in case of this error. This will take care of the flaky CVE test in the ecosystem tools.

### DIFF
--- a/pkg/cli/cve_cmd.go
+++ b/pkg/cli/cve_cmd.go
@@ -7,11 +7,17 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"strings"
+	"time"
 
 	"github.com/briandowns/spinner"
 	"github.com/spf13/cobra"
 
 	zotErrors "zotregistry.io/zot/errors"
+)
+
+const (
+	cveDBRetryInterval = 3
 )
 
 func NewCveCommand(searchService SearchService) *cobra.Command {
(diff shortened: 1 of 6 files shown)


pkg/extensions/search/cve/trivy/scanner.go
```diff
@@ -265,7 +266,17 @@ func (scanner Scanner) ScanImage(image string) (map[string]cvemodel.CVE, error)
 
 	scanner.dbLock.Lock()
```
rchincha, line 267: Why no just add a defer unlock right here? and remove the other lock/unlock pairs below

andaaron: I thought other scanning goroutines should not be blocked while waiting for unrelated reports to be processed.

andaaron:
Still looking into this. Do not merge.

pkg/extensions/search/cve/trivy/scanner.go
andaaron: I no longer modify this function.

andaaron:
This is now ready for review/merge.

peusebiu (dismissed):
lgtm

pkg/cli/cve_cmd_test.go
```diff
@@ -287,6 +288,48 @@ func TestSearchCVECmd(t *testing.T) {
...
+		cveCmd := NewCveCommand(&mockService)
+		buff := bytes.NewBufferString("")
+		cveCmd.SetOut(buff)
+		cveCmd.SetErr(buff)
+		cveCmd.SetArgs(args)
+		err := cveCmd.Execute()
+		space := regexp.MustCompile(`\s+`)
+		str := space.ReplaceAllString(buff.String(), " ")
+		t.Logf("Output: %s", str)
+		So(strings.TrimSpace(str), ShouldContainSubstring,
```
laurentiuNiculae, line 305: The function that prints this is defined:
```go
fmt.Fprintln(searchConfig.resultWriter, "[warning] CVE DB is not ready [", i, "] - retry in 3 seconds")
```
I was thinking we can define a constant "RetryPeriod" that it would be 3 (seconds) and add to this test the remaining part of the printed string. This would apply to the rest of the tests too

```suggestion
		So(strings.TrimSpace(str), ShouldContainSubstring, 
		"[warning] CVE DB is not ready [ 0 ] - retry in " + strconv.Itoa(RetryPeriod) + " seconds")
```

If you think it's worth adding it

pkg/cli/cve_cmd.go
```diff
@@ -148,13 +150,40 @@ func searchCve(searchConfig searchConfig) error {
...
+			}
+
+			if i+1 >= maxAttempts {
+				// searcher matches search config but there are errors
+				// this is the last attempt and we cannot retry
 				return err
 			}
 
-			return nil
+			if strings.Contains(err.Error(), zotErrors.ErrCVEDBNotFound.Error()) {
```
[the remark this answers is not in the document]
andaaron: Not sure unwrapping the error would work client side if the wrapping was done server side. It's not the same object, I suspect that is the reason.
I tried in the previous implementation when I was wrapping 2 errors, and it didn't work. I left the Contains() in case we want to add a new string explaining which store / path has the issue, but sending that information to zli seems like a potential security issue to me.

rchincha (approved):
lgtm

PR #1559, opened 2023-06-26
fix: don't allow blobs to be deleted if in use

peusebiu:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/.github/workflows/oci-conformance-action.yml
+++ b/.github/workflows/oci-conformance-action.yml
@@ -34,8 +34,19 @@ jobs:
           RUNNER_TRACKING_ID="" && ./bin/zot-linux-amd64 serve examples/config-conformance.json &
           IP=`hostname -I | awk '{print $1}'`
           echo "SERVER_URL=http://${IP}:8080" >> $GITHUB_ENV
-    - name: Run OCI Distribution Spec conformance tests
-      uses: opencontainers/distribution-spec@main
+    - uses: actions/checkout@v3
+      with:
+        # TODO: change to upstream once the foloowing PR is merged:
+        # https://github.com/opencontainers/distribution-spec/pull/436
+        repository: sudo-bmitch/distribution-spec
+        ref: pr-conformance-index-subject
+        path: distribution-spec
+    - name: build conformance binary from main
+      run: |
+        (cd distribution-spec/ && make conformance-binary)
+        mv distribution-spec/output/conformance.test .
+        rm -rf distribution-spec/
+    - name: run conformance
       env:
         OCI_ROOT_URL: ${{ env.SERVER_URL }}
         OCI_NAMESPACE: oci-conformance/distribution-test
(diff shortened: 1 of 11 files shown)


pkg/api/routes.go
```diff
@@ -1052,6 +1052,10 @@ func (rh *RouteHandler) DeleteBlob(response http.ResponseWriter, request *http.R
 			zcommon.WriteJSON(response,
 				http.StatusNotFound,
 				apiErr.NewErrorList(apiErr.NewError(apiErr.BLOB_UNKNOWN, map[string]string{".String()": digest.String()})))
+		} else if errors.Is(err, zerr.ErrBlobBusy) {
+			zcommon.WriteJSON(response,
+				http.StatusConflict,
+				apiErr.NewErrorList(apiErr.NewError(apiErr.BLOB_UNKNOWN, map[string]string{".String()": digest.String()})))
```
andaaron, line 1058: Add a new error under pkg/api/errors?

pkg/storage/common/common.go
```diff
@@ -457,6 +457,81 @@ func PruneImageManifestsFromIndex(imgStore storageTypes.ImageStore, repo string,
...
+
+		if found {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func BlobInUse(imgStore storageTypes.ImageStore, repo string,
```
andaaron, line 519: Function name should contain a verb.
Same applies for the other functions defined in this file.

andaaron (commented):
What is the blob is in 'use' meaning it is tagged? Can we delete it? What is the spec saying about this case?

pkg/api/routes.go
peusebiu: can't do it, api errors should be compliant with dist-spec.

peusebiu:
> What is the blob is in 'use' meaning it is tagged? Can we delete it? What is the spec saying about this case?

they talked about this here:
https://github.com/opencontainers/distribution-spec/issues/134

pkg/api/routes.go
andaaron: I think we should use 405 Method Not Allowed, in the spec they mention 405 as a valid response for failure to delete.
I think we should use the `DENIED` from the dist-spec.

```

```

pkg/api/routes.go
```diff
@@ -1054,6 +1054,10 @@ func (rh *RouteHandler) DeleteBlob(response http.ResponseWriter, request *http.R
 			zcommon.WriteJSON(response,
 				http.StatusNotFound,
 				apiErr.NewErrorList(apiErr.NewError(apiErr.BLOB_UNKNOWN, map[string]string{".String()": digest.String()})))
+		} else if errors.Is(err, zerr.ErrBlobReferenced) {
+			zcommon.WriteJSON(response,
+				http.StatusConflict,
```
andaaron, line 1059: This is not a valid response code per spec.

rchincha:
https://github.com/project-zot/zot/pull/1526

pkg/api/routes.go
rchincha: 405 status code?

andaaron: yes

andaaron (dismissed):
This is good from my side. Let's wait for approval from @rchincha

examples/config-conformance.json
```diff
@@ -2,8 +2,8 @@
   "distSpecVersion": "1.1.0-dev",
   "storage": {
     "rootDirectory": "/tmp/zot",
-    "gc": false,
-    "dedupe": false
+    "gc": true,
```
rchincha, line 5: Revert this file. I will change this in a separate PR.

.github/workflows/oci-conformance-action.yml
```diff
@@ -34,8 +34,16 @@ jobs:
           RUNNER_TRACKING_ID="" && ./bin/zot-linux-amd64 serve examples/config-conformance.json &
           IP=`hostname -I | awk '{print $1}'`
           echo "SERVER_URL=http://${IP}:8080" >> $GITHUB_ENV
-    - name: Run OCI Distribution Spec conformance tests
-      uses: opencontainers/distribution-spec@main
+    - uses: actions/checkout@v3
+      with:
+        repository: opencontainers/distribution-spec
```
rchincha, line 39: Point to Brandon's tree like in Josh's PR.

rchincha: I will fix in my next PR.

andaaron (approved):

PR #1562, opened 2023-06-27
feat(sync): sync can include self url in registry.URLs

### LINKED ISSUE #1335: [Feat]: sync "registry URLs" can include self

Mainly to achieve a simple (not necessarily consistent) clustering solution so that the configuration becomes simpler.
Let the registry URLs for sync config contain the self address, which should be ignored. So when we have a cluster with many replicas, the same configuration can be applied to all replicas and yet if we push an image to one replica, the other replicas can "sync" from that one.

_No response_

_No response_

### Additional context

_No response_

peusebiu:
sync now ignores self referencing urls, this will help in clustering mode where we can have the same config for multiple zots

closes #1335

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/extension_sync.go
+++ b/pkg/extensions/extension_sync.go
@@ -4,7 +4,13 @@
 package extensions
 
 import (
+	"net"
+	"net/url"
+	"strings"
+
+	zerr "zotregistry.io/zot/errors"
 	"zotregistry.io/zot/pkg/api/config"
+	syncconf "zotregistry.io/zot/pkg/extensions/config/sync"
 	"zotregistry.io/zot/pkg/extensions/sync"
 	"zotregistry.io/zot/pkg/log"
 	"zotregistry.io/zot/pkg/meta/repodb"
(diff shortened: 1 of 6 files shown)


pkg/extensions/extension_sync.go
```diff
@@ -19,6 +23,13 @@ func EnableSyncExtension(config *config.Config, repoDB repodb.RepoDB,
 		onDemand := sync.NewOnDemand(log)
 
 		for _, registryConfig := range config.Extensions.Sync.Registries {
+			// in order to help configure sync in cluster mode, ignore URLs pointing to this server
+			selfAddress := net.JoinHostPort(config.HTTP.Address, config.HTTP.Port)
```
andaaron, line 27: What if zot is configured to listen in 0.0.0.0? Or in one place of the config it is listed with the hostname and in the other with the IP?

peusebiu: yeah, I'm not sure how to solve this...

andaaron (changes requested):
See my other comment.

pkg/extensions/extension_sync.go
```diff
@@ -19,6 +25,19 @@ func EnableSyncExtension(config *config.Config, repoDB repodb.RepoDB,
...
 		for _, registryConfig := range config.Extensions.Sync.Registries {
+			registryConfig := registryConfig
+			if len(registryConfig.URLs) > 1 {
+				if err := removeSelfURLs(config, &registryConfig, log); err != nil {
+					return nil, err
+				}
+			}
+
+			if len(registryConfig.URLs) == 0 {
+				log.Error().Err(zerr.ErrSyncNoURLsLeft).Msg("unable to start sync extensions")
```
andaaron, line 36:
```suggestion
				log.Error().Err(zerr.ErrSyncNoURLsLeft).Msg("unable to start sync extension")
```

andaaron (dismissed):

rchincha:
Can we add a bats test for this pls? This is user-facing, so we need to be able to point to working examples.

rchincha:
Also, what happens in the k8s case. We can use kind for this.

peusebiu:
> Also, what happens in the k8s case. We can use kind for this.

added bats test for clustering, but for this one can you give more details please?

andaaron:
Let's merge after rebase

Issue #1486, opened 2023-05-31
Refactor task generator to support a configurable worker count and rename the Generator interface


andaaron:
Refactor the task scheduler:
- use a default number of workers of `runtime.NumCPU()*4`
- make the number of workers configurable by adding a `scheduler` key in the zot config, and another key `numWorkers` underneath it
- rename the `Generator` interface to `TaskGenerator` and the `GenerateTask()` method to `Next()`

PR #1563, opened 2023-06-28
fix: changing default numWorkers, making it customizable and refactor…

aokirisaki:
…ing scheduler

**What type of PR is this?**

#1486 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/extension_search_test.go
+++ b/pkg/extensions/extension_search_test.go
@@ -5,6 +5,7 @@ package extensions_test
 
 import (
 	"context"
+	"io"
 	"os"
 	"testing"
 	"time"
@@ -12,6 +13,7 @@ import (
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
 	. "github.com/smartystreets/goconvey/convey"
 
+	"zotregistry.io/zot/pkg/api/config"
 	. "zotregistry.io/zot/pkg/extensions"
 	cveinfo "zotregistry.io/zot/pkg/extensions/search/cve"
 	"zotregistry.io/zot/pkg/log"
@@ -30,8 +32,13 @@ func TestTrivyDBGenerator(t *testing.T) {
 
 		defer os.Remove(logFile.Name()) // clean up
 
-		logger := log.NewLogger("debug", logFile.Name())
-		sch := scheduler.NewScheduler(logger)
+		logger := log.NewLogger("debug", logPath)
+		writers := io.MultiWriter(os.Stdout, logFile)
+		logger.Logger = logger.Output(writers)
+
+		cfg := config.New()
+		cfg.Scheduler = &config.SchedulerConfig{NumWorkers: 3}
+		sch := scheduler.NewScheduler(cfg, logger)
 
 		repoDB := &mocks.RepoDBMock{
 			GetRepoMetaFn: func(repo string) (repodb.RepoMetadata, error) {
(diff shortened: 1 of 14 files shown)


pkg/extensions/extension_search_test.go
```diff
@@ -56,6 +57,9 @@ func TestTrivyDBGenerator(t *testing.T) {
 		sch.SubmitGenerator(generator, 12000*time.Millisecond, scheduler.HighPriority)
 
 		ctx, cancel := context.WithCancel(context.Background())
+		cfg := config.New()
```
andaaron, line 60: Is this config object used?

pkg/scheduler/scheduler.go
```diff
@@ -68,13 +70,15 @@ type Scheduler struct {
...
+	NumWorkers        int
 }
 
-func NewScheduler(logC log.Logger) *Scheduler {
+func NewScheduler(cfg *config.Config, logC log.Logger) *Scheduler {
 	chLow := make(chan Task, rateLimiterScheduler)
 	chMedium := make(chan Task, rateLimiterScheduler)
 	chHigh := make(chan Task, rateLimiterScheduler)
 	generatorPQ := make(generatorsPriorityQueue, 0)
+	NumWorkers := getNumWorkers(cfg)
```
andaaron, line 81:
```suggestion
	numWorkers := getNumWorkers(cfg)
```

pkg/extensions/extension_search_test.go
```diff
@@ -31,7 +32,7 @@ func TestTrivyDBGenerator(t *testing.T) {
 		defer os.Remove(logFile.Name()) // clean up
 
 		logger := log.NewLogger("debug", logFile.Name())
```
andaaron, line 34: Please use a logger which sends messages to both stdout and the logfile?
````
writers := io.MultiWriter(os.Stdout, logFile)
logger.Output(writers)
```
There are some tests using similar logic in search_test.go.
It's hard to figure out why the test is failing if the logs are not shown in the console.

andaaron (commented):
There's a test case failure and it's hard to troubleshoot without seeing the logs.

andaaron (approved):
LGTM

pkg/extensions/extension_search_test.go
```diff
@@ -53,16 +60,17 @@ func TestTrivyDBGenerator(t *testing.T) {
 		cveInfo := cveinfo.NewCVEInfo(storeController, repoDB, "ghcr.io/project-zot/trivy-db", "", logger)
 		generator := NewTrivyTaskGenerator(time.Minute, cveInfo, logger)
 
-		sch.SubmitGenerator(generator, 12000*time.Millisecond, scheduler.HighPriority)
+		sch.SubmitGenerator(generator, 1200*time.Millisecond, scheduler.HighPriority)
```
andaaron, line 63: This seems too often. Let's not download the DB every 1.2 seconds? It doesn't even have enough time to complete the download before the new one starts.

andaaron (dismissed):

andaaron (approved):

rchincha (commented):
This looks good. However, is there is a task scheduler stress test we can add?

andaaron:
> This looks good. However, is there is a task scheduler stress test we can add?

Lets track this separately. The initial task is complete.

PR #1566, opened 2023-06-29
fix(storage): do not open/download blobs when validating manifests

peusebiu:
when pushing manifests, zot will validate blobs (layers + config blob) are present in repo, currently it opens(in case of filesystem storage) or download( in case of cloud storage) each blob.

fixed that by adding a new method ImageStore.CheckBlobPresence() on storage to check blobs presence without checking the cache like ImageStore.CheckBlob() method does.

**What type of PR is this?**
bug

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/local/local.go
+++ b/pkg/storage/local/local.go
@@ -1074,15 +1074,17 @@ func (is *ImageStoreLocal) BlobPath(repo string, digest godigest.Digest) string 
 	return path.Join(is.rootDir, repo, "blobs", digest.Algorithm().String(), digest.Encoded())
 }
 
-// CheckBlob verifies a blob and returns true if the blob is correct.
+/*
+	CheckBlob verifies a blob and returns true if the blob is correct
+
+If the blob is not found but it's found in cache then it will be copied over.
+*/
 func (is *ImageStoreLocal) CheckBlob(repo string, digest godigest.Digest) (bool, int64, error) {
 	var lockLatency time.Time
 
 	if err := digest.Validate(); err != nil {
 		return false, -1, err
 	}
-
-	blobPath := is.BlobPath(repo, digest)
 
 	if is.dedupe && fmt.Sprintf("%v", is.cache) != fmt.Sprintf("%v", nil) {
 		is.Lock(&lockLatency)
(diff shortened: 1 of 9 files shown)


pkg/storage/local/local.go
```diff
@@ -1074,32 +1074,31 @@ func (is *ImageStoreLocal) BlobPath(repo string, digest godigest.Digest) string
...
-
-	blobPath := is.BlobPath(repo, digest)
-
 	if is.dedupe && fmt.Sprintf("%v", is.cache) != fmt.Sprintf("%v", nil) {
 		is.Lock(&lockLatency)
 		defer is.Unlock(&lockLatency)
 	} else {
 		is.RLock(&lockLatency)
-		defer is.RUnlock(&lockLatency)
+		defer is.RLock(&lockLatency)
```
andaaron, line 1090: Is this a typo?

pkg/storage/s3/s3.go
```diff
@@ -1023,6 +1027,44 @@ func (is *ObjectStorage) CheckBlob(repo string, digest godigest.Digest) (bool, i
 	return true, blobSize, nil
 }
 
+// CheckBlobPresence verifies if a blob is present inside a repository. SHOULD lock from outside.
```
andaaron, line 1030:
```suggestion
// CheckBlobPresence verifies if a blob is present inside a repository. The caller function SHOULD lock from outside.
```

pkg/storage/s3/s3.go
```diff
@@ -1023,6 +1027,44 @@ func (is *ObjectStorage) CheckBlob(repo string, digest godigest.Digest) (bool, i
...
+
+	binfo, err := is.store.Stat(context.Background(), blobPath)
+	if err != nil {
+		is.log.Error().Err(err).Str("blob", blobPath).Msg("failed to stat blob")
+
+		return false, -1, zerr.ErrBlobNotFound
+	}
+
+	// is a 'deduped' blob?
+	if binfo.Size() == 0 {
```
andaaron, line 1046: The approach in CheckBlob seems cleaner. `if err == nil && binfo.Size() > 0` return and then check the error, and later you don't need this if block.

andaaron (approved):

andaaron (dismissed):

rchincha:
s/CheckBlobPresence/StatBlob/

rchincha:
Also look at https://github.com/project-zot/zot/pull/1223

peusebiu:
> Also look at #1223

I can add those changes in another PR

andaaron (approved):

rchincha (approved):
lgtm

PR #1577, opened 2023-07-04
feat(cli): add sort flag for query search

laurentiuNiculae:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


pkg/cli/searcher.go
```diff
@@ -725,10 +725,35 @@ func (search referrerSearcher) search(config searchConfig) (bool, error) {
...
+	default:
+		return "", zerr.ErrSortCriteriaNotSupported
+	}
+}
+
 type globalSearcherGQL struct{}
 
 func (search globalSearcherGQL) search(config searchConfig) (bool, error) {
-	if !canSearch(config.params, newSet("query")) {
+	if query := config.params["query"]; query == nil || *query == "" {
```
andaaron, line 756: I think we need a more generic way to handle mandatory vs optional parameters. There are other place where we could use this.

laurentiuNiculae: I think we should specify the required parameters that determine a subcommand command. The can search command will now return true if the flag set contains at least the required parameters and it doesn't contains unsupported flags. The optional parameters will be given a default value if the flag is not present and if the flag is not part of the optional or required parameters then we'll return an error.

andaaron: The tricky part is in case a specific set of mandatory parameters match one search, and there's a different searcher which is support to run if only one of those is specified.

Your explanation sounds good, but we need to be careful.

rchincha:
@laurentiuNiculae pls rebase?

andaaron:
This is probably superseded by #1674.

PR #1609, opened 2023-07-11
feat: add additional manifest validations

peusebiu:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -5443,6 +5443,239 @@ func TestImageSignatures(t *testing.T) {
 	})
 }
 
+func TestManifestValidation(t *testing.T) {
+	Convey("Validate manifest", t, func() {
+		// start a new server
+		port := test.GetFreePort()
+		baseURL := test.GetBaseURL(port)
+
+		conf := config.New()
+		conf.HTTP.Port = port
+
+		dir := t.TempDir()
+		ctlr := makeController(conf, dir, "")
+		cm := test.NewControllerManager(ctlr)
+		// this blocks
+		cm.StartServer()
+		time.Sleep(1000 * time.Millisecond)
+		defer cm.StopServer()
+
+		repoName := "validation"
+		blobContent := []byte("this is a blob")
+		blobDigest := godigest.FromBytes(blobContent)
+		So(blobDigest, ShouldNotBeNil)
+
+		cfg, layers, manifest, err := test.GetImageComponents(2)
+		So(err, ShouldBeNil)
+
+		err = test.UploadImage(
+			test.Image{
+				Config:    cfg,
+				Layers:    layers,
+				Manifest:  manifest,
+				Reference: "1.0",
+			}, baseURL, repoName)
+		So(err, ShouldBeNil)
+
+		content, err := json.Marshal(manifest)
+		So(err, ShouldBeNil)
+		digest := godigest.FromBytes(content)
+		So(digest, ShouldNotBeNil)
+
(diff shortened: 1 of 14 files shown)


rchincha:
2023-07-11T16:32:58.4637139Z Will run [1m79[0m of [1m79[0m specs
2023-07-11T16:32:58.5333546Z 
2023-07-11T16:32:58.5334539Z     [38;5;13mWARNING: image manifest with no layers is not supported
2023-07-11T16:32:58.5334962Z     [0m
2023-07-11T16:32:58.5336781Z     /home/runner/work/zot/zot/distribution-spec/conformance/02_push_test.go:366

rchincha:
image-spec is completely independent and the upstream schema validation requires a minimum of one layer

We happen to use image-spec for our layout but dist-spec allows zero layers, which we should also allow.

This is the conflicting requirement.

pkg/storage/common/common.go
```diff
@@ -71,27 +73,36 @@ func ValidateManifest(imgStore storageTypes.ImageStore, repo, reference, mediaTy
 	switch mediaType {
 	case ispec.MediaTypeImageManifest:
 		var manifest ispec.Manifest
+
+		// validate manifest
+		if err := schema.ValidatorMediaTypeManifest.Validate(bytes.NewBuffer(body)); err != nil {
```
rchincha, line 78: If the "err" says that there is only ONE error and that is missing layers, then we should still pass.

rchincha: See the "warning" in the comments. We want to support that case in `zot`

rchincha:
FYI, https://github.com/opencontainers/distribution-spec/pull/447

pkg/storage/common/common.go
```diff
@@ -777,6 +771,19 @@ func IsNonDistributable(mediaType string) bool {
 		mediaType == ispec.MediaTypeImageLayerNonDistributableZstd //nolint:staticcheck
 }
 
+func IsEmptyLayersError(err error) bool {
+	var validationErr schema.ValidationError
+	if errors.As(err, &validationErr) {
+		if len(validationErr.Errs) == 1 && strings.Contains(err.Error(), "layers: Array must have at least 1 items") {
```
andaaron, line 777: Define this error separately. Also can we add a test (if we don't have one already) where there are multiple validation errors, and this is just 1 of them?

peusebiu: added in TestArtifact in controller_test.go

what do you mean to define it separately? it's defined in schema package

andaaron: I see a hardcoded string `"layers: Array must have at least 1 items"`.
That's what I meant.

pkg/api/controller_test.go
```diff
@@ -5557,7 +5557,86 @@ func TestArtifactReferences(t *testing.T) {
...
+					Digest:    digest,
+					Size:      int64(len(mcontent)),
+				})
+
+				mcontent, err = json.Marshal(manifest)
+				So(err, ShouldBeNil)
+				resp, err = resty.R().SetHeader("Content-Type", ispec.MediaTypeImageManifest).
+					SetBody(mcontent).Put(baseURL + fmt.Sprintf("/v2/%s/manifests/1.0", repoName))
+				So(err, ShouldBeNil)
+				// modified from http.StatusBadRequest tot http.StatusCreated
```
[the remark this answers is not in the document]
peusebiu: ok

andaaron (approved):

PR #1610, opened 2023-07-11
fix: remove inline GC and schedule a background task instead

Andreea-Lupu:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -323,7 +323,7 @@ func (c *Controller) StartBackgroundTasks(reloadCtx context.Context) {
 	taskScheduler.RunScheduler(reloadCtx)
 
 	// Enable running garbage-collect periodically for DefaultStore
-	if c.Config.Storage.GC && c.Config.Storage.GCInterval != 0 {
+	if c.Config.Storage.GC {
 		c.StoreController.DefaultStore.RunGCPeriodically(c.Config.Storage.GCInterval, taskScheduler)
 	}
 
@@ -339,7 +339,7 @@ func (c *Controller) StartBackgroundTasks(reloadCtx context.Context) {
 	if c.Config.Storage.SubPaths != nil {
 		for route, storageConfig := range c.Config.Storage.SubPaths {
 			// Enable running garbage-collect periodically for subImageStore
-			if storageConfig.GC && storageConfig.GCInterval != 0 {
+			if storageConfig.GC {
 				c.StoreController.SubStore[route].RunGCPeriodically(storageConfig.GCInterval, taskScheduler)
 			}
 
--- a/pkg/storage/common/common.go
+++ b/pkg/storage/common/common.go
@@ -4,8 +4,11 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"io"
+	"math/rand"
 	"path"
 	"strings"
+	"time"
 
 	notreg "github.com/notaryproject/notation-go/registry"
 	godigest "github.com/opencontainers/go-digest"
@@ -853,6 +856,10 @@ func (gen *DedupeTaskGenerator) IsDone() bool {
 	return gen.done
 }
 
+func (gen *DedupeTaskGenerator) IsReady() bool {
+	return true
+}
+
 func (gen *DedupeTaskGenerator) Reset() {
 	gen.lastDigests = []godigest.Digest{}
 	gen.duplicateBlobs = []string{}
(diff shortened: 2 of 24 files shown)


pkg/api/routes.go
```diff
@@ -712,6 +712,8 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
 		return
 	}
 
+	rh.c.SubmitBackgroundTask(imgStore, name)
```
rchincha, line 715: I prefer calling out specific tasks. The call here could be a wrapper around it, for example, SubmitRepoTasks() etc which could schedule all tasks. More obvious, less magical.

andaaron (commented):
Remove the inline GC (and inline GC task scheduling).
Run the GC periodically by default with a default value.

A different requirement: the GCs for different repos should not all run at the same time.

We should add some randomness to the timing we generate the GC tasks at. Not wait in the GC task itself, as that would block a scheduler goroutine.

pkg/storage/common/common.go
```diff
@@ -886,3 +893,86 @@ func (dt *dedupeTask) DoWork() error {
...
+	randSec += minDelay
+
+	return randSec
+}
+
+func (gen *GCTaskGenerator) Next() (scheduler.Task, error) {
+	if gen.lastRepo == "" && gen.nextRun.IsZero() {
+		gen.rand = rand.New(rand.NewSource(time.Now().UTC().UnixNano())) //nolint: gosec
+
+		delay := gen.getRandomDelay()
```
andaaron, line 925: This code seems to be duplicated with line 925. Move it outside the `if`?

pkg/storage/common/common_test.go
```diff
@@ -104,6 +105,9 @@ func TestValidateManifest(t *testing.T) {
 			So(err, ShouldBeNil)
 
 			_, _, err = imgStore.PutImageManifest("test", "1.0", ispec.MediaTypeImageManifest, body)
```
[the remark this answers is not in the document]
peusebiu: let's remove it.

andaaron (approved):

andaaron (dismissed):

pkg/storage/common/common.go
```diff
@@ -886,3 +893,82 @@ func (dt *dedupeTask) DoWork() error {
...
+*/
+type GCTaskGenerator struct {
+	ImgStore storageTypes.ImageStore
+	lastRepo string
+	nextRun  time.Time
+	done     bool
+	rand     *rand.Rand
+}
+
+func (gen *GCTaskGenerator) getRandomDelay() int {
```
andaaron, line 911: 0-30 seconds

andaaron (approved):

andaaron (dismissed):

andaaron (approved):

rchincha:
"Test - gc with short interval / Test - gc with short interval"

Call this "GC stress test"

andaaron (approved):

pkg/api/controller.go
```diff
@@ -339,7 +339,7 @@ func (c *Controller) StartBackgroundTasks(reloadCtx context.Context) {
 	if c.Config.Storage.SubPaths != nil {
 		for route, storageConfig := range c.Config.Storage.SubPaths {
 			// Enable running garbage-collect periodically for subImageStore
-			if storageConfig.GC && storageConfig.GCInterval != 0 {
+			if storageConfig.GC {
```
rchincha, line 342: This is changing behavior right?
storageConfig.GCInterval != 0 means we can turn off GC with this set to zero? We don't want this anymore?

andaaron: I though we decided in one of our previous discussions periodic GC would always be enabled if inline GC is removed.

Andreea-Lupu: I tested this case now and it looks like gc will be run continuously

andaaron: You still want the user to be able to disabled GC?

Andreea-Lupu: > I tested this case now and it looks like gc will be run continuously

Sorry, tested again and I was wrong first time(I think I used another config). If GCInterval=0 set explicitly, then because of this check from scheduler https://github.com/project-zot/zot/blob/main/pkg/scheduler/scheduler.go#L344 gc is executed only one time

andaaron: In this case I think we can leave the implementation as is in this PR:
- enable/disabled GC periodically by using the 'enable' setting
- if unspecified, the GC is enabled with 1h interval by default
- if GC is enabled and 0 is specified as interval, run GC only once
- if GC is enabled an a non-0 interval value is used, run GC periodically at the specified interval

rchincha (approved):
lgtm

pkg/api/controller.go
rchincha: This is a good doc item
cc: @mbshields

Issue #1598, opened 2023-07-07
[Bug]: zli images --output json does not produce a valid json document.


raharper:
### zot version

{"level":"info","distribution-spec":"1.1.0-dev","commit":"v2.0.0-rc5-0-g8cc5137","binary-type":"-lint-metrics-mgmt-scrub-search-sync-ui-userprefs","go version":"go1.20.5","time":"2023-07-07T14:57:48-05:00","message":"version"}

### Describe the bug

```
rharper@saaz:~$ zli images myRepo --name ubuntu --output json > test.json 
rharper@saaz:~$ python3 -m json.tool test.json 
Expecting value: line 1 column 1 (char 0)
     ..:(╯°□°)╯彡┻━━┻

rharper@saaz:~$ head test.json 
IMAGE NAME  TAG               OS/ARCH         DIGEST      SIGNED      SIZE        
{
  "repoName": "ubuntu",
  "tag": "rolling",
  "digest": "sha256:72e75626c5068b9d9a462c4fc80a29787d0cf61c8abc81bfd5ea69f6248d56fc",
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "manifests": [
    {
      "digest": "sha256:72e75626c5068b9d9a462c4fc80a29787d0cf61c8abc81bfd5ea69f6248d56fc",
      "configDigest": "sha256:ec226db0c27e773193775e96b7ad88efd68e1584d6fe6f93f3fd764fcbd4874b",
```

Let's remove that header and try again

```
rharper@saaz:~$ tail -n +2 test.json | head 
{
  "repoName": "ubuntu",
  "tag": "rolling",
  "digest": "sha256:72e75626c5068b9d9a462c4fc80a29787d0cf61c8abc81bfd5ea69f6248d56fc",
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "manifests": [
    {
      "digest": "sha256:72e75626c5068b9d9a462c4fc80a29787d0cf61c8abc81bfd5ea69f6248d56fc",
      "configDigest": "sha256:ec226db0c27e773193775e96b7ad88efd68e1584d6fe6f93f3fd764fcbd4874b",
      "lastUpdated": "0001-01-01T00:00:00Z",
rharper@saaz:~$ tail -n +2 test.json > test-no-header.json
rharper@saaz:~$ python3 -m json.tool test-no-header.json 
Extra data: line 52 column 2 (char 1318)
[...]

Issue #1592, opened 2023-07-07
[Bug]: zli config add --help missing newline after output


raharper:
### zot version

{"level":"info","distribution-spec":"1.1.0-dev","commit":"v2.0.0-rc5-0-g8cc5137","binary-type":"-lint-metrics-mgmt-scrub-search-sync-ui-userprefs","go version":"go1.20.5","time":"2023-07-07T10:44:31-05:00","message":"version"}

### Describe the bug

```
$ zli config add --help
Add configuration for a zot registry

Usage:
  zli config add <config-name> <url> [flags]

Flags:
  -h, --help   help for add

Useful variables:
  url		zot server URL
  showspinner	show spinner while loading data [true/false]
  verify-tls	enable TLS certificate verification of the server [default: true]rharper@saaz:
  ```
  

### To reproduce

1. Configuration
2. Client tool used
3. Seen error

### Expected behavior

```
$ zli config add --help
Add configuration for a zot registry

Usage:
  zli config add <config-name> <url> [flags]

Flags:
  -h, --help   help for add

Useful variables:
  url		zot server URL
  showspinner	show spinner while loading data [true/false]
  verify-tls	enable TLS certificate verification of the server [default: true]
rharper@saaz:
```

Issue #1591, opened 2023-07-07
[Bug]: zli image output column headers are not aligned with output and unfriendly to cli parsing


raharper:
### zot version

{"level":"info","distribution-spec":"1.1.0-dev","commit":"v2.0.0-rc5-0-g8cc5137","binary-type":"-lint-metrics-mgmt-scrub-search-sync-ui-userprefs","go version":"go1.20.5","time":"2023-07-07T09:57:33-05:00","message":"version"}

### Describe the bug

`zli image repoName --name foo` produces hard-to-read output due to the header columns not being aligned with the actual output data.

```
IMAGE NAME  TAG       OS/ARCH	      DIGEST	  SIGNED      SIZE
aaaaaa/bbbbbbb  8-dev17-213-g40ecfd036  linux/amd64	 9dd8c9ed    false	 264MB
aaaaaa/bbbbbbb  8-dev17-201-gda3c8bebc-squashfs  linux/amd64	  e6267429    false	  249MB
```

Using something like `column -t` should just work but the 'IMAGE NAME'  has whitespace and breaks the alignment

```
IMAGE            NAME                             TAG          OS/ARCH   DIGEST  SIGNED  SIZE
aaaaaa/bbbbbbbb  8-dev17-184-g7305cd43d-squashfs  linux/amd64  3f0900f9  false   249MB
aaaaaa/bbbbbbbb  8-dev17-201-gda3c8bebc-squashfs  linux/amd64  e6267429  false   249MB
```

Further, the response from zot is not sorted, so re-running the same query results in different output order.

```


### To reproduce

1. Configuration
2. Client tool used
3. Seen error


### Expected behavior

```
IMAGE            TAG                              OS/ARCH      DIGEST    SIGNED  SIZE
aaaaaa/bbbbbbbb  8-dev17-186-gf8fcb1a8c           linux/amd64  a4a4db0c  false   264MB
aaaaaa/bbbbbbbb  8-dev17-188-gc957bbb39           linux/amd64  7375c7ce  false   264MB
aaaaaa/bbbbbbbb  8-dev17-188-gc957bbb39-squashfs  linux/amd64  f897f638  false   249MB

```

### Screenshots

_No response_

### Additional context

_No response_

rchincha:
Rename "IMAGE NAME" to "REPOSITORY" similar to docker output.

andaaron:
@raharper some notes:
[...]

PR #851, opened 2022-10-06
fix: images command not truncating image name/tag 

aokirisaki:
Signed-off-by: Lisca Ana-Roberta <ana.kagome@yahoo.com>

**What type of PR is this?**

**Which issue does this PR fix**:
#823 

**What does this PR do / Why do we need it**:

**If an issue # is not available please add repro steps and logs showing the issue**:

**Testing done on this change**:

**Automation added to e2e**:

**Will this break upgrades or downgrades?**

**Does this PR introduce any user-facing change?**:

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1612, opened 2023-07-12
Fix multiple zli issues

andaaron:
feat(cli): Rename IMAGE NAME to REPOSITORY"

https://github.com/project-zot/zot/issues/1591
    - I will rename "IMAGE NAME" to "REPOSITORY" in order to make the header easier to parse
    - The order of the images cannot be predicted if zot is getting them 1 by 1 using the REST API for manifests, so they cannot be sorted when printed. We could wait on all calls to return but that may take minutes, and printing partial results as they become available is better.
    - The order of the images can be predicted when relying on the zot specific search API, but that is not available in all zot servers depending on build options. I will add sorting ascending by default. We are planning to implement configurable sorting in a separate PR - see the work under https://github.com/project-zot/zot/pull/1577
    - With regards to the column widths/alignments that was discussed before, and the issue is we don't know the values beforehand for the REST API based responses. As mentioned above printing partial results as they become available is better.
    - The column widths/alignments are partially fixed in this PR for the search API, but we should properly fix this in - see https://github.com/project-zot/zot/pull/851 - and we'll need server side changes for other tag sort criteria besides sorting ASC

https://github.com/project-zot/zot/issues/1592
    - Fix missing space after help message

https://github.com/project-zot/zot/issues/1598
    - Fix table headers showing for json/yaml format
    - Fix spacing shown with json format, use 1 row per shown entry in order to be compatible with json lines format: https://jsonlines.org/
    - Add document header `---` to every image shown in yaml format to separate the entries


rchincha:
Update our docs? @mbshields

rchincha (approved):
lgtm

Pls make sure docs are updated.

PR #1633, opened 2023-07-18
fix(authn): fix several issues with authn, closes #1632

### LINKED ISSUE #1632: [Bug] zot-server side issues related to zui login
1. In case Auth is disabled - do not apply headers `Access-Control-Allow-Credentials` in the CORS headers.
2. `isAuthnEnabled` doesn't check all authentication methods, we need to make sure it is not called with the assumption it verifies bearer or openid auth. Maybe rename it because it is misleading.
3. The logout route needs to be enabled in all cases if Auth is enabled, not just if openID is enabled
4. Fix the `Access-Control-Allow-Methods` header on the Logout route to specify only POST and OPTIONS
5. Fix the Logout godoc to mention the correct logout route path
6. Add a zot config example with the `AllowOrigin` setting.

peusebiu:
- apply Access-Control-Allow-Credentials only if authn is enabled
- enable Logout route for basic auth
- fixed Logout godoc
- fix Access-Control-Allow-Methods on Logout route
- added allowOrigin option in config example

**What type of PR is this?**
bug

- apply Access-Control-Allow-Credentials only if authn is enabled
- enable Logout route for basic auth
- fixed Logout godoc
- fix Access-Control-Allow-Methods on Logout route
- added allowOrigin option in config example
see #1632 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/config/config.go
+++ b/pkg/api/config/config.go
@@ -16,6 +16,10 @@ var (
 	ReleaseTag string //nolint: gochecknoglobals
 	BinaryType string //nolint: gochecknoglobals
 	GoVersion  string //nolint: gochecknoglobals
+
+	openIDSupportedProviders = [...]string{"google", "gitlab", "dex"} //nolint: gochecknoglobals
+	oauth2SupportedProviders = [...]string{"github"}                  //nolint: gochecknoglobals
+
 )
 
 type StorageConfig struct {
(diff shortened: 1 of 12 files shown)


pkg/api/config/config.go
```diff
@@ -227,3 +228,79 @@ func (c *Config) Sanitize() *Config {
 
 	return sanitizedConfig
 }
+
+func (c *Config) IsLdapAuthEnabled() bool {
+	if c.HTTP.Auth != nil && c.HTTP.Auth.HTPasswd.Path != "" {
```
andaaron, line 233: The lines 233 and 241 should be switched in between?

pkg/api/config/config.go
```diff
@@ -227,3 +228,79 @@ func (c *Config) Sanitize() *Config {
...
+
+func (c *Config) IsHtpasswdAuthEnabled() bool {
+	if c.HTTP.Auth != nil && c.HTTP.Auth.LDAP != nil {
+		return true
+	}
+
+	return false
+}
+
+func (c *Config) IsBearerAuthEnabled() bool {
```
andaaron, line 248: Very good idea to add these as methods.

pkg/api/constants/consts.go
```diff
@@ -20,3 +20,28 @@ const (
 	APIKeysPrefix                = "zak_"
 	CallbackUIQueryParam         = "callback_ui"
 )
+
+var (
+	openIDSupportedProviders = [...]string{"google", "gitlab", "dex"} //nolint: gochecknoglobals
+	oauth2SupportedProviders = [...]string{"github"}                  //nolint: gochecknoglobals
+)
+
+func IsOpenIDSupported(provider string) bool {
```
andaaron, line 29: Maybe these methods should move out of the constants package? They are not constants, maybe move back to pkg/api?

pkg/api/routes.go
```diff
@@ -63,37 +64,42 @@ func (rh *RouteHandler) SetupRoutes() {
 
 	applyCORSHeaders := getCORSHeadersHandler(rh.c.Config.HTTP.AllowOrigin)
 
-	if isOpenIDAuthEnabled(rh.c.Config) {
+	if rh.c.Config.IsBasicAuthnEnabled() {
```
andaaron, line 67: Shouldn't this check only OpenID and OAuth2?

pkg/api/routes.go
```diff
@@ -63,37 +64,42 @@ func (rh *RouteHandler) SetupRoutes() {
...
 		}
 	}
 
+	/* on every route which may be used by UI we set OPTIONS as allowed METHOD
+	to enable preflight request from UI to backend */
+	if rh.c.Config.IsBasicAuthnEnabled() {
+		// logout path for openID
+		rh.c.Router.HandleFunc(constants.LogoutPath,
+			getUIHeadersHandler(rh.c.Config, "POST", "OPTIONS")(applyCORSHeaders(rh.Logout))).
+			Methods("POST", "OPTIONS")
```
andaaron, line 89: Let's use https://pkg.go.dev/net/http#pkg-constants

pkg/api/routes.go
```diff
@@ -103,40 +109,47 @@ func (rh *RouteHandler) SetupRoutes() {
...
+			getUIHeadersHandler(rh.c.Config, http.MethodGet, http.MethodOptions)(
+				applyCORSHeaders(rh.ListTags))).Methods(http.MethodGet, http.MethodOptions)
 		prefixedDistSpecRouter.HandleFunc(fmt.Sprintf("/{name:%s}/manifests/{reference}", zreg.NameRegexp.String()),
-			applyCORSHeaders(rh.CheckManifest)).Methods(zcommon.AllowedMethods("HEAD")...)
+			getUIHeadersHandler(rh.c.Config, http.MethodHead, http.MethodGet, http.MethodOptions)(
+				applyCORSHeaders(rh.CheckManifest))).Methods(http.MethodHead, http.MethodOptions)
 		prefixedDistSpecRouter.HandleFunc(fmt.Sprintf("/{name:%s}/manifests/{reference}", zreg.NameRegexp.String()),
-			applyCORSHeaders(rh.GetManifest)).Methods(zcommon.AllowedMethods("GET")...)
+			getUIHeadersHandler(rh.c.Config, http.MethodHead, http.MethodGet, http.MethodOptions)(
+				applyCORSHeaders(rh.GetManifest))).Methods(http.MethodGet, http.MethodOptions)
```
andaaron, line 119:
```suggestion
				applyCORSHeaders(rh.GetManifest))).Methods(http.MethodGet)
```
Since line 116 takes care of method OPTIONS.

examples/config-openid.json
```diff
@@ -8,6 +8,7 @@
     "address": "127.0.0.1",
     "port": "8080",
     "realm": "zot",
+    "allowOrigin": "*",
```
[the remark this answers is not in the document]
rchincha: Don't like adding arbitrary things under config. We should be more disciplined about this.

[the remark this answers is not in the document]
andaaron: We can remove this change from the commit, but this has been an (undocumented?) field for about a year, with no example configuration.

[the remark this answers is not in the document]
andaaron: Also this setting is for the headers in the REST API responses, not calls to the UI endpoint.

rchincha (approved):
lgtm

PR #1636, opened 2023-07-19
refactor(extensions)!: refactor the extensions URLs and errors

andaaron:
BREAKING CHANGE: The functionality provided by the mgmt endpoint has beed redesigned - see details below
BREAKING CHANGE: The API keys endpoint has been moved -  see details below
BREAKING CHANGE: The mgmt extension config has been removed - endpoint is now enabled by having both the search and the ui extensions enabled
BREAKING CHANGE: The API keys configuration has been moved from extensions to http>auth>apikey

mgmt and imagetrust extensions:
- separate the _zot/ext/mgmt into 3 separate endpoints: _zot/ext/auth, _zot/ext/notation, _zot/ext/cosign
- signature verification logic is in a separate `imagetrust` extension
- better hanling or errors in case of signature uploads: logging and error codes (more 400 and less 500 errors)
- add authz on signature uploads (and add a new middleware in common for this purpose)
- remove the mgmt extension configuration - it is now enabled if the UI and the search extensions are enabled

userprefs estension:
- userprefs are enabled if both search and ui extensions are enabled (as opposed to just search)

apikey extension is removed and logic moved into the api folder
- Move apikeys code out of pkg/extensions and into pkg/api
- Remove apikey configuration options from the extensions configuration and move it inside the http auth section
- remove the build label apikeys

other changes:
- move most of the logic adding handlers to the extensions endpoints out of routes.go and into the extensions files.
- add warnings in case the users are still using configurations with the obsolete settings for mgmt and api keys
- add a new function in the extension package which could be a single point of starting backgroud tasks for all extensions
- more clear methods for verifying specific extensions are enabled
- fix http methods paired with the UI handlers
- rebuild swagger docs


rchincha:
mgmt -> zot/_ext/config
// mgmt/auth -> zot/_ext/auth
sign -> zot/_ext/notation
sign -> zot/_ext/cosign

rchincha:
Some "extensions" are **core** such as those related to Authentication, Signatures.
They will be enabled by default and have no configuration backing them ("extensions": { "enable": true/false ... etc

_zot/ext/config <- get and post/patch/put (in future) config based on credentials
^ there is "206 Partial Content"
_zot/ext/notation <- _notation_ tool interactions
_zot/ext/cosign <- _cosign_ tool interactions

rchincha (approved):
lgtm

Issue #1642, opened 2023-07-24
[Bug] User information is not populated in the request context if anonymous access is permitted


andaaron:
If someone provides credentials, the user information should be set regardless of anonymous access settings.

Right now if session authentication is used with a valid session and anonymous access is permitted, the code at https://github.com/project-zot/zot/blob/main/pkg/api/authn.go#L126 will skip setting user information, resulting in 403 errors on calls requiring a username (such as setting bookmarks).

PR #1647, opened 2023-07-24
fix(authn): session authn is skipped when anonymous policy is configured

peusebiu:
closes: #1642

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -62,17 +62,10 @@ func AuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 	return authnMiddleware.TryAuthnHandlers(ctlr)
 }
 
-func (amw *AuthnMiddleware) sessionAuthn(ctlr *Controller, next http.Handler, response http.ResponseWriter,
-	request *http.Request, delay int,
-) {
-	clientHeader := request.Header.Get(constants.SessionClientHeaderName)
-	if clientHeader != constants.SessionClientHeaderValue {
-		authFail(response, request, ctlr.Config.HTTP.Realm, delay)
-
-		return
-	}
-
-	identity, ok := common.GetAuthUserFromRequestSession(ctlr.CookieStore, request, ctlr.Log)
+func (amw *AuthnMiddleware) sessionAuthn(ctlr *Controller, response http.ResponseWriter,
+	request *http.Request,
+) (bool, error) {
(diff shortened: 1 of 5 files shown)


pkg/api/authn.go
```diff
@@ -394,22 +353,65 @@ func (amw *AuthnMiddleware) TryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
...
+					next.ServeHTTP(cloneResp, cloneReq)
+
+					return
+				}
+			} else {
+				// try anonymous auth only if basic auth/session was not given
+				// we want to bypass auth for mgmt route
+				isMgmtRequested := request.RequestURI == constants.FullMgmtPrefix
+				if ctlr.Config.HTTP.AccessControl.AnonymousPolicyExists() || isMgmtRequested {
+					// do we really need this? acCtx will be non nil!
```
andaaron, line 405: Not sure what to say, I would keep the existing logic of creating the context, unless you also update all the places where it is checked for nil.

pkg/api/controller_test.go
```diff
@@ -3559,151 +3559,190 @@ func TestAuthorizationWithBasicAuth(t *testing.T) {
...
+
+			client := resty.New()
+			client.SetBasicAuth(username, passphrase)
+
+			RunAuthorizationTests(t, client, baseURL, conf)
+		})
+	})
+}
+
+func TestGetUsername(t *testing.T) {
```
andaaron, line 3635: Shouldn't we have a positive test besides the negative ones?

pkg/api/authn.go
```diff
@@ -800,3 +803,39 @@ func hashUUID(uuid string) string {
...
+	// at this point we should have a session set on cookie.
+	// if created in the earlier Get() call then user is not logged in with sessions.
+	if session.IsNew {
+		return "", false
+	}
+
+	authenticated := session.Values["authStatus"]
+	if authenticated != true {
+		log.Error().Msg("can not get `user` session value")
+
```
peusebiu, line 829: this should be covered by the existing tests, check what happened!

pkg/api/authn.go
```diff
@@ -800,3 +803,39 @@ func hashUUID(uuid string) string {
...
+		return "", false
+	}
+
+	identity, ok := session.Values["user"].(string)
+	if !ok {
+		log.Error().Msg("can not get `user` session value")
```
peusebiu, line 835: same as above

rchincha (approved):
lgtm

PR #1656, opened 2023-07-27
refactor(storage): refactor storage into a single ImageStore

peusebiu:
refactor(gc): drop umoci dependency, implemented internal gc

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


pkg/storage/imagestore/imagestore.go
```diff
@@ -0,0 +1,2232 @@
...
+			monitoring.SetStorageUsage(is.metrics, is.rootDir, repo)
+		}
+	}()
+
+	err = is.deleteImageManifest(repo, reference, detectCollisions)
+	if err != nil {
+		return err
+	}
+
+	if is.gc {
```
andaaron, line 628: Do we want to keep this? @rchincha @Andreea-Lupu

pkg/storage/imagestore/imagestore.go
```diff
@@ -0,0 +1,2232 @@
...
+				})
+
+				continue
+			}
+		}
+	}
+
+	is.log.Info().Msg("gc: untagged manifests")
+
+	if err := gcUntaggedManifests(is, &index, repo, referencedByImageIndex); err != nil {
```
andaaron, line 1627: This function does not return manifests. 
It should be renamed, for example `AddUntagged...`

pkg/storage/imagestore/imagestore.go
```diff
@@ -0,0 +1,2232 @@
...
+		return err
+	}
+
+	is.log.Info().Msg("gc: cosign references")
+
+	if err := gcCosignReferences(is, &index, repo, cosignDescriptors); err != nil {
```
andaaron, line 1633: Same as above

pkg/storage/imagestore/imagestore.go
```diff
@@ -0,0 +1,2232 @@
...
+		return err
+	}
+
+	is.log.Info().Msg("gc: notation signatures")
+
+	if err := gcNotationSignatures(is, &index, repo, notationManifests); err != nil {
```
andaaron, line 1639: Same as above

pkg/storage/imagestore/imagestore.go
```diff
@@ -0,0 +1,2232 @@
...
+
+				err = json.Unmarshal(buf, &manifest)
+				if err != nil {
+					return err
+				}
+
+				// skip manifests which are not of type image
+				if manifest.Config.MediaType != ispec.MediaTypeImageConfig {
+					imgStore.log.Info().Str("config mediaType", manifest.Config.MediaType).
+						Msg("skipping gc untagged manifest, because config blob is not application/vnd.oci.image.config.v1+json")
```
andaaron, line 1684: I think you should use the constant in the msg as well.

pkg/storage/local/driver.go
```diff
@@ -0,0 +1,459 @@
...
+
+	return os.Link(src, dest)
+}
+
+type fileInfo struct {
+	os.FileInfo
+	path string
+}
+
+var _ storagedriver.FileInfo = fileInfo{}
```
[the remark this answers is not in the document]
peusebiu: see https://github.com/uber-go/guide/issues/25

pkg/storage/imagestore/imagestore.go
```diff
@@ -0,0 +1,2232 @@
...
+		if filename == "blobs" && !fileInfo.IsDir() {
+			return false, nil
+		}
+
+		found[filename] = true
+	}
+
+	// check blobs dir exists only for filesystem, in s3 we can't have empty dirs
+	if is.store.Name() == "filesystem" {
+		if !is.store.DirExists(path.Join(dir, "blobs")) {
```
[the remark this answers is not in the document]
peusebiu: seem they don't have one...

pkg/storage/imagestore/imagestore.go
peusebiu: it's gc not get :D, garbageCollectUntaggedManifests. same for below comments.

andaaron: Makes total sense, sorry.

rchincha:
PR #1610 has been merged, pls rebase.

pkg/storage/local/driver.go
```diff
@@ -0,0 +1,456 @@
...
+	}
+
+	sort.Stable(sort.StringSlice(children))
+
+	for _, child := range children {
+		// Calling driver.Stat for every entry is quite
+		// expensive when running against backends with a slow Stat
+		// implementation, such as s3. This is very likely a serious
+		// performance bottleneck.
+		fileInfo, err := driver.Stat(child)
```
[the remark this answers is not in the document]
peusebiu: yes, between driver.List and for loop

pkg/storage/imagestore/imagestore.go
```diff
@@ -0,0 +1,2177 @@
...
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+func isBlobOlderThan(imgStore storageTypes.ImageStore, repo string,
+	digest godigest.Digest, delay time.Duration, log zerolog.Logger,
```
andaaron, line 1773: Why are we using zerolog.Logger instead of our own implementation of logger? It's used in multiple places in the storage package..

pkg/storage/imagestore/imagestore.go
```diff
@@ -0,0 +1,2177 @@
...
+	"sync"
+	"time"
+	"unicode/utf8"
+
+	"github.com/docker/distribution/registry/storage/driver"
+	guuid "github.com/gofrs/uuid"
+	godigest "github.com/opencontainers/go-digest"
+	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+	artifactspec "github.com/oras-project/artifacts-spec/specs-go/v1"
+	"github.com/rs/zerolog"
```
andaaron, line 22: Why do we need this?

andaaron (dismissed):
Looks good from my side.

.github/workflows/gc-stress-test.yaml
```diff
@@ -27,11 +27,52 @@ jobs:
...
-            ./bin/zot-linux-amd64 serve examples/config-gc-bench.json &
+            ./bin/zot-linux-amd64 serve test/gc-stress/config-gc-bench-local.json &
             sleep 10
             bin/zb-linux-amd64 -c 10 -n 100 -o ci-cd http://localhost:8080
 
             killall -r zot-*
 
             # clean zot storage
             sudo rm -rf /tmp/zot
+  gc-stress-s3:
```
rchincha, line 38: Needs to move to nightly, because it is so slow!

pkg/api/routes.go
```diff
@@ -814,6 +814,10 @@ func (rh *RouteHandler) DeleteManifest(response http.ResponseWriter, request *ht
 		} else if errors.Is(err, zerr.ErrBadManifest) {
 			zcommon.WriteJSON(response, http.StatusBadRequest,
 				apiErr.NewErrorList(apiErr.NewError(apiErr.UNSUPPORTED, map[string]string{"reference": reference})))
+		} else if errors.Is(err, zerr.ErrBlobReferenced) {
```
rchincha, line 817: Can you pls check why Blob is referenced here? This is a manifest delete path right? not blob?

rchincha: ErrManifestReferenced

rchincha: Also add a comment that this manifest could be inside an index hence the conflict

pkg/storage/constants/constants.go
```diff
@@ -22,4 +22,5 @@ const (
 	DefaultGCDelay          = 1 * time.Hour
 	DefaultGCInterval       = 1 * time.Hour
 	S3StorageDriverName     = "s3"
+	LocalStorageDriverName  = "filesystem"
```
rchincha, line 25: "local"

rchincha:
When an image is deleted, and a "reference" is dangling, then we should gc that reference (opinionated implementation!)
A <- R1, R2, R3 ... if A is deleted, gc R1, R2, R3 ...

Add this test in blackbox tests also.

pkg/storage/common/common.go
```diff
@@ -529,7 +531,101 @@ func IsBlobReferenced(imgStore storageTypes.ImageStore, repo string,
...
+		log.Error().Err(err).Str("repository", repo).Str("digest", mdigest.String()).
+			Msg("gc: failed to read manifest image")
+
+		return err
+	}
+
+	refBlobs[mdigest.String()] = true
+	refBlobs[manifestContent.Config.Digest.String()] = true
+
+	// if there is a Subject, it may not exist yet and that is ok
```
andaaron, line 553: I do have a concern with regards to this.

Let's say we release an image, we tag it with `latest`, and sign it.
At some point we want to push a newer image and retag that one as `latest`.
Since the older image has a digest referenced in the signature, it will never be GCed, is that correct? Do we want that? I don't think so, because we'd never GC anything as they will be referenced forever? 

My understanding is it is good to GC images if they are only referenced as subject by artifacts, and are not tagged, and they are not referenced in an index.
We should also be ok with GCing artifacts which have missing subjects after the GC delay interval passes.

pkg/storage/imagestore/imagestore.go
```diff
@@ -0,0 +1,2205 @@
...
+
+				manifest := ispec.Manifest{}
+
+				err = json.Unmarshal(buf, &manifest)
+				if err != nil {
+					return err
+				}
+
+				// skip manifests which are not of type image
+				if manifest.Config.MediaType != ispec.MediaTypeImageConfig {
```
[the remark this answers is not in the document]
rchincha: Because we don't really know how to GC this?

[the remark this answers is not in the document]
peusebiu: what if they are untagged oci artifacts which references valid images?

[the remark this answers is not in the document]
andaaron: That we should keep too. But not in the sense of `untagger oci artifacts`, in sense of `any oci manifests` which refer to valid `oci manifests`.

Whether or not the config mediatype or the manifest artifacttype has a specific value is irrelevant for GC, right? Just tags, references and created time matter, right?

pkg/storage/common/common.go
```diff
@@ -529,7 +531,104 @@ func IsBlobReferenced(imgStore storageTypes.ImageStore, repo string,
...
+	if manifestContent.Subject != nil {
+		_, err := GetImageManifest(imgStore, repo, manifestContent.Subject.Digest, log)
```
andaaron, line 555: I think the correct questions to ask are:
- is this manifest digest tagged?
- is this manifest digest referring to a tagged manifest?
- is this manifest digest referring to a manifest with is referring to a tagged manifest?
- and so on
Not sure if this is the right place to add that logic or not.

andaaron: I think one of the issues is, if you use only references and created times, like umoci does, you basically assume someone deletes unwanted manifests along with the tags in the DELETE API call, and in that case it is easy, you just search for all orphan (unreferenced) blobs and delete them.

But since there is the case you can delete a tag, but not the manifest, the umoci behavior is not enough. We want the user to delete the tag, and GC delete the manifests behind the scenes, if it is truly unused. So we need a smarter way to do this cleanup.

peusebiu: nice catches.
ok, let me think about it.

pkg/storage/imagestore/imagestore.go
```diff
@@ -0,0 +1,2194 @@
...
+			if err != nil {
+				is.log.Error().Err(err).Str("repository", repo).Str("digest", desc.Digest.String()).
+					Msg("gc: failed to read multiarch(index) image")
+
+				return err
+			}
+
+			for _, indexDesc := range indexImage.Manifests {
+				referencedByImageIndex = append(referencedByImageIndex, indexDesc.Digest.String())
+			}
```
andaaron, line 1604: Also check for a potential subject the image index may have.

pkg/storage/imagestore/imagestore.go
```diff
@@ -0,0 +1,2194 @@
...
+	referencedByImageIndex := []string{}
+	cosignDescriptors := []ispec.Descriptor{}
+	artifactManifests := []extendedManifest{}
+
+	/* gather manifests references by multiarch images (to skip gc)
+	gather cosign and notation signatures descriptors */
+	for _, desc := range index.Manifests {
+		switch desc.MediaType {
+		case ispec.MediaTypeImageIndex:
+			indexImage, err := common.GetImageIndex(is, repo, desc.Digest, is.log)
```
andaaron, line 1594: Check if the index is tagged, same as we do for manifests.

pkg/storage/imagestore/imagestore.go
```diff
@@ -0,0 +1,2194 @@
...
+	// gc untagged manifests and signatures
+	index, err := common.GetIndex(is, repo, is.log)
+	if err != nil {
+		return err
+	}
+
+	referencedByImageIndex := []string{}
+	cosignDescriptors := []ispec.Descriptor{}
+	artifactManifests := []extendedManifest{}
+
```
andaaron, line 1588: Maybe we should keep a set of digests which are either tagged, or referenced by tagged manifests, or tagged indexes, or indexes referenced by tagged indexes or indexes referenced by indexes referenced by tagged indexes and so on. 
By referenced I mean as part of a manifest list inside an index, as part of blob list, or config for a manifest, and as "subject" for either an index or a manifest.

What about cosign signatures? Even though they are tagged, we we need to check if their target digest is part of this new slice.

For notation signatures, and other generic referrers in general (maybe sboms in the future), we need to check if their subject is part of this new slice.

Note: a single pass though the index.Manifests may not be enough to identify all the relationships between digests.

pkg/storage/imagestore/imagestore.go
```diff
@@ -0,0 +1,2194 @@
...
+			manifestContent, err := common.GetImageManifest(is, repo, desc.Digest, is.log)
+			if err != nil {
+				is.log.Error().Err(err).Str("repo", repo).Str("digest", desc.Digest.String()).
+					Msg("gc: failed to read manifest image")
+
+				return err
+			}
+
+			if manifestContent.Subject != nil {
+				artifactManifests = append(artifactManifests, extendedManifest{
```
andaaron, line 1626:
```suggestion
				referrerManifests = append(referrerManifests, extendedManifest{
```
In case we still need this code after refactor.

pkg/storage/common/common.go
andaaron: We need an image retention policy, separate from GC for cleaning up images without tags.

pkg/storage/imagestore/imagestore.go
```diff
@@ -0,0 +1,2253 @@
...
+func (is *ImageStore) garbageCollectReferrer(repo string, index ispec.Index, manifestDesc ispec.Descriptor,
+	subject *ispec.Descriptor,
+) (bool, error) {
+	var gced bool
+
+	var err error
+
+	if subject != nil {
+		// try to find subject in index.json
+		if ok := isManifestReferencedInIndex(index, subject.Digest); !ok {
```
andaaron, line 1679: I'm not convinced this is enough. Don't we need to look in all possible indexes? Not just the same index where the manifest is found? Or am I misunderstanding?

pkg/storage/imagestore/imagestore.go
```diff
@@ -0,0 +1,2253 @@
...
+	if err := is.storeDriver.Delete(blobPath); err != nil {
+		is.log.Error().Err(err).Str("blobPath", blobPath).Msg("unable to remove blob path")
+
+		return err
+	}
+
+	return nil
+}
+
+func (is *ImageStore) garbageCollectIndexReferrers(repo string, index ispec.Index) (bool, error) {
```
andaaron, line 1574: Can we move this function just after the declaration of garbageCollect? It's hard to follow the code as the functions are not called in the order of which they are defined in the file.

andaaron (commented):
The code to GC untagged manifests and blobs looks good.
I am not clear on the code to GC referrers.

andaaron (approved):
Looks good from my side.

andaaron (approved):

pkg/api/config/config.go
```diff
@@ -23,15 +23,17 @@ var (
...
-	StorageDriver map[string]interface{} `mapstructure:",omitempty"`
-	CacheDriver   map[string]interface{} `mapstructure:",omitempty"`
+	RootDirectory               string
+	Dedupe                      bool
+	RemoteCache                 bool
+	GC                          bool
+	Commit                      bool
+	GCDelay                     time.Duration
+	GCInterval                  time.Duration
+	GCReferrers                 bool
```
rchincha, line 33: Thanks for adding this.

test/gc-stress/config-gc-bench-local.json
```diff
@@ -1,9 +1,11 @@
 {
     "distSpecVersion": "1.1.0-dev",
     "storage": {
-        "rootDirectory": "/tmp/zot",
+        "rootDirectory": "/tmp/zot/local",
         "gc": true,
-        "gcDelay": "10s",
+        "gcReferrers": true,
```
rchincha, line 6: @peusebiu can you add/copy the same (both local and s3) test with gcReferrers=false also?

peusebiu:
added stress test with and without referrers, also tested the nightly builds, should be ok to be merged.

andaaron (approved):

rchincha (approved):
lgtm

PR #1637, opened 2023-07-20
Refactor metadb sorting pagination

laurentiuNiculae:
**What type of PR is this?**

**Which issue does this PR fix**:

**What does this PR do / Why do we need it**:

**If an issue # is not available please add repro steps and logs showing the issue**:

**Testing done on this change**:

**Automation added to e2e**:

**Will this break upgrades or downgrades?**

**Does this PR introduce any user-facing change?**:

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1658, opened 2023-07-27
fix(artifacts): unmarshaling error

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


pkg/meta/hooks.go
```diff
@@ -66,7 +67,7 @@ func OnUpdateManifest(repo, reference, mediaType string, digest godigest.Digest,
 	}
 
 	if !metadataSuccessfullySet {
-		log.Info().Str("tag", reference).Str("repository", repo).Msg("uploding image meta was unsuccessful for tag in repo")
+		log.Info().Str("tag", reference).Str("repository", repo).Msg("uploading image meta was unsuccessful for tag in repo")
 
 		if err := imgStore.DeleteImageManifest(repo, reference, false); err != nil {
```
andaaron, line 72: Maybe initialize a new error variable in order to keep the original error to return at 78.

pkg/meta/parse.go
```diff
@@ -332,9 +332,12 @@ func NewManifestData(repoName string, manifestBlob []byte, imageStore storageTyp
 		return mTypes.ManifestData{}, err
 	}
 
-	err = json.Unmarshal(configBlob, &configContent)
-	if err != nil {
-		return mTypes.ManifestData{}, err
+	// with artifacts support, config can be used to store the actual artifact so wont unmarshal
+	if manifestContent.Config.MediaType == ispec.MediaTypeImageConfig {
```
andaaron, line 336: Thinking there could be issues if the non-image-config is persisted in the db and unmarshalled later when making the search reads. I think this case should be handled when reading from the DB as well. If we're not doing it already.

examples/config-conformance.json
```diff
@@ -11,5 +11,10 @@
   },
   "log": {
     "level": "debug"
+  },
+  "extensions": {
+    "search": {
+      "enable": true
```
andaaron, line 17: Why do we need this change? Search should be optional. Aren't we checking the DB is initialized before attempting to write to it?

pkg/meta/hooks.go
```diff
@@ -122,8 +123,8 @@ func OnDeleteManifest(repo, reference, mediaType string, digest godigest.Digest,
 			manageRepoMetaSuccessfully = false
 		}
 
-		if refferredDigest, hasSubject := common.GetReferredSubject(manifestBlob); hasSubject {
-			err := metaDB.DeleteReferrer(repo, refferredDigest, digest)
+		if referredDigest, hasSubject := common.GetReferredSubject(manifestBlob); hasSubject {
+			err := metaDB.DeleteReferrer(repo, referredDigest, digest)
```
andaaron, line 127: Please use a different variable, so the original err is reported. And update the value to wrap both errors in case a new one is returned here.

rchincha:
PR #1637 supersedes this.

PR #1662, opened 2023-07-29
fix(auth): fix anonymous auth for ui

andaaron:
The ui sends the header X-ZOT-API-CLIENT=zot-ui regardless of session authentication status. In case of new sessions zot would reject the unauthenticated call on /v2 (which is used to determine if anonymous access is allowed by the server when the header was set) expecting all users sending this header to be already authenticated.

Since the ui received 401 from the server, it would not show the option for anonymous login.


rchincha (approved):
lgtm

PR #1671, opened 2023-08-02
feat(scheduler): pass the shutdown/reload context to running tasks

peusebiu:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/scrub/scrub.go
+++ b/pkg/extensions/scrub/scrub.go
@@ -4,6 +4,7 @@
 package scrub
 
 import (
+	"context"
 	"fmt"
 	"path"
 
@@ -13,11 +14,11 @@ import (
 )
 
 // Scrub Extension for repo...
-func RunScrubRepo(imgStore storageTypes.ImageStore, repo string, log log.Logger) error {
+func RunScrubRepo(ctx context.Context, imgStore storageTypes.ImageStore, repo string, log log.Logger) error {
 	execMsg := fmt.Sprintf("executing scrub to check manifest/blob integrity for %s", path.Join(imgStore.RootDir(), repo))
 	log.Info().Msg(execMsg)
 
-	results, err := storage.CheckRepo(repo, imgStore)
+	results, err := storage.CheckRepo(ctx, repo, imgStore)
 	if err != nil {
 		errMessage := fmt.Sprintf("error while running scrub for %s", path.Join(imgStore.RootDir(), repo))
 		log.Error().Err(err).Msg(errMessage)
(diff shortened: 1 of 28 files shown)


andaaron (approved):
Looks ok from my side

pkg/extensions/scrub/scrub.go
```diff
@@ -58,6 +59,6 @@ func NewTask(imgStore storageTypes.ImageStore, repo string, log log.Logger) *Tas
 	return &Task{imgStore, repo, log}
 }
 
-func (scrubT *Task) DoWork() error {
-	return RunScrubRepo(scrubT.imgStore, scrubT.repo, scrubT.log)
+func (scrubT *Task) DoWork(ctx context.Context) error {
+	return RunScrubRepo(scrubT.imgStore, scrubT.repo, scrubT.log) //nolint: contextcheck
```
rchincha, line 63: why not pass it to the scrub task as well?

pkg/extensions/sync/service.go
```diff
@@ -243,7 +243,7 @@ func (service *BaseService) SyncImage(repo, reference string) error {
 	service.log.Info().Str("remote", remoteURL).Str("repo", repo).Str("reference", reference).
 		Msg("sync: syncing image")
 
-	manifestDigest, err := service.syncTag(repo, remoteRepo, reference)
+	manifestDigest, err := service.syncTag(context.Background(), repo, remoteRepo, reference)
```
rchincha, line 246: Need to get that context here as well

rchincha:
This is a good PR overall, but needs some changes.

pkg/extensions/sync/service.go
peusebiu: this is for on demand, we don't want to cancel, it will run in background and if a new client request the same image it will wait on image channel to be downloaded.

pkg/extensions/scrub/scrub.go
peusebiu: done

pkg/extensions/sync/service.go
rchincha: But does zot wait for this then?

rchincha: The main idea, being from a top-level context, we either force cancel or we wait for done. We have to pick one.

rchincha: And ideally all foreground/background tasks are accounted for.

peusebiu: yes, my bad, updated, thank you!

andaaron:
Related to #1651

andaaron (approved):

andaaron (approved):
lgtm

andaaron (approved):

rchincha (approved):
lgtm

PR #1674, opened 2023-08-03
refactor(cli): added equivalent subcommands for each flag combination under every command

laurentiuNiculae:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


pkg/cli/images_sub_cmd.go
```diff
@@ -0,0 +1,300 @@
...
+	var defaultVal T
+
+	if err != nil {
+		return defaultVal
+	}
+
+	return out
+}
+
+func GetConfigOptions(cmd *cobra.Command) (bool, bool) {
```
peusebiu, line 237: this name is too broad for its purpose

pkg/cli/image_cmd.go
```diff
@@ -129,11 +129,12 @@ func setupImageFlags(imageCmd *cobra.Command, searchImageParams map[string]*stri
...
-	imageCmd.Flags().BoolVar(debug, "debug", false, "Show debug output")
+	imageCmd.PersistentFlags().StringVar(servURL, "url", "", "Specify zot server URL if config-name is not mentioned")
+	imageCmd.PersistentFlags().StringVarP(user, "user", "u", "",
+		`User Credentials of zot server in "username:password" format`)
+	imageCmd.PersistentFlags().StringVarP(outputFormat, "output", "o", "", "Specify output format [text/json/yaml]")
+	imageCmd.PersistentFlags().BoolVar(verbose, "verbose", false, "Show verbose output")
+	imageCmd.PersistentFlags().BoolVar(debug, "debug", false, "Show debug output")
 }
 
 func searchImage(searchConfig searchConfig) error {
```
peusebiu, line 140: is this still used?

laurentiuNiculae: Yes they are used by the command, we are keeping it until removed

pkg/cli/cves_sub_cmd.go
```diff
@@ -0,0 +1,205 @@
...
+
+	return fixedTagsCmd
+}
+
+func RetryWithContext(ctx context.Context, operation func() error, maxRetries int,
+	delay time.Duration,
+) error {
+	err := operation()
+
+	for attempt := 0; err != nil && attempt < maxRetries; attempt++ {
```
rchamarthy, line 191: shouldnt attempt be initialized to 1 as an attempt has already been made? Or we actually attempt first and then check error and retries inside the loop. As i read this code it looks like we will actually attempt maxRetries+1 times.

pkg/cli/cves_sub_cmd.go
```diff
@@ -0,0 +1,205 @@
...
+		timer := time.NewTimer(delay)
+
+		select {
+		case <-timer.C:
+			break
```
rchamarthy, line 196: Instead of creating a new timer in each iteration its probably best if we use a Ticker instead for this use case. Look at: https://gobyexample.com/tickers

pkg/cli/cves_sub_cmd.go
```diff
@@ -0,0 +1,205 @@
...
+
+				return err
+			}, maxRetries, cveDBRetryInterval*time.Second)
+		},
+	}
+
+	return fixedTagsCmd
+}
+
+func RetryWithContext(ctx context.Context, operation func() error, maxRetries int,
```
rchamarthy, line 186: It probably is best to move this into a library so that anyone can use this to do a Retry operation.

pkg/cli/discover.go
```diff
@@ -25,7 +29,7 @@ type schemaList struct {
 			} `json:"queryType"` //nolint:tagliatelle // graphQL schema
 		} `json:"__schema"` //nolint:tagliatelle // graphQL schema
```
rchamarthy, line 30: avoid nolint or disable this linter.

pkg/cli/discover.go
```diff
@@ -99,3 +135,68 @@ func checkExtEndPoint(config searchConfig) bool {
...
+
+	searchEndPoint, _ := combineServerAndEndpointURL(*config.servURL, constants.FullSearchPrefix)
+
+	schemaQuery := `
+        {
+            __schema() {
+                queryType {
+                    fields {
+                        name
+						args {
```
rchamarthy, line 181: is this indentation intentional?

rchamarthy (changes requested):

pkg/cli/cmdflags/flags.go
```diff
@@ -0,0 +1,13 @@
...
+const (
+	URLFlag          = "url"
+	ConfigFlag       = "config"
+	UserFlag         = "user"
+	OutputFormatFlag = "output"
+	FixedFlag        = "fixed"
+	VerboseFlag      = "verbose"
+	VersionFlag      = "version"
+	DebugFlag        = "debug"
+	SearchedCVEID    = "id"
```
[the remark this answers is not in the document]
laurentiuNiculae: Changing with flag would mean breaking compatibility with the old version. This is why I didn't update the name.

pkg/cli/discover.go
andaaron: I believe the comment was about mixing tabs with spaces in this specific case

pkg/cli/client.go
```diff
@@ -136,7 +136,7 @@ func doHTTPRequest(req *http.Request, verifyTLS bool, debug bool,
 
 		bodyBytes, _ := io.ReadAll(resp.Body)
 
-		return nil, errors.New(string(bodyBytes)) //nolint: goerr113
+		return nil, errors.New(resp.Status + " " + string(bodyBytes)) //nolint: goerr113
```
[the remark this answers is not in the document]
laurentiuNiculae: Yes. A new error needed to be created. I've wrapped the extra information around it and removed the `//nolint: goerr113`

pkg/cli/cves_sub_cmd.go
laurentiuNiculae: Yes, I agree. There are many useful pieces of code the inside `cli` package that can be moved into their own packages but that we can't yet easily factor out. When we'll drop support for the old command format (using the flags and the searchers in the backend) it will be possible to move them.

pkg/cli/cves_sub_cmd.go
laurentiuNiculae: In case we'll want to add a exponential backoff delay I think we should use the `time.After(delay)` function. It's conveying the idea of a delay a bit better in this context, a ticker makes me think of a small operation repeated every tick but our operation can take several seconds.

pkg/cli/root.go
```diff
@@ -186,6 +187,11 @@ func NewServerRootCmd() *cobra.Command {
 	return rootCmd
 }
 
+const (
```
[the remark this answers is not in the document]
laurentiuNiculae: There weren't used, I removed them

pkg/cli/discover.go
```diff
@@ -99,3 +135,68 @@ func checkExtEndPoint(config searchConfig) bool {
...
+	_, err = makeGETRequest(ctx, discoverEndPoint, username, password, *config.verifyTLS,
+		*config.debug, &discoverResponse, config.resultWriter)
+	if err != nil {
+		return err
+	}
+
+	searchEnabled := false
+
+	for _, extension := range discoverResponse.Extensions {
+		if extension.Name == "_zot" {
```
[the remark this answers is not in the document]
laurentiuNiculae: I've added one

pkg/cli/images_sub_cmd.go
```diff
@@ -0,0 +1,297 @@
...
+
+	"github.com/briandowns/spinner"
+	"github.com/spf13/cobra"
+
+	zerr "zotregistry.io/zot/errors"
+	"zotregistry.io/zot/pkg/cli/cmdflags"
+	zcommon "zotregistry.io/zot/pkg/common"
+)
+
+func NewImageListCommand(searchService SearchService) *cobra.Command {
```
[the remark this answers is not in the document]
laurentiuNiculae: From what I saw in different codebases that use cobra they tend to call the subcommands without the prefix "sub". I don't know if it adds much more clarity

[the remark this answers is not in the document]
andaaron: OK

pkg/cli/cves_sub_cmd.go
rchamarthy: Ack.

pkg/cli/cves_sub_cmd.go
laurentiuNiculae: https://github.com/project-zot/zot/issues/1741

pkg/cli/discover.go
andaaron: @laurentiuNiculae can you please double check this?

Issue #1653, opened 2023-07-26
[Feat]: ability to run behind reverse proxy


dmaes:
### Is your feature request related to a problem? Please describe.

First of all, thanks for the fantastic work!

I would like to be able to run zot behind a reverse proxy/loadbalancer. I'm only just trying out zot, and at first it seemed to run fine behind a reverse proxy, until I tried enabling openid, and it used `http.address:http.port` in it's redirect_uri, which will obviously fail if you're using something like `http: { address: 0.0.0.0, port: 8443 }` and you're actually accessing it via `https://zot.example.com` via loadbalancer.

### Describe the solution you'd like

Either a config variable for zot to use when building url's, or zot using the `X-Forwarded-Host` header would be nice.

dmaes:
I was looking at the `getRelyingPartyArgs` method in `pkg/api/authn.go` (where to openid redirect uri is build), and it seems it's also basing the scheme on the tls config, which also would not work if one would run zot on http, and proxy on https.

rchincha:
@dmaes Thanks for trying out `zot` and the feedback.

As an alternative, you can deploy `zot` via helm with an ingress and see if that is an option for you.

Meanwhile, @peusebiu can you look at this?

dmaes:
I'm using ldap for authentication in the meantime. That doesn't need to build any (redirect) uri's, and is working fine for the moment (I'm also running bare-metal, not on k8s).

I also saw there are docs an running zot behind a loadbalancer (https://github.com/project-zot/project-zot.github.io/blob/main/docs/articles/clustering.md), so actually this is more a bug in the oidc setup (and possibly other places) then a feature request.

rchincha:
@dmaes pls take a look at PR #1675 and see if it would solve your issue.

dmaes:
@rchincha looks good on first sight, I'll test it next week. Thanks.

andaaron:
This issue has been fixed in #1675, if it not, please reopen.

PR #1675, opened 2023-08-03
fix(authn): handle the case where zot with openid runs behind a proxy

peusebiu:
added a new config option under 'http' called externalURL which is used by openid/oauth2 clients to redirect back to zot

**What type of PR is this?**
bug

#1653 

add the ability for zot, configured with openid auth, to run behind a proxy

**Automation added to e2e**:

```release-note

```


andaaron (approved):

examples/config-openid.json
```diff
@@ -7,6 +7,7 @@
   "http": {
     "address": "127.0.0.1",
     "port": "8080",
+    "externalUrl": "http://127.0.0.1:8080",
```
rchincha, line 10: Should this be "externalAddress", can we not detect and set the http scheme `http` or `https`?

andaaron: Let's not detect. I'm thinking this could be the address of a load balancer, or some kubernetes node.

andaaron (approved):

Issue #1418, opened 2023-05-03
[Feat]: propagate error msgs to the client (but be mindful of OCI dist-spec)


rchincha:
### Is your feature request related to a problem? Please describe.

`zot` performs several validations and when validations fail, we rely on the HTTP status code, so far so good. However, would be more useful if we propagate the detailed error msg back to client.

https://github.com/opencontainers/distribution-spec/blob/main/spec.md#error-codes

https://github.com/project-zot/zot/issues/1411
^ also see

tripathysagar:
Please, assign me this issue. And suggest me a rough roadmap to follow.

rchincha:
Thanks @tripathysagar 

All "external" interactions happen via pkg/api/routes.go. So would do a deep dive from there and improve the error msgs instead of just returning the high-level error code.

tripathysagar:
Hello @rchincha , I am not able to do much progress on this issue due to prior commitments. Please mark the ticket as unassigned so that anybody else can pick it up.

andaaron:
This should be resolved starting with https://github.com/project-zot/zot/pull/1681

PR #1681, opened 2023-08-07
feat: propagate detailed error msgs to client (OCI dist-spec format)

adodon2go:
**What type of PR is this?**
feature

https://github.com/project-zot/zot/issues/1418

Adds jsonSchema validation errors from image-spec to errors.details returned to the client: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#error-codes

Added UTs

**Automation added to e2e**:

No

No

### DIFF
--- a/.github/workflows/oci-conformance-action.yml
+++ b/.github/workflows/oci-conformance-action.yml
@@ -36,8 +36,6 @@ jobs:
           echo "SERVER_URL=http://${IP}:8080" >> $GITHUB_ENV
     - uses: actions/checkout@v3
       with:
-        # TODO: change to upstream once the foloowing PR is merged:
-        # https://github.com/opencontainers/distribution-spec/pull/436
         repository: opencontainers/distribution-spec
         ref: main
         path: distribution-spec
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -1,16 +1,57 @@
 package errors
 
-import "errors"
+import (
+	"errors"
+)
+
+type Error struct {
+	err     error
+	details map[string]string
+}
+
+func (e *Error) Error() string {
+	return e.err.Error()
+}
+
+func (e *Error) Is(target error) bool {
+	return errors.Is(e.err, target)
+}
+
+func (e *Error) AddDetail(key, value string) *Error {
+	e.details[key] = value
+
+	return e
+}
+
+func (e *Error) GetDetails() map[string]string {
+	return e.details
+}
+
+func NewError(err error) *Error {
+	return &Error{
+		err:     err,
+		details: GetDetails(err), // preserve details if chained error
+	}
+}
+
+func GetDetails(err error) map[string]string {
+	var internalErr *Error
+	details := make(map[string]string)
+
+	if errors.As(err, &internalErr) {
+		details = internalErr.GetDetails()
+	}
+
+	return details
+}
 
 var (
 	ErrBadConfig                      = errors.New("config: invalid config")
 	ErrCliBadConfig                   = errors.New("cli: bad config")
 	ErrRepoNotFound                   = errors.New("repository: not found")
-	ErrRepoIsNotDir                   = errors.New("repository: not a directory")
 	ErrRepoBadVersion                 = errors.New("repository: unsupported layout version")
 	ErrManifestNotFound               = errors.New("manifest: not found")
 	ErrBadManifest                    = errors.New("manifest: invalid contents")
-	ErrBadIndex                       = errors.New("index: invalid contents")
 	ErrUploadNotFound                 = errors.New("uploads: not found")
 	ErrBadUploadRange                 = errors.New("uploads: bad range")
 	ErrBlobNotFound                   = errors.New("blob: not found")
--- a/pkg/api/errors/errors.go
+++ b/pkg/api/errors/errors.go
@@ -5,16 +5,10 @@ import (
 )
 
 type Error struct {
-	Code        string      `json:"code"`
-	Message     string      `json:"message"`
-	Description string      `json:"description"`
-	Detail      interface{} `json:"detail,omitempty"`
-}
-
-func (e Error) WithMessage(msg string) Error {
-	e.Message = msg
-
-	return e
+	Code        string            `json:"code"`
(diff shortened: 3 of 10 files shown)


pkg/storage/common/common.go
```diff
@@ -80,7 +83,7 @@ func ValidateManifest(imgStore storageTypes.ImageStore, repo, reference, mediaTy
 		if err := ValidateManifestSchema(body); err != nil {
 			log.Error().Err(err).Msg("OCIv1 image manifest schema validation failed")
 
-			return "", zerr.ErrBadManifest
+			return "", zerr.ErrWrapKey(zerr.ErrBadManifest, SchemaErrKey, err.Error())
```
andaaron, line 86: How about using 
```suggestion
			return "", fmt.Errorf(%w (%w)", zerr.ErrBadManifest, err)
```

rchincha:
https://github.com/opencontainers/distribution-spec/blob/main/spec.md#error-codes
^ per this spec, "details" is a single string right?

adodon2go:
> https://github.com/opencontainers/distribution-spec/blob/main/spec.md#error-codes ^ per this spec, "details" is a single string right?

What I understood from the spec is that details is an unstructured object (can be in every format- not just a string). So we can use a JSON format for output too.
What we currently have in the code and does not respect the spec output is the [description field](https://github.com/project-zot/zot/blob/main/pkg/api/errors/errors.go#L10). I think we need to remove that and merge _message+description_ into a single _message_ field

pkg/storage/common/common.go
adodon2go: This can work, although not extensible (like custom wrap and unwrap on demand).

peusebiu:
you also need to apply same logic for linter error.

this is where we wrap lint error with a message: https://github.com/project-zot/zot/blob/main/pkg/extensions/lint/lint.go#L108

this is where we take the message out and write it to the http response: https://github.com/project-zot/zot/blob/main/pkg/api/routes.go#L705

rchincha:
In pkg/api/routes.go, we are basically looking for:

IsErrBadManifest(err) -> if err is in the class of ErrBadManifest errors, then raise this code
and details are in err

In pkg/..., we are looking at this pattern:
err := someLibraryCallForSomeTask()
return errors.New("some validation failed") ... the actual error is probably not to be exposed.

.github/workflows/oci-conformance-action.yml
```diff
@@ -36,7 +36,7 @@ jobs:
           echo "SERVER_URL=http://${IP}:8080" >> $GITHUB_ENV
     - uses: actions/checkout@v3
       with:
-        # TODO: change to upstream once the foloowing PR is merged:
+        # TODO: change to upstream once the folowing PR is merged:
```
andaaron, line 39: Actually these comments should be removed, that PR was merged and we are already pointing to upstream.

pkg/storage/common/common.go
```diff
@@ -83,7 +85,7 @@ func ValidateManifest(imgStore storageTypes.ImageStore, repo, reference, mediaTy
 		if err := ValidateManifestSchema(body); err != nil {
 			log.Error().Err(err).Msg("OCIv1 image manifest schema validation failed")
 
-			return "", zerr.ErrBadManifest
+			return "", zerr.ErrBadManifest.AddDetail("jsonSchemaValidation", err.Error())
```
andaaron, line 88: OK, but if we do this, don't we modify the global variable `zerr.ErrBadManifest`? I think we would need a new instance with the same error string, and we should also be able to compare that new error instance with `zerr.ErrBadManifest` in order to see if they have the same return value for .Error(). Maybe overwrite the `Is()` function for the comparison.

A separate note: We should have a similar return at line 127 in case of index validation failures. Also line 791 would need an update.

pkg/api/errors/errors.go
```diff
@@ -102,9 +96,7 @@ func NewError(code ErrorCode, detail ...interface{}) Error {
 		MANIFEST_INVALID: {
 			Message: "manifest invalid",
 			Description: `During upload, manifests undergo several checks ensuring
-			validity. If those checks fail, this error MAY be returned, unless a more
-			specific error is included. The detail will contain information the failed
-			validation.`,
+			validity. If those checks fail, this error MAY be returned.`,
```
andaaron, line 99: Why change the text?

pkg/api/routes.go
```diff
@@ -1088,22 +1109,23 @@ func (rh *RouteHandler) DeleteBlob(response http.ResponseWriter, request *http.R
...
-			zcommon.WriteJSON(response,
-				http.StatusNotFound,
-				apiErr.NewErrorList(apiErr.NewError(apiErr.NAME_UNKNOWN, map[string]string{"name": name})))
+			details["name"] = name
+			e := apiErr.NewError(apiErr.NAME_UNKNOWN).AddDetail(map[string]string{"name": name})
+			zcommon.WriteJSON(response, http.StatusNotFound, apiErr.NewErrorList(e))
 		} else if errors.Is(err, zerr.ErrBlobNotFound) {
-			zcommon.WriteJSON(response,
-				http.StatusNotFound,
-				apiErr.NewErrorList(apiErr.NewError(apiErr.BLOB_UNKNOWN, map[string]string{".String()": digest.String()})))
```
adodon2go, line 1102: Here and below I introduced the side effect of changing key from ".String()" -> digest.
Let me know if for some reason this is not needed.

errors/errors.go
```diff
@@ -1,16 +1,57 @@
...
+
+	return e
+}
+
+func (e *Error) GetDetails() map[string]string {
+	return e.details
+}
+
+func NewError(err error) *Error {
+	return &Error{
```
andaaron, line 31: Maybe just verify the `err` parameter isn't already of our `Error` type, and if it is reuse the details value? I'm thinking some developers may start using this function everywhere not knowing the object they are passing already has a details property set.

andaaron (commented):
I think this solution is just what we were looking for.

andaaron (dismissed):

andaaron (approved):

rchincha (approved):
lgtm

PR #1682, opened 2023-08-08
refactor(authz): use a struct for user access control info operations

peusebiu:
fix(authz): fix isAdmin not using groups to determine if a user is admin.

**What type of PR is this?**
bug && refactor

**Automation added to e2e**:

```release-note

```


pkg/api/authz.go
```diff
@@ -268,24 +259,24 @@ func BaseAuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
...
 					authFail(response, request, ctlr.Config.HTTP.Realm, ctlr.Config.HTTP.Auth.FailDelay)
 
 					return
 				}
 
-				identity = acCtx.Username
+				identity = userAc.GetUsername()
 			}
 
 			if request.TLS != nil {
```
andaaron, line 282: Is there any reason for this code to be in authz.go instead of authn.go? Authz should read already existing user information from userAc.

pkg/api/authz.go
```diff
@@ -161,33 +157,28 @@ func (ac *AccessController) getUserGroups(username string) []string {
 	return groupNames
 }
 
-// getContext updates an AccessControlContext for a user/anonymous and returns a context.Context containing it.
-func (ac *AccessController) getContext(acCtx *localCtx.AccessControlContext, request *http.Request) context.Context {
-	readGlobPatterns := ac.getGlobPatterns(acCtx.Username, acCtx.Groups, Read)
-	dmcGlobPatterns := ac.getGlobPatterns(acCtx.Username, acCtx.Groups, DetectManifestCollision)
+// getContext updates an UserAccessControl with admin status and specific permissions on repos.
+func (ac *AccessController) updateUserAccessControl(userAc *reqCtx.UserAccessControl) {
+	readGlobPatterns := ac.getGlobPatterns(userAc.GetUsername(), userAc.GetGroups(), Read)
```
andaaron, line 162: I think all types of patterns and actions should be available to the NewUserAccessControl() in case we need other types of permissions per extensions.

pkg/api/authn.go
```diff
@@ -396,19 +407,11 @@ func (amw *AuthnMiddleware) TryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
...
 				if allowAnonymous || isMgmtRequested {
-					ctx := getReqContextWithAuthorization("", []string{}, request)
-					*request = *request.WithContext(ctx) //nolint:contextcheck
-
 					next.ServeHTTP(response, request)
 
 					return
 				}
 			} else if allowAnonymous || isMgmtRequested {
-				// try anonymous auth only if basic auth/session was not given
```
andaaron, line 407: I would keep the comment.

pkg/api/authz.go
peusebiu: yes, this is an edge case where authn is basically skipped, and user is authenticated using tls, by gorilrla package.

pkg/extensions/search/resolver_test.go
```diff
@@ -3584,15 +3584,13 @@ func TestExpandedRepoInfo(t *testing.T) {
...
-		authzCtxKey := localCtx.GetContextKey()
-		acCtxUser := localCtx.AccessControlContext{
-			ReadGlobPatterns: map[string]bool{
-				"repo": false,
-			},
-			Username: "user",
-		}
+		userAc := reqCtx.NewUserAccessControl()
+		userAc.SetUsername("user")
+		userAc.SetReadGlobPatterns(map[string]bool{
```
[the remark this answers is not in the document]
peusebiu: yes, currently we care only about reads, we don't need writes, because we don't modify resources in extensions. we just read them.

pkg/api/authz.go
peusebiu: no :(

andaaron: Why not?

peusebiu: because in this case there is no authn.go.

this is the case where we have authz and tls mutual authentication and no authn(ldap/htpasswd/openid enabled.
auth is made by gorrilla, and in authz we take the identity from verified chains from the request.

our authn.go is not USEd as a middleware at all.

pkg/requestcontext/user_access_control.go
```diff
@@ -0,0 +1,225 @@
...
+	return allowed
+}
+
+func (uac *UserAccessControl) SaveOnRequest(request *http.Request) {
+	uacContext := context.WithValue(request.Context(), GetContextKey(), *uac)
+
+	*request = *request.WithContext(uacContext)
+}
+
+func (uac *UserAccessControl) SaveOnContext(ctx context.Context) context.Context {
```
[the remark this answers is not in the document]
peusebiu: it takes a context as argument, and on top of it it saves a new value on it. in our case userAc struct.

in case we have other values on the arg context, they will not be lost.

pkg/api/authz.go
andaaron: It's still not clear to me why we can't do add and additional check to see if certificate authentication is enabled somewhere in https://github.com/project-zot/zot/blob/main/pkg/api/authn.go#L230, and if it is, and use this logic there.

And if certificate authentication is enabled apply the authentication handler on the router as in all other cases when authentication is enabled.

andaaron:
I re-reviewed the PR in current state, and don't have additional comments.

andaaron (dismissed):

andaaron (approved):
Looks good.

rchincha (approved):
lgtm

PR #1690, opened 2023-08-10
test(refactor): refactor tests that use zot-test to use smaller images

### LINKED ISSUE #1664: [Test]:  replace large Centos based images used in CVE testing with smaller alpine-based images
This will save disk space when running tests, and also make copying images faster during the tests.

### LINKED ISSUE #1665: [Test] Review all places where we copy images from test/data, and use the new image builders instead
This will decrease the test duration.

laurentiuNiculae:
**What type of PR is this?**

Closes #1664 
Closes #1665 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/routes_test.go
+++ b/pkg/api/routes_test.go
@@ -27,6 +27,7 @@ import (
 	"zotregistry.io/zot/pkg/api"
 	"zotregistry.io/zot/pkg/api/config"
 	"zotregistry.io/zot/pkg/api/constants"
+	"zotregistry.io/zot/pkg/log"
 	mTypes "zotregistry.io/zot/pkg/meta/types"
 	localCtx "zotregistry.io/zot/pkg/requestcontext"
 	storageTypes "zotregistry.io/zot/pkg/storage/types"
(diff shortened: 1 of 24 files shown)


pkg/api/routes_test.go
```diff
@@ -1134,7 +1135,7 @@ func TestRoutes(t *testing.T) {
 
 			status = testUpdateBlobUpload(
 				[]struct{ k, v string }{
```
andaaron, line 1137: Are you sure we can hardcode these digests? In the past we've had problems since different skopeo versions on different developer environments did the conversion to oci format differently, resulting in different hashes.

pkg/extensions/search/cve/trivy/scanner_test.go
```diff
@@ -24,6 +25,45 @@ import (
 	"zotregistry.io/zot/pkg/test/mocks"
 )
 
+func TestMain(t *testing.T) {
```
andaaron, line 28: What is this function testing besides what is already tested?

andaaron (approved):

pkg/extensions/search/cve/trivy/scanner_test.go
laurentiuNiculae: I've updated the name. Since I've removed zot-test from pretty much every test, it's not scanned anymore and there was a line that was missing from coverage, it's related to the output of Trivy's scan report. To be sure everything was as before, i've added this scan separately.

pkg/api/routes_test.go
laurentiuNiculae: I think in these tests the digest is just formally put there for the digest check, it's not actually used. The image store is being mocked. I tried running with changes to the digest and it passes.

andaaron (approved):

andaaron (approved):
Let's merge this when the jobs complete.

PR #1691, opened 2023-08-11
feat(authn): add generic oidc and allow customizable name

babs:
**What type of PR is this?**

feature

none

This PR allow admin to add generic OIDC provider and customize it's name (linked with zui: https://github.com/project-zot/zui/pull/375)

Won't break on upgrade / downgrades

It allows the display of a customizable login button with the proper patch on zui (PR https://github.com/project-zot/zui/pull/375)

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -228,7 +228,7 @@ To configure zot as a client in dex (assuming zot is hosted at 127.0.0.1:8080), 
 staticClients:
   - id: zot-client
     redirectURIs:
-      - 'http://127.0.0.1:8080/auth/callback/dex'
+      - 'http://127.0.0.1:8080/auth/callback/oidc'
     name: 'zot'
     secret: ZXhhbXBsZS1hcHAtc2VjcmV0
 ```
@@ -240,7 +240,8 @@ zot can be configured to use dex with:
     "auth": {
       "openid": {
         "providers": {
-          "dex": {
+          "oidc": {
+            "name": "Corporate SSO",
             "clientid": "zot-client",
             "clientsecret": "ZXhhbXBsZS1hcHAtc2VjcmV0",
             "keypath": "",
@@ -253,7 +254,7 @@ zot can be configured to use dex with:
   }
 ```
 
-To login using openid dex provider use http://127.0.0.1:8080/auth/login?provider=dex
+To login using openid dex provider use http://127.0.0.1:8080/auth/login?provider=oidc
 
 NOTE: Social login is not supported by command line tools, or other software responsible for pushing/pulling
 images to/from zot.
@@ -313,7 +314,9 @@ To activate API keys use:
 ```
   "http": {
     "auth": {
-      "apikey: true
+      "apikey": true
+    }
+  }
 ```
 
 ##### How to create an API Key
@@ -384,6 +387,8 @@ Should authentication fail, to prevent automated attacks, a delayed response can
   "http": {
     "auth": {
       "failDelay": 5
+    }
+  }
 ```
 
 ## Identity-based Authorization
@@ -473,7 +478,7 @@ The number of workers for the task scheduler has the default value of runtime.Nu
 ```
  "scheduler": {
         "numWorkers": 3
-    
+ }
 ```
 
 ## Logging
--- a/examples/config-openid.json
+++ b/examples/config-openid.json
@@ -34,7 +34,8 @@
             "clientsecret": "client_secret",
             "scopes": ["openid", "read_api", "read_user", "profile", "email"]
           },
-          "dex": {
+          "oidc": {
+            "name": "Corporate SSO",
             "issuer": "http://127.0.0.1:5556/dex",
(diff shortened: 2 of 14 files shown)


pkg/api/config/config.go
```diff
@@ -64,6 +64,7 @@ type OpenIDConfig struct {
 }
 
 type OpenIDProviderConfig struct {
+	Name         string
```
andaaron, line 67: Where is this set? Can we have an example configuration file added to the PR?
For example an update of https://github.com/project-zot/zot/blob/main/examples/config-openid.json

babs: Done, I've added an example in [config-openid.json](https://github.com/project-zot/zot/blob/main/examples/config-openid.json)

babs: and yes @rchincha  it's related to https://github.com/project-zot/zui/pull/375 :)

babs:
fix commit message (doc(authn) -> docs(authn))

pkg/api/config/config.go
```diff
@@ -17,8 +17,8 @@ var (
 	BinaryType string //nolint: gochecknoglobals
 	GoVersion  string //nolint: gochecknoglobals
 
-	openIDSupportedProviders = [...]string{"google", "gitlab", "dex"} //nolint: gochecknoglobals
-	oauth2SupportedProviders = [...]string{"github"}                  //nolint: gochecknoglobals
+	openIDSupportedProviders = [...]string{"google", "gitlab", "dex", "oidc"} //nolint: gochecknoglobals
```
rchincha, line 20: @peusebiu any reason we need to keep "dex" around. It could be just another "oidc" instance.

babs: The reason was to facilitate transition

rchincha: @babs we are in a rc window, so changes to be expected. Thanks for the thoughtful code though.

babs: @rchincha would you like me to remove dex ?
I'll have to update zui's PR as well.
Not a big deal at all, I'll just mention this conversation as reference

peusebiu: yes, please remove dex, we don't need it, we can just configure it inside oidc.

There are a bunch of tests which should be updated to use 'oidc' instead of 'dex'

rchincha: @babs pls fix the above comment, we can merge your PR after.

rchincha: @babs let us know if you plan to make the change, else we can merge this and patch on top also.

babs: I didn't had time to setup a vm dedicated to tests as described in the github action.
Do you have an expected schedule ?

babs: I wanted to play all the integration test to make sure I don't break anything but I think I won't have the time for.
I think merge this and patch on top is the wisest option.

examples/config-openid.json
```diff
@@ -33,11 +33,18 @@
             "clientsecret": "client_secret",
             "scopes": ["openid", "read_api", "read_user", "profile", "email"]
           },
-          "dex": {
+          "oidc": {
```
rchincha, line 36: @peusebiu oh! are we allowed to do this?

rchincha: If mutiple oidc backends to be supported, then this config model wont work for us.
We need a "name": { "type": <>...} model

babs: I agree but this would be heavier rework and generalize in zui too.

That said in lots of cases(*) you have one source of external truth (or a mix of local + one external provider). [* not all cases of courses]
Having multiple might lead to identity overlapping [depends if you use email or username as reference in your oidc], unless you add the auth source in the acl too.

babs: > @peusebiu oh! are we allowed to do this?

Did I forgot to remove my other example entry ?

babs: fixed, my bad

examples/README.md
```diff
@@ -238,7 +238,7 @@ zot can be configured to use dex with:
     "auth": {
       "openid": {
         "providers": {
-          "dex": {
+          "oidc": {
             "clientid": "zot-client",
```
andaaron, line 242: Add the new "name" key to the doc as well.

andaaron (commented):
@peusebiu please review this PR

pkg/api/config/config.go
andaaron: Please rebase and squash these into a single commit (note you need to sign all your commits after rebase). We need the branch up to date in order for GH to run the tests.

babs: rebased and squashed

babs:
rebased and squashed

rchincha:
@babs thank you.

andaaron (approved):

peusebiu (approved):
Lgtm

PR #1711, opened 2023-08-18
feat: automated detection of OS and ARCH if unset

adodon2go:
**What type of PR is this?**
feature

When user tries "make binary" or "make binary-minimal" on darwin for example, the Makefile correctly detects OS and ARCH and creates a proper binary instead of default one for linux/amd64

Manual testing on Linux amd64, Mac OS Ventura with Intel chipset,  FreeBSD v13.2 amd64.
To be tested on Mac OS with M1 chipset.

No

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,8 +29,9 @@ SWAGGER_VERSION := v1.8.12
 STACKER := $(TOOLSDIR)/bin/stacker
 BATS := $(TOOLSDIR)/bin/bats
 TESTDATA := $(TOP_LEVEL)/test/data
-OS ?= linux
-ARCH ?= amd64
+OS ?= $(shell go env GOOS)
+ARCH ?= $(shell go env GOARCH)
+
 BENCH_OUTPUT ?= stdout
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -29,8 +29,11 @@ SWAGGER_VERSION := v1.8.12
 STACKER := $(TOOLSDIR)/bin/stacker
 BATS := $(TOOLSDIR)/bin/bats
 TESTDATA := $(TOP_LEVEL)/test/data
-OS ?= linux
-ARCH ?= amd64
+#OS ?= $(shell case "$$(uname)" in (Darwin) echo "darwin";; (FreeBSD) echo "freebsd";; (*) echo "linux";; esac;)
```
adodon2go, line 32: @rchincha , I've left two ways of detecting OS & ARCH (one is commented).
We should discuss which one is preferred.

andaaron: For the record, my choice would be to use `go env`

rchincha: remove the commented autodetect code and push.

rchincha (approved):
lgtm

PR #1719, opened 2023-08-21
fix(parse): lock storage while reading using image store

laurentiuNiculae:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/meta/parse.go
+++ b/pkg/meta/parse.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"time"
 
 	godigest "github.com/opencontainers/go-digest"
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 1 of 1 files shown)


pkg/meta/parse.go
```diff
@@ -44,13 +45,19 @@ func ParseStorage(metaDB mTypes.MetaDB, storeController storage.StoreController,
...
+	imageStore.Lock(&lockLatency)
+
 	indexBlob, err := imageStore.GetIndexContent(repo)
 	if err != nil {
 		log.Error().Err(err).Str("repository", repo).Msg("load-repo: failed to read index.json for repo")
 
 		return err
 	}
 
+	imageStore.Unlock(&lockLatency)
```
andaaron, line 59: We should defer and unlock even in case of error.

pkg/meta/parse.go
```diff
@@ -247,6 +254,10 @@ func getCosignSignatureLayersInfo(
 	}
 
 	for _, layer := range manifestContent.Layers {
+		var lockLatency time.Time
+
+		imageStore.Lock(&lockLatency)
```
andaaron, line 259: same comment as above

pkg/meta/parse.go
andaaron: I talked with Laurentiu, and it looks like in this case we want to unlock right away, not when the function exits.

andaaron (approved):

pkg/meta/parse.go
```diff
@@ -44,13 +45,20 @@ func ParseStorage(metaDB mTypes.MetaDB, storeController storage.StoreController,
 func ParseRepo(repo string, metaDB mTypes.MetaDB, storeController storage.StoreController, log log.Logger) error {
 	imageStore := storeController.GetImageStore(repo)
 
+	var lockLatency time.Time
```
peusebiu, line 48: why not lock the entire function? and not just parts, repo can be modified between reading its index.json here and checking its signatures leading to errors, right?

peusebiu: If it's running before server, and before Task scheduler, then it's fine :D

andaaron (approved):

PR #1727, opened 2023-08-24
fix(authz): return 401 instead of 403

rchincha:
403 is correct as per HTTP spec
However authz is not part of dist-spec and clients know only about 401 So this is a compromise.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


rchincha:
@raulkele thoughts on this change? basically if you provide creds and access is not allowed you get a 403 else if no creds were provided, you get a 401

raulkele:
From zui's point of view, I think the change is fine, better to follow the spec more correctly.

peusebiu (dismissed):
lgtm

test/blackbox/pushpull_authn.bats
```diff
@@ -0,0 +1,145 @@
...
+    [ "$status" -eq 0 ]
+    run regctl image copy ocidir://${TEST_DATA_DIR}/golang:1.20 localhost:8080/test-regclient
+    [ "$status" -eq 0 ]
+}
+
+@test "pull image with regclient" {
+    run regctl image copy localhost:8080/test-regclient ocidir://${TEST_DATA_DIR}/golang:1.20
+    [ "$status" -eq 0 ]
+}
+@test "pull image with regclient" {
```
andaaron, line 73: This seems to be a duplicate test.

test/blackbox/pushpull_authn.bats
```diff
@@ -0,0 +1,145 @@
...
+    run regctl artifact list localhost:8080/artifact-ref:demo --format raw-body
+    [ "$status" -eq 0 ]
+    [ $(echo "${lines[-1]}" | jq '.manifests | length') -eq 0 ]
+    run regctl artifact put --annotation  demo=true --annotation format=oci --artifact-type "application/vnd.example.icecream.v1" --subject localhost:8080/artifact-ref:demo << EOF
+test reference
+EOF
+    [ "$status" -eq 0 ]
+}
+
+@test "pull OCI artifact references with regclient" {
```
andaaron, line 119: This seems to just list artifacts, not pull them.

pkg/api/controller_test.go
```diff
@@ -3867,7 +3867,7 @@ func TestAuthorizationWithOnlyAnonymousPolicy(t *testing.T) {
 			Put(baseURL + "/v2/" + TestRepo + "/manifests/0.0.2")
 		So(err, ShouldBeNil)
 		So(resp, ShouldNotBeNil)
-		So(resp.StatusCode(), ShouldEqual, http.StatusForbidden)
+		So(resp.StatusCode(), ShouldEqual, http.StatusUnauthorized)
```
andaaron, line 3870: Update comment above the request too, please.

pkg/api/controller_test.go
```diff
@@ -3757,7 +3757,7 @@ func TestAuthorizationWithOnlyAnonymousPolicy(t *testing.T) {
 		resp, err = resty.R().Post(baseURL + "/v2/" + TestRepo + "/blobs/uploads/")
 		So(err, ShouldBeNil)
 		So(resp, ShouldNotBeNil)
-		So(resp.StatusCode(), ShouldEqual, http.StatusForbidden)
+		So(resp.StatusCode(), ShouldEqual, http.StatusUnauthorized)
```
andaaron, line 3760: Update comment above the request too, please.

andaaron (commented):
I can't say I agree 100%, but it make some sense that anonymous access without matching the configured permissions for anonymous users could also result in 401 instead of 403.

Also the bats tests cover authenticated users if I understood correctly. While the go code changes I understood are for the case I mentioned above.

test/blackbox/pushpull_authn.bats
```diff
@@ -0,0 +1,145 @@
...
+  }
+}
+EOF
+    zot_serve ${zot_config_file}
+    wait_zot_reachable "http://127.0.0.1:8080/v2/_catalog"
+}
+
+@test "push image with regclient" {
+    run regctl registry set localhost:8080 --tls disabled
+    run regctl registry login localhost:8080 -u test -p test123
```
rchincha, line 63: ^ this is the big change, making sure **with** client auth enabled, things don't break

andaaron: Understood.

test/blackbox/pushpull_authn.bats
```diff
@@ -0,0 +1,145 @@
...
+  "distSpecVersion":"1.1.0-dev",
+  "storage":{
+    "dedupe": true,
+    "gc": true,
+    "gcDelay": "1h",
+    "gcInterval": "6h",
+    "rootDirectory": "${zot_root_dir}"
+  },
+  "http": {
+		"address": "127.0.0.1",
```
adodon2go, line 30: format json file

test/blackbox/pushpull_authn.bats
```diff
@@ -0,0 +1,145 @@
...
+        }
+      },
+      "adminPolicy": {
+        "users": ["admin"],
+        "actions": ["read", "create", "update", "delete"]
+      }
+    }
+  },
+  "log":{
+    "level":"debug",
```
adodon2go, line 53: remove trailing ","

test/blackbox/pushpull_authn.bats
```diff
@@ -0,0 +1,145 @@
...
+    "gcInterval": "6h",
+    "rootDirectory": "${zot_root_dir}"
+  },
+  "http": {
+		"address": "127.0.0.1",
+		"port": "8080",
+    "realm":"zot",
+    "auth": {
+      "htpasswd": {
+        "path": ${zot_htpasswd_file}
```
peusebiu, line 35:
```suggestion
        "path": "${zot_htpasswd_file}"
```

test/blackbox/pushpull_authn.bats
```diff
@@ -0,0 +1,145 @@
+load helpers_cloud
+
+function setup() {
```
peusebiu, line 3:
```suggestion
function setup_file() {
```
just to make sure login is used in all tests?

test/blackbox/pushpull_authn.bats
```diff
@@ -0,0 +1,145 @@
...
+    # Verify prerequisites are available
+    if ! verify_prerequisites; then
+        exit 1
+    fi
+
+    # Setup zot server
+    local zot_root_dir=${BATS_FILE_TMPDIR}/zot
+    local zot_config_file=${BATS_FILE_TMPDIR}/zot_config.json
+    local zot_htpasswd_file=${BATS_FILE_TMPDIR}/zot_htpasswd
+    htpasswd -Bbn test test123 >> ${zot_htpasswd_file}
```
peusebiu, line 13: should be added into verify_prerequisites()

test/blackbox/pushpull_authn.bats
andaaron: Actually this is starting a new zot server for every test. Is this intended? If not, I think Sebi is right.
And the change should also be done in the `teardown()` -> `teardown_file()`

test/blackbox/pushpull_authn.bats
```diff
@@ -0,0 +1,145 @@
+load helpers_cloud
+
```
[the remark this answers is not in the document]
andaaron: Let's update the CI make target in https://github.com/project-zot/zot/pull/1746, if we can merge that before this PR.

PR #1728, opened 2023-08-26
feat(mgmt): mgmt extention no longer depends on UI being enabled

andaaron:
It is now enabled based only on search configuration


rchincha (approved):
lgtm

PR #1731, opened 2023-08-28
fix: bats test refactoring

adodon2go:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


test/blackbox/anonymous_policy.bats
```diff
@@ -1,8 +1,17 @@
-load helpers_pushpull
+load helpers_zot
+
+function verify_prerequisites {
+    if [ ! -f ${ZOT_PATH} ]; then
+        echo "you need to build ${ZOT_PATH} before running the tests" >&3
+        return 1
+    fi
+
+    return 0
```
andaaron, line 9: Don't we need to add a check for skopeo here, and in the other files using the skopeo command?

test/blackbox/annotations.bats
```diff
@@ -1,21 +1,44 @@
...
+    if [ ! $(command -v jq) ]; then
+        echo "you need to install jq as a prerequisite to running the tests" >&3
+        return 1
+    fi
+
+    if [ ! $(command -v podman) ]; then
+        echo "you need to install podman as a prerequisite to running the tests" >&3
+        return 1
+    fi
+
```
andaaron, line 23: Don't we need to add checks or skopeo and stacker?

test/blackbox/pushpull.bats
```diff
@@ -1,8 +1,27 @@
...
+
+    if [ ! $(command -v curl) ]; then
+        echo "you need to install curl as a prerequisite to running the tests" >&3
+        return 1
+    fi
+
+    if [ ! $(command -v jq) ]; then
+        echo "you need to install jq as a prerequisite to running the tests" >&3
+        return 1
+    fi
```
andaaron, line 17: There's the missing tooling we should check for, helm, regctl, oras and so on?

test/blackbox/pushpull.bats
```diff
@@ -146,6 +167,9 @@ function teardown_file() {
     local chart_version=$(awk '/version/{printf $2}' ${BATS_FILE_TMPDIR}/helm-charts/charts/zot/Chart.yaml)
     run helm pull oci://localhost:8080/zot-chart/zot --version ${chart_version}
     [ "$status" -eq 0 ]
+    local chart_version=$(awk '/version/{printf $2}' ${BATS_FILE_TMPDIR}/helm-charts/charts/zot/Chart.yaml)
```
andaaron, line 170: Is is a duplicate line to 167.

test/blackbox/metadata.bats
```diff
@@ -1,8 +1,27 @@
...
+    if [ ! $(command -v curl) ]; then
+        echo "you need to install curl as a prerequisite to running the tests" >&3
+        return 1
+    fi
+
+    if [ ! $(command -v jq) ]; then
+        echo "you need to install jq as a prerequisite to running the tests" >&3
+        return 1
+    fi
+
```
andaaron, line 18: We should add a check for skopeo.

test/blackbox/cve.bats
```diff
@@ -1,9 +1,33 @@
...
+    if [ ! $(command -v curl) ]; then
+        echo "you need to install curl as a prerequisite to running the tests" >&3
+        return 1
+    fi
 
+    if [ ! $(command -v jq) ]; then
+        echo "you need to install jq as a prerequisite to running the tests" >&3
+        return 1
+    fi
+
```
andaaron, line 23: We should check for skopeo.

test/blackbox/detect_manifest_collision.bats
```diff
@@ -1,8 +1,17 @@
-load helpers_pushpull
+load helpers_zot
+
+function verify_prerequisites {
+    if [ ! -f ${ZOT_PATH} ]; then
+        echo "you need to build ${ZOT_PATH} before running the tests" >&3
+        return 1
+    fi
+
```
andaaron, line 8: We should check for skopeo and regctl.

test/blackbox/referrers.bats
```diff
@@ -1,8 +1,27 @@
...
+    if [ ! $(command -v curl) ]; then
+        echo "you need to install curl as a prerequisite to running the tests" >&3
+        return 1
+    fi
+
+    if [ ! $(command -v jq) ] &>/dev/null; then
+        echo "you need to install jq as a prerequisite to running the tests" >&3
+        return 1
+    fi
+
```
andaaron, line 18: Check for skopeo and oras?

test/blackbox/pushpull_running_dedupe.bats
```diff
@@ -1,8 +1,27 @@
...
+    if [ ! $(command -v curl) ]; then
+        echo "you need to install curl as a prerequisite to running the tests" >&3
+        return 1
+    fi
+
+    if [ ! $(command -v jq) ]; then
+        echo "you need to install jq as a prerequisite to running the tests" >&3
+        return 1
+    fi
+
```
andaaron, line 18: Check for skopeo, oras, helm, regctl.

test/blackbox/sync_replica_cluster.bats
```diff
@@ -20,11 +20,11 @@ function setup_file() {
 
     cat >${zot_sync_one_config_file} <<EOF
 {
-    "distSpecVersion": "1.1.0",
+    "distSpecVersion": "1.1.0-dev",
     "storage": {
         "rootDirectory": "${zot_sync_one_root_dir}"
     },
-    "http": {
+    "http": {config
```
andaaron, line 27: Is this correct?

andaaron (commented):
I agree with consolidating the helpers libraries, but I think it would be simpler to add all the checks for all required tooling there instead of individual .bats files. Why move it to the .bats files?

test/blackbox/helpers_pushpull.bash
```diff
@@ -1,84 +0,0 @@
...
-TEST_DATA_DIR=${ROOT_DIR}/test/data/
-OS="${OS:-linux}"
-ARCH="${ARCH:-amd64}"
-ZOT_PATH=${ROOT_DIR}/bin/zot-${OS}-${ARCH}
-
-mkdir -p ${TEST_DATA_DIR}
-
-function verify_prerequisites {
-    if [ ! -f ${BATS_RUN_TMPDIR}/.firstrun ]; then
-        env | grep proxy >&3
```
andaaron, line 11: Used to show the proxy, can you put it with `-i`

test/blackbox/annotations.bats
adodon2go: This tools are already checked/installed if run through a Makefile target. If run as a standalone BATS script we need to make sure versions (ex: cosign.notation) are the same as the one installed by Makefile. This adds extra complexity that is probably not needed. We can add a note at the beginning of each file that states just that. Besides that there are some tests that did not pass locally anyway (that ones that requires crictl for ex or the sync one that syncs images from dockerhub - probably limitation of unpaid version)

andaaron (approved):

rchincha (approved):
lgtm

PR #1747, opened 2023-09-01
fix: add manifest validation checks

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/common/common.go
+++ b/pkg/storage/common/common.go
@@ -95,7 +95,17 @@ func ValidateManifest(imgStore storageTypes.ImageStore, repo, reference, mediaTy
 			return "", zerr.ErrBadManifest
 		}
 
-		if manifest.Config.MediaType == ispec.MediaTypeImageConfig {
+		// validate blobs only for known media types
+		if manifest.Config.MediaType == ispec.MediaTypeImageConfig ||
+			manifest.Config.MediaType == ispec.MediaTypeEmptyJSON {
+			// validate config blob - a lightweight check if the blob is present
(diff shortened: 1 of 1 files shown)


pkg/storage/common/common.go
```diff
@@ -95,7 +95,17 @@ func ValidateManifest(imgStore storageTypes.ImageStore, repo, reference, mediaTy
 			return "", zerr.ErrBadManifest
 		}
 
-		if manifest.Config.MediaType == ispec.MediaTypeImageConfig {
+		// validate blobs only for known media types
+		if manifest.Config.MediaType == ispec.MediaTypeImageConfig ||
+			manifest.Config.MediaType == ispec.MediaTypeEmptyJSON {
```
andaaron, line 100: I thought there is no blob uploaded for ispec.MediaTypeEmptyJSON, as the data is in the descriptor itself.
What's the point of having images with ispec.MediaTypeEmptyJSON if they still have a `{}` config blob separately?

rchincha: That is still an actual blob of size=2. Hence nothing special in terms of uploads and existence.

andaaron: If that was correct, why go though all the overhead of including it in the spec? It doesn't add any benefit, compared to me defining my own {} config blob media type and it would be ingested just the same according to the spec.

andaaron (approved):

pkg/storage/common/common.go
rchincha: Convention still matters.

Issue #1412, opened 2023-05-02
[Bug]: "minimal" zot is very heavy dependency-heavy


rogpeppe:
### zot version

5fcf01d8b8c3fa62fac87681e40120aec27fc17b

### Describe the bug

Building a minimal zot binary is slower and more fragile than it could be. This is largely due to a large number of dependencies used by the zot extensions.

On my machine, running `make binary-minimal` with a hot cache (no binary but otherwise all dependencies already installed), takes ~6s elapsed time and produces a 65MB binary. Cgo is required, making cross-compilation not possible.  It's not possible to build with just `go build ./cmd/zot`. Building from scratch took over 2 minutes.

The `go.mod` file lists 58 direct module dependencies and 389 indirect dependencies. Most of those dependencies are not necessary to run the core zot functionality. The extra modules also affect the security profile of zot through use of the `unsafe` package and cgo.

As an experiment, I removed all the extensions. This resulted in a (working) binary that builds in ~1s elapsed time and is 38MB. No cgo dependence. Building from scratch, including downloading all modules, takes under 30s elapsed time.

Perhaps it would be possible to create a more minimal zot module that only has the dependencies strictly needed for the core functionality, and a module that builds on that to add extensions. That could also have a nice side effect of ensuring a strict separation between "extension" and "core" where currently there are quite a few places in the core that depend directly on extension packages, and also potentially making it possible for third parties to implement their own independent extensions.

### To reproduce

```sh
make binary-minimal
```

### Expected behavior

Smaller binaries; less dependencies.

rchincha:
@rogpeppe This is fantastic work. We will take a closer look at pruning minimal build further.

rchincha:
_full_ binary:
* has 416 package deps
* 136MB size

_minimal_ binary:
* has 264 package deps
* 64MB size

rchincha:
Curious about your comment about cgo - we build with "env CGO_ENABLED=0"

rogpeppe:
> Curious about your comment about cgo - we build with "env CGO_ENABLED=0", and we do cross-compile and that is how we publish our binaries.

That's a good point, sorry! I hadn't noticed the `-tags containers_image_openpgp` flag on the build command. It's a shame that the `github.com/containers/image/v5/signature` package doesn't use the `cgo` tag to build correctly when `CGO_ENABLED=0`.

> Do you have a PR you can share?

[...]

PR #1758, opened 2023-09-04
refactor: Reduce binary size of zot-minimal; Added CI check for binar…

adodon2go:
…y size

**What type of PR is this?**
refactoring,cleanup.ci-cd

https://github.com/project-zot/zot/issues/1412

1. Reduce binary size of zot minimal
2. Add a check in ci-cd if binary size increases with more than 1%, then PR sanity will fail.
Then we can decide if binary increase is legit or not.

**Automation added to e2e**:

No

### DIFF
--- a/.github/workflows/zot-minimal-size.yml
+++ b/.github/workflows/zot-minimal-size.yml
@@ -1,0 +1,40 @@
+name: "zot minimal binary size"
+on:
+  pull_request:
+    branches: [main]
+
+permissions: read-all
+
+jobs:
+  zot-minimal-size:
+    runs-on: ubuntu-latest
+    name: compare-binary-size
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
+        with:
+          cache: false
+          go-version: 1.20.x
+      - name: Check if zot-minimal binary increased with more than 1%
+        run: |
+          echo "Building zot-minimal and check size"
+          cd $GITHUB_WORKSPACE
+          make binary-minimal
+          BINSIZE=$(stat -c%s "bin/zot-linux-amd64-minimal")
+
+          echo "Building zot-minimal on main branch and check size"
+          mkdir -p zot_main
+          git clone https://github.com/project-zot/zot zot_main/
+          cd zot_main
+          make binary-minimal
+          BINSIZE_MAIN=$(stat -c%s "bin/zot-linux-amd64-minimal")
+          cd $GITHUB_WORKSPACE && rm -rf zot_main
+
+          [[ $BINSIZE -gt $BINSIZE_MAIN ]] || exit 0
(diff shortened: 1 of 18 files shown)


adodon2go:
Example of failed check (increased binary size of zot minimal):
https://github.com/project-zot/zot/actions/runs/6076437060/job/16484441353?pr=1758

```
PR changes increased size of zot-minimal binary
PR binary size: 65757184 Bytes
main branch binary size: 33316864 Bytes
zot minimal binary increased by 97.36% comparing with main
Error: Process completed with exit code 1.
```

pkg/api/config/config.go
```diff
@@ -221,17 +222,32 @@ func SameFile(str1, str2 string) (bool, error) {
 	return os.SameFile(sFile, tFile), nil
 }
 
+func deepCopy(src, dst interface{}) error {
+	bytes, err := json.Marshal(src)
+	if err := inject.Error(err); err != nil {
+		return err
```
andaaron, line 228: It is strange that this line and the other check above are reported as not covered.

For the record I think `zotregistry.io/zot/pkg/test/inject` is one big hack and I would avoid importing test code from under `zotregistry.io/zot/pkg/test/` into production code.

Can you try to avoid using `zotregistry.io/zot/pkg/test/inject`? Can you try using a channel to produce errors? See https://github.com/project-zot/zot/blob/main/pkg/common/common_test.go#L34

andaaron (approved):

.github/workflows/zot-minimal-size.yml
```diff
@@ -0,0 +1,40 @@
+name: "zot minimal binary size"
+on:
+  pull_request:
+    branches: [main]
+
+permissions: read-all
+
+jobs:
+  zot-minimal-size:
```
rchincha, line 9: Why not just make this a step in ci-cd.yml itself?

rchincha: I like this check btw.

adodon2go: > Why not just make this a step in ci-cd.yml itself?

It was a step in ci-cd.yml initialy. Team decision was to add a new workflow instead for two reasons:
1. Fail faster in case zot minimal binary increases in PR instead of waiting other steps from ci-cd.yml to complete
2. Not add an extra 2 min in ci-cd workflow and wait longer for it to finish

andaaron (approved):

.github/workflows/zot-minimal-size.yml
andaaron: An additional argument to keep it separate.
3. If it fails in ci-cd.yaml the `build-test / Build and test ZOT (linux, amd64) (pull_request)` GH check would be red, Having a separate check makes it more easy for reviewer to assess the state of the PR.

PR #1759, opened 2023-09-05
[POC] feat(graphql): return CVE score and data souce of score

andaaron:
CVE scores and severities depend on the data source which can be nvd, redhat, and so on.
We are now exposing this data, so the UI can consume it.

There are issues with the data itself, which need to be addressed before we decide to introduce this API.


andaaron:
Will open a new PR if needed.

PR #1770, opened 2023-09-06
refactor: Review metrics endpoints

adodon2go:
**What type of PR is this?**
refactor

Using existing metrics tests

**Automation added to e2e**:

Export metrics for minimal binary on /metrics


pkg/api/routes.go
```diff
@@ -179,17 +179,11 @@ func (rh *RouteHandler) SetupRoutes() {
...
-		// This logic needs to be reviewed, it should depend on build options
-		// not the general presence of the extensions in config
-		if rh.c.Config.Extensions == nil {
-			// minimal build
-			prefixedRouter.HandleFunc("/metrics", rh.GetMetrics).Methods("GET")
-		} else {
-			// extended build
-			ext.SetupMetricsRoutes(rh.c.Config, rh.c.Router, authHandler, rh.c.Log)
-		}
+	if rh.c.Config.BinaryType == "minimal" {
```
andaaron, line 182: We don't need to make this check here if the route is enabled by ext.SetupMetricsRoutes(), which will be selected from the correct file at build time.

pkg/extensions/extension_metrics_disabled.go
andaaron, line 1: Equivalent of `prefixedRouter.HandleFunc("/metrics", rh.GetMetrics).Methods("GET")` would need to be moved to this file, for the minimal build. And it wouldn't be conditioned by a config, I guess. Unless we want to start looking at the extensions config for the minimal build.

andaaron (approved):

andaaron (approved):

PR #1777, opened 2023-09-08
fix(api): Fix 'last' query param for <repo>/tags/list to work without param 'n'

andaaron:
Also fix additional issues:
- sorting of tags on calls without pagination parameters ('n' or 'last')
- if 'n' is 0 we should return an empty list and not error

Added tests accordingly


rchincha (approved):
lgtm

PR #1778, opened 2023-09-08
feat(ldap): add option to load ldap from file

laurentiuNiculae:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/config/config.go
+++ b/pkg/api/config/config.go
@@ -121,19 +121,45 @@ type SchedulerConfig struct {
 	NumWorkers int
 }
 
+type LDAPCredentials struct {
+	BindDN       string
+	BindPassword string
+}
+
 type LDAPConfig struct {
+	CredentialsFile    string
 	Port               int
 	Insecure           bool
 	StartTLS           bool // if !Insecure, then StartTLS or LDAPs
 	SkipVerify         bool
 	SubtreeSearch      bool
 	Address            string
-	BindDN             string
+	bindDN             string `json:"-"`
+	bindPassword       string `json:"-"`
 	UserGroupAttribute string
-	BindPassword       string
 	BaseDN             string
 	UserAttribute      string
 	CACert             string
+}
+
+func (ldapConf *LDAPConfig) BindDN() string {
+	return ldapConf.bindDN
+}
+
+func (ldapConf *LDAPConfig) SetBindDN(bindDN string) *LDAPConfig {
+	ldapConf.bindDN = bindDN
+
+	return ldapConf
+}
+
+func (ldapConf *LDAPConfig) BindPassword() string {
+	return ldapConf.bindPassword
+}
+
+func (ldapConf *LDAPConfig) SetBindPassword(bindPassword string) *LDAPConfig {
+	ldapConf.bindPassword = bindPassword
+
+	return ldapConf
 }
 
 type LogConfig struct {
@@ -266,14 +292,14 @@ func (c *Config) Sanitize() *Config {
 		panic(err)
 	}
 
-	if c.HTTP.Auth != nil && c.HTTP.Auth.LDAP != nil && c.HTTP.Auth.LDAP.BindPassword != "" {
+	if c.HTTP.Auth != nil && c.HTTP.Auth.LDAP != nil && c.HTTP.Auth.LDAP.bindPassword != "" {
 		sanitizedConfig.HTTP.Auth.LDAP = &LDAPConfig{}
 
 		if err := DeepCopy(c.HTTP.Auth.LDAP, sanitizedConfig.HTTP.Auth.LDAP); err != nil {
 			panic(err)
 		}
 
-		sanitizedConfig.HTTP.Auth.LDAP.BindPassword = "******"
+		sanitizedConfig.HTTP.Auth.LDAP.bindPassword = "******"
 	}
 
 	return sanitizedConfig
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -53,6 +53,7 @@ import (
 	"zotregistry.io/zot/pkg/api/config"
 	"zotregistry.io/zot/pkg/api/constants"
 	apiErr "zotregistry.io/zot/pkg/api/errors"
+	"zotregistry.io/zot/pkg/cli/server"
 	"zotregistry.io/zot/pkg/common"
 	extconf "zotregistry.io/zot/pkg/extensions/config"
 	"zotregistry.io/zot/pkg/log"
@@ -1985,8 +1986,13 @@ func (l *testLDAPServer) Stop() {
 }
 
 func (l *testLDAPServer) Bind(bindDN, bindSimplePw string, conn net.Conn) (vldap.LDAPResultCode, error) {
-	if bindDN == "" || bindSimplePw == "" {
-		return vldap.LDAPResultInappropriateAuthentication, errors.ErrRequireCred
+	if bindSimplePw == "" {
+		switch bindDN {
+		case "bad-user", "cn=fail-user-bind,ou=test":
+			return vldap.LDAPResultInvalidCredentials, errors.ErrInvalidCred
+		default:
+			return vldap.LDAPResultSuccess, nil
+		}
 	}
 
 	if (bindDN == LDAPBindDN && bindSimplePw == LDAPBindPassword) ||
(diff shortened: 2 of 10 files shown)


pkg/api/config/config.go
```diff
@@ -102,6 +102,7 @@ type SchedulerConfig struct {
 }
 
 type LDAPConfig struct {
+	LDAPConfigFile     string
```
rchincha, line 105: LDAPCredentialsFile only pls, no need to put all config in that file.

andaaron: To clarify, credentials being "BindDN" and "BindPassword".

rchincha: yup, and that can be in JSON

pkg/api/config/config.go
```diff
@@ -102,18 +102,19 @@ type SchedulerConfig struct {
...
-	SkipVerify         bool
-	SubtreeSearch      bool
-	Address            string
-	BindDN             string
-	UserGroupAttribute string
-	BindPassword       string
-	BaseDN             string
-	UserAttribute      string
-	CACert             string
+	LDAPCredentialsFile string
```
andaaron, line 105:
```suggestion
	CredentialsFile string
```

pkg/cli/root.go
```diff
@@ -775,6 +780,47 @@ func LoadConfiguration(config *config.Config, configPath string) error {
...
+		return err
+	}
+
+	conf.HTTP.Auth.LDAP.BindDN = newLDAPCredentials.BindDN
+	conf.HTTP.Auth.LDAP.BindPassword = newLDAPCredentials.BindPassword
+
+	return nil
+}
+
+type LdapCredentials struct {
```
andaaron, line 803: Maybe this should be defined in config.go.

pkg/api/config/config.go
```diff
@@ -101,7 +101,13 @@ type SchedulerConfig struct {
...
 }
 
+type LDAPCredentials struct {
+	BindDN       string
+	BindPassword string
+}
+
 type LDAPConfig struct {
+	CredentialsFile    string
 	Port               int
```
rchincha, line 111: BindPassword string below should be removed right? In fact, isn't that the only credential?

pkg/cli/root.go
```diff
@@ -771,6 +776,42 @@ func LoadConfiguration(config *config.Config, configPath string) error {
 	return nil
 }
 
+func updateLDAPConfig(conf *config.Config) error {
+	if conf.HTTP.Auth == nil || conf.HTTP.Auth.LDAP == nil {
+		return nil
+	}
+
+	if conf.HTTP.Auth.LDAP.CredentialsFile == "" {
```
rchincha, line 784: This credentials file may be empty (or not specified in config)
But the rest is still valid. We could have anonymous LDAP binds?

pkg/api/config/config.go
andaaron: That is the password, the username is the other property.

And why remove it from here? I thought this feature was optional?

pkg/cli/root.go
andaaron: Why would the current implementation need changes? It doesn't make an update to the credentials in the config if the file is not specified?

pkg/api/config/config.go
andaaron: Optional in the sense the user may choose to keep his credentials in the config, or he may choose to save them in the credentials file.

rchincha: No, we always want separate credentials file

pkg/cli/root.go
rchincha: Similar to sync credentials or all other credentials

rchincha: CredentialsFile can be nil and add a test case to test LDAP code works after that.

pkg/api/config/config.go
```diff
@@ -121,16 +121,23 @@ type SchedulerConfig struct {
...
+	CredentialsFile    string
+	Unauthenticated    bool
 	Port               int
 	Insecure           bool
 	StartTLS           bool // if !Insecure, then StartTLS or LDAPs
 	SkipVerify         bool
 	SubtreeSearch      bool
 	Address            string
-	BindDN             string
+	BindDN             string `json:"-"`
```
rchincha, line 138: Remove these completely from here.

pkg/api/config/config.go
```diff
@@ -121,16 +121,23 @@ type SchedulerConfig struct {
 	NumWorkers int
 }
 
+type LDAPCredentials struct {
+	BindDN       string
+	BindPassword string
+}
+
 type LDAPConfig struct {
+	CredentialsFile    string
```
rchincha, line 130: This can be "" so expect that and test for it.

pkg/api/config/config.go
laurentiuNiculae: We can't remove them. We need a way to access their value from the config. How can we transmit their values through the config if we don't set them here?

rchincha: Can it become lowercase so limited in scope (or) put it behind Getter methods.

andaaron:
Can you also add an example configuration file?

pkg/api/ldap.go
```diff
@@ -215,11 +232,20 @@ func (lc *LDAPClient) Authenticate(username, password string) (bool, map[string]
...
-		lc.Log.Error().Err(err).Str("bindDN", userDN).Msg("user bind failed")
+	if password != "" {
+		err = lc.Conn.Bind(userDN, password)
+		if err != nil {
+			lc.Log.Error().Err(err).Str("bindDN", userDN).Msg("user bind failed")
 
-		return false, user, userGroups, err
+			return false, user, userGroups, err
+		}
+	} else {
```
andaaron, line 242: What use case are we supporting here? All users we are authenticating should have valid passwords?

pkg/api/ldap.go
```diff
@@ -127,7 +128,7 @@ func (lc *LDAPClient) Authenticate(username, password string) (bool, map[string]
 	lc.lock.Lock()
 	defer lc.lock.Unlock()
 
-	if password == "" {
+	if password == "" && !lc.Unauthenticated {
```
andaaron, line 131: Why do we need this change?

pkg/api/ldap.go
andaaron, line 1: I don't understand the changes in this file.

The 2 cases I would expect are:
A) The config can have a bind user and password mentioned in the credentials file. In that case we should use it to search for the user we intend to authenticate with. If that user exists, we should bind with his username and his password to verify the password as well.

B) If we don't have a credentials file, so no bind user or password (there may be a case for bind user without a bind password, do we need to support it?), we can do an anonymous bind to search for the user, and then we should bind with his username and his password to verify the password.

Maybe I misunderstand the changes, but why would we fail if the bind username or password is not specified? And in a separate case why should we allow authentication of LDAP users without passwords?

pkg/cli/server/root.go
```diff
@@ -752,6 +756,54 @@ func LoadConfiguration(config *config.Config, configPath string) error {
 	return nil
 }
 
+func updateLDAPConfig(conf *config.Config) error {
+	if conf.HTTP.Auth == nil || conf.HTTP.Auth.LDAP == nil {
+		return nil
+	}
+
+	if conf.HTTP.Auth.LDAP.Unauthenticated {
+		conf.HTTP.Auth.LDAP.SetBindDN("anonym-user")
```
andaaron, line 765: Why do we need this?

pkg/cli/server/root.go
```diff
@@ -752,6 +756,54 @@ func LoadConfiguration(config *config.Config, configPath string) error {
...
+
+		return nil
+	}
+
+	if conf.HTTP.Auth.LDAP.CredentialsFile == "" {
```
andaaron, line 770: If the credentials file is empty we should assume we need to do an anonymous bind to search for users instead?

pkg/api/controller_test.go
```diff
@@ -2077,6 +2099,332 @@ func TestBasicAuthWithLDAP(t *testing.T) {
...
+			"Auth": {
+				"LDAP": {
+					"CredentialsFile":    "%s",
+					"BaseDN":             "%v",
+					"UserAttribute":      "%v",
+					"UserGroupAttribute": "memberOf",
+					"Insecure":           true,
+					"Address":            "%v",
+					"Port":               %v,
+					"BindDN":             "%v",
```
andaaron, line 2342:
```suggestion
```

pkg/api/controller_test.go
```diff
@@ -2077,6 +2099,332 @@ func TestBasicAuthWithLDAP(t *testing.T) {
...
+					"BindPassword":       "%v"
```
andaaron, line 2343:
```suggestion
```

pkg/cli/server/root_test.go
```diff
@@ -1447,6 +1448,134 @@ func TestScrub(t *testing.T) {
...
+		server.SetArgs([]string{"serve", configPath})
+		So(func() { err = server.Execute() }, ShouldPanic)
+	})
+
+	Convey("unauthenticated LDAP config", t, func() {
+		tempDir := t.TempDir()
+		ldapConfigContent := "{}"
+		ldapConfigPath := filepath.Join(tempDir, "ldap.json")
+
+		err := os.WriteFile(ldapConfigPath, []byte(ldapConfigContent), 0o000)
```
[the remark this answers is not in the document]
laurentiuNiculae: We don't need it anymore

andaaron (approved):

rchincha (approved):
lgtm

Issue #1769, opened 2023-09-06
[Bug]: `zli` and `zb` are very large binaries


rchincha:
### zot version

v2.0.0-rc6

### Describe the bug

Currently, both `zli` and `zb` are very large binaries. Can we triage and trim them?

Filed based on external finding.
cc: @erikschul 

```
(https://cloud-native.slack.com/archives/C03EGRE4QGH/p1694009648800789)
What's the reason that the binaries are so large? e.g. zli 113MB, zot 133 MB
```

### To reproduce

1. Configuration
2. Client tool used
3. Seen error

rchincha:
zli changes fixed in https://github.com/project-zot/zot/pull/1805

rchincha:
Closing this issue.

PR #1783, opened 2023-09-11
refactor: Reduce zb binary size

adodon2go:
What type of PR is this?
cleanup

Which issue does this PR fix:
https://github.com/project-zot/zot/issues/1769

What does this PR do / Why do we need it:

If an issue # is not available please add repro steps and logs showing the issue:

Testing done on this change:

Automation added to e2e:

Will this break upgrades or downgrades?
No

Does this PR introduce any user-facing change?:
No


cmd/zb/helper.go
```diff
@@ -14,15 +16,20 @@ import (
...
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"gopkg.in/resty.v1"
 
 	zerr "zotregistry.io/zot/errors"
-	"zotregistry.io/zot/pkg/api"
-	"zotregistry.io/zot/pkg/test"
+	testc "zotregistry.io/zot/pkg/test/common"
 )
 
+type ImageTags struct {
```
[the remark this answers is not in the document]
adodon2go: Binary increases with only 4k. So we can use this instead.

cmd/zb/main.go
```diff
@@ -3,13 +3,14 @@ package main
...
 	"os"
 
-	distspec "github.com/opencontainers/distribution-spec/specs-go"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
 
 	"zotregistry.io/zot/pkg/api/config"
 )
 
+const DistSpecVersion = "1.1.0-dev"
```
rchincha, line 12: ./pkg/api/constants/consts.go

^ would put all this in one place

pkg/test/common/utils.go
```diff
@@ -0,0 +1,25 @@
+package common
+
+import (
+	"net/url"
+
+	"gopkg.in/resty.v1"
+)
+
+func Location(baseURL string, resp *resty.Response) string {
```
rchincha, line 9: Yes, we need to start pulling in only needed things. Although this goes against golang development model but then upstream folks don't necessarily follow pkg export discipline.

andaaron (approved):

PR #1788, opened 2023-09-11
chore(dependabot): increase the limit of dependabot PRs

andaaron:
Given the default limit is 5 and some packages like AWS release daily, having a weekly schedule results in the same packages being flagged for update most of the time.

Let's increase to 10 to make sure there are no issues in other libraries


rchincha (approved):
lgtm

PR #1790, opened 2023-09-11
fix(ci): Do not initialize image trust store on every convey

andaaron:
Should miticage failures such as https://github.com/project-zot/zot/actions/runs/6148316168/job/16681757801?pr=1788

### DIFF
--- a/pkg/meta/dynamodb/dynamodb_test.go
+++ b/pkg/meta/dynamodb/dynamodb_test.go
@@ -138,7 +138,7 @@ func TestWrapperErrors(t *testing.T) {
 
 	uuid, err := guuid.NewV4()
 	if err != nil {
-		panic(err)
+		t.Fatal(err)
 	}
 
 	repoMetaTablename := "RepoMetadataTable" + uuid.String()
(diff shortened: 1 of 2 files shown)


pkg/meta/dynamodb/dynamodb_test.go
```diff
@@ -151,6 +151,11 @@ func TestWrapperErrors(t *testing.T) {
 
 	log := log.NewLogger("debug", "")
 
+	imgTrustStore, err := imagetrust.NewAWSImageTrustStore(region, endpoint)
+	if err != nil {
+		panic(err)
```
rchincha, line 156: We have a logger here right? Can we log.Fatal() instead?

andaaron: This is a test, it needs to exit if it cannot create the object.

andaaron: Switch to using the test framework specific methods.

andaaron:
Closing this as nobody has experienced that failure since Andreea's fix.

PR #1799, opened 2023-09-12
fix(ci): print zot log on failure

peusebiu:
**What type of PR is this?**
bug

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/.github/workflows/gc-stress-test.yaml
+++ b/.github/workflows/gc-stress-test.yaml
@@ -24,6 +24,7 @@ jobs:
           go-version: 1.20.x
 
       - name: Run zb
+        id: bench
         run: |
             make binary
             make bench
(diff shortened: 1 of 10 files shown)


.github/workflows/gc-stress-test.yaml
```diff
@@ -35,6 +35,13 @@ jobs:
 
             # clean zot storage
             sudo rm -rf /tmp/zot
+        continue-on-error: true
+
+      - name: Check on failures
+        if: steps.bench.outcome != 'success' || steps.bench.outcome != 'success'
```
andaaron, line 41: Why are all these ifs double checking `steps.bench.outcome != 'success'`?

andaaron: Also in this specifi case, doesn't `bench` need to be initialized?

andaaron (approved):
LGTM.

PR #1803, opened 2023-09-12
bug: can't build zot with empty EXTENSIONS

adodon2go:
**What type of PR is this?**
bug

Makes possible to compile zot binary without any EXTENSIONS set. 
Size of that binary should equal size of zot minimal binary

> EXTENSIONS= make binary

Observe that makefile target fails

1. Tested binary-minimal & exporter-minimal targets
2. tested binary target with/without EXTENSIONS

**Automation added to e2e**:

No

No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,13 +46,13 @@ endif
 comma:= ,
 space := $(null) #
 hyphen:= -
-extended-name:=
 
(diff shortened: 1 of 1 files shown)


andaaron (approved):

rchincha (approved):
lgtm

PR #1809, opened 2023-09-13
feat: add image retention policy

rchincha:
It is very common for CI pipelines to emit large amounts of both tagged and untagged images. A image retention policy is valuable.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/examples/config-retention.json
+++ b/examples/config-retention.json
@@ -1,0 +1,17 @@
+{
+    "distSpecVersion": "1.1.0-dev",
+    "storage": {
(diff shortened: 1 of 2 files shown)


hallyn:
Support for a storage retention policy would be good.  However, what's here is, I believe, too simplistic to be useful anywhere.  Anyone storing images will want to do something like "for release:N.M, for any N, keep the latest M around forever.  for any release:N.M-cixxxxx, delete after  days.  Anything else keep for 3 weeks."

Issue #1125, opened 2023-01-19
[Bug]: Bearer auth doesn't work with docker and skopeo login


sagikazarmark:
### zot version

v1.4.3

### Describe the bug

When trying to use the bearer authn feature with an external authorization server and docker/skopeo login, I noticed that login always fails.

First, I suspected an incompatibility between Zot and the custom authorization server (https://github.com/distribution-auth/auth), but the authz server actually returned a valid token.

Then I started some manual testing checking out the code I think I figured out an incompatibility between the bearer auth implementation and docker/skopeo login.

Both docker/skopeo login go to the registry without a token first. That results in the following HTTP response from Zot:

```
❯ curl -vvv http://127.0.0.1:5001/v2/
*   Trying 127.0.0.1:5001...
* Connected to 127.0.0.1 (127.0.0.1) port 5001 (#0)
> GET /v2/ HTTP/1.1
> Host: 127.0.0.1:5001
> User-Agent: curl/7.85.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 401 Unauthorized
< Access-Control-Allow-Headers: Authorization,content-type
< Access-Control-Allow-Methods: HEAD,GET,POST,OPTIONS
< Access-Control-Allow-Origin: *
< Content-Type: application/json
< Www-Authenticate: Bearer realm="http://localhost:8080/token",service="localhost:5000",scope="repository::pull"
< Date: Thu, 19 Jan 2023 19:19:42 GMT
< Content-Length: 283
< 
* Connection #0 to host 127.0.0.1 left intact
{"errors":[{"code":"UNAUTHORIZED","message":"authentication required","description":"The access controller was unable to authenticate the client.\n\t\t\tOften this will be accompanied by a Www-Authenticate HTTP response header\n\t\t\tindicating how to authenticate.","detail":null}]}
```

Note the `Www-Authenticate` header.

Docker/skopeo are supposed to go to the authorization service (realm) and send an authorization request. The authorization service running at `localhost:8080` receives the request, but it looks like the scope parameter is not sent to the service upon login. When comparing Zot to the official Docker registry that's in fact the primary difference between the response of the two:

```
❯ curl -vvv localhost:5000/v2/
*   Trying 127.0.0.1:5000...
* Connected to localhost (127.0.0.1) port 5000 (#0)
> GET /v2/ HTTP/1.1
> Host: localhost:5000
> User-Agent: curl/7.85.0
> Accept: */*
> 
[...]

PR #1814, opened 2023-09-15
DO NOT MERGE: fix: allow empty scope on /v2

sagikazarmark:
**What type of PR is this?**

bug

#1125

This PR fixes the bearer auth implementation allowing registry clients to authenticate with an empty scope when calling /v2.

As said in [this](https://github.com/project-zot/zot/issues/1125#issuecomment-1397508479) comment, according to the distribution specification that most registry clients implement, any empty scope in the token is valid when a non-resource URL is called (essentially /v2).

However, the current implementation (that relies on chartmuseum/auth) always requires a scope to be present in order to accept a token. (Under the hood the library tries to compare the empty access request contained in the token to a bogus `repository::pull` scope. The repository name is empty and Zot [falls back to pull](https://github.com/project-zot/zot/blob/8e18917b070f3934aed31c4b43cddb923a27abea/pkg/api/authn.go#L473-L476))

After reading through the Chartmuseum auth spec [here](https://github.com/chartmuseum/auth-server-example), I came to the conclusion that it's fundamentally incompatible with the Docker auth spec.

The chartmuseum spec says the first step is always pulling or pushing a specific artifact, whereas the Docker spec does not require that. The detailed nature of that spec also suggests that it may build on the original docker spec, it's actually a different one.

This PR is just a demonstration that shows how Zot should behave correctly.

There are several potential solutions:

- If chartmuseum is willing to "fix" their spec/library to accept empty access list, we can easily fix the problem. However, I'm not sure that's in the best interest on Chartmuseum if they ever plan to use their auth spec (but it essentially seems unmaintained to me)
- Copy the relevant code from the distribution code base
- Write a whole new library

**Automation added to e2e**:

NO

```release-note

```

### DIFF
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -469,19 +470,53 @@ func bearerAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 
 				return
 			}
-
+			
 			action := auth.PullAction
 			if m := request.Method; m != http.MethodGet && m != http.MethodHead {
 				action = auth.PushAction
 			}
+			
+			var permissions *auth.Permission
 
-			permissions, err := authorizer.Authorize(header, action, name)
-			if err != nil {
-				ctlr.Log.Error().Err(err).Msg("issue parsing Authorization header")
-				response.Header().Set("Content-Type", "application/json")
-				zcommon.WriteJSON(response, http.StatusInternalServerError, apiErr.NewError(apiErr.UNSUPPORTED))
+			// Empty scope should be allowed according to the distribution auth spec
+			// This is only necessary for the bearer auth type
+			if request.RequestURI == "/v2/" && authorizer.Type == auth.BearerAuthAuthorizerType {
+				if header == "" {
+					permissions = &auth.Permission{
+						WWWAuthenticateHeader: fmt.Sprintf("Bearer realm=\"%s\",service=\"%s\",scope=\"\"", authorizer.Realm, authorizer.Service),
+					}
+				} else {
+					bearerTokenMatch := regexp.MustCompile("(?i)bearer (.*)")
 
-				return
+					signedString := bearerTokenMatch.ReplaceAllString(header, "$1")
+
+					// If the token is valid, our job is done
+					// Since this is the /v2 base path and we didn't pass a scope to the auth header in the previous step
+					// there is no access check to enforce
+					_, err := authorizer.TokenDecoder.DecodeToken(signedString)
+					if err != nil {
+						ctlr.Log.Error().Err(err).Msg("issue parsing Authorization header")
+						response.Header().Set("Content-Type", "application/json")
+						zcommon.WriteJSON(response, http.StatusInternalServerError, apiErr.NewError(apiErr.UNSUPPORTED))
+
+						return
+					}
+
+					permissions = &auth.Permission{
+						Allowed: true,
+					}
+				}
+			} else {
+				var err error
+
+				permissions, err = authorizer.Authorize(header, action, name)
+				if err != nil {
+					ctlr.Log.Error().Err(err).Msg("issue parsing Authorization header")
+					response.Header().Set("Content-Type", "application/json")
+					zcommon.WriteJSON(response, http.StatusInternalServerError, apiErr.NewError(apiErr.UNSUPPORTED))
+
+					return
+				}
 			}
 
 			if !permissions.Allowed {


seniorquico:
I encountered this testing zot with skopeo, and I agree- it seems the chartmuseum/auth library is broken (although, maybe as you suggest, it was simply designed for a different spec). Another callout- The chartmuseum/auth library _only_ supports RSA keys, which is small bummer for our use case (we've fully transitioned to ECC keys).

Taking inspiration from your comment, "copy the relevant code from the distribution code base", I went spelunking... It doesn't seem very complicated (although I'm sure dragons may lurk somewhere in there). The auth context is setup and the auth handler invoked here (which is invoked by their HTTP request middleware):

[func (app *App) authorized(w http.ResponseWriter, r *http.Request, context *Context) error](https://github.com/distribution/distribution/blob/4772604ae973031ab32dd9805a4bccf61d94909f/registry/handlers/app.go#L805-L870) ...

The function above calls two simple functions to generate the auth requirements from the URL:

1. [func appendAccessRecords(records []auth.Access, method string, repo string) []auth.Access](https://github.com/distribution/distribution/blob/4772604ae973031ab32dd9805a4bccf61d94909f/registry/handlers/app.go#L904-L936) ...
1. [func appendCatalogAccessRecord(accessRecords []auth.Access, r *http.Request) []auth.Access](https://github.com/distribution/distribution/blob/4772604ae973031ab32dd9805a4bccf61d94909f/registry/handlers/app.go#L938-L956) ...

Notably, when the request path is the base `/v2/`, there are no auth requirements added to the slice.

The bearer auth handler is contained in the `registry/auth/token` directory. When there is no scope, you can see the parameter is completely omitted from the challenge [(ref)](https://github.com/distribution/distribution/blob/4772604ae973031ab32dd9805a4bccf61d94909f/registry/auth/token/accesscontroller.go#L110-L112):

```go
if scope := ac.accessSet.scopeParam(); scope != "" {
	str = fmt.Sprintf("%s,scope=%q", str, scope)
}
```

I probably need to spend some more time reading through the Token Authentication Specification, but nothing jumps out to me to suggest the chartmuseum/auth library can't be replaced by distribution implementation. There would be behavioral changes for consumers, and that might be considered a breaking change. But given how the chartmuseum/auth library's implementation seems to be broken for modern distribution clients... 🤔

After some more reading, I might try to take a stab at pulling in the distribution implementation and run some tests. I wonder if the conformance workflow in this repo with the bearer auth configured would detect these issues?

seniorquico:
Before diving in to copying bits from the distribution code base, I thought I'd go searching for an alternative library (as an alternative to the "write a whole new library" suggestion)...

https://github.com/portward/registry-auth/

 😉👍

<details><summary>OCI research notes (for posterity)</summary>
For those following along that aren't in-sync with all of the OCI happenings...

The OCI Distribution Spec does _not_ address authentication/authorization. This is a known omission:

- https://github.com/opencontainers/distribution-spec/issues/240
- https://github.com/opencontainers/distribution-spec/issues/338

A working group was formed to attempt to define a standard (and a zot maintainer is a member):

- https://github.com/opencontainers/tob/pull/119

The meeting notes are available in a dedicated repo:

- [opencontainers/wg-auth](https://github.com/opencontainers/wg-auth)

Today, the Docker v2 Token Authentication Specification is the de facto standard, and the working group appears to be starting with it as the base. The working group appears to be exploring use cases/limitations and might expand on the Docker v2 Token Authentication Specification accordingly.

Notably, in the last meeting dated 2023-08-08, the question of creating a reference authentication library was raised. That would be cool. 👍
</details>

A question... Why "DO NOT MERGE" on this PR? If the OCI working group needs time to develop the long-term solution, couldn't the workaround proposed with this PR be sufficient to temporarily get zot working again?

sagikazarmark:
@seniorquico a couple thoughts:

This PR serves two purposes: it allowed me to try Zot with Portward (which is a project I created actually) and it also allowed to demonstrate what's wrong with the current implementation.

I suspect the solution either lies in "fixing" chartmuseum or not using it at all, but this PR is probably not the right answer (ie. I made this change in roughly 5 minutes without thinking about potential security issues it may introduce).

WG auth: it's just been formed and although there is progress, it's gonna take some time to come up with a proposal that addresses all the issues and requirements, gain feedback from the wider community and get it accepted as a formal spec. Even then, it's gonna take years to adopt it. So yes, a future OCI auth spec is the answer, but it's not going to happen over night.

Portward (the registry-auth lib) sits primarily on the other side of the aisle, service as a tool for authorization services. Although it _may_ be used in a registry as well, I haven't considered that use case when I wrote the library and it may not be the best fit for this use case. There may be bits and pieces though that could be reused. Just know that it's primarily targeted at authorization services and NOT at registries (which generally have different requirements, like verifying tokens instead of issuing them)

@rchincha might be able to provide some guidance here on what we should do next.

Should we attempt "fixing" the chartmuseum implementation? It's certainly worth a try talking to them, but my gut tells me (as I said before) that it's actually a different spec and therefore not required to be compatible with the registry auth spec.

Admittedly, this would result in a simpler resolution for Zot itself.

The alternative (as I said above) is replacing the implementation with a different one.

andaaron:
An updated version of this fix was merged in https://github.com/project-zot/zot/pull/2176.
You should be able to us the latest zot release.

PR #1816, opened 2023-09-18
Add cross-compile to dockerfile

sagikazarmark:
**What type of PR is this?**

feature

This PR adds cross-compilation to the Dockerfile.

Usage:

```shell
docker buildx build --platform linux/amd64 --platform linux/arm64 -f build/Dockerfile .
```

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,14 +1,16 @@
 # ---
 # Stage 1: Install certs, build binary, create default config file
 # ---
-FROM ghcr.io/project-zot/golang:1.20 AS builder
+FROM --platform=$BUILDPLATFORM ghcr.io/project-zot/golang:1.20 AS builder
+
+ARG TARGETOS
+ARG TARGETARCH
 ARG COMMIT
-ARG OS
-ARG ARCH
+
 RUN mkdir -p /go/src/github.com/project-zot/zot
 WORKDIR /go/src/github.com/project-zot/zot
 COPY . .
-RUN make COMMIT=$COMMIT OS=$OS ARCH=$ARCH clean binary
+RUN make COMMIT=$COMMIT OS=$TARGETOS ARCH=$TARGETARCH clean binary
 RUN echo '{\n\
     "storage": {\n\
         "rootDirectory": "/var/lib/registry"\n\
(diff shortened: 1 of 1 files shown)


rchincha:
@sagikazarmark I would clean up the Makefile. Pls take a look at https://github.com/project-zot/zot/pull/1817
The final solution may be a mix of both I suppose.

sagikazarmark:
@rchincha the problem is that you cannot build multi-arch images that way. It will always be a single-arch image which is kind of annoying:

- You can't distribute a single docker compose file for developers working on Linux and macOS at the same time
- You can't deploy Zot on Kube node groups with mixed instance types

rchincha:
@sagikazarmark pls take a look at some of the CI failures.

sagikazarmark:
@rchincha done

rchincha (dismissed):
lgtm

build/Dockerfile
```diff
@@ -1,14 +1,17 @@
 # ---
 # Stage 1: Install certs, build binary, create default config file
 # ---
-FROM ghcr.io/project-zot/golang:1.20 AS builder
+FROM --platform=$BUILDPLATFORM ghcr.io/project-zot/golang:1.20 AS builder
+
+ARG TARGETOS
+ARG TARGETARCH
+ARG TARGETPLATFORM
```
andaaron, line 8: I see TRAGETPLATFORM and a BUILDTLATFORM variables.
What is TRAGETPLATFORM used for, and how are we passing BUILDPLATFORM?
Didn't we need changes in the makefile?

sagikazarmark: They are passed by Docker/Buildkit. Buildplatform is where the image is being built (otherwise the builder would try emulating the image for the target platform), Targetplatform is well....the target platform used for cross-compilation.

No need to pass these build args manually.

sagikazarmark: I guess it can be removed, I don't see it used either.

andaaron: Looks ok, if it builds locally for your use case, I agree with merging this.

andaaron (approved):

rchincha (approved):
lgtm

Issue #1815, opened 2023-09-17
[Feat]: Add a ci test so UX queries are under 200ms


rchincha:
### Is your feature request related to a problem? Please describe.

200ms is the upper limit of acceptable UX response times.

When zot is deployed with UI enabled, ignoring the network latencies, zot itself cannot exceed this.
So the test would be to load zot with many repos and images and then run each test.

PR #1833, opened 2023-09-19
feat(cve): implement CVE scanning as background tasks

andaaron:
1. Move existing CVE DB download generator/task login under the cve package
2. Add a new CVE scanner task generator and task type to run in the background, as well as tests for it
3. Move the CVE cache in its own package
4. Add a CVE scanner methods to check if an entry is present in the cache, and to retreive the results
5. Modify the FilterTags MetaDB method to not exit on first error
This is needed in order to pass all tags to the generator,
instead of the generator stopping at the first set of invalid data
6. Integrate the new scanning task generator with the existing zot code.
7. Fix an issue where the CVE scan results for multiarch images was not cached
8. Rewrite some of the older CVE tests to use the new image-utils test package
9. Use the CVE scanner as attribute of the controller instead of CveInfo.
Remove functionality of CVE DB update from CveInfo, it is now responsible,
as the name states, only for providing CVE information.
10. The logic to get maximum severity and cve count for image sumaries now uses only the scanner cache.
11. Removed the GetCVESummaryForImage method from CveInfo as it was only used in tests

### DIFF
--- a/pkg/api/config/config.go
+++ b/pkg/api/config/config.go
@@ -348,6 +348,10 @@ func (c *Config) IsSearchEnabled() bool {
 	return c.Extensions != nil && c.Extensions.Search != nil && *c.Extensions.Search.Enable
 }
 
+func (c *Config) IsCveScanningEnabled() bool {
+	return c.IsSearchEnabled() && c.Extensions.Search.CVE != nil
+}
+
 func (c *Config) IsUIEnabled() bool {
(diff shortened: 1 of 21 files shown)


andaaron:
In today's implementation the scanner populates the cache, but the API calls are still blocking. 
An improvement would be to make the API calls non-blocking and to have them push new tasks for scanning images not present in cache.

andaaron:
Related to https://github.com/project-zot/zot/issues/1815

rchincha (approved):
lgtm

PR #1834, opened 2023-09-20
ci: update golangci-lint version

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


pkg/debug/gqlplayground/gqlplayground.go
```diff
@@ -27,7 +27,7 @@ func SetupGQLPlaygroundRoutes(router *mux.Router,
 
 	templ, err := template.ParseFS(playgroundHTML, "index.html.tmpl")
 	if err != nil {
-		log.Fatal().Err(err)
+		log.Fatal().Msg(err.Error())
```
andaaron, line 30:
```suggestion
		log.Fatal().Err(err).Msg("")
```
We want the err in a separate field for it to be consumed independently by 3rd party tooling.

andaaron (approved):

PR #1844, opened 2023-09-22
fix(dedupe): run dedupe only for repositories found at startup

peusebiu:
no need to run dedupe/restore blobs for images being pushed or synced while running dedupe task, they are already deduped/restored inline.

**What type of PR is this?**
bug

#1843 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/common/common.go
+++ b/pkg/storage/common/common.go
@@ -869,14 +869,37 @@ type DedupeTaskGenerator struct {
 	and generating a task for each unprocessed one*/
 	lastDigests []godigest.Digest
 	done        bool
+	repos       []string // list of repos on which we run dedupe
 	Log         zlog.Logger
 }
 
 func (gen *DedupeTaskGenerator) Next() (scheduler.Task, error) {
 	var err error
 
+	/* at first run get from storage currently found repositories so that we skip the ones that gets synced/uploaded
+	while this generator runs, there are deduped/restored inline, no need to run dedupe/restore again */
+	if len(gen.repos) == 0 {
+		gen.repos, err = gen.ImgStore.GetRepositories()
+		if err != nil {
+			//nolint: dupword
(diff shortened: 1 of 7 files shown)


pkg/storage/common/common.go
```diff
@@ -971,14 +971,36 @@ type DedupeTaskGenerator struct {
...
+		gen.repos, err = gen.ImgStore.GetRepositories()
+		if err != nil {
+			//nolint: dupword
+			gen.Log.Error().Err(err).Msg("dedupe rebuild: unable to to get list of repositories")
+
+			return nil, err
+		}
+	}
+
+	if len(gen.repos) == 0 {
```
andaaron, line 993: Bth `if`s have the same condition, Can we make it a single if?

peusebiu: I still need to check if there are still no repositories after running GetRepositories(), I moved it inside the first if.

andaaron: Understood. You should also set an empty slice as value for repos in the Reset() method

andaaron (dismissed):

rchincha:
Pls fix the conflict.

rchincha:
Pls fix conflicts.

andaaron (approved):

pkg/storage/common/common.go
```diff
@@ -869,14 +869,37 @@ type DedupeTaskGenerator struct {
...
 	Log         zlog.Logger
 }
 
 func (gen *DedupeTaskGenerator) Next() (scheduler.Task, error) {
 	var err error
 
+	/* at first run get from storage currently found repositories so that we skip the ones that gets synced/uploaded
+	while this generator runs, there are deduped/restored inline, no need to run dedupe/restore again */
+	if len(gen.repos) == 0 {
+		gen.repos, err = gen.ImgStore.GetRepositories()
```
rchincha, line 882: This list could be very large potentially

rchincha (approved):
lgtm

Issue #1685, opened 2023-08-10
[Feat] implement `Scrub` for S3 storage


andaaron:
At the moment `scrub` is only supported on local storage because it depends on the `umoci` library. We should also implement `scrub` for S3.

As part of #1582 `umoci` is getting replaced by logic internal to zot.
We should remove the dependency on `umoci` entirely.

rchincha:
Closing this since PR #1845 has been merged.

Issue #734, opened 2022-08-20
scrub: also create a media-type scrub method instead of just repo


rchincha:
**Is your feature request related to a problem? Please describe.**

As we add more media-types (now we also support manifest index which is a container of images), it needs to be scrubbed differently.

**Describe the solution you'd like**

foreach manifest in a repo:
   switch manifest.mediaType:
        case ImageManifest: // we already support this
        case ImageIndex: // unpack the manifest, iterate over each manifest and scrub them, this is NEW
        case ArtifactType: // something else, this is future

rchincha:
Related to issue https://github.com/project-zot/zot/issues/640

andaaron:
The image index case was covered in https://github.com/project-zot/zot/pull/1031.
The artifact case is postponed for now.

andaaron:
Since the artifact manifest was removed, the only remaining requirement is to also take the image manifest 'Subject' into consideration when running scrub.

andaaron:
Done in https://github.com/project-zot/zot/pull/1845

PR #1845, opened 2023-09-22
refactor(scrub): replace umoci logic in scrub implementation

Andreea-Lupu:
- implement scrub also for S3 storage by replacing umoci
- change scrub implementation for ImageIndex
- take the `Subject` into consideration when running scrub
- remove test code relying on the umoci library. Since we started
relying on images in test/data, and we create our own images using
go code we can obtain digests by other means. (cherry picked from commit https://github.com/andaaron/zot/commit/489d4e2d23c1b4e48799283f8281024bbef6123f)

**What type of PR is this?**

#1685 #734 

**Automation added to e2e**:

```release-note

```


pkg/storage/scrub.go
```diff
@@ -92,84 +88,171 @@ func CheckRepo(ctx context.Context, imageName string, imgStore storageTypes.Imag
...
+		results = append(results, imageResult)
+	}
+
+	return results, nil
+}
+
+func getManifestResult(
+	ctx context.Context, manifest ispec.Descriptor, imgStore storageTypes.ImageStore, imageName, tag string,
+	listOfManifests map[godigest.Digest]ScrubImageResult,
+) ScrubImageResult {
```
andaaron, line 131: Why do we need to return the results, if we add them to the map anyway?

pkg/storage/scrub.go
```diff
@@ -92,84 +88,171 @@ func CheckRepo(ctx context.Context, imageName string, imgStore storageTypes.Imag
...
-		results = append(results, imageResult)
-	}
+		return okResult
+	case ispec.MediaTypeImageManifest:
+		imgRes := CheckIntegrity(ctx, imageName, tag, manifest, imgStore)
+		listOfManifests[manifest.Digest] = imgRes
 
-	return results, nil
+		// if integrity ok then check subject if exists
+		if imgRes.Error == "" {
```
andaaron, line 187: This code should be cleaner if we don't return results, just populate the map

andaaron:
Why do we still have umoci in go.mod?

pkg/storage/scrub.go
```diff
@@ -92,153 +91,234 @@ func CheckRepo(ctx context.Context, imageName string, imgStore storageTypes.Imag
...
+
+			var m ispec.Manifest
+
+			_ = json.Unmarshal(manifestContent, &m)
+
+			if m.Subject != nil {
+				imgRes = getManifestResult(ctx, *m.Subject, imgStore, imageName, tag, scrubbedManifests)
+				scrubbedManifests[manifest.Digest] = imgRes
+
+				return imgRes
```
andaaron, line 202: Is this correct? We should be returning result for current image, not for the referred to image?

andaaron: We are returning the result for the current image based on the imageName and tag parameters sent to getManifestResult at 199.

pkg/storage/scrub.go
Andreea-Lupu: Fixed. Now this function is called `scrubManifest` and it just adds the results to the map without returning them.

pkg/storage/scrub.go
Andreea-Lupu: Fixed.

andaaron (approved):

rchincha (approved):
lgtm

PR #1856, opened 2023-09-25
build: fix container image targets

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -354,8 +354,8 @@ run-container:
 	${CONTAINER_RUNTIME} run --rm --security-opt label=disable -v $$(pwd):/go/src/github.com/project-zot/zot \
 		zot-build:latest
 
-.PHONY: binary-stacker
-binary-stacker: $(STACKER)
+.PHONY: oci-image
+oci-image: $(STACKER)
 	${STACKER} --debug build \
 		-f build/stacker.yaml \
 		--substitute COMMIT=$(COMMIT) \
@@ -364,9 +364,9 @@ binary-stacker: $(STACKER)
 		--substitute RELEASE_TAG=$(RELEASE_TAG) \
 		--substitute REPO_NAME=zot-$(OS)-$(ARCH)
 
-.PHONY: image
-image:
-	${CONTAINER_RUNTIME} build ${BUILD_ARGS} -f build/Dockerfile -t zot:latest .
+.PHONY: docker-image
+docker-image:
+	${CONTAINER_RUNTIME} buildx build --platform $(OS)/$(ARCH) -f build/Dockerfile .
 
 $(BATS):
 	rm -rf bats-core; \
@@ -438,7 +438,7 @@ fuzz-all:
 	bash test/scripts/fuzzAll.sh ${fuzztime}; \
 	rm -rf pkg/storage/testdata; \
 
-$(STACKER):
+$(STACKER): check-linux
 	mkdir -p $(TOOLSDIR)/bin; \
 	curl -fsSL https://github.com/project-stacker/stacker/releases/latest/download/stacker -o $@; \
 	chmod +x $@
--- a/build/Dockerfile-conformance
+++ b/build/Dockerfile-conformance
@@ -1,13 +1,17 @@
 # ---
 # Stage 1: Install certs, build binary, create default config file
 # ---
-FROM ghcr.io/project-zot/golang:1.20 AS builder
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -338,8 +338,8 @@ run-container:
 	${CONTAINER_RUNTIME} run --rm --security-opt label=disable -v $$(pwd):/go/src/github.com/project-zot/zot \
 		zot-build:latest
 
-.PHONY: binary-stacker
-binary-stacker: $(STACKER)
+.PHONY: oci-image
+oci-image: $(STACKER)
```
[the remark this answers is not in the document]
adodon2go: Do we need binary-container target (see above)? 
Seams that it is not really used (in CI for example).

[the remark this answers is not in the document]
rchincha: Fixed. Yes not really used - this was from early days ...

build/Dockerfile-conformance
```diff
@@ -2,7 +2,11 @@
 # Stage 1: Install certs, build binary, create default config file
 # ---
 FROM ghcr.io/project-zot/golang:1.20 AS builder
```
andaaron, line 4: Don't we need `--platform=$BUILDPLATFORM` here, and in the other Dockerfiles?

rchincha: good catch! fixed

Makefile
```diff
@@ -348,9 +348,9 @@ binary-stacker: $(STACKER)
 		--substitute RELEASE_TAG=$(RELEASE_TAG) \
 		--substitute REPO_NAME=zot-$(OS)-$(ARCH)
 
-.PHONY: image
-image:
-	${CONTAINER_RUNTIME} build ${BUILD_ARGS} -f build/Dockerfile -t zot:latest .
+.PHONY: docker-image
+docker-image:
+	${CONTAINER_RUNTIME} buildx build --platform $(OS)/$(ARCH) -f build/Dockerfile .
```
adodon2go, line 353: Does this work with podman? Locally (on my side) this target fails.
Should we use docker directly?

Makefile
```diff
@@ -338,8 +338,8 @@ run-container:
 	${CONTAINER_RUNTIME} run --rm --security-opt label=disable -v $$(pwd):/go/src/github.com/project-zot/zot \
 		zot-build:latest
 
-.PHONY: binary-stacker
-binary-stacker: $(STACKER)
+.PHONY: oci-image
```
adodon2go, line 341: This target can be run only on linux. In fact $(STACKER) depends on check-linux target.

> aiandu@aiandus-MacBook-Pro zot % git diff .
> diff --git a/Makefile b/Makefile
> index f2ef779..d108907 100644
> --- a/Makefile
> +++ b/Makefile
> @@ -421,7 +421,7 @@ fuzz-all:
>         bash test/scripts/fuzzAll.sh ${fuzztime}; \
>         rm -rf pkg/storage/testdata; \
>  
> -$(STACKER):
> +$(STACKER): check-linux
>         mkdir -p $(TOOLSDIR)/bin; \
>         curl -fsSL https://github.com/project-stacker/stacker/releases/latest/download/stacker -o $@; \
>         chmod +x $@

Makefile
rchincha: $ make CONTAINER_RUNTIME=podman docker-image

Makefile
rchincha: Correct, `stacker` is currently supported only on host platform linux/amd64. Target will depend on the toolchain used inside stacker.

build/Dockerfile-zb
```diff
@@ -1,10 +1,12 @@
...
+
+ARG TARGETOS
+ARG TARGETARCH
 ARG COMMIT
-ARG OS
-ARG ARCH
+
 RUN mkdir -p /go/src/github.com/project-zot/zot
 WORKDIR /go/src/github.com/project-zot/zot
 COPY . .
```
andaaron, line 12: The make command below needs an update the replace the old OS and ARCH variables with the new ones.

Makefile
rchincha: updated the Makefile rule

rchincha:
Added a nightly test just in case

Makefile
rchincha: https://github.com/rchincha/zot/actions/runs/6396870937/job/17363595069

podman works

adodon2go (approved):

Issue #1859, opened 2023-09-26
[Bug]: metrics endpoint must be secured behind authN


rchincha:
### zot version

v1.4.3 and above

### Describe the bug

Currently, prometheus scraping can be done without authN. This could become a security issue.

### To reproduce

1. Configuration
2. Client tool used
3. Seen error

adodon2go:
Fixed by #1864

PR #1864, opened 2023-09-27
fix: metrics endpoint must be secured behind authN

adodon2go:
**What type of PR is this?**
bug

https://github.com/project-zot/zot/issues/1859

**Automation added to e2e**:

### DIFF
--- a/pkg/api/errors/errors.go
+++ b/pkg/api/errors/errors.go
@@ -120,8 +120,8 @@ func NewError(code ErrorCode) *Error {
 
 		UNAUTHORIZED: {
 			Message: "authentication required",
-			Description: "The access controller was unable to authenticate the client." +
-				"Often this will be accompanied by a Www-Authenticate HTTP response header " +
+			Description: "The access controller was unable to authenticate the client. " +
+				"Often this will be accompanied by a WWW-Authenticate HTTP response header " +
 				"indicating how to authenticate.",
 		},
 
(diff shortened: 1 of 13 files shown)


peusebiu (dismissed):
lgtm

pkg/api/errors/errors.go
```diff
@@ -136,7 +136,7 @@ func NewError(code ErrorCode) *Error {
 
 		UNAUTHORIZED: {
 			Message: "authentication required",
-			Description: "The access controller was unable to authenticate the client." +
+			Description: "The access controller was unable to authenticate the client. " +
 				"Often this will be accompanied by a Www-Authenticate HTTP response header " +
```
rchincha, line 140: nitpick ... "WWW-Authenticate" let's fix this as we catch them.

test/blackbox/metrics.bats
```diff
@@ -52,11 +64,35 @@ function setup_file() {
...
+        "address": "0.0.0.0",
+        "port": "9000",
+        "auth": {
+            "htpasswd": {
+                "path": "${zot_htpasswd_file}"
+            }
+        }
+    },
+    "log": {
+        "level": "debug"
```
andaaron, line 85: Maybe add a log file and print the output on failure in teardown()

test/blackbox/metrics.bats
```diff
@@ -19,14 +25,14 @@ function setup_file() {
         exit 1
     fi
 
-    # Download test data to folder common for the entire suite, not just this file
-    skopeo --insecure-policy copy --format=oci docker://ghcr.io/project-zot/golang:1.20 oci:${TEST_DATA_DIR}/golang:1.20
```
andaaron, line 23: Is no longer necessary?

andaaron (approved):

PR #1865, opened 2023-09-28
fix(cve): trivy is leaking FDs for the Java DB file - disable the feature by default

andaaron:
And use a trivy fork which allows us to disable analyzers if we need to. In our case Java specific analyzers

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -514,3 +514,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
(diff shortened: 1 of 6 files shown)


adodon2go (approved):
LGTM

rchincha:
Let's wait for a bit for this to be fixed.
https://github.com/aquasecurity/trivy/issues/5272

Maybe consider building trivy with this patch and use that.

andaaron:
Looks like it was merged. Let's wait on a release https://github.com/aquasecurity/trivy/commit/18d168769c16a4f780561919140fa72608e348fb?

andaaron:
Let's use https://github.com/project-zot/zot/pull/1901. Unless they release.

PR #1866, opened 2023-09-28
feat(retention): added image retention policies

peusebiu:
feat(metaDB): add more image statistics info

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


examples/config-gc-periodic.json
```diff
@@ -2,15 +2,23 @@
     "distSpecVersion": "1.1.0-dev",
     "storage": {
         "rootDirectory": "/tmp/zot",
-        "gc": true,
-        "gcDelay": "1h",
-        "gcInterval": "24h",
+        "gc": {
```
rchincha, line 5: Don't put this under "gc"
^ This is mainly understood for cleanup after user-triggered DELETEs

Create a new section called "retention" 
^ This is mainly understood for internally-triggered DELETEs.

rchincha:
Also call this "feat(retention)"

rchincha:
sync policies (are per-repo, regex, semver, etc) are a good fit here IMO. Thoughts?

pkg/meta/boltdb/boltdb.go
```diff
@@ -507,7 +514,13 @@ func (bdw *BoltDB) SetRepoReference(repo string, reference string, manifestDiges
 		}
 
 		if _, ok := repoMeta.Statistics[manifestDigest.String()]; !ok {
-			repoMeta.Statistics[manifestDigest.String()] = mTypes.DescriptorStatistics{DownloadCount: 0}
+			repoMeta.Statistics[manifestDigest.String()] = mTypes.DescriptorStatistics{
+				DownloadCount:     0,
+				LastPullTimestamp: time.Time{},
+				PushTimestamp:     time.Now().Local(),
+				// Q do we care if it was pushed by sync or by metaDB.ParseStorage()?
```
andaaron, line 521: What happens right now if the image was added by sync? What would this variable value be?

pkg/storage/gc/gc.go
```diff
@@ -120,11 +125,18 @@ func (gc GarbageCollect) cleanRepo(repo string) error {
...
 	}
 
-	// update repos's index.json in storage
-	if err := gc.imgStore.PutIndexContent(repo, index); err != nil {
+	// apply tags retention
+	if err := gc.retainTags(repo, &index); err != nil {
 		return err
 	}
 
+	if !gc.opts.DryRun {
```
andaaron, line 133: Do we have enough logging for the case in which DryRun is true? To see exactly what would be deleted?

pkg/storage/gc/gc.go
```diff
@@ -338,9 +372,9 @@ func (gc GarbageCollect) removeManifest(repo string, index *ispec.Index,
 				return false, err
 			}
 		} else {
-			err := gc.metaDB.RemoveRepoReference(repo, desc.Digest.String(), desc.Digest)
+			err := gc.metaDB.RemoveRepoReference(repo, reference, desc.Digest)
 			if err != nil {
-				gc.log.Error().Err(err).Msg("gc, metadb: unable to remove repo reference in metaDB")
+				gc.log.Error().Err(err).Msg("gc,metadb: unable to remove repo reference in metaDB")
```
andaaron, line 377: Maybe update all the logs in package `gc` to add the `module` value:
```suggestion
				gc.log.Error().Err(err).Str("module","gc").Msg("unable to remove repo reference in metaDB")
```

pkg/storage/gc/gc.go
```diff
@@ -596,6 +636,36 @@ func (gc GarbageCollect) addORASImageManifestBlobsToReferences(repo string, mdig
 	return nil
 }
 
+func (gc GarbageCollect) retainTags(repo string, index *ispec.Index) error {
```
andaaron, line 639:
```suggestion
func (gc GarbageCollect) removeTagsPerRetentionPolicy(repo string, index *ispec.Index) error {
```

pkg/storage/gc/gc_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+		// Create ImageStore
+		store := local.NewImageStore(rootDir, false, false, log, metrics, nil, nil)
+
+		storeController := storage.StoreController{}
+		storeController.DefaultStore = store
+
+		// for image retention testing
+		// old images
+		gc_old1 := CreateRandomImage()
```
andaaron, line 33: Please update names according to the linter.

pkg/storage/gc/policy.go
```diff
@@ -0,0 +1,198 @@
...
+
+	repoPolicy, err := p.getRepoPolicy(repo)
+	if err != nil {
+		// if repo policy not found, then do not apply any tag retention policy
+		return rules
+	}
+
+	tagPolicy, err := p.getTagPolicy(tag, repoPolicy.TagsRetention)
+	if err != nil {
+		// if tag policy not found, gc all tags by default
```
andaaron, line 82:
```suggestion
		// if tag policy not found, retain all tags by default
```

Correct? The specified tags in the config are the ones we want to keep, not the ones to delete, right?

pkg/storage/gc/retention/rules/latestpull/latestpull.go
```diff
@@ -0,0 +1,35 @@
...
+func (r rule) Perform(candidates []candidate.Candidate) []candidate.Candidate {
+	sort.Slice(candidates, func(i, j int) bool {
+		return candidates[i].PullTimestamp.Before(candidates[j].PullTimestamp)
+	})
+
+	// take top count candidates
+	upper := r.count
+	if r.count > len(candidates) {
+		upper = len(candidates)
+	}
```
andaaron, line 32: `retained by` needs to be updated too?

pkg/storage/gc/retention/rules/dayspush/dayspush.go
```diff
@@ -0,0 +1,38 @@
...
+	filtered := make([]candidate.Candidate, 0)
+
+	timestamp := time.Now().Add(time.Duration(-r.count) * day)
+
+	for _, candidate := range candidates {
+		if candidate.PushTimestamp.After(timestamp) {
+			filtered = append(filtered, candidate)
+		}
+	}
+
```
andaaron, line 36: `retained by` needs to be updated too?

pkg/storage/gc/retention/rules/retainall/retainall.go
```diff
@@ -0,0 +1,27 @@
...
+	return rule{retain}
+}
+
+func (r rule) Name() string {
+	return ruleName
+}
+
+func (r rule) Perform(candidates []candidate.Candidate) []candidate.Candidate {
+	if r.retain {
+		return candidates
```
andaaron, line 23: Retained by needs to be updated?

andaaron (commented):
We are interested in both tags which would be retained and tags which would be removed, in case of dry run.
Can we show both? Are we showing both?

examples/config-gc.json
```diff
@@ -3,10 +3,48 @@
...
+        "retention": {
+            "dryRun": false,
+            "policies": [
+                {
+                    "repoNames": ["infra/*", "prod/*"],
+                    "deleteReferrers": false,
+                    "deleteUntagged": true,
+                    "tagsRetention": [{
+                        "names": ["v2.*", "*-prod"],
+                        "retainAlways": true
```
peusebiu, line 18: retains all tags, is true by default, and can not be used together with the other rules.

examples/config-gc.json
```diff
@@ -3,10 +3,48 @@
...
+                        "pulledWithinLastNrDays": 7,
+                        "pushedWithinLastNrDays": 7   
+                    }]
+                },
+                {
+                    "repoNames": ["**"],
+                    "deleteReferrers": true,
+                    "deleteUntagged": true,
+                    "tagsRetention": [{
+                        "mostRecentlyPushedCount": 10,
```
peusebiu, line 40: tag is retained if any of these 4 rules is met.

examples/config-gc.json
```diff
@@ -3,10 +3,48 @@
...
+        "gcInterval": "1h",
+        "retention": {
+            "dryRun": false,
+            "policies": [
+                {
+                    "repoNames": ["infra/*", "prod/*"],
+                    "deleteReferrers": false,
+                    "deleteUntagged": true,
+                    "tagsRetention": [{
+                        "names": ["v2.*", "*-prod"],
```
peusebiu, line 17: any tag that doesn't match any tagRetention[].names will be removed,
if names rule is empty, then rules are applied for all tags.

examples/config-gc.json
```diff
@@ -3,10 +3,48 @@
...
         "gcDelay": "2h",
         "untaggedImageRetentionDelay": "4h",
-        "gcInterval": "1h"
+        "gcInterval": "1h",
+        "retention": {
+            "dryRun": false,
+            "policies": [
+                {
+                    "repoNames": ["infra/*", "prod/*"],
+                    "deleteReferrers": false,
```
peusebiu, line 14: weather or not to delete manifests with missing subjects.

examples/config-gc.json
```diff
@@ -3,10 +3,48 @@
...
+                    "deleteUntagged": true,
```
peusebiu, line 15: whether or not delete manifests without tags.

examples/config-gc-periodic.json
```diff
@@ -4,13 +4,13 @@
         "rootDirectory": "/tmp/zot",
         "gc": true,
         "gcDelay": "1h",
-        "gcInterval": "24h",
+        "gcInterval": "1h",
```
rchamarthy, line 7: I wouldn't call this gc. Garbage is something that is deleted and not used and we are doing cleanup. That cleanup should happen on a continuous basis, not on a time period. For a given repository, we should define retention rules. This is explicitly at a repository level or at an image level.

examples/config-gc.json
```diff
@@ -3,10 +3,48 @@
     "storage": {
         "rootDirectory": "/tmp/zot",
         "gc": true,
-        "gcReferrers": true,
         "gcDelay": "2h",
         "untaggedImageRetentionDelay": "4h",
-        "gcInterval": "1h"
+        "gcInterval": "1h",
+        "retention": {
+            "dryRun": false,
```
rchamarthy, line 10: Is dryRun going to just produce a log that this image retention period is over and can be deleted?

examples/config-gc.json
rchamarthy, line 15: How does retention work? Say if repo1 has a retention policy for an image and repo2 uses an image from repo1 and has a much larger retention policy. The image from repo1 goes away but the layers are retained from a blob pov?

examples/config-gc.json
rchamarthy, line 18: Isnt this default behavior why do i need to configure a knob if am going to retain?

examples/config-gc.json
```diff
@@ -3,10 +3,48 @@
...
+                        "names": ["v1.*"],
+                        "pulledWithinLastNrDays": 7,
+                        "pushedWithinLastNrDays": 7   
+                    }]
+                },
+                {
+                    "repoNames": ["**"],
+                    "deleteReferrers": true,
+                    "deleteUntagged": true,
+                    "tagsRetention": [{
```
rchamarthy, line 39: Does tagsRetention mean retain if this matches and delete otherwise? Is it any of these metrics match or all of them have to match?

pkg/api/config/config.go
```diff
@@ -27,15 +27,36 @@ type StorageConfig struct {
...
+	RepoNames       []string
+	DeleteReferrers bool
+	DeleteUntagged  bool
+	TagsRetention   []TagsRetentionPolicy
+}
+
+type TagsRetentionPolicy struct {
+	Names                   []string
+	RetainAlways            bool // default is true
+	PulledWithinLastNrDays  *int
```
rchamarthy, line 54: why "*int"

examples/config-gc.json
hallyn: That was what we had discussed, yes.

examples/config-gc.json
peusebiu: what do  you mean by "uses", if by that you mean it has some blobs deduped from image1, then yes.

examples/config-gc.json
peusebiu: if any of the rules matches, then it will be retained.

examples/config-gc.json
peusebiu: because all the others tags will be removed, if you have a list of tagsRetention, only those matching its policies/rules will be retained, the rest will be removed.

pkg/api/config/config.go
peusebiu: so I can check for nil value, but I guess I can also skip 0 values.

pkg/meta/boltdb/boltdb.go
peusebiu: it will be "", anonymous

pkg/storage/gc/gc.go
peusebiu: yes, we have logs in the functions that removes manifests. here we don't update the index.json of repo so that manifests won't get removed by gc.blobs() (unreferenced blobs)

examples/config-gc.json
peusebiu: Now I understand the question, yes I think we can remove it.

andaaron: It is still unclear. The `names` key is optional right, if not present it matches all tags in the repo?
1. What happens if a tag retention entry has no keys at all, no "names", no "retainAlways", no other keys?
2. How do I specify a retention rule to keep everything in a repo? I used to think I need to add an entry having "retainAlways" as a single key with value "true". If this is not the case, do I need to use a regex to match any tag?

To clarify what I personally understand by "retain" being the default action:

1. If a repo matches the repo name pattern, but no retention rule, everything in that repo will be deleted correct?
Thinking the act of someone specifying a matcher for that repo is a commitment from that user that he wants the cleanup to be applied to that repo. So delete everything which does not match a tag retention rule for that repo.

2. If a repo doesn't match any repo name pattern, it means the user doesn't want to use this feature to clean up those repos, so nothing will be removed. -> this is my understanding of "retain by default".

examples/config-gc.json
andaaron: These rules are for manifest and tags. This knob only deletes manifests. Manifest do no "use" other manifests (unless they are signatures or some other kind of artifacts), but that is in the context of the same repo.

Manifests could reuse layers also referenced by other manifests (base image).
Layers are GCed later, they are not "retained", and like Sebi mentioned, the dedupe functionality is cross-repo, and takes care internally of keeping a valid copy of the layer.

examples/config-gc.json
peusebiu: I removed that key.

1.1) it retains everything
1.2) if you want to keep everything for a repo you need to make sure your repo doesn't match any repo policy, in other words you don't write a policy for it.
or you put a default with repoNames: ["**"] where deleteReferrers is false, deleteUntagged is false, and no retention set.

2.1) if there are no retention rules at all, then it will retain everything, if there is at least one retention rule, and it doesn't match it then it will remove those tags that don't match

2.2) true, and it's true for the current logic.

andaaron (dismissed):

examples/config-gc-periodic.json
hallyn: > That cleanup should happen on a continuous basis, not on a time period.

That doesn't really work for this.  We would have to, at startup, set a timer for every image and possibly reset the timer on every access, then delete the image when the timer goes off.  That could be hundreds of thousands of timers.  Doing a scan every 24h makes a lot more sense here.

hallyn:
You have dryRun as an all-or-nothing for all policies, IIUC.  It seems to me I might want to say "all image produced by pull requests (as signified by a CI-XXX tag) are deleted (dryRun: false) after 5 days.  But for any unknown namespaces, simply report (dryRun: true), because someone may have pushed a new namespace which I don't know about yet."  This would send me notification of unhandled namespaces, which simply having the catch-all be to not delete would not do.

Maybe that's too much complication, though.  Just a thought.

hallyn:
One problem I see here - which admins should at least be warned about - is, IIUC, if I start with expiration rules set, then run for a week without expiration rules set, then re-enable them, the Metadb will not have been used during that week, and images pushed in the meantime will not have pushed/pulled times set.  Am I right about that?  That could lead to very wrong behavior (from admin's POV).

pkg/cli/server/root.go
```diff
@@ -861,6 +861,37 @@ func validateGC(config *config.Config, log zlog.Logger) error {
...
+				log.Error().Err(glob.ErrBadPattern).Str("pattern", pattern).
+					Msg("retention repo glob pattern could not be compiled")
+
+				return zerr.ErrBadConfig
+			}
+		}
+
+		for _, tagRule := range policy.TagsRetention {
+			for _, regex := range tagRule.Names {
+				_, err := regexp.Compile(regex)
```
hallyn, line 886: Performance-wise, would it pay off to actually cache the compiled regexp for each rule and re-use it later instead of re-compiling it?

peusebiu: yes, that would be nice.

peusebiu:
> You have dryRun as an all-or-nothing for all policies, IIUC. It seems to me I might want to say "all image produced by pull requests (as signified by a CI-XXX tag) are deleted (dryRun: false) after 5 days. But for any unknown namespaces, simply report (dryRun: true), because someone may have pushed a new namespace which I don't know about yet." This would send me notification of unhandled namespaces, which simply having the catch-all be to not delete would not do.
> 
> Maybe that's too much complication, though. Just a thought.

Would be nice, but it's not clear what the definition of unknown namespace is.
If I start a zot server with empty rootDir, all namespaces are unknown, until I restart zot?

when do I switch from "unknown namespace" to "old namespace".

examples/config-gc.json
```diff
@@ -3,10 +3,47 @@
     "storage": {
         "rootDirectory": "/tmp/zot",
         "gc": true,
-        "gcReferrers": true,
         "gcDelay": "2h",
         "untaggedImageRetentionDelay": "4h",
```
rchincha, line 7: Why is this outside the "retention" section?

examples/config-gc.json
```diff
@@ -3,10 +3,48 @@
...
-        "gcInterval": "1h"
+        "gcInterval": "1h",
+        "retention": {
+            "dryRun": false,
+            "policies": [
+                {
+                    "repoNames": ["infra/*", "prod/*"],
```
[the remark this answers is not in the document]
rchincha: Not longest prefix match?

examples/config-gc.json
rchincha: Also make a **new** example file called config-retention.json.

gc != retention

examples/config-gc.json
```diff
@@ -3,10 +3,47 @@
...
+                    "deleteReferrers": false,
+                    "deleteUntagged": true,
+                    "tagsRetention": [{
+                        "names": ["v2.*", ".*-prod"]
```
rchincha, line 17: "pattern"?

examples/config-gc.json
```diff
@@ -3,10 +3,47 @@
...
+                    },
+                    {
+                        "names": ["v3.*", ".*-prod"],
+                        "pulledWithinLastNrDays": 7
```
rchincha, line 21: "pulledWithin": "7d"?

examples/config-gc.json
peusebiu: if they are older than gcDelay, yes

examples/config-gc.json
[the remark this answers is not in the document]
andaaron: Because it would be very tricky to get the configuration right as a user if the entries in the policy list can be in any order.
You would need to look at potentially hundreds such entries in any order to figure out which one would apply for the repo you are interested in.

If the first one matches you don't lose anything functionality-wise.
You can place your longest prefix match first, and your shorter prefix match afterwards and get the same result you would have gotten if we matched always on longest prefix.

examples/config-gc.json
peusebiu: https://pkg.go.dev/time#ParseDuration

the biggest unit is h(our) 
it gets hard to express weeks/months, I changed it, but I can revert it back

examples/config-retention.json
```diff
@@ -0,0 +1,55 @@
...
+        "rootDirectory": "/tmp/zot",
+        "gc": true,
+        "gcDelay": "2h",
+        "gcInterval": "1h",
+        "retention": {
+            "dryRun": false,
+            "delay": "24h",
+            "policies": [
+                {
+                    "repoNames": ["infra/*", "prod/*"],
```
andaaron, line 13: `repositories`

Follow sync configuration naming patterns.

examples/config-retention.json
```diff
@@ -0,0 +1,55 @@
...
+                    "deleteReferrers": false,
+                    "deleteUntagged": true,
```
andaaron, line 15: This should be `true` by default.

examples/config-retention.json
```diff
@@ -0,0 +1,55 @@
...
+                    "tagsRetention": [{
```
andaaron, line 16:
```suggestion
                    "keepTags": [{
```

pkg/api/controller.go
```diff
@@ -261,7 +261,8 @@ func (c *Controller) InitImageStore() error {
 
 func (c *Controller) InitMetaDB(reloadCtx context.Context) error {
 	// init metaDB if search is enabled or we need to store user profiles, api keys or signatures
-	if c.Config.IsSearchEnabled() || c.Config.IsBasicAuthnEnabled() || c.Config.IsImageTrustEnabled() {
+	if c.Config.IsSearchEnabled() || c.Config.IsBasicAuthnEnabled() || c.Config.IsImageTrustEnabled() ||
+		c.Config.IsMetaDBNeededForRetention() {
```
andaaron, line 265:
```suggestion
		c.Config.IsRetentionEnabled() {
```

andaaron (commented):
Notes following the call on 11 of October.

pkg/storage/gc/gc.go
```diff
@@ -112,28 +119,39 @@ func (gc GarbageCollect) cleanRepo(repo string) error {
...
 	index, err := common.GetIndex(gc.imgStore, repo, gc.log)
 	if err != nil {
+		gc.log.Error().Err(err).Str("module", "gc").Str("repository", repo).Msg("unable to read index.json in repo")
+
 		return err
 	}
 
 	// gc referrers manifests with missing subject and untagged manifests
-	if err := gc.cleanManifests(repo, &index); err != nil {
+	if err := gc.removeManifestsPerRepoPolicy(repo, &index); err != nil {
```
andaaron, line 128: Lines 128 should be after line 144?
First remove tags, and then remove all untagged manifests per policy. And at the end delete orphaned blobs.

andaaron (dismissed):
LGTM

andaaron (approved):

examples/config-retention.json
```diff
@@ -0,0 +1,68 @@
...
+        "rootDirectory": "/tmp/zot",
+        "gc": true,
+        "gcDelay": "2h",
+        "gcInterval": "1h",
+        "retention": {
+            "dryRun": false,
+            "delay": "24h",
+            "policies": [
+                {
+                    "repositories": ["infra/*", "prod/*"],
```
rchincha, line 13: Can we add a check if same repository prefix patterns overlap in other rules? So that folks are informed about possibly "surprising" behaviors.

Would add this check/feedback for `zot verify` as well.

Just issue a warning that "found overlaps...etc" but move on.

pkg/storage/gc/retention/candidate/candidate.go
```diff
@@ -0,0 +1,40 @@
+package candidate
```
rchincha, line 1: Why make so many subdirs for this? Just add all the files under the pkg/storage/retention.go? instead? It is a storage feature right?

peusebiu: the rules implements an interface, that's why I made each one a package.
candidate package is both used by gc package and by rules (avoid circular import)
same with regex.

yes, currently retention is implemented inside gc, because some functionalities were already in gc, like deleteReferrers, deleteUntagged.

pkg/storage/gc/policy.go
```diff
@@ -0,0 +1,265 @@
...
+	log.Info().Str("module", "retention").
+		Bool("dry-run", dryRun).
+		Str("repository", repo).
+		Str("mediaType", candidate.MediaType).
+		Str("digest", candidate.DigestStr).
+		Str("tag", candidate.Tag).
+		Str("lastPullTimestamp", candidate.PullTimestamp.String()).
+		Str("pushTimestamp", candidate.PushTimestamp.String()).
+		Str("decision", decision).
+		Str("reason", reason).Msgf("will %s tag", decision)
```
andaaron, line 264:
```suggestion
		Str("reason", reason).Msg("applied policy")
```

pkg/storage/gc/retention/candidate/candidate.go
rchincha: That is why we have pkg/storage/types.go etc

All interfaces go there.

andaaron: There are multiple implementations each in a package, what does pkg/storage/types.go have to do with this topic?

andaaron (dismissed):
LGTM

hallyn:
Thanks - feature-wise looks good to me.

pkg/retention/types/types.go
```diff
@@ -0,0 +1,30 @@
+package types
```
rchincha, line 1: Much better, thank you.

rchincha:
Pls resolve the conflicts and push again

rchincha:
cc: @mbshields this needs documentation as well

lictw:
Good interface, nice options for any case - great job, thanks! I will build and test it for sure, I need this!

I have one suggestion, I need it and I think it will be nice option too! I don't know how to name it exactly, let it be "dynamic tags":

Imagine tags like environment-{1,2,..}-* and I want to keep 2 latest **for each environment** (1, 2, etc), so I must write rules for each environment, but it's a dynamic thing, and there are many of them, I suggest to simplify it somehow.

So many groups of tags and single rule for each group, something like regex capture group (maybe named) I think the best solution cause our group value must be in any matched tag and we should extract&compare it for each tag, and it's also very "flat":
```
names: [(environment-\d+)-.*]
```
.., so for each matched tag we should compare $1 and interpret tags with same group as unique list/rule and do the things.

Maybe you imagine better interface, but I really think it's cool feature, and it will be very unique feature!

lictw:
Update after first testing:

1. Please update main from upstream to pull changes about "delete repositories without manifests", now in your fork repositories are left after policies delete all tags.
2. Help me understand what is `deleteReferrers`, how it works? Also `deleteUntagged` about unfinished blobs, when it's uploaded, but not tagged (canceled push for ex.)?

3. I'm right that if some tags matched to rule, than only this rule defines who will be retained and this tags can't be now affected by rules below?
```
        {
          "repositories": ["**"],
          "deleteUntagged": true,
          "deleteReferrers": true,
          "keepTags": [{
            "patterns": ["main$"]
          },{
            "patterns": ["main"],
            "mostRecentlyPushedCount": 3
          },{
            "pushedWithin": "72h",
            "pulledWithin": "720h"
          }]
        }
```
For ex. here for tags like: main (meta-tag, always points to the last build of this branch), main.1, main.2, branch-foo (meta), branch-foo.13 I want:
1. `main` meta-tag always presented 
2. 3 latest build from main branch not counting `main` meta-tag, cause it already matched by rule above 
3. for builds from other branches and it will not affect any `main` builds cause all of them already matched by above rules

So the main point that this rules will not preserve `main` builds for 3 days how builds from other branches and will keep only 3 latest how it says, right? Then it's great implementation, thank you very much!

Also about above feature `dynamic tags`, in real world for complete retention I need something like this:
```
        {
          "repositories": ["**"],
          "deleteUntagged": true,
          "deleteReferrers": true,
          "keepTags": [{
            "patterns": ["main$"]
          },{
            "patterns": ["main\\..+"],
            "mostRecentlyPushedCount": 8
          },{
            "patterns": ["environments-[a-z]+$"]
          },{
            "patterns": ["(environments-[a-z]+)\\..+"],
            "mostRecentlyPushedCount": 1
          },{
            "pushedWithin": "72h",
            "pulledWithin": "720h"
          }]
        }
```
Everything like above, but also keep meta-tag for any environment and 1 latest build for each, and also, as for `main`, not match them by last rule (so not delay deletion for 72h).

In general, everything looks working, I use build of fork with retention for 2 days in my main zot installation, also seems that all works as I described in 3, and so as I said - it's great!

lictw:
And also, I see there is a **no** difference between `"patterns": ["main", "master"]` and `"patterns": ["(main|master)"]`, so it generates a single list to which the rule applies (pushedCount for example). I think it will be great if each pattern will generate personal list and when we write like `"patterns": ["main", "master"]` then we just want apply the single rule, but for each match, they are not the same, and my `dynamic tags` in this case will just automatic propagation of the patterns, when single `"patterns": ["(environment-[a-z]+)\\..+"]` will be processed (based on existing tags) as `"patterns": ["(environment-foo\\..+", "(environment-bar\\..+", "(environment-baz\\..+"]`.

peusebiu:
> Update after first testing:
> 
> 1. Please update main from upstream to pull changes about "delete repositories without manifests", now in your fork repositories are left after policies delete all tags.
> 2. Help me understand what is `deleteReferrers`, how it works? Also `deleteUntagged` about unfinished blobs, when it's uploaded, but not tagged (canceled push for ex.)?
> 3. I'm right that if some tags matched to rule, than only this rule defines who will be retained and this tags can't be now affected by rules below?
> 
> ```
>         {
>           "repositories": ["**"],
>           "deleteUntagged": true,
>           "deleteReferrers": true,
>           "keepTags": [{
>             "patterns": ["main$"]
>           },{
>             "patterns": ["main"],
>             "mostRecentlyPushedCount": 3
>           },{
>             "pushedWithin": "72h",
>             "pulledWithin": "720h"
>           }]
>         }
> ```
> 
> For ex. here for tags like: main (meta-tag, always points to the last build of this branch), main.1, main.2, branch-foo (meta), branch-foo.13 I want:
> 
> 1. `main` meta-tag always presented
> 2. 3 latest build from main branch not counting `main` meta-tag, cause it already matched by rule above
> 3. for builds from other branches and it will not affect any `main` builds cause all of them already matched by above rules
> 
> So the main point that this rules will not preserve `main` builds for 3 days how builds from other branches and will keep only 3 latest how it says, right? Then it's great implementation, thank you very much!
> 
> Also about above feature `dynamic tags`, in real world for complete retention I need something like this:
> 
> ```
>         {
>           "repositories": ["**"],
>           "deleteUntagged": true,
>           "deleteReferrers": true,
>           "keepTags": [{
>             "patterns": ["main$"]
>           },{
>             "patterns": ["main\\..+"],
>             "mostRecentlyPushedCount": 8
>           },{
>             "patterns": ["environments-[a-z]+$"]
>           },{
>             "patterns": ["(environments-[a-z]+)\\..+"],
>             "mostRecentlyPushedCount": 1
>           },{
>             "pushedWithin": "72h",
>             "pulledWithin": "720h"
>           }]
>         }
> ```
> 
> Everything like above, but also keep meta-tag for any environment and 1 latest build for each, and also, as for `main`, not match them by last rule (so not delay deletion for 72h).
> 
> In general, everything looks working, I use build of fork with retention for 2 days in my main zot installation, also seems that all works as I described in 3, and so as I said - it's great!

Hello, and thanks for trying it!

1. done
2. deleteReferrers, will remove manifests for which their subject is missing.(was deleted, or was never uploaded)
3. deleteUntagged, will remove manifests without tags
4. yes, that's right, only the first tag policy is applied so your "main$" should be safe.

```
Imagine tags like environment-{1,2,..}-* and I want to keep 2 latest for each environment (1, 2, etc), so I must write rules for each environment, but it's a dynamic thing, and there are many of them, I suggest to simplify it somehow.
```
^here, I don't understand why you need rules per each environment, you want different rules for each one? 

Let me think about the rest of the points, thanks for raising them.

rchincha:
@peusebiu pls rebase

lictw:
> I don't understand why you need rules per each environment, you want different rules for each one?

The same, the same to keep last N, but N for each, 1 latest for env-A, 1 latest for env-B, not 1 for any. It seems to me that the best implementation that I have reached is described [here](https://github.com/project-zot/zot/pull/1866#issuecomment-1783317952).

rchincha:
> > deleteReferrers, will remove manifests for which their subject is missing.(was deleted, or was never uploaded)
> 
> PS what is subject?)

The OCI distribution spec is working towards allowing uploading of arbitrary content and add "soft links" to each other.
https://github.com/opencontainers/distribution-spec/blob/main/spec.md#enabling-the-referrers-api

rchincha (approved):
lgtm

rchincha:
@lictw Thanks for all the feedback and testing.
This PR has been merged, so build from main and try.
We will take up further enhancements in future PRs.

lictw:
@rchincha yep, nice job! Please ping me in future PRs to watch.

lictw:
@peusebiu Hi! I'm back. I'm faced with a situation that I can't seem to solve with the current implementation. But a task seems so basic that I can't believe in it, so I decided to ask if I understand everything correctly.

I want to keep 3 latest tags, but only for some time, something like:
````
          "repositories": ["foo/**"],
          "keepTags": [{
            "mostRecentlyPushedCount": 3,
            "pushedWithin": "100h"
          }]
````
but there is OR logic, so above will keep latest 3 tags for forever. It's a DEV builds, I want to delete them fully after some time, but I also what to have only some latest builds if there is many, so it's the reason.

Also, I also have the latest tag, so I want to always keep the latest and 3 latest pushed tags, and delete all of them after some time.

And as side question about latest pulled, is it tracking for a specific tag or for whole image? If an image has 2 tags, but was pulled only by one, the second will be deleted?

lictw:
@peusebiu up (@rchincha maybe?)

andaaron:
@lictw

To answer your last paragraph, only the tags are deleted, the actual image/manifest is kept/deleted afterwards based on the separate setting you used for `deleteUntagged`. If the manifest is still referenced by at least one other remaining tag it should not be deleted.

I'm not sure there is a possibility to do an intersection between mostRecentlyPushedCount and pushedWithin. They were designed to independently apply for the same regex pattern identifying the tag strings. And I believe this constraint was because the `patterns` were meant to be unique, so there wouldn't be a possibility to have the same `patterns` value in multiple entries inside `keepTags` if you wanted to implement the OR.
I may be mistaking though. @peusebiu , @rchincha , what do you think?

lictw:
@peusebiu why 4? Just to clarify that I understand how it equally works. Latest already done after first rule and should be excluded, is not it? So 3 from of the remaining. 

After all, does my case is actual? Do you have any thoughts on feature improvements?

PR #1868, opened 2023-09-29
fix: errors returned by zot should match the dist-spec errors

adodon2go:
**What type of PR is this?**
bug

Error codes that zot sends to clients should match https://specs.opencontainers.org/distribution-spec/?v=v1.0.0#error-codes

**Automation added to e2e**:


rchincha (approved):
lgtm

PR #1877, opened 2023-10-03
fix(nightly): increase nightly delays and timeouts

peusebiu:
change the startup dedupe blackbox test source image with a lighter one

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/test/blackbox/pushpull_running_dedupe.bats
+++ b/test/blackbox/pushpull_running_dedupe.bats
@@ -24,7 +24,7 @@ function setup_file() {
         exit 1
     fi
     # Download test data to folder common for the entire suite, not just this file
-    skopeo --insecure-policy copy --format=oci docker://ghcr.io/project-zot/golang:1.20 oci:${TEST_DATA_DIR}/golang:1.20
+    skopeo --insecure-policy copy --format=oci docker://ghcr.io/project-zot/test-images/alpine:3.17.3 oci:${TEST_DATA_DIR}/alpine:1
     # Setup zot server
     local zot_root_dir=${BATS_FILE_TMPDIR}/zot
     local zot_config_file=${BATS_FILE_TMPDIR}/zot_config.json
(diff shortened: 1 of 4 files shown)


test/blackbox/restore_s3_blobs.bats
```diff
@@ -124,10 +125,12 @@ function teardown_file() {
 
     # start with dedupe disabled
     zot_serve ${zot_config_file_nodedupe}
-    wait_zot_reachable 8080
```
peusebiu, line 127: no need to wait, just read the log file

test/blackbox/pushpull_running_dedupe.bats
```diff
@@ -119,15 +119,19 @@ function teardown_file() {
     sed -i 's/false/true/g' ${zot_config_file}
 
     zot_serve ${ZOT_PATH} ${zot_config_file}
+
+    # give time to repoDB to parse storage
+    sleep 10
```
andaaron, line 124: As far as I know zot won't start serving requests until repoDB finishes parsing the storage.

rchincha: yes, that should be the preferred way IMO.
Do a curl GET /v2/ check, sleep loop until 200.

test/blackbox/restore_s3_blobs.bats
peusebiu: wait_for_string uses tail to catch the log message, if we wait for zot to be reachable startup dedupe may have finished in the meantime. 

I moved the wait_zot_reachable in the second test

test/blackbox/pushpull_running_dedupe.bats
peusebiu: then I'll increase the timeouts of curl, it's not enough for this test

test/blackbox/restore_s3_blobs.bats
andaaron: Looks good.

andaaron (approved):

PR #1878, opened 2023-10-03
refactor(test): make sure cli tests are not internal unless they need to be

andaaron:
As part of this change searchConfig needed to be exported, as it was passed as a parameter to exported functions

At this moment most of the tests remaining internal depend on the mock service. The interface it implements has un-exported methods.


rchincha (approved):
lgtm

PR #1879, opened 2023-10-03
fix: running tests locally fails

adodon2go:
**What type of PR is this?**
bug

**Automation added to e2e**:

### DIFF
--- a/pkg/meta/dynamodb/dynamodb_internal_test.go
+++ b/pkg/meta/dynamodb/dynamodb_internal_test.go
@@ -14,13 +14,14 @@ import (
 
 	"zotregistry.io/zot/pkg/log" //nolint:go-staticcheck
 	"zotregistry.io/zot/pkg/meta/version"
+	tskip "zotregistry.io/zot/pkg/test/skip"
 )
 
 func TestWrapperErrors(t *testing.T) {
-	const (
-		endpoint = "http://localhost:4566"
-		region   = "us-east-2"
-	)
+	tskip.SkipDynamo(t)
+
+	const region = "us-east-2"
+	endpoint := os.Getenv("DYNAMODBMOCK_ENDPOINT")
 
 	uuid, err := guuid.NewV4()
 	if err != nil {
--- a/pkg/meta/parse_test.go
+++ b/pkg/meta/parse_test.go
@@ -28,6 +28,7 @@ import (
 	. "zotregistry.io/zot/pkg/test/image-utils"
 	"zotregistry.io/zot/pkg/test/mocks"
 	"zotregistry.io/zot/pkg/test/signature"
+	tskip "zotregistry.io/zot/pkg/test/skip"
 )
 
 const repo = "repo"
@@ -367,7 +368,7 @@ func TestParseStorageWithBoltDB(t *testing.T) {
 }
 
 func TestParseStorageDynamoWrapper(t *testing.T) {
-	skipIt(t)
+	tskip.SkipDynamo(t)
 
 	Convey("Dynamodb", t, func() {
 		rootDir := t.TempDir()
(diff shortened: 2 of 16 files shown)


pkg/meta/parse_test.go
```diff
@@ -367,7 +368,7 @@ func TestParseStorageWithBoltDB(t *testing.T) {
 }
 
 func TestParseStorageDynamoWrapper(t *testing.T) {
-	skipIt(t)
+	tskip.SkipS3(t)
```
andaaron, line 371: This should probably be a Dynamo check. Or both? At least Dynamo.

pkg/meta/dynamodb/dynamodb_internal_test.go
```diff
@@ -14,13 +15,18 @@ import (
...
 	"zotregistry.io/zot/pkg/meta/version"
+	tskip "zotregistry.io/zot/pkg/test/skip"
 )
 
 func TestWrapperErrors(t *testing.T) {
-	const (
-		endpoint = "http://localhost:4566"
-		region   = "us-east-2"
-	)
+	tskip.SkipS3(t)
```
andaaron, line 22: As far as I can tell the check and endpoint should be for DynamoDB

peusebiu: yes

pkg/meta/parse_test.go
peusebiu: just Dynamo, test uses local storage.

pkg/storage/s3/s3_test.go
```diff
@@ -1239,7 +1231,8 @@ func TestNegativeCasesObjectsStorage(t *testing.T) {
 }
 
 func TestS3Dedupe(t *testing.T) {
-	skipIt(t)
+	tskip.SkipS3(t)
```
peusebiu, line 1234: this is the only place where we need both, the rest should be just S3. they are using boltdb

pkg/test/skip/skip_test.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// for code coverage.
+func TestSkipS3(t *testing.T) {
+	Convey("Test if S3MOCK_ENDPOINT env is uset", t, func() {
+		envName := "S3MOCK_ENDPOINT"
+		envVal := os.Getenv(envName)
+		if len(envVal) > 0 {
+			So(os.Setenv(envName, ""), ShouldBeNil)
+			defer So(os.Setenv(envName, envVal), ShouldBeNil)
```
andaaron, line 19: The defers should be before setting the new value, right?
Otherwise the variable would not be reset if line 18 errors.

I think same applies to the Dynamo test

pkg/test/skip/skip_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"testing"
+
+	. "github.com/smartystreets/goconvey/convey"
+
+	tskip "zotregistry.io/zot/pkg/test/skip"
+)
+
+// for code coverage.
+func TestSkipS3(t *testing.T) {
+	Convey("Test if S3MOCK_ENDPOINT env is uset", t, func() {
```
andaaron, line 14:
```suggestion
	Convey("Test if S3MOCK_ENDPOINT env is unset", t, func() {
```

pkg/meta/dynamodb/dynamodb_internal_test.go
```diff
@@ -14,13 +15,18 @@ import (
...
 
 func TestWrapperErrors(t *testing.T) {
-	const (
-		endpoint = "http://localhost:4566"
-		region   = "us-east-2"
-	)
+	tskip.SkipDynamo(t)
+
+	const region = "us-east-2"
+	endpoint := os.Getenv("S3MOCK_ENDPOINT")
```
andaaron, line 25:
```suggestion
	endpoint := os.Getenv("DYNAMODBMOCK_ENDPOINT")
```
And remove the if

pkg/test/skip/skip_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+	. "github.com/smartystreets/goconvey/convey"
+
+	tskip "zotregistry.io/zot/pkg/test/skip"
+)
+
+// for code coverage.
+func TestSkipS3(t *testing.T) {
+	Convey("Test if S3MOCK_ENDPOINT env is unset", t, func() {
+		os.Clearenv()
```
andaaron, line 15: I think this one removes everything from the environment, including stuff we need. I think we need to set everything back in a defer.

andaaron (approved):

PR #1881, opened 2023-10-04
fix(cache): make dynamoDB aware of orignal/deduped blobs

peusebiu:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/cache/dynamodb.go
+++ b/pkg/storage/cache/dynamodb.go
@@ -26,11 +26,11 @@ type DynamoDBDriverParameters struct {
 }
 
 type Blob struct {
-	Digest   string   `dynamodbav:"Digest,string"`
-	BlobPath []string `dynamodbav:"BlobPath,stringset"`
+	Digest            string   `dynamodbav:"Digest,string"`
+	DuplicateBlobPath []string `dynamodbav:"DuplicateBlobPath,stringset"`
+	OriginalBlobPath  string   `dynamodbav:"OriginalBlobPath,string"`
 }
 
-// Use ONLY for tests.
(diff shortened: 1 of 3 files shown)


pkg/storage/cache/dynamodb.go
```diff
@@ -198,24 +199,28 @@ func (d *DynamoDBDriver) HasBlob(digest godigest.Digest, path string) bool {
...
 
 		return err
 	}
 
-	result, _ := d.GetBlob(digest)
+	originBlob, _ := d.GetBlob(digest)
+	// if original blob is the one deleted
+	if originBlob == path {
+		// move duplicate blob to original, storage will move content here
+		originBlob, _ = d.getDuplicateBlob(digest)
```
andaaron, line 215: My understanding is the original blob is also found in the duplicate list.
Seems like this method always returns the first blob in the duplicates list.
Is there a chance the original blob is the first one in the duplicate list?

peusebiu: no, there's no chance because I already removed it from the duplicate list earlier, and in duplicate list there can not be the same path multiple times, because it's a set with unique strings. right?

andaaron: I don't know if the are unique. Are you sure?

peusebiu: yes, https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html

I'll test it.

andaaron (approved):

rchincha (approved):
lgtm

PR #1884, opened 2023-10-04
docs: add logging guidelines

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


andaaron (approved):
I think there should be an easy way to find this file. Maybe mention it in the contributing guide?
There are also other readmes in the same situation, "hidden" under different packages. Maybe we should have a way to advertise them

Issue #1876, opened 2023-10-02
[Bug]: Metrics should be protected behind authN


rchincha:
### zot version

v2.0.0-rc6

### To reproduce

1. Configuration
2. Client tool used
3. Seen error

rchincha:
A "scraping" username/identity under extensions/metrics who is allowed to scrape.

PR #1895, opened 2023-10-05
fix: metrics should be protected behind authZ

adodon2go:
**What type of PR is this?**
bug

#1876

In case anonymous policy is present in access control,we allow anonymous users to access to access /metrics endpoints.
To prevent this in accessControl section of config you can specify a list of metrics users that are allowed to access /metrics endpoint.
In case accessControl is not specified in config but auth is present than only authenticated users can access /metrics endpoint

**Automation added to e2e**:

### DIFF
--- a/examples/config-metrics-authz.json
+++ b/examples/config-metrics-authz.json
@@ -1,0 +1,39 @@
+{
+    "distSpecVersion": "1.1.0-dev",
+    "storage": {
+        "rootDirectory": "/tmp/zot"
+    },
+    "http": {
+        "address": "127.0.0.1",
+        "port": "8080",
+        "auth": {
+            "htpasswd": {
+              "path": "test/data/htpasswd"
+            }
+        },
+        "accessControl": {
+            "metrics":{
+                "users": ["metrics"]
+            },
+            "repositories": {
+                "**": {
+                    "anonymousPolicy": [
+                        "read"
+                    ],
+                    "defaultPolicy": ["read","create"]
+                }
+            }
+        }
+    },
+    "log": {
+        "level": "debug"
+    },
+    "extensions": {
+        "metrics": {
+            "enable": true,
+            "prometheus": {
+                "path": "/metrics"
+            }
+        }
+    }
+}
--- a/pkg/api/authz.go
+++ b/pkg/api/authz.go
@@ -191,14 +191,10 @@ func (ac *AccessController) getAuthnMiddlewareContext(authnType string, request 
 func (ac *AccessController) isPermitted(userGroups []string, username, action string,
 	policyGroup config.PolicyGroup,
 ) bool {
-	var result bool
-
 	// check repo/system based policies
 	for _, p := range policyGroup.Policies {
 		if common.Contains(p.Users, username) && common.Contains(p.Actions, action) {
-			result = true
-
-			return result
+			return true
 		}
 	}
 
@@ -207,9 +203,7 @@ func (ac *AccessController) isPermitted(userGroups []string, username, action st
 			if common.Contains(p.Actions, action) {
 				for _, group := range p.Groups {
 					if common.Contains(userGroups, group) {
-						result = true
-
-						return result
+						return true
 					}
 				}
 			}
@@ -217,20 +211,16 @@ func (ac *AccessController) isPermitted(userGroups []string, username, action st
 	}
 
 	// check defaultPolicy
-	if !result {
-		if common.Contains(policyGroup.DefaultPolicy, action) && username != "" {
-			result = true
-		}
+	if common.Contains(policyGroup.DefaultPolicy, action) && username != "" {
+		return true
 	}
 
 	// check anonymousPolicy
-	if !result {
-		if common.Contains(policyGroup.AnonymousPolicy, action) && username == "" {
-			result = true
-		}
+	if common.Contains(policyGroup.AnonymousPolicy, action) && username == "" {
+		return true
 	}
 
-	return result
+	return false
 }
 
 func BaseAuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
@@ -340,6 +330,43 @@ func DistSpecAuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
 			} else {
 				next.ServeHTTP(response, request) //nolint:contextcheck
 			}
+		})
+	}
+}
+
+func MetricsAuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(response http.ResponseWriter, request *http.Request) {
+			if ctlr.Config.HTTP.AccessControl == nil {
+				// allow access to authenticated user as anonymous policy does not exist
+				next.ServeHTTP(response, request)
+
+				return
+			}
+			if len(ctlr.Config.HTTP.AccessControl.Metrics.Users) == 0 {
+				log := ctlr.Log
+				log.Warn().Msg("auth is enabled but no metrics users in accessControl: /metrics is unaccesible")
+				common.AuthzFail(response, request, "", ctlr.Config.HTTP.Realm, ctlr.Config.HTTP.Auth.FailDelay)
+
+				return
+			}
+
+			// get access control context made in authn.go
+			userAc, err := reqCtx.UserAcFromContext(request.Context())
+			if err != nil { // should never happen
+				common.AuthzFail(response, request, "", ctlr.Config.HTTP.Realm, ctlr.Config.HTTP.Auth.FailDelay)
+
+				return
+			}
+
+			username := userAc.GetUsername()
+			if !common.Contains(ctlr.Config.HTTP.AccessControl.Metrics.Users, username) {
+				common.AuthzFail(response, request, username, ctlr.Config.HTTP.Realm, ctlr.Config.HTTP.Auth.FailDelay)
+
+				return
+			}
+
+			next.ServeHTTP(response, request) //nolint:contextcheck
 		})
 	}
 }
--- a/pkg/extensions/README.md
+++ b/pkg/extensions/README.md
@@ -30,9 +30,9 @@ package extensions
     IsAdmin         bool
     Username        string
     Groups          []string
-    } 
+    }
     ```
-  This data can then be accessed from the request context so that <b>every extension can apply its own authorization logic, if needed </b>. 
+  This data can then be accessed from the request context so that <b>every extension can apply its own authorization logic, if needed </b>.
 
 - when a new extension comes out, the developer should also write some blackbox tests, where a binary that contains the new extension should be tested in a real usage scenario. See [test/blackbox](test/blackbox/sync.bats) folder for multiple extensions examples.
 
@@ -40,6 +40,6 @@ package extensions
 
 - with every new extension, you should modify the EXTENSIONS variable in Makefile by adding the new extension. The EXTENSIONS variable represents all extensions and is used in Make targets that require them all (e.g make test).
 
-- the available extensions that can be used at the moment are: <b>sync, scrub, metrics, search </b>.
+- the available extensions that can be used at the moment are: <b>sync, search, scrub, metrics, lint, ui, mgmt, userprefs, imagetrust </b>.
 NOTE: When multiple extensions are used, they should be listed in the above presented order.
 
--- a/pkg/extensions/extension_metrics.go
+++ b/pkg/extensions/extension_metrics.go
@@ -26,13 +26,14 @@ func EnableMetricsExtension(config *config.Config, log log.Logger, rootDir strin
 }
 
 func SetupMetricsRoutes(config *config.Config, router *mux.Router,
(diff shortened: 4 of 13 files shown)


pkg/api/authz.go
```diff
@@ -10,6 +10,7 @@ import (
 	"zotregistry.io/zot/pkg/api/config"
 	"zotregistry.io/zot/pkg/api/constants"
 	"zotregistry.io/zot/pkg/common"
+	zcommon "zotregistry.io/zot/pkg/common"
```
andaaron, line 13: This is already imported 1 line above.

pkg/api/authz.go
```diff
@@ -343,3 +334,46 @@ func DistSpecAuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
...
+			userAc, err := reqCtx.UserAcFromContext(request.Context())
+			if err != nil { // should never happen
+				authFail(response, request, ctlr.Config.HTTP.Realm, ctlr.Config.HTTP.Auth.FailDelay)
+
+				return
+			}
+
+			username := userAc.GetUsername()
+			if !zcommon.Contains(ctlr.Config.HTTP.AccessControl.Metrics.Users, username) {
+				authFail(response, request, ctlr.Config.HTTP.Realm, ctlr.Config.HTTP.Auth.FailDelay)
```
andaaron, line 371: Maybe use zcommon.AuthzFail here?

pkg/api/authz.go
```diff
@@ -343,3 +334,46 @@ func DistSpecAuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
...
+				log.Warn().Msg("auth is enabled but no metrics users in accessControl: /metrics is unaccesible")
+				authFail(response, request, ctlr.Config.HTTP.Realm, ctlr.Config.HTTP.Auth.FailDelay)
+
+				return
+			}
+
+			// get access control context made in authn.go
+			userAc, err := reqCtx.UserAcFromContext(request.Context())
+			if err != nil { // should never happen
+				authFail(response, request, ctlr.Config.HTTP.Realm, ctlr.Config.HTTP.Auth.FailDelay)
```
andaaron, line 364: Maybe use zcommon.AuthzFail here?

pkg/api/authz.go
```diff
@@ -343,3 +334,46 @@ func DistSpecAuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
...
+			if ctlr.Config.HTTP.AccessControl == nil {
+				// allow access to authenticated user as anonymous policy does not exist
+				next.ServeHTTP(response, request)
+
+				return
+			}
+			if len(ctlr.Config.HTTP.AccessControl.Metrics.Users) == 0 {
+				log := ctlr.Log
+				log.Warn().Msg("auth is enabled but no metrics users in accessControl: /metrics is unaccesible")
+				authFail(response, request, ctlr.Config.HTTP.Realm, ctlr.Config.HTTP.Auth.FailDelay)
```
andaaron, line 356: Maybe use zcommon.AuthzFail here?

pkg/extensions/README.md
```diff
@@ -30,16 +30,16 @@ package extensions
...
+  This data can then be accessed from the request context so that <b>every extension can apply its own authorization logic, if needed </b>.
 
 - when a new extension comes out, the developer should also write some blackbox tests, where a binary that contains the new extension should be tested in a real usage scenario. See [test/blackbox](test/blackbox/sync.bats) folder for multiple extensions examples.
 
 - newly added blackbox tests should have targets in Makefile. You should also add them as Github Workflows, in [.github/workflows/ecosystem-tools.yaml](.github/workflows/ecosystem-tools.yaml)
 
 - with every new extension, you should modify the EXTENSIONS variable in Makefile by adding the new extension. The EXTENSIONS variable represents all extensions and is used in Make targets that require them all (e.g make test).
 
-- the available extensions that can be used at the moment are: <b>sync, scrub, metrics, search </b>.
+- the available extensions that can be used at the moment are: <b>sync, search, scrub, metrics, lint, ui, mgmt, profile, userprefs, imagetrust </b>.
```
andaaron, line 43: `profile` is not an extensions, just like `debug` is not extension

pkg/extensions/extension_metrics.go
```diff
@@ -26,13 +26,14 @@ func EnableMetricsExtension(config *config.Config, log log.Logger, rootDir strin
 }
 
 func SetupMetricsRoutes(config *config.Config, router *mux.Router,
-	authFunc mux.MiddlewareFunc, log log.Logger, metrics monitoring.MetricServer,
+	authFunc, autzFunc mux.MiddlewareFunc, log log.Logger, metrics monitoring.MetricServer,
```
andaaron, line 29:
```suggestion
	authFunc, authzFunc mux.MiddlewareFunc, log log.Logger, metrics monitoring.MetricServer,
```

andaaron (dismissed):

andaaron (dismissed):

pkg/test/common/utils.go
```diff
@@ -15,9 +15,13 @@ import (
 )
 
 const (
-	BaseURL       = "http://127.0.0.1:%s"
-	BaseSecureURL = "https://127.0.0.1:%s"
-	SleepTime     = 100 * time.Millisecond
+	BaseURL               = "http://127.0.0.1:%s"
+	BaseSecureURL         = "https://127.0.0.1:%s"
+	SleepTime             = 100 * time.Millisecond
+	Username              = "test"
```
andaaron, line 21: If you ask me these are values used in very specific test cases. Why leak them in a library?
IMHO they shouldn't have been globals in the first place.

adodon2go: We have code duplication in a lot of places(including functions that are copy-pasted). If they do not rely on some extensions libraries I think it is better to keep common code under test/common

andaaron: This is not reusable test code, this is test **data**, specific to some tests, which just happens to be repeated.
If there are functions we should move to the libraries, we can look at moving those.

adodon2go: Disagree on this. If those are test data than it should be defined (as variables) inside tests. This are constants that we reuse all over again due to some copy-paste and if something is not available in the test package, we copy-paste them and import there. Not a good practice.

andaaron: Then they need to become test variables. The username "test", password "test", repo name `everyone/isallowed` are specific to some tests. There's nothing general enough for these 3 to be package level variables or constants. The only one which could make some sense to be reused is `AuthorizationAllRepos`.

If someone made a bad decision by making them constants, it doesn't mean we should make that worse by moving them to another package and exporting them.

I don't want people to start thinking:
- Do I need to use this test.Username value, or can I use any usernames in my new tests? Would my new test break if I use a different username?
- Are there any dependencies in the test library on these variables?
- I want to use a different username, do I need to add it to the package test?

test/blackbox/metrics.bats
```diff
@@ -32,6 +32,7 @@ function setup_file() {
     zot_config_file=${BATS_FILE_TMPDIR}/zot_config.json
     zot_htpasswd_file=${BATS_FILE_TMPDIR}/zot_htpasswd
     htpasswd -Bbn ${AUTH_USER} ${AUTH_PASS} >> ${zot_htpasswd_file}
+    htpasswd -Bbn ${METRICS_USER} ${METRICS_PASS} >> ${zot_htpasswd_file}
```
rchincha, line 35: Make sure we add a test to cover no authn case also, meaning if authn is not configured, then anyone can still access.

adodon2go: This scenario is covered in TestMetricsAuthentication in monitoring_test.go

pkg/test/common/utils.go
adodon2go: Will update the PR by removing almost all constants from every test files and replacing them with local variables.

pkg/test/common/utils.go
```diff
@@ -15,9 +16,11 @@ import (
 )
 
 const (
-	BaseURL       = "http://127.0.0.1:%s"
-	BaseSecureURL = "https://127.0.0.1:%s"
-	SleepTime     = 100 * time.Millisecond
+	BaseURL               = "http://127.0.0.1:%s"
+	BaseSecureURL         = "https://127.0.0.1:%s"
+	SleepTime             = 100 * time.Millisecond
+	AuthorizedNamespace   = "everyone/isallowed"
```
andaaron, line 22: This repo name is specific to some tests.

pkg/test/common/utils.go
```diff
@@ -177,3 +180,16 @@ func CustomRedirectPolicy(noOfRedirect int) resty.RedirectPolicy {
 		return nil
 	})
 }
+
+func GenerateRandomString() string {
```
andaaron, line 184: Given the character constraints, maybe have different functions such as GenerateRandomRepoName(), GenerateRandomUsername(), GenerateRandomPassword(). Just a suggestion.

pkg/extensions/search/userprefs_test.go
```diff
@@ -818,17 +817,17 @@ func TestGlobalSearchWithUserPrefFiltering(t *testing.T) {
...
+		port := test.GetFreePort()
+		baseURL := test.GetBaseURL(port)
 		conf := config.New()
 		conf.HTTP.Port = port
 		conf.Storage.RootDirectory = dir
 
 		simpleUser := "simpleUser"
 		simpleUserPassword := "simpleUserPass"
-		credTests := fmt.Sprintf("%s\n\n", getCredString(simpleUser, simpleUserPassword))
+		credTests := fmt.Sprintf("%s\n\n", test.GetCredString(simpleUser, simpleUserPassword))
```
andaaron, line 828: fmt.Sprintf() may no longer be necessary, if test.GetCredString() already adds `\n`

pkg/extensions/search/userprefs_test.go
```diff
@@ -622,17 +621,17 @@ func TestChangingRepoState(t *testing.T) {
...
```
andaaron, line 632: fmt.Sprintf may no longer be necessary.

pkg/extensions/search/userprefs_test.go
```diff
@@ -468,9 +467,9 @@ func TestChangingRepoState(t *testing.T) {
 	forbiddenRepo := "forbidden"
 	accesibleRepo := "accesible"
 
-	credTests := fmt.Sprintf("%s\n\n", getCredString(simpleUser, simpleUserPassword))
+	credTests := fmt.Sprintf("%s\n\n", test.GetCredString(simpleUser, simpleUserPassword))
```
andaaron, line 470: fmt.Sprintf() may no longer be necessary.

pkg/api/controller_test.go
```diff
@@ -73,31 +72,23 @@ import (
...
-	if err != nil {
-		panic(err)
-	}
-
-	usernameAndHash := fmt.Sprintf("%s:%s", username, string(hash))
-
-	return usernameAndHash
-}
+var (
+	username         = test.GenerateRandomString() //nolint: gochecknoglobals
```
andaaron, line 84: Are all these variables used?

pkg/api/controller_test.go
```diff
@@ -1453,8 +1456,10 @@ func TestMutualTLSAuthWithUserPermissions(t *testing.T) {
 		So(err, ShouldBeNil)
 		caCertPool := x509.NewCertPool()
 		caCertPool.AppendCertsFromPEM(caCert)
-		htpasswdPath := test.MakeHtpasswdFile()
-		defer os.Remove(htpasswdPath)
+		/* username := test.GenerateRandomString()
```
andaaron, line 1459: To clarify what the change needs to be here before merge.

pkg/api/controller_test.go
adodon2go: Yes. There are used by LDAP related tests that rely on some common username between different tests.
In a different PR it will make sense to separate those tests into some ldap_tests.go

andaaron (approved):

examples/config-metrics-authn.json
```diff
@@ -0,0 +1,26 @@
+{
+    "distSpecVersion": "1.1.0-dev",
+    "storage": {
+        "rootDirectory": "/tmp/zot"
+    },
+    "http": {
```
rchincha, line 6: don't we need a AccessControl Policy for metrics also?

examples/config-metrics-authz.json
```diff
@@ -0,0 +1,40 @@
...
+            "metrics":{
+                "users": ["metrics"]
+            },
+            "repositories": {
+                "**": {
+                    "anonymousPolicy": [
+                        "read",
+                        "create"
+                    ],
+                    "defaultPolicy": ["read"]
```
andaaron, line 24: Nitpicking: I think it would make more sense for "defaultPolicy" to be "read" and "create", since these are logger in users, while "anonymousPolicy" to be "read", since the user identify is unknown in this case.

rchincha:
There is a test failure?

andaaron (approved):

PR #1897, opened 2023-10-06
fix: quick metrics setup in kind with prometheus

adodon2go:
**What type of PR is this?**
bug

We have a script in examples/metrics to quickly setup a kind cluster with prometheus support.
That can be used together with a running grafana instance to quickly display some graphs.

**Automation added to e2e**:


andaaron (approved):
lgtm

examples/metrics/Makefile
```diff
@@ -2,7 +2,7 @@ CONTAINER_RUNTIME := docker
 
 .PHONY: binary-container
 binary-container:
-	${CONTAINER_RUNTIME} build -f build/Dockerfile -t zot-build:latest ../../.
+	${CONTAINER_RUNTIME} build -f Dockerfile -t zot-build:latest ../../.
```
rchincha, line 5: Our Dockerfiles are located under build/ right?

examples/metrics/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # ---
 # Stage 1: Install certs, build binary, create default config file
 # ---
-FROM ghcr.io/project-zot/golang:1.19 AS builder
+FROM ghcr.io/project-zot/golang:1.20 AS builder
```
rchincha, line 4: Ideally, this should be build-arg'ed out but that also means whoever uses this needs to be aware.

examples/metrics/Makefile
adodon2go: Not anymore. Not sure who/why removed them from there (can be investigated)

andaaron: Are we discussing https://github.com/project-zot/zot/blob/main/build/Dockerfile? Or others?

rchincha: https://github.com/project-zot/zot/tree/main/build
^ i still see them?

rchincha:
Also can this be added as a nightly test?

andaaron (commented):
This looks good from my side.

adodon2go:
> Also can this be added as a nightly test?

I will add a new CI workflow to address this.

andaaron:
> > Also can this be added as a nightly test?
> 
> I will add a new CI workflow to address this.

Let's track this in https://github.com/project-zot/zot/issues/1930

andaaron (approved):

Issue #1914, opened 2023-10-09
[Bug]: delay on auth testing


aritas1:
### zot version

v2.0.0-rc6

### Describe the bug

My Watchtower clients (see [source](https://github.com/containrrr/watchtower/blob/284066a39d083fe4865ae43119a41b42d450b178/pkg/registry/auth/auth.go#L151)) and one of my docker/24.0.6 machines are probing the `/v2/` endpoint for every manifest request to see if an authentication is required.
With the the defaut settings, these requests get delayed for 5 seconds, so pulling layers gets very slow.

Some logs (note the lower latency due to my config):
```
zot-zot-1  | {"level":"error","error":"auth: failed parsing authorization header","goroutine":265496,"caller":"zotregistry.io/zot/pkg/api/authn.go:137","time":"2023-10-09T19:02:26.244222203Z","message":"failed to parse authorization header"}
zot-zot-1  | {"level":"info","module":"http","clientIP":"172.16.0.33:56702","method":"GET","path":"/v2/","statusCode":401,"latency":"1s","bodySize":283,"headers":{"Accept":["*/*"],"Accept-Encoding":["gzip"],"Connection":["close"],"User-Agent":["Watchtower (Docker)"],"X-Forwarded-For":["::ffff:xx.xx.xx.xx"],"X-Forwarded-Port":["443"],"X-Forwarded-Proto":["https"],"X-Real-Ip":["::ffff:xx.xx.xx.xx"]},"goroutine":265496,"caller":"zotregistry.io/zot/pkg/api/session.go:131","time":"2023-10-09T19:02:27.244453749Z","message":"HTTP API"}
zot-zot-1  | {"level":"info","identity":"nas","goroutine":265498,"caller":"zotregistry.io/zot/pkg/api/authn.go:177","time":"2023-10-09T19:02:27.413767392Z","message":"user profile successfully set"}
[...]

PR #1920, opened 2023-10-10
fix(authn): apply fail delay only if credentials/sessions are supplied

peusebiu:
**What type of PR is this?**

https://github.com/project-zot/zot/issues/1914

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -675,7 +675,9 @@ func getRelyingPartyArgs(cfg *config.Config, provider string) (
 }
 
 func authFail(w http.ResponseWriter, r *http.Request, realm string, delay int) {
-	time.Sleep(time.Duration(delay) * time.Second)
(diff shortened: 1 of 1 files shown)


andaaron (dismissed):

pkg/api/authn.go
```diff
@@ -675,7 +675,9 @@ func getRelyingPartyArgs(cfg *config.Config, provider string) (
 }
 
 func authFail(w http.ResponseWriter, r *http.Request, realm string, delay int) {
-	time.Sleep(time.Duration(delay) * time.Second)
+	if !isAuthorizationHeaderEmpty(r) || !hasSessionHeader(r) {
```
rchincha, line 678: Shouldn't it be a && instead here?

rchincha, line 678: hasSessionHeader - remove the ! negation

andaaron (approved):

PR #1927, opened 2023-10-12
feat(api): repair corrupted blobs when pushed again

### LINKED ISSUE #1922: [Feat]: Repair corrupted blobs when pushed again

Push an image to zot
truncate -s 0 a blob
Push the image again.

Ideally, on the HEAD/GET requests for the second push, return that the blob doesn't exist so that it gets pushed again.

`scrub` will catch these errors, but we should also do this ^

_No response_

_No response_

### Additional context

_No response_

peusebiu:
CheckBlob() returns ErrBlobNotFound on corrupted blobs

closes #1922

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/imagestore/imagestore.go
+++ b/pkg/storage/imagestore/imagestore.go
@@ -866,20 +866,11 @@ func (is *ImageStore) FinishBlobUpload(repo, uuid string, body io.Reader, dstDig
 		return err
 	}
 
-	fileReader, err := is.storeDriver.Reader(src, 0)
-	if err != nil {
-		is.log.Error().Err(err).Str("blob", src).Msg("failed to open file")
-
-		return zerr.ErrUploadNotFound
-	}
-
-	defer fileReader.Close()
-
-	srcDigest, err := godigest.FromReader(fileReader)
+	srcDigest, err := getBlobDigest(is, src)
 	if err != nil {
 		is.log.Error().Err(err).Str("blob", src).Msg("failed to open blob")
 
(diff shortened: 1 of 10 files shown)


andaaron (commented):
Code looks ok. 

Maybe we should make this a configurable feature?
Computing these digests when a file is downloaded may be expensive.

andaaron (dismissed):

pkg/storage/imagestore/imagestore.go
```diff
@@ -1630,6 +1646,74 @@ func (is *ImageStore) deleteBlob(repo string, digest godigest.Digest) error {
...
+func getBlobSizeFromRepoManifests(imgStore *ImageStore, repo string, blobDigest godigest.Digest, log zlog.Logger,
+) (int64, error) {
+	index, err := common.GetIndex(imgStore, repo, log)
+	if err != nil {
+		return -1, err
+	}
+
+	for _, desc := range index.Manifests {
+		switch desc.MediaType {
+		case ispec.MediaTypeImageManifest:
```
peusebiu, line 1659: will add oras manifest

peusebiu:
needed to increase a bit the delays on minio because it failed multiple times. The added overhead of checking manifests is big enough to interfere, but small enough to not impact the Minimum performance regression check.

When I used the digest to calculate a blob's digest, and validate, the Minimum performance regression check failed.

andaaron (approved):

rchincha (approved):
tested this.

lgtm

PR #1929, opened 2023-10-16
fix(tests): update imagetrust tests to use mock service

Andreea-Lupu:
- use secretsManagerMock and secretsManagerCacheMock to avoid failing because of "already exists" error when running multiple times image_trust_test on the same localstack instance

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


andaaron (approved):

pkg/extensions/imagetrust/cosign.go
```diff
@@ -256,11 +257,11 @@ func (cloud *PublicKeyAWSStorage) StorePublicKey(name godigest.Digest, publicKey
 	}
 
 	_, err := cloud.secretsManagerClient.CreateSecret(context.Background(), secretInputParam)
-	if err != nil {
-		return err
+	if err != nil && strings.Contains(err.Error(), "already exists") {
```
rchincha, line 260: can we not use errors.Is() instead?

rchincha: Same in notation.go

andaaron (approved):

peusebiu (approved):

PR #1946, opened 2023-10-18
ci: use runners provided by CNCF

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


andaaron (approved):

rchincha:
Not sure if this PR is helping speed up ci though!

adodon2go:
It definitely improves speed time: 3x faster running test-extended using this (from 48m->15m)
Also it helps running kind in CI: my prometheus script fails without this change

PR #1950, opened 2023-10-19
fix: tests refactoring

adodon2go:
**What type of PR is this?**
refactoring

**Automation added to e2e**:

### DIFF
--- a/pkg/test/common/utils.go
+++ b/pkg/test/common/utils.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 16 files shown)


andaaron (dismissed):
LGTM

pkg/test/common/utils.go
```diff
@@ -177,3 +179,31 @@ func CustomRedirectPolicy(noOfRedirect int) resty.RedirectPolicy {
 		return nil
 	})
 }
+
+// Generates a random string with length 10 from lower case & upper case characters.
+func GenerateRandomString() string {
+	//nolint: gosec
+	seededRand := rand.New(rand.NewSource(time.Now().UnixNano()))
```
rchincha, line 186: Log this seed so that if there is a test failure that I need to reproduce, I can manually set it to that seed value.

Andreea-Lupu (approved):

Issue #1965, opened 2023-10-25
[Bug]: zot storage metrics needs to be more accurate with what is actually saved on disk


adodon2go:
### zot version

v2.0.0-rc6

### To reproduce

1. Configuration
2. Client tool used
3. Seen error

PR #1972, opened 2023-10-30
fix: more accurate storage metrics after zot restart

adodon2go:
**What type of PR is this?**
bug

#1965 

**Automation added to e2e**:


pkg/api/controller.go
```diff
@@ -372,6 +372,10 @@ func (c *Controller) StartBackgroundTasks(reloadCtx context.Context) {
 		ext.EnableMetricsExtension(c.Config, c.Log, c.Config.Storage.RootDirectory)
 		ext.EnableSearchExtension(c.Config, c.StoreController, c.MetaDB, taskScheduler, c.CveScanner, c.Log)
 	}
+	// runs once if metrics are enabled
+	if c.Config.IsMetricsEnabled() {
+		c.StoreController.DefaultStore.PopulateStorageMetrics(time.Duration(0), taskScheduler)
```
peusebiu, line 377: here we run storage metric for both local and s3, but getDirSize() func supports only local imagestore.

Either update that function to take the storage/types.go.Driver implementation as an arg and use the driver to get dir size, or run it only for local storage, and we fixed this later.

otherwise lgtm

rchincha: Try to fix this now so we are done with this.

pkg/test/mocks/image_store_mock.go
```diff
@@ -55,6 +55,7 @@ type MockedImageStore struct {
 	GetAllBlobsFn                func(repo string) ([]string, error)
 	CleanupRepoFn                func(repo string, blobs []godigest.Digest, removeRepo bool) (int, error)
 	PutIndexContentFn            func(repo string, index ispec.Index) error
+	RunStorageMetricsFn          func(interval time.Duration, sch *scheduler.Scheduler)
```
peusebiu, line 58: rename this to PopulateStorageMetricsFn :D

peusebiu (commented):
otherwise lgtm

peusebiu (approved):

PR #1988, opened 2023-10-30
feat(sha512): add support for sha512 digest

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/imagestore/imagestore.go
+++ b/pkg/storage/imagestore/imagestore.go
@@ -3,9 +3,11 @@ package imagestore
 import (
 	"bytes"
 	"crypto/sha256"
+	"crypto/sha512"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"hash"
 	"io"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


rchincha:
cc: @ktarplee

pkg/storage/imagestore/imagestore.go
```diff
@@ -933,7 +935,14 @@ func (is *ImageStore) FullBlobUpload(repo string, body io.Reader, dstDigest godi
 
 	uuid := u.String()
 	src := is.BlobUploadPath(repo, uuid)
-	digester := sha256.New()
+
+	var digester hash.Hash
+	if dstDigest.Algorithm() == godigest.SHA256 {
```
adodon2go, line 940: we need to add support for SHA384 too:
```
else if dstDigest.Algorithm() == godigest.SHA384 {
   digester = sha512.New384()
}
```

ktarplee: You should not need to reference sha256, sha384, or sha512 in your code.  The godigest library does that for you.

```go
digester := dstDigest.Algorithm().Hash() # hash.Hash is returned
```

ktarplee: Of course you should check that the algorithm is available with `algDigest.Algorithm().Available()` (otherwise it will panic).

ktarplee: The main branch of godigest even allows one to register new algorithms if they choose.  See [this](https://github.com/opencontainers/go-digest/blob/d50d2fec9c98606a94f0620bf09ed8a81a8aa237/algorithm.go#L78C11-L78C11).  So having the list of algorithms in Zot as a switch statement is not ideal.  Let the library from open containers handle it for you. :)

andaaron:
Let's close this in favor of https://github.com/project-zot/zot/pull/2075
Note there's some clarifications needed on the spec, see: https://github.com/opencontainers/distribution-spec/issues/494

PR #2030, opened 2023-11-13
build(zui): add new makefile variables for the zui repo

andaaron:
To enable building zui from forks in case of local testing


rchincha (approved):
lgtm, this is useful if we want to switch repos to test changes.

PR #2041, opened 2023-11-13
feat(cluster): initial commit for scale-out cluster

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -50,6 +50,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/secretsmanager v1.28.6
 	github.com/aws/aws-secretsmanager-caching-go v1.1.3
 	github.com/containers/image/v5 v5.30.0
+	github.com/dchest/siphash v1.2.3
 	github.com/google/go-github/v52 v52.0.0
 	github.com/gorilla/securecookie v1.1.2
 	github.com/gorilla/sessions v1.2.2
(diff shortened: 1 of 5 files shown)


rchincha:
Proposal:

1. Single nodes have an upper limit on CPU, memory and storage. So eventually, have to grow out of single nodes.
2. Repository "name" based _static_ routing.
3. Each node either can handle the request locally, or must route.
4. HTTP redirect based (3xx) or proxy?
5. Choose proxy in 4. since not clear if clients handle 3xx, however, there is an extra cost in proxy/networking.
6. But nice in the sense that we can expose all nodes via DNS or a stateless ingress can hide and spray to these nodes.
7. The cluster size can grow or shrink, which means the new lookup may not find data locally.
8. However, since content-based lookups, we can add the other nodes as an implicit "sync" rule for a particular "name". If the data is not found locally, "sync" will get it from other members, so the cluster re-balances over time.

elee1766:
previously i have worked with seaweedfs, which allows for side-adjustable clusters with discovery similar to how you describe. they use raft for this: https://github.com/seaweedfs/seaweedfs/blob/master/weed/server/raft_server.go  

that said, what is the rationale to implement your own cluster system? i feel like this is something better solved separately, since as a user i would probably feel more comfortable using zot on top of redis/ceph instead of making zot do the replication/sharding.

andaaron:
Let's close this, as https://github.com/project-zot/zot/pull/2385 accomplishes what this PR was supposed to.

PR #2042, opened 2023-11-14
fix: TestPopulateStorageMetrics fails occasionally in CI

adodon2go:
**What type of PR is this?**
bug

**Automation added to e2e**:

### DIFF
--- a/pkg/storage/common/common.go
+++ b/pkg/storage/common/common.go
@@ -1095,7 +1095,7 @@ func (gen *StorageMetricsInitGenerator) Next() (scheduler.Task, error) {
 	}
 	gen.lastRepo = repo
 
-	return NewStorageMetricsTask(gen.ImgStore, gen.Metrics, repo), nil
+	return NewStorageMetricsTask(gen.ImgStore, gen.Metrics, repo, gen.Log), nil
 }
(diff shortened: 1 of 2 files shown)


andaaron (approved):

pkg/storage/common/common.go
```diff
@@ -1116,16 +1116,19 @@ type smTask struct {
...
+	log zlog.Logger,
 ) *smTask {
-	return &smTask{imgStore, metrics, repo}
+	return &smTask{imgStore, metrics, repo, log}
 }
 
 func (smt *smTask) DoWork(ctx context.Context) error {
 	// run task
 	monitoring.SetStorageUsage(smt.metrics, smt.imgStore.RootDir(), smt.repo)
+	smt.log.Debug().Msg("monitoring: computed storage usage for repo " + smt.repo)
```
rchincha, line 1131: .Str("repo", smt.repo).Msg("...")

adodon2go: Agree, but for testing purposes where we wait for some specific string messages in the logs before proceeding further and asserting, we need the message in that form unless we will start using some json go libraries for better log parsing

PR #2044, opened 2023-11-14
build: test both go 1.20 and 1.21

andaaron:

### DIFF
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -14,11 +14,14 @@ jobs:
   golangci:
     name: lint
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


andaaron:
Looks like the go.mod requires an update when used with Go 1.21.
To keep compatibility with 1.20 we would need to `go mod tidy -go=1.20`
But we get the error `go: github.com/letsencrypt/boulder@v0.0.0-20231026200631-000cd05d5491 requires go@1.21, but 1.20 is requested`

Seem unlikely we can keep a single committed version of the go.mod file for both 1.20 and 1.21.

I will close this PR, and will investigate using only version 1.21 for zot release 2.0.1.

Issue #1710, opened 2023-08-17
[Bug]: Handle parameters in "Content-Type=<oci-media-types>; charset=utf-8"


rchincha:
### zot version

v1.4.3 and above

### Describe the bug

Expect and handle parameters. Although the spec says exact media-type, but clients may add params.

### To reproduce

1. Configuration
2. Client tool used
3. Seen error

andaaron:
@rchincha can you please update the description to be more explicit?

andaaron:
Related to https://github.com/opencontainers/distribution-spec/issues/408

rchincha:
https://github.com/opencontainers/distribution-spec/pull/469
^ we should probably expect paramaters (to be safe) but ignore everything except params[0]

PR #2055, opened 2023-11-16
fix: handle parameters in Content-Type header

Andreea-Lupu:
**What type of PR is this?**

#1710 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -663,7 +663,7 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
 		return
 	}
 
-	mediaType := request.Header.Get("Content-Type")
+	mediaType := zcommon.GetContentType(request)
 	if !storageCommon.IsSupportedMediaType(mediaType) {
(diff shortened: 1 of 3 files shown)


andaaron (approved):

pkg/api/routes.go
```diff
@@ -663,7 +663,7 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
 		return
 	}
 
-	mediaType := request.Header.Get("Content-Type")
+	mediaType := zcommon.GetContentType(request)
```
rchincha, line 666: There are two places in pkg/api/routes.go?
666:    mediaType := request.Header.Get("Content-Type")
1219:           if contentType := request.Header.Get("Content-Type"); contentType != constants.BinaryMediaType {

Andreea-Lupu: I wasn't sure about L1219. In spec it is specified for `/v2/{name}/manifests/{reference} [put]` (L666) that the parameters on the Content-Type header should be ignore , but for `/v2/{name}/blobs/uploads [post]` (L1219) there is no info about that. Do you think I should make the change also for L1219?

PR #2073, opened 2023-11-21
Sync s3

peusebiu:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


rchincha:
@peusebiu pls add an example as well so folks know what to try

rchincha:
cc: @elee1766

pkg/extensions/config/sync/config.go
```diff
@@ -15,6 +15,7 @@ type Credentials struct {
 type Config struct {
 	Enable          *bool
 	CredentialsFile string
+	TmpDir          string
```
rchincha, line 18: Call this SyncDir and add a comment that this is needed only for cloud case.
Also, can we add a check in LoadConfig that this is indeed the case.

andaaron (approved):

andaaron (approved):

rchincha (approved):
lgtm

PR #2075, opened 2023-11-22
fix(digests): do not mandate sha256 as the only algorithm used for hashing blobs

andaaron:

### DIFF
--- a/pkg/storage/imagestore/imagestore.go
+++ b/pkg/storage/imagestore/imagestore.go
@@ -2,7 +2,6 @@ package imagestore
 
 import (
 	"context"
-	"crypto/sha256"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -490,9 +489,9 @@ func (is *ImageStore) PutImageManifest(repo, reference, mediaType string, //noli
 		refIsDigest = false
 	}
 
-	dig, err := common.ValidateManifest(is, repo, reference, mediaType, body, is.log)
+	err = common.ValidateManifest(is, repo, reference, mediaType, body, is.log)
 	if err != nil {
-		return dig, "", err
+		return mDigest, "", err
 	}
 
 	index, err := common.GetIndex(is, repo, is.log)
@@ -547,11 +546,11 @@ func (is *ImageStore) PutImageManifest(repo, reference, mediaType string, //noli
 	}
 
 	if !updateIndex {
-		return desc.Digest, subjectDigest, nil
+		return mDigest, subjectDigest, nil
 	}
 
 	// write manifest to "blobs"
-	dir := path.Join(is.rootDir, repo, "blobs", mDigest.Algorithm().String())
+	dir := path.Join(is.rootDir, repo, ispec.ImageBlobsDir, mDigest.Algorithm().String())
 	manifestPath := path.Join(dir, mDigest.Encoded())
 
 	if _, err = is.storeDriver.WriteFile(manifestPath, body); err != nil {
@@ -584,7 +583,7 @@ func (is *ImageStore) PutImageManifest(repo, reference, mediaType string, //noli
 		return "", "", err
 	}
 
-	return desc.Digest, subjectDigest, nil
+	return mDigest, subjectDigest, nil
 }
 
 // DeleteImageManifest deletes the image manifest from the repository.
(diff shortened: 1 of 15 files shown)


andaaron:
See also https://github.com/opencontainers/distribution-spec/issues/494

pkg/storage/imagestore/imagestore.go
```diff
@@ -1689,40 +1692,54 @@ func (is *ImageStore) deleteBlob(repo string, digest godigest.Digest) error {
...
-	dir := path.Join(is.rootDir, repo, "blobs", "sha256")
+func (is *ImageStore) GetAllBlobs(repo string) ([]godigest.Digest, error) {
+	blobsDir := path.Join(is.rootDir, repo, ispec.ImageBlobsDir)
 
-	files, err := is.storeDriver.List(dir)
-	if err != nil {
-		if errors.As(err, &driver.PathNotFoundError{}) {
-			is.log.Debug().Msg("empty rootDir")
+	algorithms := []godigest.Algorithm{
+		godigest.SHA256,
```
ktarplee, line 1716: Do you really want to list the approved algorithms or let godigest do that for you.  When you create a `godigest.Algorithm` you can call `alg.Available()` to see if the digest algorithm has been registered.  I do not believe the spec limits the algorithms allowed (it only requires sha256).   Then other algorithms can be added without modifying this code.   godigest allows other algorithms to be registered with [godigest.RegisterAlgorithm](https://github.com/opencontainers/go-digest/blob/master/algorithm.go#L78).  There is a PR to add blake3 to the list of algorithms and that should ideally not require any changes to the Zot code based except a version bump on godigest.

andaaron: The issue is the godigest library doesn't have an API to retrieve a slice of all registered algorithms. So I ended up listing them by hand. 

I also made the assumption these algorithms here are always available, as they are defined as constants there in the library. Is this correct?

andaaron: Worked around this issue by checking the names of the existing folders and checking if they are available, as opposed to obtaining the list of available algorithms first.

peusebiu (dismissed):
lgtm

ktarplee:
- @andaaron What else is needed on this PR?  What you have looks good however I did not see the REST API changes.  I expected there to be some changes necessary there.

andaaron:
> * @andaaron What else is needed on this PR?  What you have looks good however I did not see the REST API changes.  I expected there to be some changes necessary there.

We need the spec for specifying a non-canonical digest algorithm for the manifest in case of pushing by tag.
Once we have the spec we can implement that.

Otherwise this PR should be good for the blobs use case. But I don't know if there's any case in which the blobs use non-canonical  digest algorithms while the manifest uses the canonical (sha256) digest algorithm.

andaaron:
@rchincha I propose we merge this as is, and the API changes in a separate PR after https://github.com/opencontainers/distribution-spec/pull/543 is agreed upon.

rchincha:
```
Summarizing 6 Failures:
  [FAIL] OCI Distribution Conformance Tests Pull Pull manifests [It] HEAD request to sha512 manifest (digest) should yield 200 response                                     /data/hdd/rchincha/go/src/github.com/opencontainers/distribution-spec/conformance/01_pull_test.go:236                                                                     [FAIL] OCI Distribution Conformance Tests Pull Pull manifests [It] HEAD request to sha512 manifest (tag) should yield 200 response
  /data/hdd/rchincha/go/src/github.com/opencontainers/distribution-spec/conformance/01_pull_test.go:264                                                                     [FAIL] OCI Distribution Conformance Tests Pull Pull manifests [It] GET request to sha512 manifest (digest) should yield 200 response                                      /data/hdd/rchincha/go/src/github.com/opencontainers/distribution-spec/conformance/01_pull_test.go:301                                                                     [FAIL] OCI Distribution Conformance Tests Pull Teardown [It] Delete sha512 manifest created in setup                                                                      /data/hdd/rchincha/go/src/github.com/opencontainers/distribution-spec/conformance/01_pull_test.go:383                                                                     [FAIL] OCI Distribution Conformance Tests Push Manifest Upload [It] GET request to sha512 manifest URL (digest) should yield 200 response                                 /data/hdd/rchincha/go/src/github.com/opencontainers/distribution-spec/conformance/02_push_test.go:452                                                                     [FAIL] OCI Distribution Conformance Tests Push Teardown [It] Delete manifest created in tests                                                                             /data/hdd/rchincha/go/src/github.com/opencontainers/distribution-spec/conformance/02_push_test.go:474   
  ```
  

https://github.com/sudo-bmitch/distribution-spec/tree/pr-digest-on-tag-push
^ building against this

rchincha (approved):
lgtm

PR #2076, opened 2023-11-22
feat(metrics): add scheduler related metrics

adodon2go:
**What type of PR is this?**
feature

#1870 

**Automation added to e2e**:


pkg/scheduler/scheduler.go
```diff
@@ -402,12 +497,48 @@ func (scheduler *Scheduler) SubmitGenerator(taskGenerator TaskGenerator, interva
...
+
+	switch p {
+	case LowPriority:
+		priority = "low"
+	case MediumPriority:
+		priority = "medium"
+	case HighPriority:
+		priority = "high"
+	default:
+		priority = "InvalidPriority"
```
andaaron, line 523: Why is this capitalized, and the other values are not?

pkg/scheduler/scheduler.go
```diff
@@ -402,12 +497,48 @@ func (scheduler *Scheduler) SubmitGenerator(taskGenerator TaskGenerator, interva
...
+
+	switch s {
+	case ready:
+		status = "ready"
+	case waiting:
+		status = "waiting"
+	case done:
+		status = "done"
+	default:
+		status = "InvalidState"
```
andaaron, line 540: Why is this capitalized, and the other values are not?

pkg/scheduler/scheduler.go
```diff
@@ -217,13 +310,15 @@ func (scheduler *Scheduler) generateTasks() {
...
-	// check if the generator with highest prioriy is ready to run
+	// check if the generator with highest priority is ready to run
 	if scheduler.generators[0].getState() == ready {
 		gen = scheduler.generators[0]
 	} else {
 		gen, _ = heap.Pop(&scheduler.generators).(*generator)
 		if gen.getState() == waiting {
 			scheduler.waitingGenerators = append(scheduler.waitingGenerators, gen)
+		} else if gen.getState() == done {
+			scheduler.doneGenerators = append(scheduler.doneGenerators, gen)
```
andaaron, line 321: Is there a mechanism to remove the generators from this slice?

pkg/api/cookiestore.go
```diff
@@ -157,3 +158,13 @@ func (cleanTask *CleanTask) DoWork(ctx context.Context) error {
...
 	return nil
 }
+
+func (cleanTask *CleanTask) String() string {
+	return fmt.Sprintf("{SessionCleanup: %s, sessions: %s}",
+		"session clean task", // description of generator's task purpose
+		cleanTask.sessions)
+}
+
+func (cleanTask *CleanTask) Name() string {
```
andaaron, line 168: Are the "Name()" methods called somewhere? I looked for the code and could not find the calls.

pkg/scheduler/scheduler.go
```diff
@@ -352,6 +445,8 @@ func (gen *generator) generate(sch *Scheduler) {
 
 			return
 		}
+
+		task = nextTask
```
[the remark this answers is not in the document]
andaaron: Discussed this over IM, it's cosmetic only.

pkg/scheduler/scheduler.go
andaaron: I suggest it to be "invalid", for consistency

pkg/api/cookiestore.go
andaaron: So it is only called by the monitoring logic.
I noticed the "Name()" method in some of the task implementations are not covered.
Wouldn't it be better if "Name()" was also called in all the "String()" implementations?

```
	return fmt.Sprintf("{%s: %s, sessions: %s}",
		cleanTask.Name(),
		"session clean task", // description of generator's task purpose
		cleanTask.sessions)
```

This way we make we'll both increase coverage and have one less place where to hardcode the task/generator name

PR #2077, opened 2023-11-23
feat(metadb): search images by blob digest

laurentiuNiculae:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


andaaron (dismissed):

andaaron (approved):

rchincha:
@laurentiuNiculae so what is the ux for this?
In the same search box, I type as usual and if at some point it detects a sha256 it interprets as a blob sha?

laurentiuNiculae:
@rchincha yes, we can search by digests by writing the following in the search bar:

> sha256:abc123...

So we can write a full or partial hash, but we need to specify the hashing algorithm. The results will match hashes that start with the given string

PR #2120, opened 2023-12-08
fix(proto): the size of the repo should be int64, since that is the same type used for the manifest/config/index/digest sizes it sums up.

andaaron:
Using int32 may result in negative size values when returned by the graphql API


peusebiu (approved):

rchincha (approved):
lgtm

PR #2147, opened 2023-12-20
feat(config): handle config files with no explicit extension

peusebiu:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/cli/server/root.go
+++ b/pkg/cli/server/root.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/cli/server/root.go
```diff
@@ -736,12 +738,38 @@ func LoadConfiguration(config *config.Config, configPath string) error {
...
+	switch ext {
+	case "":
+		log.Info().Msg("config file with no extension, trying all supported config types")
+
+		for _, configType := range viper.SupportedExts {
+			viperInstance.SetConfigType(configType)
+			viperInstance.SetConfigFile(configPath)
+
+			err := viperInstance.ReadInConfig()
+			if err == nil {
```
andaaron, line 759: Don't we need to track if:
- at least one extension there was no error, so the log at line 763 should not show at all
- all the extensions failed we should return early from this function
?

pkg/cli/server/root.go
```diff
@@ -738,12 +740,44 @@ func LoadConfiguration(config *config.Config, configPath string) error {
...
+			viperInstance.SetConfigFile(configPath)
+
+			err = viperInstance.ReadInConfig()
+			if err == nil {
+				break
+			}
+		}
+
+		if err != nil {
+			log.Error().Err(err).Str("path", configPath).Msg("failed to read configuration, tried all suppoerted config types")
```
andaaron, line 769:
```suggestion
			log.Error().Err(err).Str("path", configPath).Msg("failed to read configuration, tried all supported config types")
```

andaaron (approved):

PR #2150, opened 2023-12-20
fix: enable panic backtraces

rchincha:
We expect panics in the server/datapath to be few and far between. So the backtraces are more valuable now.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -117,8 +117,9 @@ func (c *Controller) Run() error {
 
 	engine.Use(
 		SessionLogger(c),
-		handlers.RecoveryHandler(handlers.RecoveryLogger(c.Log),
-			handlers.PrintRecoveryStack(false)))
+		handlers.RecoveryHandler(
+			handlers.PrintRecoveryStack(true),
+		))
 
 	if c.Audit != nil {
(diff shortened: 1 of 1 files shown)


pkg/api/controller.go
```diff
@@ -117,8 +117,9 @@ func (c *Controller) Run() error {
 
 	engine.Use(
 		SessionLogger(c),
-		handlers.RecoveryHandler(handlers.RecoveryLogger(c.Log),
```
andaaron, line 120: Why remove the handlers.RecoveryLogger(c.Log)?

rchincha: Keeping that causes panic backtrace to not show.

andaaron: So where should it be shown, in stdout or in the logfile?
I assume in whatever place is configured via zot config, so where is that reference to the logger in this specific case?

andaaron: Shouldn't changing true to false be enough to solve the issue?

rchincha: Doesn't appear to be enough.

andaaron (approved):

PR #2151, opened 2023-12-20
fix: npe if ldap query doesn't return attributes

rchincha:
We cannot assume the LDAP server will have group attributes programmed everytime. So handle it accordingly.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/ldap.go
+++ b/pkg/api/ldap.go
@@ -173,8 +173,11 @@ func (lc *LDAPClient) Authenticate(username, password string) (bool, map[string]
 	}
 
 	attributes := lc.Attributes
+
 	attributes = append(attributes, "dn")
-	attributes = append(attributes, lc.UserGroupAttribute)
+	if lc.UserGroupAttribute != "" {
+		attributes = append(attributes, lc.UserGroupAttribute)
+	}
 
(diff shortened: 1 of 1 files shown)


pkg/api/ldap.go
```diff
@@ -216,8 +216,11 @@ func (lc *LDAPClient) Authenticate(username, password string) (bool, map[string]
 	}
 
 	userDN := search.Entries[0].DN
-	userAttributes := search.Entries[0].Attributes[0]
-	userGroups := userAttributes.Values
+	var userGroups []string
+	if len(search.Entries[0].Attributes) > 0 {
```
andaaron, line 220: My understanding is this a configuration issue which would happen in all cases of authentication if the zot or ldap server configuration is not changed by the zot admin or on the other side the ldap admin.

There's no way right now for the zot to know why permissions are not working because of this mismatched configuration. We should not let this pass silently.

I think there should be a warning in the logs in case the user group attribute set in the zot config does not match the actual ldap configuration.

andaaron (commented):
Maybe we should also take in consideration not adding the user group attribute in it if the config has empty string value, at line 177?

andaaron (approved):

PR #2158, opened 2024-01-08
fix: the scheduler is now fair

andaaron:
Generators are now ordered by rank in the priority queue.

    The rank computation formula is:
    - 100/(1+generated_task_count) for high priority tasks
    - 10/(1+generated_task_count) for medium priority tasks
    - 1/(1+generated_task_count) for low priority tasks

    Note the ranks are used when comparing generators both with the same priority and with different priority.
    So now we are:
    - giving an opportunity to all generators with the same priority to take turns generating tasks
    - giving roughly 1 low priority and 10 medium priority tasks the opportunity to run for every 100 high priority tasks running.

    After a generator generates a task, the generators are reordered in the priority queue based on rank.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -3,6 +3,7 @@ package scheduler
 import (
 	"container/heap"
 	"context"
+	"math"
 	"runtime"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


rchincha:
So this is a dynamic heuristic - 
Between generators of constant output rate, numerator determines priority and when one of them has a higher rate, priority decreases.

PR #2164, opened 2024-01-12
fix: excessive memory usage

peusebiu:
instead of reading entire files before calculating their digests stream them by using their Reader method.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/scrub.go
+++ b/pkg/storage/scrub.go
@@ -279,17 +279,8 @@ func CheckLayers(
 	}
 
 	for _, layer := range man.Layers {
-		layerContent, err := imgStore.GetBlobContent(imageName, layer.Digest)
-		if err != nil {
+		if err := imgStore.VerifyBlobDigestValue(imageName, layer.Digest); err != nil {
 			imageRes = getResult(imageName, tagName, layer.Digest, err)
-
-			break
-		}
-
-		computedDigest := godigest.FromBytes(layerContent)
-
-		if computedDigest != layer.Digest {
-			imageRes = getResult(imageName, tagName, layer.Digest, errors.ErrBadBlobDigest)
 
 			break
 		}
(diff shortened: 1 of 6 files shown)


pkg/storage/scrub.go
```diff
@@ -279,21 +279,28 @@ func CheckLayers(
...
+			break
+		}
 
 		if computedDigest != layer.Digest {
 			imageRes = getResult(imageName, tagName, layer.Digest, errors.ErrBadBlobDigest)
 
 			break
 		}
 
+		_ = layerReader.Close()
```
andaaron, line 302: We need to make sure this is called, even on the `break` cases above.
Why not move the entire code block to a separate function?

rchincha:
@peusebiu can we see the effect of this change when a image with several large blobs need to be scrubbed?

peusebiu:
> @peusebiu can we see the effect of this change when a image with several large blobs need to be scrubbed?

Ok I needed to handle a lock case, so I made an update to add a new storage method which doesn't lock.

Yes tested it and no memory spike at all, on my machine is stays on a steady 1.3% memory, on main it has a lot of spikes when it's digesting big blobs.

pkg/storage/types/types.go
```diff
@@ -57,6 +57,7 @@ type ImageStore interface { //nolint:interfacebloat
 	PutIndexContent(repo string, index ispec.Index) error
 	StatIndex(repo string) (bool, int64, time.Time, error)
 	GetBlobContent(repo string, digest godigest.Digest) ([]byte, error)
+	GetBlobReader(repo string, digest godigest.Digest) (io.ReadCloser, int64, error)
```
andaaron, line 60: It is not clear how this is different from the method at line 51

```suggestion
        VerifyBlobDigestValue(repo string, digest godigest.Digest) error
```

Let's call this function from the scrub logic.

PR #2176, opened 2024-01-22
fix(bearer): fixed /v2/ route not implementing token spec

peusebiu:
We use chartmuseum lib for handling bearer requests, which is not implementing the token spec, mainly it expects "scope" parameter to be given on every request, even for /v2/ route which doesn't represent a resource.

Handle this /v2/ route inside our code.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


andaaron (approved):

pkg/api/cookiestore.go
```diff
@@ -110,6 +110,10 @@ func getExpiredSessions(dir string) ([]string, error) {
 		return nil
 	})
 
+	if os.IsNotExist(err) {
```
peusebiu, line 113: I saw some errors about this one in our tests, if rootDir/_sessions is empty then generator should finished for the moment to not keep the workers busy.

rchincha (approved):
lgtm

PR #2183, opened 2024-01-25
feat: add changing user password route for htpasswd

onidoru:
**What type of PR is this?**
feature

It adds a feature of changing user password for htpasswd-file authentication.

Issue: https://github.com/project-zot/zot/issues/702

Unit testing.

No.

No.


pkg/api/routes.go
```diff
@@ -2204,6 +2211,80 @@ func (rh *RouteHandler) RevokeAPIKey(resp http.ResponseWriter, req *http.Request
...
+	// leave OPTIONS method
+	if req.Method == http.MethodOptions {
+		return
+	}
+
+	body, err := io.ReadAll(req.Body)
+	if err != nil {
+		rh.c.Log.Error().Msg("failed to read req body")
+		resp.WriteHeader(http.StatusInternalServerError)
+		_, _ = resp.Write([]byte("internal server error"))
```
andaaron, line 2237: Generally in case of zot we've just returned error code 500 without any body.
If we want to return a body, I think we should be consistent with the rest of the APIs and return it as a json.

Applies to all the errors returned.

pkg/api/routes.go
```diff
@@ -2204,6 +2211,80 @@ func (rh *RouteHandler) RevokeAPIKey(resp http.ResponseWriter, req *http.Request
...
+
+		return
+	}
+
+	userAc, err := reqCtx.UserAcFromContext(req.Context())
+	if err != nil {
+		return
+	}
+
+	username := userAc.GetUsername()
```
andaaron, line 2256: What happens if the call does not contain the information needed to return the user?
For example I allow both local user access and anonymous access on the same server, and someone make a call to this API without providing a username.

pkg/api/htpasswd_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+			gotPasshprase, ok := client.Get(username)
+			So(ok, ShouldBeTrue)
+			So(bcrypt.CompareHashAndPassword([]byte(gotPasshprase), []byte(password)), ShouldBeNil)
+		})
+	})
+}
+
+//
+// func Test_credMap_Get(t *testing.T) {
```
andaaron, line 151: Either uncomment or remove these lines.

pkg/api/controller_test.go
```diff
@@ -4446,6 +4446,60 @@ func TestAuthorization(t *testing.T) {
 	})
 }
 
+func TestChangePassword(t *testing.T) {
```
andaaron, line 4449: Are you testing some new logic here? It doesn't seem to be a new test case.

laurentiuNiculae (approved):
lgtm

pkg/api/htpasswd_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+		Convey("change with empty new oldPassword", func() {
+			err := client.ChangePassword(username, oldPassword, "")
+			So(err, ShouldEqual, zerr.ErrPasswordIsEmpty)
+
+			passphrase, ok := client.Get(username)
+			So(ok, ShouldBeTrue)
+			So(bcrypt.CompareHashAndPassword([]byte(passphrase), []byte(oldPassword)), ShouldBeNil)
+		})
+
```
andaaron, line 50: Maybe test with empty user. Although we should validate the user in routes.go before reaching this logic.

pkg/api/htpasswd.go
```diff
@@ -0,0 +1,184 @@
...
+	// write new content to file
+	output := strings.Join(lines, "\n")
+
+	err = os.WriteFile(hc.filepath, []byte(output), constants.DefaultDirPerms)
+	if err != nil {
+		return fmt.Errorf("error occurred while writing to creds-file: %w", err)
+	}
+
+	// set to credMap only if all file operations are successful to prevent collisions
+	hc.credMap.rw.Lock()
```
andaaron, line 163: Why not?
```suggestion
	hc.credMap.Set(login, string(newPassphrase))
```

pkg/api/htpasswd.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// ChangePassword changes the user password.
+// It accepts user login, his supposed old password for verification and new password.
+func (hc *HtpasswdClient) ChangePassword(login, supposedOldPassword, newPassword string) error {
+	if len(newPassword) == 0 {
+		return zerr.ErrPasswordIsEmpty
+	}
+
+	hc.credMap.rw.RLock()
```
andaaron, line 113: Why not?
```suggestion
	hc.credMap.Get(login)
```

pkg/api/htpasswd.go
onidoru: In this case it'll make the hash from the `newPassphrase`, which is already the hash from `newPassword`.  We could replace the code to ` hc.credMap.Set(login, newPassword)`, so using it vs doing direct map update is the matter of choice.

pkg/api/htpasswd.go
```diff
@@ -0,0 +1,184 @@
...
+
+				break
+			}
+		}
+	}
+
+	// write new content to file
+	output := strings.Join(lines, "\n")
+
+	err = os.WriteFile(hc.filepath, []byte(output), constants.DefaultDirPerms)
```
peusebiu, line 157: I suggest that you first write to a temp file then just rename it so that the operation would be more atomic.

os.Rename is atomic

pkg/api/routes.go
```diff
@@ -2204,6 +2211,79 @@ func (rh *RouteHandler) RevokeAPIKey(resp http.ResponseWriter, req *http.Request
...
+	if err := json.Unmarshal(body, &reqBody); err != nil {
+		rh.c.Log.Error().Msg("failed to unmarshal req body")
+		resp.WriteHeader(http.StatusBadRequest)
+
+		return
+	}
+
+	userAc, err := reqCtx.UserAcFromContext(req.Context())
+	if err != nil {
+		resp.WriteHeader(http.StatusNotFound)
```
peusebiu, line 2251: This should never happen, this should be StatusInternalServerError.

pkg/api/routes.go
```diff
@@ -2204,6 +2211,79 @@ func (rh *RouteHandler) RevokeAPIKey(resp http.ResponseWriter, req *http.Request
...
+
+		return
+	}
+
+	username := userAc.GetUsername()
+	if username == "" {
+		resp.WriteHeader(http.StatusNotFound)
```
peusebiu, line 2258: username can not be empty in any situation, right?
if this route was hit, then user is logged in, then we should have a username, otherwise this is an InternalServerError

andaaron: This is related to my previous comment: https://github.com/project-zot/zot/pull/2183/files#r1466598996

pkg/api/routes.go
```diff
@@ -2204,6 +2211,79 @@ func (rh *RouteHandler) RevokeAPIKey(resp http.ResponseWriter, req *http.Request
...
+			status = http.StatusBadRequest
+		}
+
+		resp.WriteHeader(status)
+
+		return
+	}
+
+	resp.WriteHeader(http.StatusOK)
+}
```
peusebiu, line 2280: we should also clear user's sessions.

peusebiu: get the session from the request.

set it's max age to 0 or -1 session.Options.MaxAge = -1
save it: session.Save()

you can look in authn.go to check how sessions are handled.
We want to clear it so that the user get's logout after changing he's password.

pkg/api/routes.go
```diff
@@ -2204,6 +2211,79 @@ func (rh *RouteHandler) RevokeAPIKey(resp http.ResponseWriter, req *http.Request
...
+// @Accept  json
+// @Produce json
+// @Param old_password body string true "Old password"
+// @Param new_password body string true "New password"
+// @Success 200 {string} string "password changed"
+// @Failure 500 {string} string "internal server error"
+// @Failure 401 {string} string "unauthorized"
+// @Failure 400 {string} string "bad request"
+// @Failure 403 {string} string "old password is incorrect".
+func (rh *RouteHandler) ChangePassword(resp http.ResponseWriter, req *http.Request) {
```
peusebiu, line 2227: Think about how this htpasswd interferes with LDAP.

pkg/api/htpasswd.go
```diff
@@ -0,0 +1,203 @@
...
+	if err := tmpfile.Close(); err != nil {
+		os.Remove(tmpfile.Name())
+		return fmt.Errorf("error occurred when closing temp htpasswd file: %w", err)
+	}
+
+	if err := os.Rename(tmpfile.Name(), hc.filepath); err != nil {
+		return fmt.Errorf("error occurred while replacing htpasswd file with new file: %w", err)
+	}
+
+	err = os.WriteFile(hc.filepath, output, constants.DefaultDirPerms)
```
peusebiu, line 176: this should not be needed, right, or am I missing something?

you first write in tmpfile then you rename it to hc.filepath, that should be enough.

pkg/api/htpasswd.go
```diff
@@ -0,0 +1,203 @@
...
+		return nil
+	}
+
+	// encrypt new password
+	newPassphrase, err := bcrypt.GenerateFromPassword([]byte(newPassword), bcrypt.DefaultCost)
+	if err != nil {
+		return fmt.Errorf("error occurred while encrypting new password: %w", err)
+	}
+
+	file, err := os.ReadFile(hc.filepath)
```
andaaron, line 135: We need a lock to block other users changing their passwords from reading/writing the file at the same time, and only one of the password changes ending up in the file.

pkg/api/constants/consts.go
```diff
@@ -19,6 +19,7 @@ const (
 	LoginPath                    = AppNamespacePath + "/auth/login"
 	LogoutPath                   = AppNamespacePath + "/auth/logout"
 	APIKeyPath                   = AppNamespacePath + "/auth/apikey"
+	ChangePasswordPath           = AppNamespacePath + "/auth/change_password"
```
rchincha, line 22: "chpasswd"

rchincha:
Cases to take care of:
1. Two competing API calls - set_admin_passwd=A, set_admin_passwd=B
2. Competing calls between API and direct file modification

PR #2185, opened 2024-01-26
feat(gql): add new query for diff of cves for 2 images

laurentiuNiculae:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/search/resolver.go
+++ b/pkg/extensions/search/resolver.go
@@ -277,6 +277,263 @@ func getCVEListForImage(
 	}, nil
 }
 
+func getCVEDiffListForImages(
+	ctx context.Context, //nolint:unparam // may be used in the future to filter by permissions
+	minuend gql_generated.ImageInput,
+	subtrahend gql_generated.ImageInput,
+	metaDB mTypes.MetaDB,
+	cveInfo cveinfo.CveInfo,
+	requestedPage *gql_generated.PageInput,
+	searchedCVE string,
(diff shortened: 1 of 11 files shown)


pkg/extensions/search/resolver.go
```diff
@@ -277,6 +277,135 @@ func getCVEListForImage(
...
+	ctx context.Context, //nolint:unparam // may be used in the future to filter by permissions
+	minuend string,
+	subtrahend string,
+	metaDB mTypes.MetaDB,
+	cveInfo cveinfo.CveInfo,
+	requestedPage *gql_generated.PageInput,
+	searchedCVE string,
+	log log.Logger, //nolint:unparam // may be used by devs for debugging
+) (*gql_generated.CVEDiffResult, error) {
+	minuendRepo, minuendTag := zcommon.GetImageDirAndTag(minuend)
```
andaaron, line 290: I think we should account for the os/architecture. Maybe get the manifest digest as part of the minuend, and take the architecture of that manifest into consideration. Or have it as a parameter.

Using minuendSummary.Manifests[0] seems very shaky, and will not work for multiarch images in case a vulnerability is for a specific architecture.

rchincha:
@laurentiuNiculae in terms of priorities, make sure cli works first and then zui

andaaron:
Superseded by https://github.com/project-zot/zot/pull/2242

Issue #2182, opened 2024-01-24
[Feat]: add support for systemd socket activation


rgl:
### Is your feature request related to a problem? Please describe.

I would like to configure zot to use a privileged port like 80 or 443 without granting it the `CAP_NET_BIND_SERVICE` capability.

### Describe the solution you'd like

Would like to configure Zot to use [systemd Socket Activation](https://0pointer.de/blog/projects/socket-activated-containers.html).

For more information, please see:

* https://0pointer.de/blog/projects/socket-activated-containers.html
* https://www.sobyte.net/post/2022-01/systemd-socket-activation/

rchincha:
@rgl thanks for filing this issue.
Do you want to give a shot at a PR?
https://github.com/project-zot/zot/blob/main/examples/zot.service

rgl:
I can!

Just so we are on the same page, this means that I will modify the following code, OK?

https://github.com/project-zot/zot/blob/v2.0.0/pkg/api/controller.go#L145-L168

rchincha:
@rgl would just add another zot.socket-service file

rgl:
@rchincha the PR is at https://github.com/project-zot/zot/pull/2186.

please be aware that I've updated the go.mod too, which I'm unsure if should be included in the PR. please advice in the PR issue.

gira0:
Still very interested in seeing systemd socket activation get added. Is #4009 going to be revived, is there a way I can help?

andaaron:
> Still very interested in seeing systemd socket activation get added. Is #4009 going to be revived, is there a way I can help?

Hi, sure, please pick it up.

PR #2049, opened 2023-11-15
build(go): switch to go 1.21

andaaron:
Also update to the latest swaggo version, as the previous one did not work with go 1.21

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2186, opened 2024-01-26
feat: add support for systemd socket activation

rgl:
**What type of PR is this?**
feature

https://github.com/project-zot/zot/issues/2182

**WARNING I'm unsure about the go.mod version changes. Should they be done? If so, should CI be upgraded too?**

It allows zot to use [systemd socket activation](https://0pointer.de/blog/projects/socket-activated-containers.html).

This was implemented using the [coreos/go-systemd library](https://github.com/coreos/go-systemd/blob/main/examples/activation/httpserver/httpserver.go) (see the [httpserver example](https://github.com/coreos/go-systemd/blob/main/examples/activation/httpserver/httpserver.go)).

With this change, we configure systemd to use a privileged port like 80 or 443 with the zot.socket file. With this, we do not need to grant zot the CAP_NET_BIND_SERVICE capability. systemd will bind the socket and will pass it to zot as a file descriptor.

This can be tested using:

```bash
systemd-socket-activate --listen=127.0.0.1:9999 ./bin/zot-linux-amd64 serve examples/config-minimal.json
```

The example systemd zot.service file was modified to depend on the zot.socket file too.

This can be installed using:

```bash
install examples/zot.service /etc/systemd/system/zot.service
install examples/zot.socket /etc/systemd/system/zot.socket
systemctl daemon-reload
systemctl enable zot.service zot.socket
systemctl restart zot.service zot.socket
```

**Automation added to e2e**:

No.

Yes and No. If you want to use this feature, the zot listening address will no longer be configured from the zot configuration file, instead it will be configured from the zot.socket file.

```release-note
add support for systemd socket activation. please see the example zot.socket and zot.service files.
```


go.mod
```diff
@@ -1,6 +1,8 @@
 module zotregistry.io/zot
 
-go 1.20
+go 1.21
```
andaaron, line 3: Please leave the go 1.21 update for a separate PR. There are multiple changes to be made for this, which I have in: https://github.com/project-zot/zot/pull/2049

examples/zot.service
andaaron, line 1: Can we have two separate examples? One with systemd socket activation and one without?

andaaron (commented):
Thank you for the PR!

examples/zot.service
rchincha: @rgl yes, would leave the original example alone. Not everyone may want this approach.

go.mod
rgl: I've removed the go.* file changes. But, I'm not able to build this with go 1.21, as it refuses to upgrade the dependencies due to:

```console
$ go mod tidy -go=1.20 
go: github.com/letsencrypt/boulder@v0.0.0-20231026200631-000cd05d5491 requires go@1.21, but 1.20 is requested
```

examples/zot.service
rgl: I've moved the example to the system-socket-activation sub-directory without touchpoing the existing zot.service file.

go.mod
rgl: BTW, I can wait for https://github.com/project-zot/zot/pull/2049 to land, and then rebase on master then.

rchincha:
@rgl sorry, could you rebase and push pls.

rgl:
@rchincha I can only build this branch with go 1.21.6. as such, this is waiting for https://github.com/project-zot/zot/pull/2049 to land in the main branch; when that happens, I'll rebase on the main branch.

rchincha:
@rgl do you rebase and try again?

rgl:
@rchincha sure. I've rebased it. please note that

rchincha:
Let's target this PR for v2.1.0

PR #2187, opened 2024-01-26
refactor: rename go module from zotregistry.io/zot to zotregistry.dev/zot

andaaron:


rchincha:
There are no projects currently doing an direct import of this. So appears to be safe to do so. Let's wait a little bit before we merge.

rchincha (approved):
lgtm

This is a very large patch. Did a careful review.

PR #2215, opened 2024-01-31
fix: Fix example configs and README.md file

onidoru:
**What type of PR is this?**
documentation

Example LDAP configs do not represent the actual state of things and are outdated. 

```bash
# could also run $ make verify-config
# it's also the same problem with `examples/config-ldap-credentials.json`, so it's excluded from the Makefile
$ ./bin/zot-linux-amd64 verify examples/config-example.json 
{"level":"error","error":"1 error(s) decoding:\n\n* 'HTTP.Auth.LDAP' has invalid keys: binddn, bindpassword","time":"2024-01-16T23:36:27+02:00","message":"failed to unmarshaling new config"}
```

### DIFF
--- a/.github/workflows/verify-config.yaml
+++ b/.github/workflows/verify-config.yaml
@@ -24,7 +24,7 @@ jobs:
         uses: actions/setup-go@v5
         with:
           cache: false
-          go-version: 1.20.x
+          go-version: 1.21.x
       - name: Cache go dependencies
         id: cache-go-dependencies
         uses: actions/cache@v4
--- a/.github/workflows/verify-config.yaml
+++ b/.github/workflows/verify-config.yaml
@@ -1,0 +1,46 @@
+name: "Verify Example Config Files"
+
+# Validate all example config files are relevant and valid.
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches: [main]
+  release:
+    types:
+      - published
+
+permissions: read-all
+
+jobs:
+  verify-config:
+    name: Verify Config Files
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install go
+        uses: actions/setup-go@v5
+        with:
+          cache: false
+          go-version: 1.20.x
+      - name: Cache go dependencies
+        id: cache-go-dependencies
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-mod-
+      - name: Install go dependencies
+        if: steps.cache-go-dependencies.outputs.cache-hit != 'true'
+        run: |
+          cd $GITHUB_WORKSPACE
+          go mod download
+      - uses: ./.github/actions/setup-localstack
+      - name: run verify-config
+        run: |
+          cd $GITHUB_WORKSPACE
+          make verify-config
--- a/.github/workflows/verify-config.yaml
+++ b/.github/workflows/verify-config.yaml
@@ -39,7 +39,6 @@ jobs:
         run: |
           cd $GITHUB_WORKSPACE
           go mod download
-      - uses: ./.github/actions/setup-localstack
       - name: run verify-config
         run: |
           cd $GITHUB_WORKSPACE
--- a/pkg/cli/server/root.go
+++ b/pkg/cli/server/root.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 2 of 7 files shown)


andaaron (commented):
Change looks good, but we should make sure we also call `make verify-config` in CI.

pkg/cli/server/root.go
```diff
@@ -861,8 +861,8 @@ func readLDAPCredentials(ldapConfigPath string) (config.LDAPCredentials, error)
 
 	var ldapCredentials config.LDAPCredentials
 
-	if err := viperInstance.Unmarshal(&ldapCredentials); err != nil {
-		log.Error().Err(err).Msg("failed to unmarshal new config")
+	if err := viperInstance.UnmarshalExact(&ldapCredentials); err != nil {
+		log.Error().Err(err).Msg("failed to unmarshal ldap credentials config")
```
andaaron, line 865: Please add a test to cover this line. It should be easier than before, since we should fail in case of invalid or missing fields.

.github/workflows/verify-config.yaml
```diff
@@ -0,0 +1,46 @@
...
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-mod-
+      - name: Install go dependencies
+        if: steps.cache-go-dependencies.outputs.cache-hit != 'true'
+        run: |
+          cd $GITHUB_WORKSPACE
+          go mod download
+      - uses: ./.github/actions/setup-localstack
```
andaaron, line 42: This should not be needed, as we don't test the S3 APIs in this workflow.

onidoru: Fixed.

pkg/cli/server/root.go
onidoru: Added new tests to ensure that credentials file contains all required fields or has any extra fields.

Please note that it does not check if [credentialsFile](https://github.com/project-zot/zot/pull/2215/files#diff-464834167072cf556f61bd7ba8a129ccde65577dc94ac2b0ce0ddecfb21d864cR21) field is present in `auth` config. Let me know if we want to change this old behavior and make it required.

pkg/cli/server/root.go
```diff
@@ -861,12 +861,34 @@ func readLDAPCredentials(ldapConfigPath string) (config.LDAPCredentials, error)
...
-		log.Error().Err(err).Msg("failed to unmarshal new config")
+	metaData := &mapstructure.Metadata{}
+	if err := viperInstance.UnmarshalExact(&ldapCredentials, metadataConfig(metaData)); err != nil {
+		log.Error().Err(err).Msg("failed to unmarshal ldap credentials config")
 
 		return config.LDAPCredentials{}, err
 	}
 
+	if len(metaData.Keys) == 0 {
+		log.Error().Err(zerr.ErrBadConfig).
```
andaaron, line 872: Can you cover this case? Should be easy enough.

pkg/cli/server/root.go
```diff
@@ -861,12 +861,34 @@ func readLDAPCredentials(ldapConfigPath string) (config.LDAPCredentials, error)
...
+			Msg("failed to load ldap credentials config due to the absence of any key:value pair")
+
+		return config.LDAPCredentials{}, zerr.ErrBadConfig
+	}
+
+	if len(metaData.Unused) > 0 {
```
andaaron, line 878: Can you cover this with a test case?

pkg/cli/server/root.go
onidoru: Covered.

pkg/cli/server/root.go
onidoru: Covered.

pkg/cli/server/root_test.go
```diff
@@ -1236,6 +1236,161 @@ storage:
...
+
+		tmpCredsFile, err := os.CreateTemp("", "zot-cred*.json")
+		So(err, ShouldBeNil)
+		defer os.Remove(tmpCredsFile.Name())
+
+		// `bindDN` key is missing
+		content := []byte(`{
+    		"bindDN":"cn=ldap-searcher,ou=Users,dc=example,dc=org",
+    		"bindPassword":"ldap-searcher-password",
+			"extraKey": "extraValue"
```
andaaron, line 1330: For some reason the test does not cover the lines we expected. Also the comment at line 1326 should probably be removed,

pkg/cli/server/root_test.go
```diff
@@ -1236,6 +1236,161 @@ storage:
...
+		tmpFile, err := os.CreateTemp("", "zot-test*.json")
+		So(err, ShouldBeNil)
+		defer os.Remove(tmpFile.Name())
+
+		tmpCredsFile, err := os.CreateTemp("", "zot-cred*.json")
+		So(err, ShouldBeNil)
+		defer os.Remove(tmpCredsFile.Name())
+
+		// `bindDN` key is missing
+		content := []byte(``)
```
andaaron, line 1368: I think this case covers the error on `viperInstance.UnmarshalExact(` which is good, but we still need a separate test case with a valid empty json `content := []byte(`{}`)`.

rchincha:
Superseded by https://github.com/project-zot/zot/pull/2322
Thanks @onidoru for the PR, closing.

PR #2220, opened 2024-02-01
fix(scheduler): the session cleanup generator is reset too often

andaaron:
This causes the "fair" scheduler to run it too often in the detriment of other generators. The intention was to run it every 2 hours but the measurement unit for 7200 was not specified.

Add more logs, including showing a generator name, in order to troubleshoot this kind of issues easier in the future.


peusebiu (approved):
lgtm

rchincha (approved):
lgtm

we should have caught this! :(

Issue #2213, opened 2024-01-30
[Bug]: with "sync" extension and when used as a cache, `zot` doesn't return a cached image when the upstream registry is down


piontec:
### zot version

v2.0.1

### To reproduce

zot's `config.json` 

```json
"extensions": {
    "sync": {
      "credentialsFile": "cred.json",
      "registries": [
        {
          "urls": ["https://gsoci.azurecr.io"],
          "onDemand": true,
          "tlsVerify": true,
          "maxRetries": 3,
          "retryDelay": "5m"
        }
      ]
    },
}
```

## Steps to reproduce:

Using `containerd` and `crictl` as clients.

1. Delete and fetch

```
# crictl rmi gsoci.azurecr.io/giantswarm/alpine;  crictl pull gsoci.azurecr.io/giantswarm/alpine:3.16.9
ERRO[0000] no such image gsoci.azurecr.io/giantswarm/alpine 
FATA[0000] unable to remove the image(s)                
Image is up to date for sha256:f535e67342d78f4f32f72d77d77046ee4cd635f2fcb0134062e55deeb265189a
```

2. Block outgoing network traffic
3. Fetch again - it fails

command
```
# crictl rmi gsoci.azurecr.io/giantswarm/alpine;  crictl pull gsoci.azurecr.io/giantswarm/alpine:3.16.9
ERRO[0000] no such image gsoci.azurecr.io/giantswarm/alpine 
FATA[0000] unable to remove the image(s)                
E0130 15:00:11.729433    4176 remote_image.go:167] "PullImage from image service failed" err="rpc error: code = Canceled desc = failed to pull and unpack image \"gsoci.azurecr.io/giantswarm/alpine:3.16.9\": failed to resolve reference \"gsoci.azurecr.io/giantswarm/alpine:3.16.9\": failed to do request: Head \"http://172.18.0.1:5000/v2/giantswarm/alpine/manifests/3.16.9?ns=gsoci.azurecr.io\": context canceled" image="gsoci.azurecr.io/giantswarm/alpine:3.16.9"
FATA[0060] pulling image: rpc error: code = Canceled desc = failed to pull and unpack image "gsoci.azurecr.io/giantswarm/alpine:3.16.9": failed to resolve reference "gsoci.azurecr.io/giantswarm/alpine:3.16.9": failed to do request: Head "http://172.18.0.1:5000/v2/giantswarm/alpine/manifests/3.16.9?ns=gsoci.azurecr.io": context canceled 
```

[...]

Issue #2212, opened 2024-01-30
[Bug]: Sync doesn't handle _catalog authentication with bearer


peusebiu:
### zot version

v2.0.1

### Describe the bug

Currently most authentication is handled by the lib sync is using to copy images, but the /v2/_catalog is handled in our code base and it only supports basic auth, this should be fixed by implementing a bearer client for the _catalog routes, or by using the lib directly.

Originally posted on Slack by Łukasz Piątkowski

### To reproduce

 Use any OCI compliant remote registry(that have bearer auth) in the sync config, enable periodically sync.

```
"extensions": {
    "sync": {
      "credentialsFile": "./cred.json",
      "registries": [
        {
          "urls": ["https://xyz.azurecr.io/"],
          "onDemand": false,
          "tlsVerify": true,
          "maxRetries": 3,
          "retryDelay": "1m",
          "pollInterval": "5m",

```
Logs:
```
{
  "level": "error",
  "errorType": "*errors.errorString",
  "remote registry": "https://xyz.azurecr.io/",
  "error": "{\"errors\":[{\"code\":\"UNAUTHORIZED\",\"message\":\"authentication required, visit https://aka.ms/acr/authorization for more information.\",\"detail\":[{\"Type\":\"registry\",\"Name\":\"catalog\",\"Action\":\"*\"}]}]}\n",
  "goroutine": 56,
  "caller": "zotregistry.io/zot/pkg/extensions/sync/service.go:206",
  "time": "2024-01-30T14:42:25.285957831+01:00",
  "message": "failed to get repository list from remote registry"
}
{
  "level": "error",
  "component": "scheduler",
  "error": "{\"errors\":[{\"code\":\"UNAUTHORIZED\",\"message\":\"authentication required, visit https://aka.ms/acr/authorization for more information.\",\"detail\":[{\"Type\":\"registry\",\"Name\":\"catalog\",\"Action\":\"*\"}]}]}\n",
  "goroutine": 56,
  "caller": "zotregistry.io/zot/pkg/scheduler/scheduler.go:462",
  "time": "2024-01-30T14:42:25.286011191+01:00",
  "message": "failed to execute generator"
}
```

### Expected behavior

periodically sync should work, currently just onDemand works.

andaaron:
@peusebiu shall we close this one as fixed in https://github.com/project-zot/zot/pull/2222?

peusebiu:
yes, closing it, fixed in: https://github.com/project-zot/zot/pull/2222

PR #2222, opened 2024-02-01
fix(sync): added bearer client for sync

peusebiu:
fixed ping function taking too much time

closes: #2213 #2212

**What type of PR is this?**
bug

Fixes the sync taking too much time to fallback to serving local cached image, in case the remote is unavailable.
Added bearer auth for http client.
Should also fix multiple unauthorized messages we saw in sync logs when trying to sync referrers.

This is WIP

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -482,7 +482,7 @@ func bearerAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 					if err != nil {
 						ctlr.Log.Error().Err(err).Msg("failed to parse Authorization header")
 						response.Header().Set("Content-Type", "application/json")
-						zcommon.WriteJSON(response, http.StatusInternalServerError, apiErr.NewError(apiErr.UNSUPPORTED))
+						zcommon.WriteJSON(response, http.StatusUnauthorized, apiErr.NewError(apiErr.UNSUPPORTED))
 
 						return
 					}
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -240,7 +240,7 @@ func (rh *RouteHandler) CheckVersionSupport(response http.ResponseWriter, reques
 	response.Header().Set(constants.DistAPIVersion, "registry/2.0")
 	// NOTE: compatibility workaround - return this header in "allowed-read" mode to allow for clients to
 	// work correctly
-	if rh.c.Config.HTTP.Auth != nil {
+	if rh.c.Config.IsBasicAuthnEnabled() || rh.c.Config.IsBearerAuthEnabled() {
 		// don't send auth headers if request is coming from UI
 		if request.Header.Get(constants.SessionClientHeaderName) != constants.SessionClientHeaderValue {
 			if rh.c.Config.HTTP.Auth.Bearer != nil {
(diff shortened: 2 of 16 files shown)


pkg/common/common.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	httpTimeout        = 5 * time.Minute
+	httpTimeout        = 30 * time.Second
```
rchincha, line 18: Does it make sense to relate this to the maxRetries and retryDelay?

pkg/api/authn.go
```diff
@@ -482,7 +482,7 @@ func bearerAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 					if err != nil {
 						ctlr.Log.Error().Err(err).Msg("failed to parse Authorization header")
 						response.Header().Set("Content-Type", "application/json")
-						zcommon.WriteJSON(response, http.StatusInternalServerError, apiErr.NewError(apiErr.UNSUPPORTED))
+						zcommon.WriteJSON(response, http.StatusUnauthorized, apiErr.NewError(apiErr.UNSUPPORTED))
```
peusebiu, line 485: I think this is more correct and inline with the spec.

pkg/common/common.go
peusebiu: Yeah, need to read more about this, looking on the comment, it seems that it can interrupt the body read if the timeout is reached, and in the case of huge blobs, I need to read and test it more in depth

pkg/api/routes.go
```diff
@@ -240,7 +240,7 @@ func (rh *RouteHandler) CheckVersionSupport(response http.ResponseWriter, reques
 	response.Header().Set(constants.DistAPIVersion, "registry/2.0")
 	// NOTE: compatibility workaround - return this header in "allowed-read" mode to allow for clients to
 	// work correctly
-	if rh.c.Config.HTTP.Auth != nil {
```
peusebiu, line 243: this was always evaluated to true(because httpassword.path which is empty string), sending an auth header with the next value:
"basic realm="

pkg/extensions/sync/sync.go
```diff
@@ -63,6 +63,9 @@ type Remote interface {
 	GetRepoTags(repo string) ([]string, error)
 	// Get manifest content, mediaType, digest given an ImageReference
 	GetManifestContent(imageReference types.ImageReference) ([]byte, string, digest.Digest, error)
+	// In the case of public dockerhub images 'library' namespace is added to the repo names of images
+	// eg: alpine -> library/alpine
```
andaaron, line 67: Maybe the commit message should indicate this changeset is breaking an existing behavior, and people should update their configs?
Do we break it? Is it possible not to break it?

peusebiu: We don't break it, it will still work as it is today, the only difference is that the bearer client would not work because docker.io/library is not the URL for docker APIs, it's a redirect to hub.docker.com.

So it will not send any WWW-Authenticate header for the client to use.

It will still sync images, but all calls to referrers (cosign, notation etc) will get html responses into the body, and it won't unmarshal the reponses..

andaaron: OK, makes sense. Maybe we should advise users on what endpoint to use. In one of the docs.

andaaron (approved):

pkg/api/authn.go
rchincha: ok

pkg/api/routes.go
rchincha: :(

rchincha (approved):
lgtm

PR #2236, opened 2024-02-08
test(blackbox): fix the scrub test sometimes deleting the image manifest from the layout

andaaron:
If the manifest is not present, scrub no longer errors, so the test looking for errors in the log was not failing.

See the related scrub changes in: https://github.com/project-zot/zot/pull/2180


peusebiu (approved):
lgtm

PR #2253, opened 2024-02-17
fix(npe): handle case where os.Stat returns different error types in DirExists

andaaron:
Also add tests to fix some of the coverage fluctuations.


peusebiu (approved):
lgtm

PR #2255, opened 2024-02-18
chore: update image-spec and dist spec to 1.1.0

andaaron:
BREAKING CHANGE: the dist spec version in the config files needs to be bumped to 1.1.0
in order for the config verification to pass without warnings.

Also fix 1 dependabot alert for helm.


rchincha (approved):
lgtm

Issue #2238, opened 2024-02-08
[Bug]: Repeated requests to remote registry when not able to load catalog


ctr49:
### zot version

v2.0.2-rc1

### Describe the bug

There seems to be a problem with handling errors when a remote registry does not respond "as expctected".
Configuring a registry that throws a 400 on registry/v2/_catalog (not OCI compliant?!) results in **repeated** requests towards that registry, regardless of retry settings.

### To reproduce

1. Configuration
Configuring the Chainguard registry in pass-through mode (on demand) in sync extension:
``` json
{
    "storage": {
        "rootDirectory": "/var/lib/registry",
        "commit": true
    },
    "http": {
        "address": "0.0.0.0",
        "port": "5000"
    },
    "log": {
        "level": "warn"
    },
    "extensions": {
        "metrics": {},
        "sync": {
            "enable": true,
            "credentialsFile": "./sync-auth-credentials.json",
            "registries": [
                {
                    "urls": [
                        "https://cgr.dev"
                    ],
                    "content": [
                        {
                            "prefix": "**",
                            "destination": "/chainguard-images"
                        }
                    ],
                    "onDemand": true,
                    "maxRetries": 3,
                    "retryDelay": "60m",
                    "pollInterval": "6h",
                    "tlsVerify": true
                }
            ]
        },
        "search": {},
        "scrub": {},
        "lint": {},
        "trust": {},
        "ui": {}
    }
}
```
2. Client tool used
Pre-compiled zot-linux-arm64 (using the provided helm Chart with container from the project's repository)
3. Seen error
Seeing about **10 per second** of those, probably also stressing the remote end:
``` json
{
	"level": "error",
	"component": "scheduler",
	"error": "{\"errors\":[{\"code\":\"NOT_IMPLEMENTED\",\"message\":\"The catalog API is not yet supported.\"}]}\n",
	"generator": "SyncGenerator",
	"goroutine": 32,
	"caller": "zotregistry.dev/zot/pkg/scheduler/scheduler.go:465",
	"time": "2024-02-08T17:22:58.965145576Z",
	"message": "failed to execute generator"
}
```

### Expected behavior

I'd expect zot to honor the 400 message from the upstream registry and not hammering it any further.

### Additional context

Not sure if the Chainguard registry is also at fault (but to me, throwing a 400 does not seem entirely wrong), but in any case the error handling on the zot side should prevent a request storm that could easily result in a DoS (on either side).

peusebiu:
[...]

Issue #1801, opened 2023-09-12
[Refactor] Sync should check features supported by upstream registry before syncing references


andaaron:
At the moment the sync logic tries to pull references without knowing if the upstream registry supports the referrers API or the oras API.
This results in a lot of calls returning responses with bad status codes, which could be avoided (for various reasons, such as spamming the log messages, hitting rate limits, slowing down zot, and so on).

At a bare minimum we can try making the referrers/oras API call once per registry (or repo?).

rchincha:
Yes, this one is important.

rchincha:
Can we close this?

andaaron:
Let's close it.

PR #2272, opened 2024-02-25
fix(sync): sync generator now backs off on errors

peusebiu:
closes: #2238

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


pkg/extensions/sync/sync.go
```diff
@@ -77,85 +87,148 @@ type Destination interface {
...
-	return fmt.Sprintf("{Name: \"%s\", repo: \"%s\"}",
+	return fmt.Sprintf("{Name: \"%s\", repository: \"%s\"}",
 		srt.Name(), srt.repo)
 }
 
 func (srt *syncRepoTask) Name() string {
 	return "SyncTask"
 }
+
+func onGeneratorError(gen *SyncRepoTaskGenerator) {
```
andaaron, line 207: Why not have this as a generator method?
Same applies for the other methods below.
Modifying the internal state of the generator object using methods outside of the struct is a bad practice.

andaaron (commented):
The generator backs off, but that should be the case only for the registry where the error has occurred. 

I was thinking the solution would also solve https://github.com/project-zot/zot/issues/1801. 
I think the implementation in the PR could make it worse in that case, since zot would continue to send "bad" requests to badly configured registries, and it would take longer to finish syncing registries with a good configuration.
Or am I missing something?

Why not just skip that repository and try to sync again when the generator is reset?

peusebiu:
> The generator backs off, but that should be the case only for the registry where the error has occurred.
> 
That is true, there is one generator per remote registry

> I was thinking the solution would also solve #1801. I think the implementation in the PR could make it worse in that case, since zot would continue to send "bad" requests to badly configured registries, and it would take longer to finish syncing registries with a good configuration. Why not simply stop sending the bad requests instead of keep sending them but with a backoff interval? Or am I missing something?
> 
I added some code for this, please review it again. Let me know if you think it's ok, for referrers api things are clear, but for the rest of them...

> Why not just skip that repository and try to sync again when the generator is reset?
agree

peusebiu:
> The generator backs off, but that should be the case only for the registry where the error has occurred.
> 
> I was thinking the solution would also solve #1801. I think the implementation in the PR could make it worse in that case, since zot would continue to send "bad" requests to badly configured registries, and it would take longer to finish syncing registries with a good configuration. Why not simply stop sending the bad requests instead of keep sending them but with a backoff interval? Or am I missing something?
> 
> Why not just skip that repository and try to sync again when the generator is reset?

pkg/extensions/sync/references/references.go
```diff
@@ -78,14 +83,32 @@ func (refs References) syncAll(ctx context.Context, localRepo, upstreamRepo,
...
+		supported, ok := refs.features.Get(ref.Name(), upstreamRepo)
+		if !supported && ok {
+			continue
+		}
+
 		syncedRefsDigests, err = ref.SyncReferences(ctx, localRepo, upstreamRepo, subjectDigestStr)
 		if err != nil {
+			// in the case of oci referrers, it will return 404 only if the repo is not found or refferers API is not supported
+			// no need to continue to make requests to the same repo
+			if ref.Name() == constants.OCI && errors.Is(err, zerr.ErrSyncReferrerNotFound) {
```
andaaron, line 95: There's no need to check zerr.ErrSyncReferrerNotFound, right? Both this if and the next one seem to do the same thing.

pkg/extensions/sync/references/references.go
```diff
@@ -78,14 +83,32 @@ func (refs References) syncAll(ctx context.Context, localRepo, upstreamRepo,
...
+				refs.features.Set(ref.Name(), upstreamRepo, false)
+			}
+
 			refs.log.Debug().Err(err).
 				Str("reference type", ref.Name()).
 				Str("image", fmt.Sprintf("%s:%s", upstreamRepo, subjectDigestStr)).
 				Msg("couldn't sync image referrer")
 		}
 
+		refs.features.Set(ref.Name(), upstreamRepo, true)
```
andaaron, line 110: Isn't this overwriting the previously set value? in case err is not nil?

pkg/extensions/sync/references/references.go
peusebiu: the next one is !negated. I moved the second block here to make things more clear.

andaaron (approved):

rchincha (approved):
lgtm

These are heuristics, let's see how well it is received.

PR #2273, opened 2024-02-25
fix(sync): better blobs cleanup

peusebiu:
cleanup blob after each sync.copyBlob()

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/sync/destination.go
+++ b/pkg/extensions/sync/destination.go
@@ -186,6 +186,12 @@ func (registry *DestinationRegistry) CommitImage(imageReference types.ImageRefer
 	return nil
 }
 
+func (registry *DestinationRegistry) CleanupImage(imageReference types.ImageReference, repo, reference string) error {
+	tmpDir := getTempRootDirFromImageReference(imageReference, repo, reference)
+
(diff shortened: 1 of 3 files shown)


pkg/extensions/sync/destination.go
```diff
@@ -248,6 +248,8 @@ func (registry *DestinationRegistry) copyManifest(repo string, manifestContent [
 func (registry *DestinationRegistry) copyBlob(repo string, blobDigest digest.Digest, blobMediaType string,
 	tempImageStore storageTypes.ImageStore,
 ) error {
+	defer os.Remove(tempImageStore.BlobPath(repo, blobDigest))
```
andaaron, line 251: This seems to be a side effect of copyBlob which doesn't have anything to do with the actual copy, which may be confusing for developers.

I suggest either a :
- name change
- move the deletion outside the method and into a new method which would be called instead of this one.

andaaron (commented):
I suggest also deleting leftover folders on zot startup, before it makes the REST API available to clients.

andaaron (approved):

rchincha (approved):
lgtm

Issue #2299, opened 2024-03-06
[Bug]: manifest download with digest reference fails with 500 'image meta not found'


akashsinghal:
### zot version

v2.0.2-rc.2

### Describe the bug

I'm trying to download a manifest using a digested reference: `localhost:5000/cosign@sha256:f092da75cf80b71e7244f64b76bcd8f3cfebda0951fd00379bea2c8289379609` but it fails with a 500 error. Looking at the zot logs I can see 

```
{"level":"error","error":"image meta not found","repository":"cosign","reference":"sha256:f092da75cf80b71e7244f64b76bcd8f3cfebda0951fd00379bea2c8289379609","goroutine":195,"caller":"zotregistry.dev/zot/pkg/meta/hooks.go:126","time":"2024-03-06T19:45:32.029070934Z","message":"failed to update stats on download image"}
{"level":"info","module":"http","username":"test_user","component":"session","clientIP":"172.17.0.1:38856","method":"GET","path":"/v2/cosign/manifests/sha256:f092da75cf80b71e7244f64b76bcd8f3cfebda0951fd00379bea2c8289379609","statusCode":500,"latency":"0s","bodySize":0,"headers":{"Accept":["application/vnd.docker.distribution.manifest.v2+json, application/vnd.docker.distribution.manifest.list.v2+json, application/vnd.oci.image.manifest.v1+json, application/vnd.oci.image.index.v1+json, application/vnd.oci.artifact.manifest.v1+json"],"Accept-Encoding":["gzip"],"Authorization":["******"],"User-Agent":["oras/1.1.0"]},"goroutine":195,"caller":"zotregistry.dev/zot/pkg/api/session.go:132","time":"2024-03-06T19:45:32.029117534Z","message":"HTTP API"}
```

I've pushed an image `localhost:5000/cosign:signed-key` and used latest cosign to sign the image. I'm using `oras manifest fetch` to download the manifest of the cosign index.  I can download the manifest if I reference it by tag but not by digest directly.

I've confirmed that the digest of the cosign index is correct:
```
> oras manifest fetch --descriptor localhost:5000/cosign:sha256-4790bc62b2852dd3a5f76e1b1866d3895d1125a487365c67e1ebee86150f6a67.sig
{"mediaType":"application/vnd.oci.image.manifest.v1+json","digest":"sha256:f092da75cf80b71e7244f64b76bcd8f3cfebda0951fd00379bea2c8289379609","size":558}
```
```
[...]

PR #2301, opened 2024-03-07
fix: ignore metadb errors if tag not found

rchincha:
Manifests cat be retrieved by digest but we maintain stats based only on tags. So ignore errors if not found.

Fixes issue #2299

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -514,7 +514,7 @@ func (rh *RouteHandler) GetManifest(response http.ResponseWriter, request *http.
 
 	if rh.c.MetaDB != nil {
 		err := meta.OnGetManifest(name, reference, mediaType, content, rh.c.StoreController, rh.c.MetaDB, rh.c.Log)
-		if err != nil {
+		if err != nil && !errors.Is(err, zerr.ErrImageMetaNotFound) {
 			response.WriteHeader(http.StatusInternalServerError)
 
 			return
--- a/pkg/meta/hooks.go
+++ b/pkg/meta/hooks.go
@@ -100,7 +100,7 @@ func OnDeleteManifest(repo, reference, mediaType string, digest godigest.Digest,
 	return nil
 }
 
-// OnDeleteManifest is called when a manifest is downloaded. It increments the download couter on that manifest.
+// OnGetManifest is called when a manifest is downloaded. It increments the download couter on that manifest.
 func OnGetManifest(name, reference, mediaType string, body []byte,
 	storeController storage.StoreController, metaDB mTypes.MetaDB, log log.Logger,
 ) error {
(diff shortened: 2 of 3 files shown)


pkg/api/routes.go
```diff
@@ -514,7 +514,7 @@ func (rh *RouteHandler) GetManifest(response http.ResponseWriter, request *http.
 
 	if rh.c.MetaDB != nil {
 		err := meta.OnGetManifest(name, reference, mediaType, content, rh.c.StoreController, rh.c.MetaDB, rh.c.Log)
-		if err != nil {
```
andaaron, line 517: This seems too wide of a net for me, and metadb errors with other root causes may go unnoticed (not even logged afaik). Do we want this? On the bright side clients would have a functioning API even in those cases.

As an alternative we could update `storage.CheckIsImageSignature` to also look for annotations. I see cosign adds the "dev.cosignproject.cosign/signature" annotation to all the layers in the manifest. Or later mediaType "application/vnd.dev.cosign.simplesigning.v1+json"

rchincha: We don't want to pollute zot with all unknown types. The case is we found the manifest (correctly) from the local storage, we didn't find it in the metadb since we didn't put it there, so do you want to fail the request for the specific error of "not found"?

andaaron: This is a cosign specific signature miss-identification, I think we should fix the cosign specify logic it there in the cosign specific logic we already have.

We could also keep the generic error handling as in this PR if returning the content is more important compared to having accurate download counters.

So, is the following scenario working correctly?
1. Push container image by tag (regular image, no special artifact cases).
2. Pull container image by digest.

rchincha:
```
 skopeo copy --format=oci --preserve-digests --dest-tls-verify=false --dest-creds=test:test docker://ghcr.io/project-zot/zot-linux-amd64@sha256:c4d8e94e78e340ac7a57d80becca26bce4d2fbd1233cb6f9280e02d100820a17 docker://localhost:5000/ghcr.io/project-zot/zot-linux-amd64@sha256:c4d8e94e78e340ac7a57d80becca26bce4d2fbd1233cb6f9280e02d100820a17
 
 oras manifest fetch localhost:5000/ghcr.io/project-zot/zot-linux-amd64@sha256:c4d8e94e78e340ac7a57d80becca26bce4d2fbd1233cb6f9280e02d100820a17
 ```
 
 ^ appears to work correctly.

rchincha: Per discussion best to take the approach in this PR.

rchincha: Added a test case.

andaaron (approved):

PR #2330, opened 2024-03-19
chore: update to go 1.22

andaaron:
1.
Only go toolchain version is updated.
We compile with go 1.22, but we allow others to compile using language version 1.21 if they wish to. If we also updated the go version in go.mod everyone would be forced to update, as that is enforced as a minimum allowed version.

This comment explains the difference well enough https://news.ycombinator.com/item?id=36455759

2.
 It looks like they made some cleanup in the logic allowing buildmode pie on various platforms.

Related to https://github.com/golang/go/issues/31544
See the code at: https://cs.opensource.google/go/go/+/master:src/internal/platform/supported.go;l=222-231;drc=d7fcb5cf80953f1d63246f1ae9defa60c5ce2d76;bpv=1;bpt=0

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -18,7 +18,7 @@ jobs:
       - uses: actions/setup-go@v5
         with:
           cache: false
-          go-version: 1.21.x
+          go-version: 1.22.x
       # Run benchmark with `go test -bench` and stores the output to a file
       - name: Run benchmark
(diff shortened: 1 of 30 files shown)


rchincha (approved):
lgtm

rchincha:
cc:@dfr fyi, some build related changes wrt freebsd

rchincha:
We should make our lives a little easier by parameterizing this information so that we change it in only one place.

PR #2382, opened 2024-04-10
fix(cli/server): serve command expected positional args

ralgozino:
Expect exactly one positional argument for the serve command with the path to the config file.

**What type of PR is this?**

bug

There's no reported issue AFAIK.

This PR adds a check to the `serve` command to expect exactly one positional argument with the config file path. Improving the user experience.

Running `zot serve` (without passing a config file path) results in an index out-of-range error:

```
$ ./bin/zot-darwin-arm64 serve
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
zotregistry.dev/zot/pkg/cli/server.NewServerRootCmd.newServeCmd.func2(0x14002026800?, {0x10aede5e0, 0x0, 0x105f211c0?})
	zotregistry.dev/zot/pkg/cli/server/root.go:65 +0x1b0
github.com/spf13/cobra.(*Command).execute(0x14000354608, {0x10aede5e0, 0x0, 0x0})
	github.com/spf13/cobra@v1.8.0/command.go:983 +0x840
github.com/spf13/cobra.(*Command).ExecuteC(0x14000354308)
	github.com/spf13/cobra@v1.8.0/command.go:1115 +0x344
github.com/spf13/cobra.(*Command).Execute(0x10aab7538?)
	github.com/spf13/cobra@v1.8.0/command.go:1039 +0x1c
main.main()
	zotregistry.dev/zot/cmd/zot/main.go:10 +0x20
```

As a new user, I found this error to be misleading. I thought I had some error on the json file instead (I assumed that the `config.json` file would be used by default).

With the change included in this PR applied, this is the output you get instead:

```
$ ./bin/zot-darwin-arm64-minimal serve
Error: accepts 1 arg(s), received 0
Usage:
  zot serve <config> [flags]

Aliases:
  serve, serve

Flags:
  -h, --help   help for serve
```

Passing 1 argument works as expected:

```
$ ./bin/zot-darwin-arm64-minimal serve non-existent.json
{"level":"error","error":"open non-existent.json: no such file or directory","path":"non-existent.json","time":"2024-04-10T10:42:52+02:00","message":"failed to read configuration"}
Error: open non-existent.json: no such file or directory
Usage:
  zot serve <config> [flags]

Aliases:
  serve, serve

Flags:
  -h, --help   help for serve
```

**Automation added to e2e**:

I've added a unit test that checks for an error when not passing the expected positional argument.

Note: I tried running `make` locally, but it is failing on my machine even when running it on the `main` branch. I don't expect this change to impact other parts of the code.

No

```release-note
Running `zot serve` without the path to a config file won't result in an index out-of-range error anymore, it will show the help message of the command instead instructing to add the missing argument.
```

### DIFF
--- a/pkg/cli/server/root_test.go
+++ b/pkg/cli/server/root_test.go
@@ -48,6 +48,18 @@ func TestServe(t *testing.T) {
 	})
 
 	Convey("Test serve config", t, func(c C) {
+		Convey("no config arg", func(c C) {
+			os.Args = []string{"cli_test", "serve"}
+			err := cli.NewServerRootCmd().Execute()
+			So(err, ShouldNotBeNil)
+		})
+
+		Convey("two args", func(c C) {
+			os.Args = []string{"cli_test", "serve", "config", "second arg"}
+			err := cli.NewServerRootCmd().Execute()
+			So(err, ShouldNotBeNil)
+		})
+
 		Convey("unknown config", func(c C) {
 			os.Args = []string{"cli_test", "serve", path.Join(os.TempDir(), "/x")}
 			err := cli.NewServerRootCmd().Execute()
--- a/pkg/cli/server/root.go
+++ b/pkg/cli/server/root.go
@@ -47,6 +47,7 @@ func newServeCmd(conf *config.Config) *cobra.Command {
 		Aliases: []string{"serve"},
 		Short:   "`serve` stores and distributes OCI images",
 		Long:    "`serve` stores and distributes OCI images",
+		Args:    cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) > 0 {
 				if err := LoadConfiguration(conf, args[0]); err != nil {


andaaron (dismissed):

pkg/cli/server/root_test.go
```diff
@@ -48,6 +48,12 @@ func TestServe(t *testing.T) {
 	})
 
 	Convey("Test serve config", t, func(c C) {
```
rchincha, line 50: Also add a test case for more than one arg

pkg/cli/server/root_test.go
```diff
@@ -1256,11 +1262,11 @@ storage:
 		err = tmpCredsFile.Close()
 		So(err, ShouldBeNil)
 
-		content = []byte(fmt.Sprintf(`{ "distSpecVersion": "1.1.0-dev", 
-			"storage": { "rootDirectory": "/tmp/zot" }, "http": { "address": "127.0.0.1", "port": "8080", 
-			"auth": { "ldap": { "credentialsFile": "%v", "address": "ldap.example.org", "port": 389, 
-			"startTLS": false, "baseDN": "ou=Users,dc=example,dc=org", 
-			"userAttribute": "uid", "userGroupAttribute": "memberOf", "skipVerify": true, "subtreeSearch": true }, 
+		content = []byte(fmt.Sprintf(`{ "distSpecVersion": "1.1.0-dev",
```
rchincha, line 1265: Are these whitespace cleanups? Would move this to a separate commit

rchincha (commented):
@ralgozino pls also GPG sign your commits.
https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

ralgozino:
> @ralgozino pls also GPG sign your commits.
> https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

Interesting, I just realized that I have it setup only for my work email address, I will fix it.

Is it enough to GPG sign or should I also sign off the commit? (I see the check for sign off in Ci is failing).

I guess I will have to force push to overwrite the commit

rchincha:
> > @ralgozino pls also GPG sign your commits.
> > https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits
> 
> Interesting, I just realized that I have it setup only for my work email address, I will fix it.
> 
> Is it enough to GPG sign or should I also sign off the commit? (I see the check for sign off in Ci is failing).
> 
> I guess I will have to force push to overwrite the commit

You would have to sign and sign-off also.

pkg/cli/server/root_test.go
ralgozino: I've removed these changes

ralgozino:
I should have addressed all your suggestions @rchincha, let me know if something else needs to be done.

Thank you for the review.

pkg/cli/server/root_test.go
ralgozino: Added the test case for 2 args

rchincha (approved):
lgtm

rchincha:
@ralgozino thanks for your PR.

Issue #125, opened 2020-08-05
zot cluster


hallyn:
We will want to support running a cluster of zot servers.

When a blob is uploaded, it should be distributed to all the nodes.

When fetching an image, the client should be able to fetch each
layer blob from a different server to load balance.

tych0:
Might be nice to do it with redirects, so you only have to talk to *a* zot in the cluster instead of the right one.

rchincha:
Couple of design considerations to consider:
1) is the client aware of the members of a cluster or we do some sort of proxying
2) given unique content-addressable blobs, routing can assume the quorum to be stable or unstable (DHT assumptions)

rchincha:
https://github.com/project-zot/zot/pull/2041

andaaron:
Considerations on storing various data - in the context of the clustering discussion.
I wrote these a while back, some may have been addressed.

- We support local file system and s3 in case of AWS for the image stores (responsible for storing image blobs)
- In case of zot sync we only support local, because the 3rd party library we use for sync-ing uses local storage as an intermediate destination for the copy - a note for kubernetes/cloud use case
- The information about dedupe is stored in a cache DB - on local file system (cache.db under root dir) or on DynamoDB in case of AWS - there are multiple such cache DBs, one per image store - these would probably need to be the same for all zot instances
- Trivy is using local disk space to store CVE information and scan results (in the folder _trivy under the root directory) - 2 DBs, one for Java scanning and one for the rest, the Java one being huge (hundreds of MB if I rememeber correctly) - right now we are not storing CVE scan results anywhere else, we rely on Trivy and in-memory cache of the latest results.
- zot user session authentication (for zui) uses a folder _sessions under the root directory to store session information - if we have multiple zot instances we need to design also for the session authentication use case
- Right now we advertize zot with an embedded UI - in case of a cluster, would we have multiple UIs, or a separate UI service?
- We have a metadata db (meta.db stored locally under the root directory, or as DynamoDB tables for cloud case) - this DB needs to be the same for all zots (we store information on manifests, configs, download counters, signature verification results)
- We store certificates and private keys for signature verification locally regardless of storage type (we want to also support AWS), these files are under the root dir (folders _notation and _cosign) -> not sure if this is still applicable today for AWS.

PR #2385, opened 2024-04-12
feat(cluster): Add support for request proxying for scale out

vrajashkr:
**What type of PR is this?**
feature

Towards #125 
Carry forward of #2041
Initial fork PR https://github.com/rchincha/zot/pull/160

This PR adds support for proxying requests to other members of a zot instance cluster for scale out.
This PR only has support for shared storage cluster scale out.

- Functional sanity tests are integrated into BATS and run as part of CI.
- Scale tests are integrated into BATS and can be triggered via manual workflow trigger, release build, and main branch merge.

**Automation added to e2e**:
N/A - no functionality change from an application point of view. The proxy is transparent to clients.

No. Scale-out is new and there wouldn't be existing setups using this config yet.

yes
```release-note
Users can now specify settings for a "scale-out" mode where multiple instances of zot can run simultaneously using the same shared reliable storage for improved scale and performance in large deployments.
```

TODO:
- [x] Unit Tests for new code
- [x] Fix for localstack bottleneck for high scale
- [x] Manual testing on personal fork for the high scale workflow trigger


test/scale-out/cloud_scale_out_basic_auth_scale.bats
```diff
@@ -0,0 +1,130 @@
+# Note: Intended to be run as "make run-scale-out-tests"
```
vrajashkr, line 1: Typo here - should be `make run-scale-out-high-scale-tests`

pkg/api/proxy.go
```diff
@@ -0,0 +1,110 @@
...
+			h := siphash.New([]byte(config.Cluster.HashKey))
+			h.Write([]byte(name))
+			sum64 := h.Sum64()
+
+			targetMember := config.Cluster.Members[sum64%uint64(len(config.Cluster.Members))]
+
+			// from the member list and our DNS/IP address, figure out if this request should be handled locally
+			localMember := fmt.Sprintf("%s:%s", config.HTTP.Address, config.HTTP.Port)
+			if targetMember == localMember {
+				logger.Debug().Msg("Target cluster member is the local member. Handling request locally.")
```
rchincha, line 42: use lowercase log msgs so s/Target/target/
everywhere ...

Also add more details such as logger.Debug().Str("name", name).Msg() etc.

pkg/api/proxy.go
```diff
@@ -0,0 +1,110 @@
...
+				return
+			}
+			logger.Debug().Msg(fmt.Sprintf("Target member is %s. Proxying the request", targetMember))
+
+			proxyQueryScheme := "http"
+			if config.HTTP.TLS != nil {
+				proxyQueryScheme = "https"
+			}
+
+			proxyHttpRequest(response, request, targetMember, proxyQueryScheme)
```
rchincha, line 53: Can this subroutine throw an error ... if so handle/log it.

rchincha:
This turned out to be a fairly simple change.
Let's focus on making all state into "single-writer" model next.

vrajashkr:
> This turned out to be a fairly simple change. Let's focus on making all state into "single-writer" model next.

Thanks for the base commit @rchincha. It set a good foundation for me to build on :)

pkg/api/proxy.go
vrajashkr: Addressed the log message case comment.

Will look at adding more relevant details to the logger.

pkg/api/routes.go
```diff
@@ -127,40 +127,60 @@ func (rh *RouteHandler) SetupRoutes() {
...
-				applyCORSHeaders(rh.GetReferrers))).Methods(http.MethodGet, http.MethodOptions)
+			clusterRouteProxy(
+				getUIHeadersHandler(rh.c.Config, http.MethodGet, http.MethodOptions)(
+					applyCORSHeaders(rh.GetReferrers),
+				),
+			),
+		).Methods(http.MethodGet, http.MethodOptions)
 		prefixedRouter.HandleFunc(constants.ExtCatalogPrefix,
 			getUIHeadersHandler(rh.c.Config, http.MethodGet, http.MethodOptions)(
 				applyCORSHeaders(rh.ListRepositories))).Methods(http.MethodGet, http.MethodOptions)
```
andaaron, line 186: How do we aggregate the list of repos for the entire cluster? Or we don't need to?

pkg/api/proxy.go
```diff
@@ -0,0 +1,123 @@
...
+				response.WriteHeader(http.StatusNotFound)
+
+				return
+			}
+
+			h := siphash.New([]byte(config.Cluster.HashKey))
+			h.Write([]byte(name))
+			sum64 := h.Sum64()
+
+			targetMember := config.Cluster.Members[sum64%uint64(len(config.Cluster.Members))]
```
andaaron, line 42: Is there a possibility to loose access to content (by being redirected to bad member) if I add or remove cluster members (number of cluster members changes)?

rchincha: Yes, a node can go down and cluster has to be re-shaped.

pkg/api/proxy.go
vrajashkr: Added "name" of the repo to the logger. Thanks for the suggestion!

pkg/api/routes.go
vrajashkr: Yes, we do need to aggregate the list of repos for the entire cluster. That's one of the next items on the list of TODOs.

examples/config-cluster-member0.json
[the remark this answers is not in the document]
vrajashkr: Thanks for the suggestion! I'll move them to the existing cluster directory in the next commit.

HAProxy would be ideal as a frontend to load balance requests to nodes, however, the load balancing would be better off done via the name rather than round-robin. This would help improve performance, but a round-robin balancing would be functionally identical as the nodes will internally proxy as required.

luisdavim:
latter wold it be possible to leverage this to scale out in situations where a shared storage is not available by also setting up syncs between the members?

examples/config-cluster-member0.json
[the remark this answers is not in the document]
vrajashkr: I've moved the examples into a dedicated directory "examples/scale-out-cluster-cloud".

pkg/api/proxy.go
vrajashkr: Thanks for the comment! I've handled this. If there is a failure, the response will return status 500 and the error will be logged.

vrajashkr:
> latter wold it be possible to leverage this to scale out in situations where a shared storage is not available by also setting up syncs between the members?

Hi @luisdavim, thank you for the question! This PR focuses on shared storage, but support for a situation without shared storage is definitely something that is being looked into.

pkg/api/routes.go
vrajashkr: For shared storage, there won't be a need to aggregate as all instances have access to all the data. In a scenario where individual instances have isolated storage, we will need to aggregate them.

pkg/api/proxy.go
```diff
@@ -0,0 +1,147 @@
...
+	clonedBody := cloneRequestBody(req)
+
+	fwdRequest, err := http.NewRequestWithContext(ctx, req.Method, cloneURL.String(), clonedBody)
+	if err != nil {
+		return nil, err
+	}
+
+	copyHeader(fwdRequest.Header, req.Header)
+
+	resp, err := getHTTPClient(ctrlr).Do(fwdRequest)
```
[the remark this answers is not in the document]
vrajashkr: This is expected as the code is proxying the client's request. The implementation ensures that the proxying only happens when the route path is matched.

examples/scale-out-cluster-cloud/config-cluster-member0.json
```diff
@@ -0,0 +1,44 @@
...
+  "http": {
+    "address": "127.0.0.1",
+    "port": "9000"
+  },
+  "log": {
+    "level": "debug"
+  },
+  "cluster": {
+    "members": [
+      "127.0.0.1:9000",
```
rchincha, line 38: The only difference should be these ports since we are running on bare-metal

pkg/api/config/config.go
```diff
@@ -121,6 +121,13 @@ type SchedulerConfig struct {
 	NumWorkers int
 }
 
+// ClusterConfig is the scale-out configuration which is identical for all
+// replicas.
+type ClusterConfig struct {
+	Members []string `json:"members" mapstructure:"members"`
+	HashKey string   `json:"hashKey" mapstructure:"hashKey"`
```
rchincha, line 128: Add a comment that we use "SipHash" as the hashing algorithm, so we need a hash key.

pkg/api/constants/consts.go
```diff
@@ -31,4 +31,6 @@ const (
 	DeletePermission = "delete"
 	// behaviour actions.
 	DetectManifestCollisionPermission = "detectManifestCollision"
+	// Log String keys.
+	RepositoryLogKey = "repository"
```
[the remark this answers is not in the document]
vrajashkr: It is used for logger statements to add context. The key is usually replicated in all the statements, so I created a re-usable constant.

For example: `logger.Debug().Str(constants.RepositoryLogKey, name).Msg("proxying the request")`

test/scale-out/cloud_scale_out_basic_auth_tls.bats
```diff
@@ -0,0 +1,159 @@
...
+function teardown() {
+    local zot_root_dir=${BATS_FILE_TMPDIR}/zot
+    haproxy_stop_all
+    zot_stop_all
+    rm -rf ${zot_root_dir}
+    teardown_cloud_services
+}
+
+@test "Check for successful zb run on haproxy frontend" {
+    zb_run "cloud-scale-out-basic-auth-tls-bats" "https://127.0.0.1:8000" 1 1 "${ZOT_AUTH_USER}:${ZOT_AUTH_PASS}"
```
rchincha, line 158: zb add more clients and requests and also use a cidr
  -s, --src-cidr string        Use specified cidr to obtain ips to make requests from, src-ips and src-cidr are mutually exclusive
  -i, --src-ips string         Use colon-separated ips to make requests from, src-ips and src-cidr are mutually exclusive
  
  so that multiple client src IPs can be used.

rchincha: Same everywhere else.

pkg/api/proxy.go
```diff
@@ -0,0 +1,160 @@
...
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/dchest/siphash"
+	"github.com/gorilla/mux"
+
+	"zotregistry.dev/zot/pkg/api/constants"
+)
+
```
rchincha, line 16: Add comments on what this does

pkg/api/proxy.go
```diff
@@ -0,0 +1,160 @@
...
+			vars := mux.Vars(request)
+			name, ok := vars["name"]
+
+			if !ok || name == "" {
+				response.WriteHeader(http.StatusNotFound)
+
+				return
+			}
+
+			h := siphash.New([]byte(config.Cluster.HashKey))
```
rchincha, line 39: Do we always have to do a "New()" here? Why not stash it and do a `if h { h.Reset()}`

rchincha: because New() could allocate and be expensive?

examples/scale-out-cluster-cloud/config-cluster-member0.json
```diff
@@ -0,0 +1,44 @@
...
+      "127.0.0.1:9001",
+      "127.0.0.1:9002"
+    ],
+    "hashKey": "loremipsumdolors"
```
rchincha, line 42: What is the size requirement for this hashKey? Can we enforce it?

rchincha: "SipHash computes a 64-bit [message authentication code](https://en.wikipedia.org/wiki/Message_authentication_code) from a variable-length message and 128-bit secret key"

pkg/api/constants/consts.go
```diff
@@ -31,4 +31,6 @@ const (
 	DeletePermission = "delete"
 	// behaviour actions.
 	DetectManifestCollisionPermission = "detectManifestCollision"
+	// Log String keys.
```
rchincha, line 34: "Log" make it lower case

rchincha:
Add detailed comments in your commit msg. This is a significant feature enhancement.

pkg/api/constants/consts.go
vrajashkr: Sure, will change this in this file.

However, for the other files, would it be alright to keep this casing? For longer comments, it helps with readability.
Here's a reference I got from golang's website: https://go.dev/doc/comment

pkg/api/proxy.go
vrajashkr: Added comments

test/scale-out/cloud_scale_out_basic_auth_tls.bats
vrajashkr: Added --src-cidr

test/scale-out/cloud_scale_out_basic_auth_tls.bats
```diff
@@ -0,0 +1,159 @@
...
+    option ssl-hello-chk
+    balance roundrobin
+EOF
+    # Populate server list
+    generate_haproxy_server_list ${NUM_ZOT_INSTANCES} >> ${haproxy_cfg_file}
+
+    haproxy_start ${haproxy_cfg_file}
+}
+
+function launch_zot_server() {
```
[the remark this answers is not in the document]
vrajashkr: Refactored the files to make it modular and re-usable.

pkg/api/config/config.go
vrajashkr: Added comments.

.github/workflows/nightly.yaml
```diff
@@ -6,11 +6,13 @@ on:
```
rchincha, line 6: Also add the following so we can trigger this (but revert it later)

on:
  push:
    branches:
      - main
  pull_request:
    branches:
      - main

test/scale-out/helpers_zot.bash
```diff
@@ -0,0 +1,239 @@
...
+    local concurrent_reqs=${3}
+    local num_requests=${4}
+    local credentials=${5}
+
+    if [ ! -d "${ZB_RESULTS_PATH}" ]; then
+        mkdir -p "${ZB_RESULTS_PATH}"
+    fi
+
+    if [ -z "${credentials}" ]; then
+        ${ZB_PATH} -c ${concurrent_reqs} -n ${num_requests} ${zot_address} --src-cidr 127.0.10.0/24 -o ci-cd --skip-cleanup
```
rchincha, line 103: perfect!

pkg/api/routes.go
```diff
@@ -127,40 +127,66 @@ func (rh *RouteHandler) SetupRoutes() {
...
+	clusterRouteProxy := ClusterProxy(rh.c)
+
 	// https://github.com/opencontainers/distribution-spec/blob/main/spec.md#endpoints
+	// dist-spec APIs that need to be proxied are wrapped in clusterRouteProxy for scale-out proxying.
+	// these are handlers that have a repository name.
 	{
 		prefixedDistSpecRouter.HandleFunc(fmt.Sprintf("/{name:%s}/tags/list", zreg.NameRegexp.String()),
-			getUIHeadersHandler(rh.c.Config, http.MethodGet, http.MethodOptions)(
-				applyCORSHeaders(rh.ListTags))).Methods(http.MethodGet, http.MethodOptions)
+			clusterRouteProxy(
```
rchincha, line 137: For the graphQL query, if a local metadb (boltdb), then we need a fan-out since state will be local.

.github/workflows/nightly.yaml
vrajashkr: Thanks for the suggestion! I do see that this file has a `workflow_dispatch:`. From what I gather, this should ideally allow us to trigger it manually as soon as we merge so we may not need the `on: push:`.

Meanwhile, I will add a `pull_request` section so that it runs in the PR.

pkg/api/routes.go
vrajashkr: I agree. We would need to fetch data from the other instances before serving results in this case.
We could probably take this up in the next PR. This one could focus on shared storage and shared metadata.

pkg/api/proxy.go
vrajashkr: Ran a quick check on the siphash repo to compare the approaches:

Added the following code to `siphash_test.go`:
```go
func BenchmarkSum64WithLongStringInput(b *testing.B) {
	h := New([]byte("loremipsumdolors"))
	h.Write([]byte("thisisaveryveryveryveryveryveryveryveryveryveryveryveryveryveryverylongreponame"))
	h.Sum64()
}

func BenchmarkHashWithLongStringInput(b *testing.B) {
	Hash(1234567890, 123143930393, []byte("thisisaveryveryveryveryveryveryveryveryveryveryveryveryveryveryverylongreponame"))
}
```
Running these gave the following results:
```
➜  siphash git:(master) ✗ go test -bench '^BenchmarkSum64WithLongStringInput$'
goos: linux
goarch: amd64
pkg: github.com/dchest/siphash
cpu: Intel Core i7 9xx (Nehalem Class Core i7)
BenchmarkSum64WithLongStringInput-6   	1000000000	         0.0000013 ns/op
PASS
ok  	github.com/dchest/siphash	0.014s
➜  siphash git:(master) ✗ go test -bench '^BenchmarkHashWithLongStringInput$'
goos: linux
goarch: amd64
pkg: github.com/dchest/siphash
cpu: Intel Core i7 9xx (Nehalem Class Core i7)
BenchmarkHashWithLongStringInput-6   	1000000000	         0.0000003 ns/op
PASS
ok  	github.com/dchest/siphash	0.015s
➜  siphash git:(master) ✗ 
```
The results were repeatable in multiple executions.

The results seem to indicate that for zot's use case, there doesn't seem to be a significant change in execution time (only on the order of about 10ns/op for very large repo names). We can probably stick to the implementation that accepts a byte so that users can supply strings in the config file as opposed to accepting a string and then converting to uint64 for the Hash function (the `newDigest` method called by `New` internally does this).

examples/scale-out-cluster-cloud/config-cluster-member0.json
vrajashkr: Yes, the buffer must be 128 bits which translates to a string of 16 character length where each character takes up 8 bits. We can enforce this. Will add a validation check for the same.

.github/workflows/nightly.yaml
vrajashkr: Added pull_request to the commit. Once the result is available from the run, will remove it again.

examples/scale-out-cluster-cloud/config-cluster-member0.json
vrajashkr: Added a config validation for this and the members list.

vrajashkr:
Currently trying to figure out what's up in the nightly job.

This works:
```
cluster members patch file
{
  "cluster": {
    "members": [
      "127.0.0.1:10000",
      "127.0.0.1:10001",
      "127.0.0.1:10002",
      "127.0.0.1:10003",
      "127.0.0.1:10004",
      "127.0.0.1:10005"
    ]
  }
}
launching zot server 127.0.0.1:10000
config file: /tmp/bats-run-v0GNci/file/0/zot_config_127.0.0.1_10000.json
log file: /tmp/bats-run-v0GNci/file/0/zot-127.0.0.1-10000.log
launching zot server 127.0.0.1:10001
config file: /tmp/bats-run-v0GNci/file/0/zot_config_127.0.0.1_10001.json
log file: /tmp/bats-run-v0GNci/file/0/zot-127.0.0.1-10001.log
launching zot server 127.0.0.1:10002
config file: /tmp/bats-run-v0GNci/file/0/zot_config_127.0.0.1_10002.json
log file: /tmp/bats-run-v0GNci/file/0/zot-127.0.0.1-10002.log
launching zot server 127.0.0.1:10003
config file: /tmp/bats-run-v0GNci/file/0/zot_config_127.0.0.1_10003.json
log file: /tmp/bats-run-v0GNci/file/0/zot-127.0.0.1-10003.log
launching zot server 127.0.0.1:10004
config file: /tmp/bats-run-v0GNci/file/0/zot_config_127.0.0.1_10004.json
log file: /tmp/bats-run-v0GNci/file/0/zot-127.0.0.1-10004.log
launching zot server 127.0.0.1:10005
config file: /tmp/bats-run-v0GNci/file/0/zot_config_127.0.0.1_10005.json
log file: /tmp/bats-run-v0GNci/file/0/zot-127.0.0.1-10005.log
root       46486   46465  3 09:37 ?        00:00:15 /usr/local/bin/zot-linux-amd64 serve /etc/zot/config.json
runner     54576   54527 15 09:44 ?        00:00:00 /home/runner/work/zot/zot/bin/zot-linux-amd64 serve /tmp/bats-run-v0GNci/file/0/zot_config_127.0.0.1_10000.json
runner     54608   54527 17 09:44 ?        00:00:00 /home/runner/work/zot/zot/bin/zot-linux-amd64 serve /tmp/bats-run-v0GNci/file/0/zot_config_127.0.0.1_10001.json
runner     54664   54527 22 09:44 ?        00:00:00 /home/runner/work/zot/zot/bin/zot-linux-amd64 serve /tmp/bats-run-v0GNci/file/0/zot_config_127.0.0.1_10002.json
runner     54694   54527 29 09:44 ?        00:00:00 /home/runner/work/zot/zot/bin/zot-linux-amd64 serve /tmp/bats-run-v0GNci/file/0/zot_config_127.0.0.1_10003.json
runner     54727   54527 43 09:44 ?        00:00:00 /home/runner/work/zot/zot/bin/zot-linux-amd64 serve /tmp/bats-run-v0GNci/file/0/zot_config_127.0.0.1_10004.json
runner     54757   54527 86 09:44 ?        00:00:00 /home/runner/work/zot/zot/bin/zot-linux-amd64 serve /tmp/bats-run-v0GNci/file/0/zot_config_127.0.0.1_10005.json
global
    log /tmp/bats-run-v0GNci/file/0/haproxy/log    local0
    log /tmp/bats-run-v0GNci/file/0/haproxy/log    local1 notice
    maxconn 20000
    stats timeout 30s
    daemon

defaults
    log     global
    mode    tcp
    option  tcplog
    option  dontlognull
    timeout connect 5000
    timeout client  50000
    timeout server  50000

frontend zot
    bind *:8000
    default_backend zot-cluster

backend zot-cluster
    balance roundrobin
    option ssl-hello-chk
    server zot0 127.0.0.1:10000 check
    server zot1 127.0.0.1:10001 check
    server zot2 127.0.0.1:10002 check
    server zot3 127.0.0.1:10003 check
    server zot4 127.0.0.1:10004 check
    server zot5 127.0.0.1:10005 check
Configuration file is valid
root        5607       1  0 09:32 ?        00:00:00 /usr/sbin/haproxy -Ws -f /etc/haproxy/haproxy.cfg -p /run/haproxy.pid -S /run/haproxy-master.sock
haproxy     5609    5607  0 09:32 ?        00:00:00 /usr/sbin/haproxy -Ws -f /etc/haproxy/haproxy.cfg -p /run/haproxy.pid -S /run/haproxy-master.sock
runner     54527   54508  1 09:44 ?        00:00:00 bash /home/runner/work/zot/zot/hack/tools/libexec/bats-core/bats-exec-test --dummy-flag --print-output-on-failure /home/runner/work/zot/zot/test/scale-out/cloud_scale_out_basic_auth_tls.bats test_Check_for_successful_zb_run_on_haproxy_frontend 1 1 1
runner     54791       1  0 09:44 ?        00:00:00 haproxy -f /tmp/bats-run-v0GNci/file/0/haproxy/haproxy-test.cfg
ok 1 Check for successful zb run on haproxy frontend
```

but this doesn't (same test, but with 100 instances, 100 concurrency, and 1000 requests):

```
cluster members patch file
{
  "cluster": {
    "members": [
      "127.0.0.1:10000",
      "127.0.0.1:10001",
      "127.0.0.1:10002",
      "<TRUNCATED>",
      "127.0.0.1:10099"
    ]
  }
}
launching zot server 127.0.0.1:10000
config file: /tmp/bats-run-qLebO3/file/0/zot_config_127.0.0.1_10000.json
log file: /tmp/bats-run-qLebO3/file/0/zot-127.0.0.1-10000.log
launching zot server 127.0.0.1:10001
config file: /tmp/bats-run-qLebO3/file/0/zot_config_127.0.0.1_10001.json
log file: /tmp/bats-run-qLebO3/file/0/zot-127.0.0.1-10001.log
launching zot server 127.0.0.1:10002
config file: /tmp/bats-run-qLebO3/file/0/zot_config_127.0.0.1_10002.json
log file: /tmp/bats-run-qLebO3/file/0/zot-127.0.0.1-10002.log
launching zot server 127.0.0.1:10003
<TRUNCATED>
launching zot server 127.0.0.1:10099
config file: /tmp/bats-run-qLebO3/file/0/zot_config_127.0.0.1_10099.json
log file: /tmp/bats-run-qLebO3/file/0/zot-127.0.0.1-10099.log
runner     34273   33971  0 09:33 ?        00:00:01 /home/runner/work/zot/zot/bin/zot-linux-amd64 serve /tmp/bats-run-qLebO3/file/0/zot_config_127.0.0.1_10000.json
runner     34303   33971  0 09:33 ?        00:00:01 /home/runner/work/zot/zot/bin/zot-linux-amd64 serve /tmp/bats-run-qLebO3/file/0/zot_config_127.0.0.1_10001.json
runner     34332   33971  0 09:33 ?        00:00:01 /home/runner/work/zot/zot/bin/zot-linux-amd64 serve /tmp/bats-run-qLebO3/file/0/zot_config_127.0.0.1_10002.json
<TRUNCATED>
runner     37519   33971 88 09:35 ?        00:00:00 /home/runner/work/zot/zot/bin/zot-linux-amd64 serve /tmp/bats-run-qLebO3/file/0/zot_config_127.0.0.1_10099.json
global
    log /tmp/bats-run-qLebO3/file/0/haproxy/log    local0
    log /tmp/bats-run-qLebO3/file/0/haproxy/log    local1 notice
    maxconn 20000
    stats timeout 30s
    daemon

defaults
    log     global
    mode    tcp
    option  tcplog
    option  dontlognull
    timeout connect 5000
    timeout client  50000
    timeout server  50000

frontend zot
    bind *:8000
    default_backend zot-cluster

backend zot-cluster
    balance roundrobin
    option ssl-hello-chk
    server zot0 127.0.0.1:10000 check
    server zot1 127.0.0.1:10001 check
    server zot2 127.0.0.1:10002 check
    server zot3 127.0.0.1:10003 check
    <TRUNCATED>
    server zot99 127.0.0.1:10099 check
Configuration file is valid
root        5626       1  0 09:30 ?        00:00:00 /usr/sbin/haproxy -Ws -f /etc/haproxy/haproxy.cfg -p /run/haproxy.pid -S /run/haproxy-master.sock
haproxy     5628    5626  0 09:30 ?        00:00:00 /usr/sbin/haproxy -Ws -f /etc/haproxy/haproxy.cfg -p /run/haproxy.pid -S /run/haproxy-master.sock
runner     33971   33952  1 09:33 ?        00:00:01 bash /home/runner/work/zot/zot/hack/tools/libexec/bats-core/bats-exec-test --dummy-flag --print-output-on-failure /home/runner/work/zot/zot/test/scale-out/cloud_scale_out_basic_auth_tls_scale.bats test_Check_for_successful_zb_run_on_haproxy_frontend 1 1 1
runner     37565       1  0 09:35 ?        00:00:00 haproxy -f /tmp/bats-run-qLebO3/file/0/haproxy/haproxy-test.cfg
not ok 1 Check for successful zb run on haproxy frontend
# (from function `zb_run' in file test/scale-out/helpers_zot.bash, line 105,
#  in test file test/scale-out/cloud_scale_out_basic_auth_tls_scale.bats, line 72)
#   `zb_run "cloud-scale-out-high-scale-bats" "https://127.0.0.1:8000" 100 1000 "${ZOT_AUTH_USER}:${ZOT_AUTH_PASS}"' failed
# make_bucket: zot-storage-test
# {
#     "TableDescription": {
#         "AttributeDefinitions": [
#             {
#                 "AttributeName": "Digest",
#                 "AttributeType": "S"
#             }
#         ],
#         "TableName": "BlobTable",
#         "KeySchema": [
#             {
#                 "AttributeName": "Digest",
#                 "KeyType": "HASH"
#             }
#         ],
#         "TableStatus": "ACTIVE",
#         "CreationDateTime": 1714901626.806,
#         "ProvisionedThroughput": {
#             "ReadCapacityUnits": 10,
#             "WriteCapacityUnits": 5
#         },
#         "TableSizeBytes": 0,
#         "ItemCount": 0,
#         "TableArn": "arn:aws:dynamodb:us-east-2:000000000000:table/BlobTable",
#         "TableId": "e935d51e-5650-48bc-854f-5fa7d9d4c22a",
#         "DeletionProtectionEnabled": false
#     }
# }
# Adding password for user poweruser
#   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
#                                  Dload  Upload   Total   Spent    Left  Speed
# 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
# curl: (7) Failed to connect to 127.0.0.1 port 10000 after 0 ms: Connection refused
# Warning: Problem : connection refused. Will retry in 1 seconds. 20 retries
# Warning: left.
# 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100    48    0    48    0     0  74188      0 --:--:-- --:--:-- --:--:-- 48000
# Client sent an HTTP request to an HTTPS server.
#   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
#                                  Dload  Upload   Total   Spent    Left  Speed
# 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
# curl: (7) Failed to connect to 127.0.0.1 port 10001 after 0 ms: Connection refused
# Warning: Problem : connection refused. Will retry in 1 seconds. 20 retries
# Warning: left.
# 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100    48    0    48    0     0  86486      0 --:--:-- --:--:-- --:--:-- 48000
# Client sent an HTTP request to an HTTPS server.
#   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
#                                  Dload  Upload   Total   Spent    Left  Speed
# 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
# curl: (7) Failed to connect to 127.0.0.1 port 10002 after 0 ms: Connection refused
# Warning: Problem : connection refused. Will retry in 1 seconds. 20 retries
# Warning: left.
# 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100    48    0    48    0     0  77922      0 --:--:-- --:--:-- --:--:-- 48000
# Client sent an HTTP request to an HTTPS server.
#   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
#                                  Dload  Upload   Total   Spent    Left  Speed
# 

<TRUNCATED>
# 2024/05/05 09:36:17 http: TLS handshake error from 127.0.0.1:35688: tls: client offered only unsupported versions: []
# 2024/05/05 09:36:17 http: TLS handshake error from 127.0.0.1:49802: tls: client offered only unsupported versions: []
# 2024/05/05 09:36:17 http: TLS handshake error from 127.0.0.1:45762: tls: client offered only unsupported versions: []
# 2024/05/05 09:36:17 http: TLS handshake error from 127.0.0.1:46754: tls: client offered only unsupported versions: []
# 2024/05/05 09:36:17 http: TLS handshake error from 127.0.0.1:48458: tls: client offered only unsupported versions: []
# 2024/05/05 09:36:17 http: TLS handshake error from 127.0.0.1:36336: tls: client offered only unsupported versions: []
# 2024/05/05 09:36:17 http: TLS handshake error from 127.0.0.1:59904: tls: client offered only unsupported versions: []
# 2024/05/05 09:36:17 http: TLS handshake error from 127.0.0.1:37558: tls: client offered only unsupported versions: []
# 2024/05/05 09:36:17 http: TLS handshake error from 127.0.0.1:47264: tls: client offered only unsupported versions: []
# 2024/05/05 09:36:17 http: TLS handshake error from 127.0.0.1:36020: tls: client offered only unsupported versions: []
# 2024/05/05 09:36:17 http: TLS handshake error from 127.0.0.1:42472: tls: client offered only unsupported versions: []
# 2024/05/05 09:36:17 http: TLS handshake error from 127.0.0.1:52352: tls: client offered only unsupported versions: []
# 2024/05/05 09:36:17 http: TLS handshake error from 127.0.0.1:52300: tls: client offered only unsupported versions: []
# 2024/05/05 09:36:17 http: TLS handshake error from 127.0.0.1:37172: tls: client offered only unsupported versions: []
```

Running the test locally with a smaller count was also successful. If it is an issue with scale, a TLS error was the last thing I was expecting :)

pkg/cli/server/root_test.go
```diff
@@ -2028,6 +2028,141 @@ func TestUpdateLDAPConfig(t *testing.T) {
...
+
+	Convey("Should successfully load example configs for cloud", t, func() {
+		for memberIdx := 0; memberIdx < 3; memberIdx++ {
+			cfgFileToLoad := fmt.Sprintf("%s/config-cluster-member%d.json", baseExamplePath, memberIdx)
+			cfg := config.New()
+			err := cli.LoadConfiguration(cfg, cfgFileToLoad)
+			So(err, ShouldBeNil)
+		}
+	})
+
```
vrajashkr, line 2042: add a test case for validating the tls configs too to verify that they are correct.

examples/scale-out-cluster-cloud/haproxy.cfg
```diff
@@ -0,0 +1,24 @@
+global
+        maxconn 2000
```
vrajashkr, line 2: add back the log statements - validation brings up a warning otherwise.

examples/scale-out-cluster-cloud/tls/haproxy.cfg
```diff
@@ -0,0 +1,22 @@
+global
+        maxconn 2000
```
vrajashkr, line 2: add back the log statements - validation brings up a warning otherwise.

pkg/api/proxy.go
```diff
@@ -0,0 +1,185 @@
...
+				return
+			}
+
+			// the target member is the only one which should do read/write for the dist-spec APIs
+			// for the given repository.
+			targetMember := computeTargetMember(config, name)
+
+			// from the member list and our DNS/IP address, calculate if this request should be handled locally.
+			// handle locally, if the "address:port" is a match, else, proxy it to the target member.
+			localMember := fmt.Sprintf("%s:%s", config.HTTP.Address, config.HTTP.Port)
```
andaaron, line 52: What if the configured local member IP address is 127.0.0.1 or 0.0.0.0? Maybe we should check if the targetMember is one of the addresses of the local node. Maybe keep a list of the local addresses in a variable.

pkg/cli/server/root_test.go
vrajashkr: Added a test for this.

examples/scale-out-cluster-cloud/haproxy.cfg
vrajashkr: added

examples/scale-out-cluster-cloud/tls/haproxy.cfg
vrajashkr: added

.github/workflows/nightly.yaml
vrajashkr: The nightly job is green now! :)
I'll be removing it in the next push.

image

pkg/api/proxy.go
vrajashkr: Thanks for the comment @andaaron 
I've addressed this by using the same approach as the sync extension - as suggested.
We now fetch the local IPs from the interfaces as well as resolve hostnames if the member list contains hostnames.

Need to write some Unit Tests for it though.

rchincha:
https://github.com/project-zot/zot/blob/a716be3bbb05072f2cb4096988fc2c1d1c2a8e61/pkg/api/proxy.go#L263-L263
resp, err := httpClient.Do(fwdRequest)

^ make sure the regex check is done on this "url" before forwarding
https://github.com/project-zot/zot/blob/main/pkg/regexp/regexp.go <- find uses of this.

pkg/api/controller.go
```diff
@@ -58,6 +58,13 @@ func NewController(config *config.Config) *Controller {
 	var controller Controller
 
 	logger := log.NewLogger(config.Log.Level, config.Log.Output)
+
+	if config.Cluster != nil {
+		// add global context of address:port for the instance's loggers only if
+		// scale-out cluster mode is enabled
+		host := fmt.Sprintf("%s:%s", config.HTTP.Address, config.HTTP.Port)
```
rchincha, line 65: This is not going to work right? Since you are likely only listening on 0.0.0.0:9000 which is same for every member. You really want to find that entry in the cluster memberlist that belongs to this node.

rchincha: But this log is super-important

pkg/api/controller_test.go
```diff
@@ -958,6 +958,434 @@ func TestBlobReferenced(t *testing.T) {
...
+		Convey("Proxying a request should fail with an error", func() {
+			// debian gets proxied to the second instance
+			resp, err := resty.R().Get(fmt.Sprintf("%s/v2/%s/tags/list", test.GetSecureBaseURL(ports[0]), "debian"))
+			So(err, ShouldBeNil)
+			So(resp, ShouldNotBeNil)
+			So(resp.StatusCode(), ShouldEqual, http.StatusInternalServerError)
+		})
+	})
+}
+
```
rchincha, line 1388: Thanks for all these test cases.

pkg/api/proxy.go
```diff
@@ -0,0 +1,310 @@
...
+			targetMemberIPs, err := getTargetMemberServerSockets(targetMember)
+			if err != nil {
+				logger.Error().Str(constants.RepositoryLogKey, name).
+					Msg(fmt.Sprintf("failed to get target member sockets. error: %s", err.Error()))
+				response.WriteHeader(http.StatusInternalServerError)
+
+				return
+			}
+
+			logger.Debug().Str(constants.RepositoryLogKey, name).Msg(fmt.Sprintf("target member sockets: %v", targetMemberIPs))
```
rchincha, line 77: Avoid %v in Msg()
Instead, use .Str("members", targetMemberIPs).Msg("cluster members")

Same approach everywhere pls.

pkg/api/controller.go
```diff
@@ -58,6 +58,13 @@ func NewController(config *config.Config) *Controller {
...
+		logger.Logger = logger.Logger.With().Str("host", host).Logger()
```
rchincha, line 66: Instead of "host", call this "clusterMember", so we know we are logging in context of a cluster.

pkg/common/common.go
```diff
@@ -145,3 +146,42 @@ func IsContextDone(ctx context.Context) bool {
 		return false
 	}
 }
+
+func GetLocalIPs() ([]string, error) {
```
rchincha, line 150: Also account for the possibility that it is DNS names and not actual IP addresses in the cluster member list. So if dns name, resolve the name and check if that IP is plumbed.

pkg/api/proxy.go
```diff
@@ -0,0 +1,310 @@
...
+			}
+
+			if isLocal {
+				logger.Debug().Str(constants.RepositoryLogKey, name).Msg("handling the request locally")
+				next.ServeHTTP(response, request)
+
+				return
+			}
+
+			// if the header contains a hop-count, return an error response as there should be no multi-hop
```
rchincha, line 96: Ok, this check is *after* the isLocal check, works.

pkg/common/common.go
```diff
@@ -145,3 +146,42 @@ func IsContextDone(ctx context.Context) bool {
...
+			if localIP, ok := addr.(*net.IPNet); ok {
+				localIPs = append(localIPs, localIP.IP.String())
+			}
+		}
+	}
+
+	return localIPs, nil
+}
+
+func GetIPFromHostName(host string) ([]string, error) {
```
andaaron, line 174: This code doesn't seem to be covered by tests. Are we actually calling it?

pkg/api/proxy.go
```diff
@@ -0,0 +1,310 @@
...
+	targetHost := target[0]
+	targetPort := target[1]
+
+	addr := net.ParseIP(targetHost)
+	if addr != nil {
+		// this is an IP address, return as is
+		return []string{targetMemberSocket}, nil
+	}
+	// this is a hostname - try to resolve to an IP
+	resolvedAddrs, err := common.GetIPFromHostName(targetHost)
```
andaaron, line 204: Looks like this code is not reached by tests. We should cover it.

andaaron (commented):
Do we want to compute the sockets on every request? Is there a performance impact?

rchincha:
> Do we want to compute the sockets on every request? Is there a performance impact?

Yes, we shouldn't.

pkg/api/controller.go
vrajashkr: Addressed. I've renamed it to "clusterMember" and also added a new field for the member index for quicker lookups.

pkg/common/common.go
vrajashkr: Yes, `GetIPFromHostName` in this file performs lookup for hostnames.

This is called in controller.go when identifying the local member's socket in order to resolve all the target Member IP addresses.

pkg/api/proxy.go
vrajashkr: Yes, if we attempt to proxy it, the process will quit.

pkg/api/proxy.go
vrajashkr: Thanks! I've added code coverage for this.

vrajashkr:
> > Do we want to compute the sockets on every request? Is there a performance impact?
> 
> Yes, we shouldn't.

I agree, this a performance hit especially since it happens for every request. I've changed to logic to perform the socket logic at the controller startup (because we need the member data for logging). This happens only once.

When proxying, we do a lookup on the index alone (since all the zots have an identical member config) which is a quicker comparison than string compare too.

pkg/api/proxy.go
vrajashkr: Addressed this. I've changed it to add all the context to the logger as a field as much as possible.

.github/workflows/nightly.yaml
vrajashkr: This has now been removed.

pkg/common/common.go
```diff
@@ -145,3 +146,93 @@ func IsContextDone(ctx context.Context) bool {
...
 }
+
+// get a list of IP addresses configured on the host's
+// interfaces.
+func GetLocalIPs() ([]string, error) {
+	var localIPs []string
+
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return []string{}, err
```
vrajashkr, line 157: @rchincha, do you happen to have any thoughts on how we could test for these error scenarios where we're querying something from the system?

pkg/api/proxy.go
```diff
@@ -0,0 +1,185 @@
...
+		clientOpts.CertOptions.ClientKeyFile = tlsConfig.Key
+		clientOpts.CertOptions.RootCaCertFile = tlsConfig.CACert
+	}
+
+	httpClient, err := common.CreateHTTPClient(&clientOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := httpClient.Do(fwdRequest)
```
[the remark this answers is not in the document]
vrajashkr: This code is reached only when it is routed. There is a path check before routing:
`prefixedDistSpecRouter.HandleFunc(fmt.Sprintf("/{name:%s}/tags/list", zreg.NameRegexp.String()),`

I'm not 100% sure on whether there can be content after list and still call this handler. TODO: need to check that.

However, regexing once more here would add a penalty to every proxied request.

vrajashkr:
> https://github.com/project-zot/zot/blob/a716be3bbb05072f2cb4096988fc2c1d1c2a8e61/pkg/api/proxy.go#L263-L263
> 
> 
> resp, err := httpClient.Do(fwdRequest)
> ^ make sure the regex check is done on this "url" before forwarding https://github.com/project-zot/zot/blob/main/pkg/regexp/regexp.go <- find uses of this.

TODO: checking if any extras after the specified path are also routed to the same handler.

Left a comment above on the fact that the regex is already done before routing. Doing it once more would add a penalty to every proxied request. If there is a chance that other paths are routed here, a regex would be needed.

rchincha:
> > https://github.com/project-zot/zot/blob/a716be3bbb05072f2cb4096988fc2c1d1c2a8e61/pkg/api/proxy.go#L263-L263
> > 
> > resp, err := httpClient.Do(fwdRequest)
> > ^ make sure the regex check is done on this "url" before forwarding https://github.com/project-zot/zot/blob/main/pkg/regexp/regexp.go <- find uses of this.
> 
> TODO: checking if any extras after the specified path are also routed to the same handler.
> 
> Left a comment above on the fact that the regex is already done before routing. Doing it once more would add a penalty to every proxied request. If there is a chance that other paths are routed here, a regex would be needed.

If the regex is already done, I suppose since clusterProxy is a wrapper around the routes, no need to do it again.

pkg/common/common.go
rchincha: https://github.com/project-zot/zot/tree/main/pkg/test/inject
It is for these kinds of scenarios, we added this facility. Take a look, however, the downside is code becomes somewhat unreadable for an uninformed reader.

vrajashkr:
Regarding the regex, I tried adding some query params to the API call, and it seems to have let it through:
```
curl -v http://localhost:9000/v2/alpine/tags/list?example=test
*   Trying 127.0.0.1:9000...
* Connected to localhost (127.0.0.1) port 9000 (#0)
> GET /v2/alpine/tags/list?example=test HTTP/1.1
> Host: localhost:9000
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Access-Control-Allow-Headers: Authorization,content-type,X-ZOT-API-CLIENT
< Access-Control-Allow-Methods: GET,OPTIONS
< Access-Control-Allow-Origin: *
< Content-Length: 35
< Content-Type: application/json
< Date: Sat, 18 May 2024 14:07:28 GMT
< 
* Connection #0 to host localhost left intact
{"name":"alpine","tags":["latest"]}
```

Logs:
```json
{"level":"info","clusterMember":"127.0.0.1:9001","clusterMemberIndex":"1","module":"http","component":"session","clientIP":"127.0.0.1:51370","method":"GET","path":"/v2/alpine/tags/list?example=test","statusCode":200,"latency":"0s","bodySize":35,"headers":{"Accept":["*/*"],"Accept-Encoding":["gzip"],"User-Agent":["curl/7.81.0"],"X-Zot-Cluster-Hop-Count":["1"]},"goroutine":27,"caller":"zotregistry.dev/zot/pkg/api/session.go:132","time":"2024-05-18T14:07:28.799382217Z","message":"HTTP API"}
```

This is no different from the usual behaviour of the application though.

Any extra path params will fail the query, as expected:
```
curl -v http://localhost:9000/v2/alpine/tags/list/example
*   Trying 127.0.0.1:9000...
* Connected to localhost (127.0.0.1) port 9000 (#0)
> GET /v2/alpine/tags/list/example HTTP/1.1
> Host: localhost:9000
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 404 Not Found
< Content-Type: text/plain; charset=utf-8
< X-Content-Type-Options: nosniff
< Date: Sat, 18 May 2024 14:38:38 GMT
< Content-Length: 19
< 
404 page not found
* Connection #0 to host localhost left intact
```

rchincha (approved):
lgtm

PR #2414, opened 2024-05-06
refactor(http): refactor http client to accept more customisable options

vrajashkr:
**What type of PR is this?**
refactor

N/A

This PR refactors the `CreateHTTPClient` function to take in a more customisable set of options for certificates and TLS. This is needed as we would like to re-use the same for scale out proxy which also needs the same functionality to automatically inherit per-host certificates and mTLS support. It also provides flexibility to allow users to customise the name and path for certificates for the sync feature in future.

Not a bug, but a helpful enhancement for the scale-out feature #2385 
The feature requires use of functionalities such as per-host certificates and mTLS support.

Existing tests should pass and new unit tests added. There should be no functional impact due to the change.

**Automation added to e2e**:
N/A

No. This is a refactor which is expected to maintain the original behaviour.

N/A


vrajashkr:
I hit an error in the earlier run with this message:
```
Failures:

  * zotregistry.dev/zot/pkg/extensions/sync/sync_internal_test.go 
  Line 166:
  Expected: nil
  Actual:   'open ca.crt: no such file or directory'
```

This was due to a mistake I made in `pkg/extensions/sync/httpclient/client.go` where the cert file paths are only set when the CertDir is specified. In the incorrect code, the files were always being set which wasn't correct :)

I've fixed that. Hopefully the tests go through completely this time round.

rchincha (approved):
lgtm

PR #2424, opened 2024-05-26
feat(sync): support for periodic repo sync in scale-out cluster

### LINKED ISSUE #2420: [Feat]: ownership-based sync in scale-out cluster

In the implementation of the scale-out clustering functionality in #2385, once the cluster is up, based on the siphash of a repository, a particular member is the designated "manager" of the repository. All reads and writes to that repository are internally proxied by zot to that member regardless of where the client's request lands.

Keeping this in mind, at this point, enabling sync with the scale out cluster will result in all the members mirroring the same repositories (even though they shouldn't).

Each zot member should:
- Detect when it is in scale-out cluster mode
- Realize that it should only sync the repositories that it is currently managing
- Selectively perform a sync for the repositories that it manages.

There are 2 cases:
1. On-demand sync - here, the list of repositories is known. zot should check whether the repository is managed by the instance and sync it / skip it.
2. Periodic sync - get a list of all repos that are there in the cluster and then sync only the repositories that are managed by that particular instance.

Additionally, all existing functionality such as repo authorization, etc. should work as they do today.

N/A

### Additional context

N/A

vrajashkr:
**What type of PR is this?**
feature

Closes #2420

This commit includes support for periodic repo sync in a scale-out cluster.
Before this commit, all cluster members would sync all the configured repos as the config is shared amongst all the members.

With this change, in periodic sync, the cluster member checks whether it manages the repo. If it does not manage the repo, it will skip the sync.

This commit also includes a unit test to test on-demand sync too, but there are no logic changes for it as it is implicitly handled by the proxying logic.

Covered by Unit Tests

**Automation added to e2e**:
N/A

No

Yes

```release-note
In scale-out cluster mode, when sync is enabled, cluster members will check to see if a repo is managed locally before syncing. If the repo is not managed by the local member, the sync for that repo will be skipped.
```

### DIFF
--- a/pkg/extensions/sync/service.go
+++ b/pkg/extensions/sync/service.go
@@ -13,6 +13,9 @@ import (
 	"github.com/opencontainers/go-digest"
 
 	zerr "zotregistry.dev/zot/errors"
+	"zotregistry.dev/zot/pkg/api/config"
+	"zotregistry.dev/zot/pkg/api/constants"
+	"zotregistry.dev/zot/pkg/cluster"
 	"zotregistry.dev/zot/pkg/common"
 	syncconf "zotregistry.dev/zot/pkg/extensions/config/sync"
 	client "zotregistry.dev/zot/pkg/extensions/sync/httpclient"
@@ -25,6 +28,7 @@ import (
 type BaseService struct {
 	config          syncconf.RegistryConfig
 	credentials     syncconf.CredentialsFile
+	clusterConfig   *config.ClusterConfig
 	remote          Remote
 	destination     Destination
 	retryOptions    *retry.RetryOptions
@@ -40,6 +44,7 @@ type BaseService struct {
 func New(
 	opts syncconf.RegistryConfig,
 	credentialsFilepath string,
+	clusterConfig *config.ClusterConfig,
 	tmpDir string,
 	storeController storage.StoreController,
 	metadb mTypes.MetaDB,
@@ -63,6 +68,10 @@ func New(
 	}
 
 	service.credentials = credentialsFile
+
+	// load the cluster config into the object
+	// can be nil if the user did not configure cluster config
+	service.clusterConfig = clusterConfig
 
 	service.contentManager = NewContentManager(opts.Content, log)
 
@@ -227,6 +236,23 @@ func (service *BaseService) GetNextRepo(lastRepo string) (string, error) {
 		lastRepo = service.getNextRepoFromCatalog(lastRepo)
 		if lastRepo == "" {
 			break
+		}
+
+		if service.clusterConfig != nil {
+			targetIdx, targetMember := cluster.ComputeTargetMember(
+				service.clusterConfig.HashKey, service.clusterConfig.Members, lastRepo)
+
+			// if the target index does not match with the local socket index,
+			// then the local instance is not responsible for syncing the repo and should skip the sync
+			if targetIdx != service.clusterConfig.Proxy.LocalMemberClusterSocketIndex {
+				service.log.Debug().
+					Str(constants.RepositoryLogKey, lastRepo).
+					Str("targetMemberIndex", fmt.Sprintf("%d", targetIdx)).
+					Str("targetMember", targetMember).
+					Msg("skipping sync of repo not managed by local instance")
+
+				continue
+			}
 		}
 
 		matches = service.contentManager.MatchesContent(lastRepo)
--- a/pkg/extensions/sync/sync_test.go
+++ b/pkg/extensions/sync/sync_test.go
@@ -227,6 +227,41 @@ func makeDownstreamServer(
 	return dctlr, destBaseURL, destDir, client
 }
 
+func makeInsecureDownstreamServerFixedPort(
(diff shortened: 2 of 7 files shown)


pkg/extensions/sync/service.go
```diff
@@ -229,12 +238,42 @@ func (service *BaseService) GetNextRepo(lastRepo string) (string, error) {
...
 
 	return lastRepo, nil
 }
 
+// computes the target member using siphash and returns the index and the member
+// siphash was chosen to prevent against hash attacks where an attacker
+// can target all requests to one given instance instead of balancing across the cluster
+// resulting in a Denial-of-Service (DOS).
+// ref: https://en.wikipedia.org/wiki/SipHash
+func computeTargetMember(clusterConfig *config.ClusterConfig, name string) (uint64, string) {
```
vrajashkr, line 268: This logic is common to `proxy.go` as well as `service.go` which are in different packages. However, it is not common enough to be placed in the `common` package.

@rchincha do you know which package might be appropriate to have this shared hash compute logic?

pkg/extensions/sync/sync_test.go
```diff
@@ -227,6 +227,69 @@ func makeDownstreamServer(
 	return dctlr, destBaseURL, destDir, client
 }
 
+func makeDownstreamServerFixedPort(
```
vrajashkr, line 230: This function is a copy of `makeDownstreamServer` above and enhanced to include support for fixed ports (which is needed for scale out cluster) and also support configuring scale out cluster config.

There is some opportunity for code re-use here.

pkg/extensions/sync/sync_test.go
```diff
@@ -729,6 +792,180 @@ func TestOnDemand(t *testing.T) {
...
+		// cluster config for both downstream instances.
+		clusterCfgDownstream1 := config.ClusterConfig{
+			Members: []string{
+				"127.0.0.1:43222",
+				"127.0.0.1:43223",
+			},
+			HashKey: "loremipsumdolors",
+		}
+
+		// cluster config for both downstream instances.
```
vrajashkr, line 834: Comment needs to be improved here.

pkg/extensions/sync/sync_test.go
```diff
@@ -729,6 +792,180 @@ func TestOnDemand(t *testing.T) {
...
+		// Tags List for test CVE image should return 404.
+		// only hit one instance as the request will get proxied anyway.
+		resp, err = dstClient1.R().Get(
+			fmt.Sprintf("%s/v2/%s/tags/list", dctrl1BaseURL, testCveImage),
+		)
+		So(err, ShouldBeNil)
+		So(resp, ShouldNotBeNil)
+		So(resp.StatusCode(), ShouldEqual, http.StatusNotFound)
+
+		// Should successfully sync test image and when trying to load manifest.
```
vrajashkr, line 891: minor typo in this comment.

pkg/extensions/sync/sync_test.go
```diff
@@ -729,6 +792,180 @@ func TestOnDemand(t *testing.T) {
...
+		// only hit one instance as the request will get proxied anyway.
+		resp, err = dstClient1.R().Get(
+			fmt.Sprintf("%s/v2/%s/manifests/%s", dctrl1BaseURL, testImage, testImageTag),
+		)
+		So(err, ShouldBeNil)
+		So(resp.StatusCode(), ShouldEqual, http.StatusOK)
+
+		// Tags list for test image should return data after the sync
```
vrajashkr, line 899: Ensure consistency in case for comments

pkg/extensions/sync/sync_test.go
```diff
@@ -729,6 +792,180 @@ func TestOnDemand(t *testing.T) {
...
+
+		var initialTags TagsList
+		err = json.Unmarshal(resp.Body(), &initialTags)
+		So(err, ShouldBeNil)
+		So(initialTags, ShouldEqual, TagsList{
+			Name: testImage,
+			Tags: []string{testImageTag},
+		})
+
+		// Should successfully sync test vulnerable image and when trying to check manifest
```
vrajashkr, line 917: minor typo in comment

pkg/extensions/sync/service.go
rchincha: This is tricky since sync is an extension which can be optionally compiled. So maybe put it in cluster pkg and make it available to sync?

pkg/extensions/sync/sync_test.go
rchincha: If this is cluster related put this also in cluster pkg

pkg/extensions/sync/service.go
vrajashkr: I moved the code to a new cluster package and generalized the parameters so that it can be used in both `api` and `sync` packages

pkg/extensions/sync/sync_test.go
vrajashkr: This is not exactly cluster specific. It just accepts a fixed port number (the existing one finds an available port) and accepts a cluster config to assign to the config struct.

There is an opportunity to refactor this, but I think I'd like to take it up later on.

PR #2533, opened 2024-07-09
fix(storage): handle dedupe disabled in GetAllDedupeReposCandidates()

peusebiu:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -879,13 +879,11 @@ func canMount(userAc *reqCtx.UserAccessControl, imgStore storageTypes.ImageStore
 ) (bool, error) {
 	canMount := true
 
-	// authz enabled
 	if userAc != nil {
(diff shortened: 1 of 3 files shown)


pkg/api/routes.go
```diff
@@ -943,9 +941,12 @@ func (rh *RouteHandler) CheckBlob(response http.ResponseWriter, request *http.Re
 		return
 	}
 
-	userCanMount, err := canMount(userAc, imgStore, digest)
-	if err != nil {
-		rh.c.Log.Error().Err(err).Msg("unexpected error")
+	userCanMount := true
+	if rh.c.Config.HTTP.AccessControl != nil {
```
andaaron, line 945: Why not use the new config method?

andaaron (approved):

Issue #2526, opened 2024-07-04
[Bug]: `failed to get state: securecookie: the value is not valid`


AndersBennedsgaard:
### zot version

v2.1.0-rc2 (default in the Helm chart)

### Describe the bug

When I set up OIDC auth with Dex yesterday, it was working fine. But when I tried to access our Zot instance this morning, I was met with `failed to get state: securecookie: the value is not valid`. I have a hard time figuring out where the issue comes from, since the only relevant log I can find is:

```text
{"level":"info","module":"http","component":"session","clientIP":"<client-ip>:57554","method":"GET","path":"/zot/auth/callback/oidc?code=<code>&state=<state>","statusCode":401,"latency":"0s","bodySize":58,"headers":{"Accept":["text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8"],"Accept-Encoding":["gzip, deflate, br, zstd"],"Accept-Language":["en-US,en;q=0.5"],"Cookie":["statecookie=<state-cookie>; state=<state>"],"Priority":["u=1"],"Referer":["<external-url>/"],"Sec-Fetch-Dest":["document"],"Sec-Fetch-Mode":["navigate"],"Sec-Fetch-Site":["cross-site"],"Sec-Fetch-User":["?1"],"Te":["trailers"],"Upgrade-Insecure-Requests":["1"],"User-Agent":["Mozilla/5.0 (X11; Linux x86_64; rv:127.0) Gecko/20100101 Firefox/127.0"],"X-Envoy-Expected-Rq-Timeout-Ms":["15000"],"X-Envoy-External-Address":["<ip>"],"X-Forwarded-For":["<ip>"],"X-Forwarded-Proto":["https"],"X-Request-Id":["58079f48-8fc9-4b15-b122-6d89873acba5"],"X-Request-Start":["t=1720077531.100"]},"goroutine":175,"caller":"zotregistry.dev/zot/pkg/api/session.go:132","time":"2024-07-04T07:18:51.101298721Z","message":"HTTP API"}
```

It seems that the cookie key is generated here:

https://github.com/project-zot/zot/blob/002ff05f6e5ab3870aca75e616e8a4ed9b4a5c68/pkg/api/authn.go#L684

The issue might also come from the fact that a random key is generated on startup, which would invalidate the cookies created from other instances of Zot (I run multiple replicas of Zot on Kubernetes, so the different instances have different cookies, and they can restart at random points in time):

[...]

PR #2536, opened 2024-07-11
fix(authn): make hashing/encryption keys used to secure cookies

peusebiu:
If they are not configured zot will generate a random hashing key at startup, invalidating all cookies if zot is restarted. closes: #2526

**What type of PR is this?**
bug

restarting zot invalidates currently active sessions.

https://github.com/project-zot/zot/issues/2526

**Automation added to e2e**:

no

```release-note
modified configuration file, client can specify hashing and encrypt keys to secure web session cookies.
```

### DIFF
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -22,7 +22,6 @@ import (
 	"github.com/google/go-github/v52/github"
 	"github.com/google/uuid"
 	"github.com/gorilla/mux"
-	"github.com/gorilla/securecookie"
 	"github.com/gorilla/sessions"
 	godigest "github.com/opencontainers/go-digest"
 	"github.com/zitadel/oidc/v3/pkg/client/rp"
@@ -334,10 +333,12 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
 
 		for provider := range ctlr.Config.HTTP.Auth.OpenID.Providers {
 			if config.IsOpenIDSupported(provider) {
-				rp := NewRelyingPartyOIDC(context.TODO(), ctlr.Config, provider, ctlr.Log)
+				rp := NewRelyingPartyOIDC(context.TODO(), ctlr.Config, provider, ctlr.Config.HTTP.Auth.SessionHashKey,
+					ctlr.Config.HTTP.Auth.SessionEncryptKey, ctlr.Log)
 				ctlr.RelyingParties[provider] = rp
 			} else if config.IsOauth2Supported(provider) {
-				rp := NewRelyingPartyGithub(ctlr.Config, provider, ctlr.Log)
+				rp := NewRelyingPartyGithub(ctlr.Config, provider, ctlr.Config.HTTP.Auth.SessionHashKey,
+					ctlr.Config.HTTP.Auth.SessionEncryptKey, ctlr.Log)
 				ctlr.RelyingParties[provider] = rp
 			}
 		}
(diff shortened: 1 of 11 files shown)


andaaron (dismissed):

pkg/api/authn.go
```diff
@@ -681,9 +681,25 @@ func getRelyingPartyArgs(cfg *config.Config, provider string, log log.Logger) (
...
+	var hashKey []byte
 
-	cookieHandler := httphelper.NewCookieHandler(key, key, httphelper.WithMaxAge(relyingPartyCookieMaxAge))
+	var blockKey []byte
+
+	envHashKey := os.Getenv("COOKIE_HASH_KEY")
+	if envHashKey != "" {
+		hashKey = []byte(envHashKey)
+	} else {
+		hashKey = securecookie.GenerateRandomKey(64) //nolint: gomnd
```
AndersBennedsgaard, line 692: Could it make sense to log a warning about random keys not being recommended?

AndersBennedsgaard:
I still get the error explained in https://github.com/project-zot/zot/issues/2526 with these changes, however, it only happens when I run multiple replicas of Zot. I tried to enable some extra logging by changing

https://github.com/peusebiu/zot/blob/c6688abf0bf7edc59db02fca3bea8994b0383dba/pkg/api/authn.go#L851

to

```go
stateCookie, err := ctlr.CookieStore.Get(r, "statecookie")                                                          
if err != nil {                                                                                                     
  ctlr.Log.Error().Err(err).Str("component", "openID").Msg("failed to get 'statecookie' from request")            
}
```

and I started to get

```json
{"level":"error","error":"securecookie: the value is not valid","component":"openID","goroutine":296,"caller":"zotregistry.dev/zot/pkg/api/authn.go:856","time":"2024-07-12T09:34:55.128562134Z","message":"failed to get 'statecookie' from request"}
```

I guess that Zot can't handle cookies from different running instances, as it saves some things in memory which the other instance does not have access to

peusebiu:
Hello @AndersBennedsgaard I have pushed a new commit, can you try your test again please? 

thank you!

rchincha:
@peusebiu pls fix the linter issues and rebase also.

rchincha:
@AndersBennedsgaard does this PR work for you?

peusebiu:
@rchincha updated, now it's using a separate file for keys.

AndersBennedsgaard:
@rchincha and @peusebiu sorry for not getting back to you before (I was on vacation) but I have now tested these changes and it seems to work perfectly with multiple replicas :partying_face: 

My only issue I've found is that while `encryptKey` is [supposed to be optional](https://github.com/peusebiu/zot/blob/0499fc43bb9c792e57faa3a463ebd0fe1ccd8042/examples/README.md?plain=1#L399), Zot fails to start up if you don't set it. I guess you would need a `mapstructure:",omitempty"` on the `EncryptKey` key?

peusebiu:
> @rchincha and @peusebiu sorry for not getting back to you before (I was on vacation) but I have now tested these changes and it seems to work perfectly with multiple replicas 🥳
> 
> My only issue I've found is that while `encryptKey` is [supposed to be optional](https://github.com/peusebiu/zot/blob/0499fc43bb9c792e57faa3a463ebd0fe1ccd8042/examples/README.md?plain=1#L399), Zot fails to start up if you don't set it. I guess you would need a `mapstructure:",omitempty"` on the `EncryptKey` key?

Fixed it, sorry for that, can you test again? :D

andaaron (approved):
lgtm

AndersBennedsgaard (approved):
It works perfectly now, thanks @peusebiu :partying_face:

Issue #2527, opened 2024-07-04
[Bug]: Sync extension panic


AndersBennedsgaard:
### zot version

v2.1.0-rc2

### Describe the bug

I have set up sync to ghcr.io, but whenever I run something like `helm pull oci://<zot-url>/ghcr.io/grafana/helm-charts/grafana --version 8.0.2` the sync extension panics with:

```text
[...]

PR #2537, opened 2024-07-12
Fix sync extension logging

AndersBennedsgaard:
**What type of PR is this?**

bug

closes https://github.com/project-zot/zot/issues/2527

It does two very small things.
First: it ensures that Zot doesn't panic with a nil pointer dereference on incorrectly configured local storage in the sync extension, and instead returns an error when `localImageStore` is nil.
Second: (and arguably more important) it enables logging for the sync extension image store.

I have manually tested this on my cluster, and can confirm that logging accurately shows my configuration error in #2527 (I had configured Zot to use a read-only filesystem, which obviously doesn't work since it needs permissions to write to the download directory used in the sync extension) and the panic instead just returns an error which is logged to stdout.

**Automation added to e2e**:

No.

```release-note

```

### DIFF
--- a/pkg/extensions/sync/destination.go
+++ b/pkg/extensions/sync/destination.go
@@ -92,7 +92,7 @@ func (registry *DestinationRegistry) GetImageReference(repo, reference string) (
 func (registry *DestinationRegistry) CommitImage(imageReference types.ImageReference, repo, reference string) error {
 	imageStore := registry.storeController.GetImageStore(repo)
 
-	tempImageStore := getImageStoreFromImageReference(imageReference, repo, reference)
+	tempImageStore := getImageStoreFromImageReference(imageReference, repo, reference, registry.log)
 
 	defer os.RemoveAll(tempImageStore.RootDir())
 
@@ -282,11 +282,11 @@ func (registry *DestinationRegistry) copyBlob(repo string, blobDigest digest.Dig
 }
 
 // use only with local imageReferences.
-func getImageStoreFromImageReference(imageReference types.ImageReference, repo, reference string,
+func getImageStoreFromImageReference(imageReference types.ImageReference, repo, reference string, log log.Logger,
 ) storageTypes.ImageStore {
 	tmpRootDir := getTempRootDirFromImageReference(imageReference, repo, reference)
 
-	return getImageStore(tmpRootDir)
+	return getImageStore(tmpRootDir, log)
 }
 
 func getTempRootDirFromImageReference(imageReference types.ImageReference, repo, reference string) string {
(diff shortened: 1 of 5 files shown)


pkg/extensions/sync/oci_layout.go
```diff
@@ -40,6 +40,9 @@ func (oci OciLayoutStorageImpl) GetContext() *types.SystemContext {
 
 func (oci OciLayoutStorageImpl) GetImageReference(repo string, reference string) (types.ImageReference, error) {
 	localImageStore := oci.storeController.GetImageStore(repo)
+	if localImageStore == nil {
+		return nil, fmt.Errorf("no image store found for repo %s", repo)
```
andaaron, line 44: Please define the new error in https://github.com/project-zot/zot/blob/main/errors/errors.go

AndersBennedsgaard: I've added a new error:

https://github.com/project-zot/zot/pull/2537/files#diff-d04adb084335cede75155605abdcc0bef92c45a740b274f985f378e955c81a13R90

I was not sure on the wording, so I reused another similarly sounding error. Feel free to suggest another wording.

I also added a simple test, since the CI was complaining about coverage :zany_face:

pkg/extensions/sync/destination.go
```diff
@@ -92,7 +92,7 @@ func (registry *DestinationRegistry) GetImageReference(repo, reference string) (
 func (registry *DestinationRegistry) CommitImage(imageReference types.ImageReference, repo, reference string) error {
 	imageStore := registry.storeController.GetImageStore(repo)
 
-	tempImageStore := getImageStoreFromImageReference(imageReference, repo, reference)
+	tempImageStore := getImageStoreFromImageReference(imageReference, repo, reference, registry.log)
```
rchincha, line 95: passing a logger around is useful

rchincha:
@AndersBennedsgaard do you want to squash your commits in the PR into a single commit?

pkg/extensions/sync/destination.go
AndersBennedsgaard: Yes. But more importantly, loggers initialized as `log.Logger{}` does not ever actually output any logs. So the two loggers used here 

https://github.com/project-zot/zot/blob/e5eacaa082c9fa9f85fb4548841e8dc9f1918590/pkg/extensions/sync/destination.go#L304-L308

did never print anything to stdout. That means that logs such as https://github.com/project-zot/zot/blob/e5eacaa082c9fa9f85fb4548841e8dc9f1918590/pkg/storage/imagestore/imagestore.go#L73 would never fire

AndersBennedsgaard:
I don't know why the CI is failing. It's complaining about code coverage  and something about licenses, but I don't get the issue..

> @AndersBennedsgaard do you want to squash your commits in the PR into a single commit?

@rchincha I could do that, but wouldn't it be easier just to do a squash merge when it gets merged to main?

peusebiu (approved):
lgtm!

PR #2554, opened 2024-07-19
feat(oci): support specifying the manifest digest when pushing it

andaaron:
See https://github.com/opencontainers/distribution-spec/issues/494 and reladed work


andaaron:
Still waiting on the dist spec update.

andaaron:
Should no longer be needed given https://github.com/opencontainers/distribution-spec/issues/494#issuecomment-4150241382
We have added support for the multi-tag push in latest zot.

PR #2532, opened 2024-07-08
build(deps): bump all dependencies

jkroepke:
# Notes
- Non AWS S3 backend now require `forcepathstyle=true` (https://github.com/distribution/distribution/pull/4291)

PR #2556, opened 2024-07-21
ci(deps): upgrade golangci-lint

jkroepke:
**What type of PR is this?**

cleanup

Cut off from #2532 

I had issues with the old golangci-lint version, see https://github.com/golangci/golangci-lint/issues/3718

**Automation added to e2e**:

```release-note

```


golangcilint.yaml
```diff
@@ -1,11 +1,38 @@
...
+    - inamedparam
+    - protogetter
+    - promlinter
+    - perfsprint
+    - testifylint
+    - wsl
+    - zerologlint
+    - canonicalheader
+    - contextcheck
+    - mnd
```
andaaron, line 32: Why is this disabled? It has specific configuration below.

golangcilint.yaml
```diff
@@ -1,11 +1,38 @@
...
+    - rowserrcheck
+    - sqlclosecheck
+    - revive
+    - musttag
+    - depguard
+    - execinquery
+    - gomnd
+    - inamedparam
+    - protogetter
+    - promlinter
```
andaaron, line 25: Why is this disabled?

golangcilint.yaml
```diff
@@ -1,11 +1,38 @@
...
+    - exhaustruct
+    - rowserrcheck
+    - sqlclosecheck
+    - revive
+    - musttag
+    - depguard
+    - execinquery
+    - gomnd
+    - inamedparam
+    - protogetter
```
andaaron, line 24: Why is this disabled?

golangcilint.yaml
```diff
@@ -1,11 +1,38 @@
...
+    - promlinter
+    - perfsprint
+    - testifylint
```
andaaron, line 27: I think we can simply remove the dependency on github.com/stretchr/testify/assert instead. Seems to be used in a single case

golangcilint.yaml
```diff
@@ -1,11 +1,38 @@
...
+    - wsl
```
andaaron, line 28: This is disabled, but its config is customized below.

golangcilint.yaml
```diff
@@ -1,11 +1,38 @@
...
+    - zerologlint
```
andaaron, line 29: What issues does this one find, why is it disabled?

golangcilint.yaml
```diff
@@ -1,11 +1,38 @@
...
+    - canonicalheader
```
andaaron, line 30: Why disable it?

golangcilint.yaml
```diff
@@ -1,11 +1,38 @@
...
+    - contextcheck
```
andaaron, line 31: This was not disabled before, please reenable

golangcilint.yaml
```diff
@@ -1,11 +1,38 @@
...
+    - mnd
+    - nilnil
```
andaaron, line 33: This would be interesting to run, does it have a lot of fails?

golangcilint.yaml
```diff
@@ -1,11 +1,38 @@
...
+    - intrange
+    - copyloopvar
```
andaaron, line 35: This would be interesting to run. Does it have a lot of failures?

andaaron (commented):
Thank you. @jkroepke for the PR. I do have some questions on the disabled linters.

jkroepke:
I enabled the mentioned linters. The results should be visible in the UI.

lint settings are very moody. From my point of view, the motivation to go over all lints, fix the lint issue or ignore should be more decided by an zot maintainer.

Some linter are supporting auto-fix which introduced a lot of white spaces changes and resulting into merge conflicts again.

There are 2 possibilities to move forward:

- 1 big PR which covers all maintainer needs.
- 1 big PR which does to golang-ci update with disabled lint settings and multiple small separate PR for each dedicated disabled linter.

jkroepke:
> contextcheck - I think we should use nolint in these specific cases, I don't understand why we would use a context in the cases it highlight.

One use-case is to pass the context from end-to-end. For example, if a request get cancels, it would also cancel all outstanding operations, too.


For perfsprint and wsl, there is an auto fix command. Do not do it by hand.

andaaron:
Why are the imports changed? Looks like these settings are not respected?
```
  gci:
    sections:
      - standard
      - default
      - prefix(zotregistry.dev/zot)     
```

The default and prefix imports are not separated in the last update.

jkroepke:
Make me sad that test failing here because of running lint fixes.

Are you interest to continue? Going through all lint issues and resolve creates a hugh amount of line diff with endless merge conflicts.

I still recommend disable some linter and resolve the issues in a upcoming PR.

rchincha:
> Are you interest to continue? Going through all lint issues and resolve creates a hugh amount of line diff with endless merge conflicts.

@jkroepke fully familiar with this situation 😞 having done this several times. And your effort here is much appreciated. Let me take a look at the list a little more carefully so we can find a good compromise.

rchincha:
> Thank you @jkroepke for looking into this.
> 
> @ram, based on the latest lint output I think we can:
> 


>     * promlinter - add to the `ignore` list, the impact of the renames is not clear to me

Add to ignored list.

> 
>     * protogetter - we should update the code to `fix` the issues, there are some valuable checks for nil pointers in the getters we don't use
>

Add to ignored list.
 
>     * canonicalheader - we should update the code to `fix` the issues, `Docker-Content-Digest` should be fixed, I am not sure of the implications of it being capitalized wrong right now.
> 

This actually nice to have. 'Docker-Content-Digest' is wrong how?

>     * contextcheck - I think we should use `nolint` in these specific cases, I don't understand why we would use a context in the cases it highlight.

Add it to ignored for now. But this is another linter we need to revisit, since we really want to be clear about propagating top-level context cancel.

> 
>     * copyloopvar - add to the `ignore` list or update the code to `fix`, theoretically the code would perform better if we don't copy the variables.

Add to ignored.

> 
>     * mnd - update the `nolint` messages to replace `gomnd` with `mnd`, I also think we can use nolint in the new places it found
> 

gomnd looks deprecated, so we need to handle this.


>     * nilnil - use `nolint`, I think we are justified to keep the code logic as is in our use case for scheduler

Hmm..add to ignored list. We must revisit though.

>    * perfsprint - I think we can update the code to `fix` these, what do you think @ram? Theoretically there should be some performance improvement.
> 

Agreed, these should be fixed.

>     * wsl - we should update the code to `fix` these

Agreed.

> 
>     * zerologlint - this is a false positive, we should use `nolint`

Ok.

andaaron:
> This actually nice to have. 'Docker-Content-Digest' is wrong how?

 'Docker-Content-Digest' is not wrong. In our code it is spelled differently, not capitalized.

andaaron:
@rchincha, I started https://github.com/project-zot/zot/pull/2571 for some of the fixes not covered in this PR.

jkroepke:
wsl issues are crazy high, since the test files seems like copy/paste and the same issues appeared over and over again.

But it looks good for now. Only the benchmark tests reports that something is way slower. But I had the same issue, in #2532  too

@andaaron Sorry I do not see your PR. But feel free to merge and I'm going to the hell trough conflicts. It's okay

andaaron:
> @andaaron Sorry I do not see your PR. But feel free to merge and I'm going to the hell trough conflicts. It's okay

My PR was merged.

andaaron:
The `Clustering test / Stateless zot with shared reliable storage (pull_request)` job should be fixed now.

jkroepke:
The golangci-lint check is green 🎉

To include #2571, I do a merge from main.

andaaron (approved):
Looks good!

rchincha:
@jkroepke thanks for getting this done. These types of tasks are very painful but clear the inevitable accumulation of tech debt!

PR #2558, opened 2024-07-22
fix(sync): search for certificates in sync certDir

eusebiu-constantin-petu-dbk:
instead of expecting certain filenames

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


andaaron (approved):
I am a little sceptical about guessing file purpose by extension, but otherwise looks good.

Why is this better than documenting the required file names?

rchincha:
Instead of looking at the extension, maybe we first assume PEM format, try and load and if that fails try DER format, if that fails, then return error.

andaaron:
@eusebiu-constantin-petu-dbk let's close this?

andaaron:
Let's close this for now. The original issue was fixed in https://github.com/project-zot/zot/pull/2747

PR #2571, opened 2024-07-24
fix: various lint issues

andaaron:
Originally found by @jkroepke in: https://github.com/project-zot/zot/actions/runs/10056774230/job/27796324933?pr=2556 And https://github.com/project-zot/zot/pull/2556 in general

This commit covers:
- (canonicalheader) capitalization of Docker-Content-Digest header
- (protogetter) the proto getters were not used, they check for nil pointers, we should switch to using them
- (zerologlint) fix the false positive


rchincha (approved):
lgtm

Issue #2434, opened 2024-06-01
[Feat]: Scale-out cluster support for independent per-instance storage deployments


vrajashkr:
### Is your feature request related to a problem? Please describe.

The current implementation of scale-out cluster is supported on shared-storage deployments - many instances and one reliable shared storage such as S3.

Currently, some REST APIs will not work for deployments where the image storage for each instance is isolated (for example, one zot instance per VM with independent disk storage).

### Describe the solution you'd like

In a scale-out cluster deployment using isolated per-instance storage, APIs such as `/v2/_catalog` as well as GraphQL endpoints will need to fan-out requests to all the members and aggregate the received data before returning results.

### Describe alternatives you've considered

N/A

vrajashkr:
`/v2/_catalog` is easier to handle as the logic is entirely implemented using data structs.
For this, we can simply detect a scale out cluster in non-shared storage mode and then fan-out + aggregate.

However, the GQL APIs are handled via a GQL server handling auto generated GQL - the implementation of which cannot be changed.
Additionally, there is a need to support pagination.

```golang
extRouter.Methods(allowedMethods...).
		Handler(gqlHandler.NewDefaultServer(gql_generated.NewExecutableSchema(resConfig)))
```

One approach might be to look into the Resolver code (resolver.go) and modify the data read to include proxied results before the response is taken for pagination.

For example:
```
func (cveinfo BaseCveInfo) GetImageListForCVE(ctx context.Context, repo, cveID string) ([]cvemodel.TagInfo, error)
```
This function must ask all other nodes for images having this CVE before returning the TagInfo objects.

That logic could be introduced as a wrapper on top of these functions that will do the following:
1. Call the function locally
2. Call the function for all other members (this can be done using GQL potentially to promote re-use)
3. Return the result to client

To facilitate this, the GraphQL requests would need to be proxied. Since the request data does not flow to the resolver, there would be a need to bring that logic to the resolver so `GetImageListForCVE` knows to call `GetImageListForCVE` on all other hosts via the GraphQL API. Each zot instance will become a GQL client as well.

rchincha:
@vrajashkr do you want to start with paginated result (aggregate) of paginated results (fanout)? The second part is what needs to redesigned? Maybe we simply add a pagination=off param.

rchincha:
Separately, keep an eye on sync as a means to "heal" after a cluster resize.

vrajashkr:
There are 2 major items to handle:
[...]

PR #2588, opened 2024-08-04
feat/scale-out: support for proxying GQL queries

vrajashkr:
**What type of PR is this?**
feature

Towards #2434

Previously, only dist-spec APIs were supported for scale-out as in a shared storage environment, the metadata was shared and any instance could correctly respond to the GQL queries as all the data is available.

In a local scale-out cluster deployment, the metadata store, in addition to the file storage is isolated to each member in the cluster. Due to this, there is a need to proxy the GQL queries as well for UI and client requests to work as expected.

This change introduces a new GQL proxy + a handler for the GlobalSearch request that proxies the request to all the members and collects them for response to the client.
Support for other GQL queries is pending.

Unit Tests added.

No, there shouldn't be any impact to upgrades and downgrades.

```release-note
A zot scale-out cluster deployed without using shared storage is now supported. Client queries from the UI and other clients will be proxied amongst the cluster members to fetch data for the request.
```

### DIFF
--- a/pkg/api/proxy.go
+++ b/pkg/api/proxy.go
@@ -1,8 +1,6 @@
 package api
 
 import (
-	"bytes"
-	"context"
 	"fmt"
 	"io"
 	"net"
@@ -13,6 +11,7 @@ import (
 	"zotregistry.dev/zot/pkg/api/constants"
 	"zotregistry.dev/zot/pkg/cluster"
 	"zotregistry.dev/zot/pkg/common"
+	"zotregistry.dev/zot/pkg/proxy"
 )
 
 // ClusterProxy wraps an http.HandlerFunc which requires proxying between zot instances to ensure
@@ -68,7 +67,7 @@ func ClusterProxy(ctrlr *Controller) func(http.HandlerFunc) http.HandlerFunc {
 
 			logger.Debug().Str(constants.RepositoryLogKey, name).Msg("proxying the request")
 
-			proxyResponse, err := proxyHTTPRequest(request.Context(), request, targetMember, ctrlr)
+			proxyResponse, err := proxy.ProxyHTTPRequest(request.Context(), request, targetMember, ctrlr.Config)
 			if err != nil {
 				logger.Error().Err(err).Str(constants.RepositoryLogKey, name).Msg("failed to proxy the request")
 				http.Error(response, err.Error(), http.StatusInternalServerError)
@@ -77,7 +76,7 @@ func ClusterProxy(ctrlr *Controller) func(http.HandlerFunc) http.HandlerFunc {
 			}
 			defer proxyResponse.Body.Close()
 
-			copyHeader(response.Header(), proxyResponse.Header)
+			proxy.CopyHeader(response.Header(), proxyResponse.Header)
 			response.WriteHeader(proxyResponse.StatusCode)
 			_, _ = io.Copy(response, proxyResponse.Body)
 		})
@@ -111,101 +110,6 @@ func getTargetMemberServerSockets(targetMemberSocket string) ([]string, error) {
 	}
 
 	return targetSockets, nil
-}
-
-// proxy the request to the target member and return a pointer to the response or an error.
-func proxyHTTPRequest(ctx context.Context, req *http.Request,
-	targetMember string, ctrlr *Controller,
-) (*http.Response, error) {
-	cloneURL := *req.URL
-
-	proxyQueryScheme := "http"
-	if ctrlr.Config.HTTP.TLS != nil {
-		proxyQueryScheme = "https"
-	}
-
-	cloneURL.Scheme = proxyQueryScheme
-	cloneURL.Host = targetMember
-
-	clonedBody := cloneRequestBody(req)
-
-	fwdRequest, err := http.NewRequestWithContext(ctx, req.Method, cloneURL.String(), clonedBody)
-	if err != nil {
-		return nil, err
-	}
-
-	copyHeader(fwdRequest.Header, req.Header)
-
-	// always set hop count to 1 for now.
-	// the handler wrapper above will terminate the process if it sees a request that
-	// already has a hop count but is due for proxying.
-	fwdRequest.Header.Set(constants.ScaleOutHopCountHeader, "1")
-
-	clientOpts := common.HTTPClientOptions{
-		TLSEnabled: ctrlr.Config.HTTP.TLS != nil,
-		VerifyTLS:  ctrlr.Config.HTTP.TLS != nil, // for now, always verify TLS when TLS mode is enabled
-		Host:       targetMember,
-	}
-
-	tlsConfig := ctrlr.Config.Cluster.TLS
-	if tlsConfig != nil {
-		clientOpts.CertOptions.ClientCertFile = tlsConfig.Cert
-		clientOpts.CertOptions.ClientKeyFile = tlsConfig.Key
-		clientOpts.CertOptions.RootCaCertFile = tlsConfig.CACert
-	}
-
-	httpClient, err := common.CreateHTTPClient(&clientOpts)
-	if err != nil {
-		return nil, err
-	}
-
-	resp, err := httpClient.Do(fwdRequest)
-	if err != nil {
-		return nil, err
-	}
-
-	var clonedRespBody bytes.Buffer
-
-	// copy out the contents into a new buffer as the response body
-	// stream should be closed to get all the data out.
-	_, _ = io.Copy(&clonedRespBody, resp.Body)
-	resp.Body.Close()
-
-	// after closing the original body, substitute it with a new reader
-	// using the buffer that was just created.
-	// this buffer should be closed later by the consumer of the response.
-	resp.Body = io.NopCloser(bytes.NewReader(clonedRespBody.Bytes()))
-
-	return resp, nil
-}
-
-func cloneRequestBody(src *http.Request) io.Reader {
-	var bCloneForOriginal, bCloneForCopy bytes.Buffer
-	multiWriter := io.MultiWriter(&bCloneForOriginal, &bCloneForCopy)
-	numBytesCopied, _ := io.Copy(multiWriter, src.Body)
-
-	// if the body is a type of io.NopCloser and length is 0,
-	// the Content-Length header is not sent in the proxied request.
-	// explicitly returning http.NoBody allows the implementation
-	// to set the header.
-	// ref: https://github.com/golang/go/issues/34295
-	if numBytesCopied == 0 {
-		src.Body = http.NoBody
-
-		return http.NoBody
-	}
-
-	src.Body = io.NopCloser(&bCloneForOriginal)
-
-	return bytes.NewReader(bCloneForCopy.Bytes())
-}
-
-func copyHeader(dst, src http.Header) {
-	for k, vv := range src {
-		for _, v := range vv {
-			dst.Add(k, v)
-		}
-	}
 }
 
 // identifies and returns the cluster socket and index.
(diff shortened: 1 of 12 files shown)


vrajashkr:
There's still quite a bit of work to be done on this change. Sharing an early draft for review on the approach and handling.

pkg/api/proxy.go
```diff
@@ -13,6 +11,7 @@ import (
 	"zotregistry.dev/zot/pkg/api/constants"
 	"zotregistry.dev/zot/pkg/cluster"
 	"zotregistry.dev/zot/pkg/common"
+	"zotregistry.dev/zot/pkg/proxy"
```
rchincha, line 14: What is this split between pkg/api/proxy.go and pkg/proxy/proxy.go?

pkg/common/model.go
```diff
@@ -15,16 +15,16 @@ type RepoInfo struct {
...
-	LastUpdated   time.Time    `json:"lastUpdated"`
-	Size          string       `json:"size"`
-	Platforms     []Platform   `json:"platforms"`
-	Vendors       []string     `json:"vendors"`
-	IsStarred     bool         `json:"isStarred"`
-	IsBookmarked  bool         `json:"isBookmarked"`
-	StarCount     int          `json:"starCount"`
-	DownloadCount int          `json:"downloadCount"`
-	NewestImage   ImageSummary `json:"newestImage"`
+	Name          string       `json:"Name"`          //nolint:tagliatelle // graphQL schema
```
rchincha, line 18: Why was this change needed?

pkg/extensions/search/gql_proxy/global_search_handler.go
```diff
@@ -0,0 +1,118 @@
...
+			log.Error().Str(LoggerFieldOperation, HandlerOperation).
+				Str(LoggerFieldTargetMember, targetMember).
+				Err(err).Msg("failed to proxy request")
+			http.Error(response, "failed to process GQL request", http.StatusInternalServerError)
+
+			return
+		}
+
+		proxyBody, err := io.ReadAll(proxyResponse.Body)
+		proxyResponse.Body.Close()
```
rchincha, line 43: close it right here? or defer this?

pkg/extensions/search/gql_proxy/global_search_handler.go
```diff
@@ -0,0 +1,118 @@
...
+				Str("responseBody", string(proxyBody)).
+				Err(err).
+				Msg("failed to unmarshal received response body")
+			http.Error(response, "failed to process GQL request", http.StatusInternalServerError)
+
+			return
+		}
+
+		// aggregate errors
+		collatedResult.Errors = append(collatedResult.Errors, proxyRespData.Errors...)
```
rchincha, line 69: So we now have a situation where we may a good result mixed with errors. This could be problematic/confusing.

rchincha:
Overall the patch is not super-complicated.

pkg/common/model.go
```diff
@@ -214,14 +214,14 @@ type GlobalSearchResultResp struct {
 }
 
 type GlobalSearchResult struct {
```
andaaron, line 216: Don't these renames mean we break backwards compatibility with older zots? Do we care about that?

pkg/api/proxy.go
vrajashkr: During development, there was a circular import for api and the new gql proxy in the extensions. This led me to break up the generic proxy logic into its own package and call it from the api package as well as from the gql proxy package.

That said, I do agree that the file naming could potentially be better.

pkg/common/model.go
vrajashkr: Previously, the GQL server was handling serialization of the data for responding to the client. Now, since the proxy handler is responding on behalf of the GQL server (after proxying), the standard Golang serialization takes place.

GQL starts keys with an uppercase, but the standard struct annotations don't encourage the same - hence the `nolint`.

Example:
```
{
    "errors": [
        {
            "message": "unable to run vulnerability scan on tag v0.0.19.231225-squashfs in repo machine/bootkit/rootfs: error: image 'machine/bootkit/rootfs@sha256:9efb1b9dd349e3fc5fa2dd658354a395a28ac7e391c85a72b39384da7a5ec7a1' scanning is not supported for given image media type",
            "path": [
                "GlobalSearch"
            ]
        },
        {
            "message": "unable to run vulnerability scan in repo machine/bootkit/rootfs: manifest digest: sha256:9efb1b9dd349e3fc5fa2dd658354a395a28ac7e391c85a72b39384da7a5ec7a1, error: image 'machine/bootkit/rootfs@sha256:9efb1b9dd349e3fc5fa2dd658354a395a28ac7e391c85a72b39384da7a5ec7a1' scanning is not supported for given image media type",
            "path": [
                "GlobalSearch"
            ]
        }
    ],
    "data": {
        "GlobalSearch": {
            "Page": {
                "TotalCount": 37,
                "ItemCount": 3
            },
            "Repos": [
                {
                    "Name": "machine/bootkit/rootfs",
                    "LastUpdated": "2023-12-25T15:31:59.110429376Z",
                    "Size": "319472431",
                    "Platforms": [
                        {
                            "Os": "linux",
                            "Arch": "amd64"
                        }
                    ],
                    "IsStarred": false,
                    "IsBookmarked": false,
                    "NewestImage": {
                        "Tag": "v0.0.19.231225-squashfs",
                        "Vulnerabilities": {
                            "MaxSeverity": "",
                            "Count": 0
                        },
                        "Description": "A minimal bootable root filesystem",
                        "IsSigned": false,
                        "SignatureInfo": [],
                        "Licenses": "GPLv2 and others",
                        "Vendor": "project-machine",
                        "Labels": ""
                    },
                    "StarCount": 0,
                    "DownloadCount": 5285
                },
                {
                    "Name": "c3/ubuntu/base-amd64",
                    "LastUpdated": "2024-03-01T00:46:16.186838886Z",
                    "Size": "273201849",
                    "Platforms": [
                        {
                            "Os": "linux",
                            "Arch": "amd64"
                        }
                    ],
                    "IsStarred": false,
                    "IsBookmarked": false,
                    "NewestImage": {
                        "Tag": "jammy",
                        "Vulnerabilities": {
                            "MaxSeverity": "",
                            "Count": 0
                        },
                        "Description": "base is a minimal glibc-based Linux system",
                        "IsSigned": false,
                        "SignatureInfo": [],
                        "Licenses": "GPL-2.0-or-later",
                        "Vendor": "Cisco Systems, Inc.",
                        "Labels": ""
                    },
                    "StarCount": 0,
                    "DownloadCount": 260
                },
                {
                    "Name": "tools/busybox",
                    "LastUpdated": "2022-10-04T18:22:45.289257759Z",
                    "Size": "773920",
                    "Platforms": [
                        {
                            "Os": "linux",
                            "Arch": "amd64"
                        }
                    ],
                    "IsStarred": false,
                    "IsBookmarked": false,
                    "NewestImage": {
                        "Tag": "1.34.1",
                        "Vulnerabilities": {
                            "MaxSeverity": "",
                            "Count": 0
                        },
                        "Description": "",
                        "IsSigned": false,
                        "SignatureInfo": [],
                        "Licenses": "",
                        "Vendor": "",
                        "Labels": ""
                    },
                    "StarCount": 0,
                    "DownloadCount": 112
                }
            ]
        }
    }
}
```

pkg/extensions/search/gql_proxy/global_search_handler.go
vrajashkr: Since it is completely read after `ReadAll`, we may not need to keep it open till the end of the function execution. We can close it right away.

pkg/common/model.go
vrajashkr: From what I gathered, the GlobalSearch and other such payloads that are part of the GQL schema are handled entirely by the GQL server.

Ideally, this implementation should send the same payload as the GQL server except aggregated.

Not sure if I fully got your question though.

andaaron: So you mean this code did not produce the same payload as the GQL server before the change?
In this case we should definitely fix this issue.

vrajashkr:
Update on this:
As an alternative solution, we were looking into having the fanout and proxy logic inside the resolver code and use closures to hold the request/response data.

However, the resolver has a signature like the following:
`pkg/extensions/search/resolver.go`

```
func globalSearch(ctx context.Context, query string, metaDB mTypes.MetaDB, filter *gql_generated.Filter,
	requestedPage *gql_generated.PageInput, cveInfo cveinfo.CveInfo, log log.Logger, //nolint:unparam
) (*gql_generated.PaginatedReposResult, []*gql_generated.ImageSummary, []*gql_generated.LayerSummary, error,
) {
```

This appears to be called by the internal resolver from the GQL server where we don't have any control of the behaviour.

Next items:
1. Explore the current GQL server to see if there is any way to expose the request/response data.
2. Explore whether the current GQL server can support proxying on its own.
3. See if there are any other projects out there trying to proxy GQL requests.

vrajashkr:
Started a discussion on the gqlgen repo: https://github.com/99designs/gqlgen/discussions/3295

While looking through their Discord, I came across the following gqlparser library that gqlgen uses internally. I'll take a look and evaluate this for parsing the query and formatting the output.

https://github.com/vektah/gqlparser

vrajashkr:
`gqlparser` appears to have worked very well for parsing the request query data and making the information accessible.
In the last commit, a change was made to take in the query name using this approach.

What was left partially handled was that the spec appears to allow for multiple GQL operations in a single request payload, but zot appears to only use one request at a time. TODO: need to verify this. If this is only one request at a time, then it will allow the current GQL proxy logic to function, otherwise, we may need to re-think the deserialization and aggregation part.

I'm currently looking into making the output available exactly as the GQL spec would.
The last commit already included a change to set the `Content-Type` header to `application/json` and indent the output.
The next commit will explore trimming the JSON to remove fields that were not in the GQL request. I explored how the GQL server does this and it is mostly handled by generated code that checks each field which won't work in our post-aggregation serialization implementation, hence the exploration into a generic field remover.

vrajashkr:
A number of months have gone by.

Here's the latest on this:
1. From a solutions point of view, it would ideally be better to recommend scale-out zot backed by a single metadata storage (such as Redis) for local scale-out. This requires no further code changes as it already works today. All the instances share the same metadata server so any server could respond to the client and the response would be correct.
2. We still desire for a zot-only solution for local scale-out without having any dependencies where metadata storage is embedded per instance and the OCI artifacts are locally stored in the instances. To this end, we will continue exploring options for this approach.

This PR demonstrated what it takes to get a reasonably function GlobalSearch working. However, maintaining this code becomes quite the hassle as there are multiple GQL queries supported by zot and each would need to be updated whenever there is a GraphQL schema change that impacts that query. An approach that would potentially address this issue to generate (the same way gqlgen is generating things) the handlers for these "scale-out" GQL queries that are derived based on the GQL schema. That way, after a schema change, gqlgen updates the normal GQL handlers and our generator updates the scale-out handlers. Or perhaps this could be integrated into gqlgen itself in some manner.

Exploration is currently in progress for this rather interesting problem.

Some optimizations can be done later:
1. Handling pagination consistently
2. Handling pagination efficiency by not querying every zot - only enough to meet the requirement of the query.

vrajashkr:
Did some research on this.

Some related findings include: Apollo Federation
refs:
- https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/federation
- https://gqlgen.com/recipes/federation/

The gist of this is that there is a use-case where different backend microservices serve different data in different forms, but a single Apollo Graph Router could stitch all that data together from a single GraphQL query and serve a consolidated response to the client according to the GraphQL schema.

Along these lines, gqlgen also supports Apollo Federation.

While this particular implementation may not be what we could use right away, there are some notable similarities between what GQL proxy is looking for and what the Apollo Graph Router is doing - stitching together responses from multiple sources.

vrajashkr:
Note: This is an older PR and can remain closed in favor of an updated approach in #3074

PR #2626, opened 2024-08-26
fix(delete manifest): distinct behaviors for delete by tag vb delete by digest

andaaron:
In case of delete by tag only the tag is removed, the manifest itself would continue to be accessible by digest. In case of delete by digest the manifest would be completely removed (provided it is not used by an index or another reference).

### DIFF
--- a/pkg/storage/common/common.go
+++ b/pkg/storage/common/common.go
@@ -330,17 +330,21 @@ func RemoveManifestDescByReference(index *ispec.Index, reference string, detectC
 ) (ispec.Descriptor, error) {
 	var removedManifest ispec.Descriptor
 
-	var found bool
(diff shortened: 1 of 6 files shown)


pkg/storage/common/common.go
```diff
@@ -361,6 +365,19 @@ func RemoveManifestDescByReference(index *ispec.Index, reference string, detectC
 		return ispec.Descriptor{}, zerr.ErrManifestNotFound
 	}
 
+	// In case of delete by digest we remove the manifest right away from storage
```
rchincha, line 368: What happens if tag1 -> digest and tag2 -> digest, if we delete by digest immediately, tag1 and tag2 are dangling?

andaaron: No, they are all deleted

PR #2716, opened 2024-10-11
test: add more tests for GCing indexes referencing other indexes referencing manifests

andaaron:
Looks like we didn't have many GC tests for retaining multiarch images. I added more data to the existing image retention tests, besides the new GC tests.


rchincha (approved):
lgtm

PR #2727, opened 2024-10-14
chore: reduce number of spurious log messages produced by GetNextDigestWithBlobPaths

andaaron:
Exit early in case of all folders and known non-blob file names. This avoids the logic for validating digests, and log message generation.


rchincha (approved):
lgtm

PR #2732, opened 2024-10-19
fix: issues with nested index processing in CVE and metaDB code

andaaron:
Also fix an issue with searching tags, which should work with case insensitive searches.


rchincha (approved):
lgtm

Issue #2744, opened 2024-10-23
[Bug]: S3 Storage Driver Config Fails Check


Exonical:
### zot version

v2.1.1

### Describe the bug

Zot S3 storage driver does not pass configuration verification checks. Service fails to start due to this. Using Zot on Bare Metal.

### To reproduce

1. Configuration
```bash
{
  "distSpecVersion": "1.1.0",
  "storage": {
    "rootDirectory": "/data/zot",
    "dedupe": true,
    "gc": true,
    "gcDelay": "1h",
    "gcInterval": "6h",
    "storageDriver": {
        "name": "s3",
        "rootdirectory": "/zot",
        "region": "us-east-2",
        "bucket": "zot-storage",
        "forcepathstyle": true,
        "secure": true,
        "skipverify": false,
        "accesskey": "accesskey",
        "secretkey": "secretkey"
    }
  },
  "http": {
    "address": "0.0.0.0",
    "externalUrl": "https://zot.dev.local",
    "port": "8443",
    "realm": "zot",
    "tls": {
      "cert": "/etc/ssl/zot/fullchain.pem",
      "key": "/etc/ssl/zot/privkey.pem"
    },
    "accessControl": {
      "groups": {
        "DEV2-ServerAdmins": {
          "users": ["testuser"]
        }
      },
      "repositories": {
        "**": {
          "policies": [{
            "groups": ["DEV-ServerAdmins"],
            "actions": ["read", "create", "update"]
          }]
        }
      }
    },
    "auth": {
      "openid": {
        "providers": {
          "oidc": {
            "issuer": "https://sso01.dev.local/realms/DEV2",
            "clientid": "zot",
            "clientsecret": "secretkey",
            "keypath": "",
            "scopes": ["openid", "profile", "email", "groups"]
          }
        }
      },
      "htpasswd": {
        "path": "/etc/zot/htpasswd"
      },
      "failDelay": 5
    }
  },
  "log":{
    "level": "debug",
    "output": "/var/log/zot/zot.log",
    "audit": "/var/log/zot/zot-audit.log"
  },
  "extensions": {
    "ui": {
      "enable": true
    },
    "search": {
      "enable": true,
      "cve": {
        "updateInterval": "24h",
        "trivy": { "javadbrepository": "harbor.dev.local/aquasec/trivy-java-db:1", "dbrepository": "harbor.dev.local/aquasec/trivy-db:2"}
      }
    },
    "sync": {
      "enable": false,
      "registries": [
        {
[...]

PR #2745, opened 2024-10-25
fix: improve output of zot verify

andaaron:
See #2744, there are 2 updates:
1. Silence the usage output in case of config verification errors
2. Wrap the error details in all of the zot configuration errors so they are propagated and shown in stdout for `zot verify` We also need to keep logging those message to the zot logs file since the same logic is called in the `zot serve` use case.


andaaron:
Example based on https://github.com/project-zot/zot/issues/2744

```
bin/zot-linux-amd64 verify examples/test.json
{"level":"error","config":"examples/test.json","time":"2024-10-25T11:15:04.494640658Z","message":"invalid config file"}
Error: invalid server config: invalid database config, dedupe set to true with remote storage and database, but no remote database configured
```

rchincha (approved):
lgtm

PR #2763, opened 2024-11-07
chore: update Trivy and Trivy dependencies

andaaron:
The Trivy library now supports multiple locations from where to download the DBs. The zot code has been updated to properly call the updated library functions. If at some point we would want to support multiple Trivy DBs in zot, we could look into it more.


rchincha (approved):
lgtm

Issue #2715, opened 2024-10-11
[Bug]: Content Discovery pagination support


uvegla:
### zot version

v2.1.1

### Describe the bug

When setting up `onDemand: false` sync extension rules, the prefix matching is done only the 1st page result of the content discovery made towards the remote registry. With large registries this results in some images never getting synced by Zot.

The pagination rules are defined by the specification here: https://github.com/opencontainers/distribution-spec/blob/v1.1.0/spec.md#content-discovery

The code that does the discovery is rooted here: https://github.com/project-zot/zot/blob/v2.1.1/pkg/extensions/sync/service.go#L221-L232

It makes a call to `<remote>/v2/_catalog` and only makes it once. Going further down the path, the `Link` response header is not checked and simply the response body is returned, thus not getting a full picture of the registry.

This could potentially be an issue for listing tags as well - as according to the docs, that supports pagination as well - but I did not check that use case. 

### To reproduce

# Configuration

You have to try this with a large registry with more than a 100 repositories. Relevant part of the Zot JSON configuration:

```json
"registries": [
  {
    "urls": [
      "https://example.org/"
    ],
    "onDemand": false,
    "pollInterval": "12h",
    "tlsVerify": true,
    "maxRetries": 3,
    "retryDelay": "30m",
    "onlySigned": false,
    "content": [
      {
        "prefix": "my-org/aaa-operator",
        "tags": {
          "regex": "^v?\\d+\\.\\d+\\.\\d+$"
        }
      },
      {
        "prefix": "my-org/zzz-operator",
        "tags": {
          "regex": "^v?\\d+\\.\\d+\\.\\d+$"
        }
      }
    ]
  }
]
```

Then simply run:

```
zot serve config.json
```

Assuming there are more than 100 images in the repository and `my-org/zzz-operator` is alphabetically beyond 100, the syncing will never occur for it, practically it is ignored.

# Manual reproduction

```shell
# 1st 100 repositories
curl -H "Authorization: Bearer $(curl -s https://<REMOTE>/oauth2/token\?scope\=registry%3Acatalog%3A%2A\&service\=<REMOTE> | jq -r .access_token)" https://<REMOTE>/v2/_catalog

# Take the contents on the Link response header and make follow up requests, for example
[...]

PR #2782, opened 2024-11-13
Catalog content discovery

eusebiu-constantin-petu-dbk:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -8145,6 +8145,439 @@ func TestRouteFailures(t *testing.T) {
 	})
 }
 
+func TestPagedRepositoriesWithAuthorization(t *testing.T) {
+	port := test.GetFreePort()
+	baseURL := test.GetBaseURL(port)
+	conf := config.New()
+	conf.HTTP.Port = port
+	username, _ := test.GenerateRandomString()
+	password, _ := test.GenerateRandomString()
+	htpasswdPath := test.MakeHtpasswdFileFromString(test.GetCredString(username, password))
+
+	defer os.Remove(htpasswdPath)
+
+	conf.HTTP.Auth = &config.AuthConfig{
+		HTPasswd: config.AuthHTPasswd{
+			Path: htpasswdPath,
+		},
+	}
+
+	readPolicyGroup := config.PolicyGroup{
+		Policies: []config.Policy{
+			{
+				Users: []string{username},
+				Actions: []string{
+					constants.ReadPermission,
+				},
+			},
+		},
+		DefaultPolicy: []string{},
+	}
+
+	conf.HTTP.AccessControl = &config.AccessControlConfig{
+		Repositories: config.Repositories{
+			test.AuthorizationAllRepos: config.PolicyGroup{
+				Policies: []config.Policy{
+					{
+						Users: []string{username},
+						Actions: []string{
+							constants.ReadPermission,
+							constants.CreatePermission,
+						},
+					},
+				},
+
+				DefaultPolicy: []string{},
+			},
+		},
+		AdminPolicy: config.Policy{
+			Users:   []string{},
+			Actions: []string{},
+		},
+	}
+
+	ctlr := api.NewController(conf)
+	ctlr.Config.Storage.RootDirectory = t.TempDir()
+
+	cm := test.NewControllerManager(ctlr)
+	cm.StartAndWait(port)
+
+	defer cm.StopServer()
+
+	client := resty.New()
+	client.SetBasicAuth(username, password)
+
+	img := CreateRandomImage()
+
+	repoNames := []string{
+		"alpine1", "alpine2", "alpine3",
+		"alpine4", "alpine5", "alpine6",
+		"alpine7", "alpine8", "alpine9",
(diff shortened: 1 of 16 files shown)


pkg/extensions/sync/remote.go
```diff
@@ -50,13 +51,33 @@ func (registry *RemoteRegistry) GetContext() *types.SystemContext {
...
+	_, header, _, err := registry.client.MakeGetRequest(ctx, &catalog, "application/json", "", //nolint: dogsled
 		constants.RoutePrefix, constants.ExtCatalogPrefix)
 	if err != nil {
 		return []string{}, err
 	}
 
-	return catalog.Repositories, nil
+	repos := catalog.Repositories
+
+	link := header.Get("Link")
```
andaaron, line 62: So these repos return the "Link" even if we don't specify the query parameter?

uvegla:
I gave it a try and found some problem I think. I updated the original issue: https://github.com/project-zot/zot/issues/2715#issuecomment-2478903426 and will take a deeper look.

pkg/extensions/sync/remote.go
```diff
@@ -50,13 +51,33 @@ func (registry *RemoteRegistry) GetContext() *types.SystemContext {
...
 
-	_, _, _, err := registry.client.MakeGetRequest(ctx, &catalog, "application/json", //nolint: dogsled
+	_, header, _, err := registry.client.MakeGetRequest(ctx, &catalog, "application/json", "", //nolint: dogsled
 		constants.RoutePrefix, constants.ExtCatalogPrefix)
 	if err != nil {
 		return []string{}, err
 	}
 
-	return catalog.Repositories, nil
+	repos := catalog.Repositories
```
uvegla, line 60: I think the problem of duplicating the 2nd page in place of the 1st page in the final list starts here. Further down `catalog` is passed down as a reference as well and the `Repositories` property is overwritten with the 2nd page result, thus `repos` get "overwritten" as well. Further iterations in the loop work, because at `append(...)` the list capacity need to expand and a new list is reserved in memory after that.

This way the behaviour is flaky. Its just a matter of luck to some extent when the list gets expanded and which page get lost or overwritten.

uvegla, line 60: I think simply constructing an empty array and keep appending to that should do the trick:

```suggestion
	var repos []string

	repos = append(repos, catalog.Repositories...)
```

pkg/extensions/sync/remote.go
uvegla: According to the specification: https://github.com/opencontainers/distribution-spec/blob/v1.1.0/spec.md#content-discovery

> When n is zero, this endpoint MUST return an empty list, and MUST NOT include a Link header.

Since we are not setting it to 0, its probably not an issue. However it also states further down that:

> A Link header MUST be included in the response when the descriptor list cannot be returned in a single manifest. Each response is an image index with different descriptors in the manifests field. The Link header MUST be set according to [RFC5988](https://www.rfc-editor.org/rfc/rfc5988.html) with the Relation Type rel="next".

So technically if there is only 1 page, it is not required to be in the response if I get it right. But according to https://pkg.go.dev/net/http#Header.Get an empty string is returned when there is no value associated with the requested header - or it does not exist.

rchincha:
Also be mindful about access control - one may not have read access to all repos.

pkg/extensions/sync/remote.go
rchincha: Note that the list can be inconsistent in general since repos may be get added/deleted/updated in between pagination results.

eusebiu-constantin-petu-dbk:
@uvegla Thank you very much for your work on this!

pkg/extensions/sync/remote.go
eusebiu-constantin-petu-dbk: some registry might have a limit hardcoded like https://github.com/distribution/distribution/tree/main/registry and it will return links to get next page.

pkg/api/controller_test.go
```diff
@@ -8145,6 +8145,439 @@ func TestRouteFailures(t *testing.T) {
...
+	cm.StartAndWait(port)
+
+	defer cm.StopServer()
+
+	rthdlr := api.NewRouteHandler(ctlr)
+
+	img := CreateRandomImage()
+
+	repoNames := []string{
+		"alpine1", "alpine2", "alpine3",
```
andaaron, line 8313: Can we mix them so the results from the default store and substores are not in alphabetical order?

Let's say substores with paths /a and /c, and images "alpine1", "alpine2", "alpine3", "a/alpine4", "a/alpine5", "a/alpine6", "b/alpine7", "b/alpine8", "b/alpine9", "c/alpine10", "c/alpine11", "c/alpine12", "d/alpine10", "d/alpine11", "d/alpine12". Also some entries not having a folder in path.

uvegla:
Tested locally, looks good to me now! 🙂  Thank you for working on this! 👍🏻

rchincha:
@eusebiu-constantin-petu-dbk can you take a look at @andaaron 's comments?

PR #2886, opened 2025-01-17
ci: nightly ci now uses ubuntu 24.04+

rchincha:
which needs user namespaces to be re-enabled so that stacker can build/run

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


rchincha:
Let's wait for https://github.com/rchincha/zot/actions/runs/12833247432

andaaron (approved):
LGMT

.github/workflows/nightly.yaml
```diff
@@ -166,6 +166,8 @@ jobs:
       - uses: ./.github/actions/clean-runner
       - name: Build image
         run: |
+          # enable username namespaces
```
andaaron, line 169: Actually not needed here.

andaaron: It is needed at line 201

andaaron (changes requested):

.github/workflows/nightly.yaml
andaaron: See https://github.com/project-zot/zot/actions/runs/12821167569/job/35767842486

rchincha: fixed

andaaron (approved):

PR #2891, opened 2025-01-18
fix(sync): fixed harbor authentication issues on _catalog endpoint

eusebiu-constantin-petu-dbk:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -174,4 +174,5 @@ var (
 	ErrInvalidSearchQuery             = errors.New("invalid search query")
 	ErrImageNotFound                  = errors.New("image not found")
 	ErrAmbiguousInput                 = errors.New("input is not specific enough")
+	ErrReceivedUnexpectedAuthHeader   = errors.New("received unexpected www-authenticate header")
 )
--- a/pkg/extensions/sync/httpclient/client.go
+++ b/pkg/extensions/sync/httpclient/client.go
@@ -160,12 +160,12 @@ func (httpClient *Client) Ping() bool {
 	defer cancel()
 
 	//nolint: bodyclose
-	resp, _, err := httpClient.get(ctx, pingURL.String(), false)
+	resp, _, err := httpClient.get(ctx, pingURL.String(), "", false)
 	if err != nil {
 		return false
 	}
 
-	httpClient.getAuthType(resp)
+	httpClient.authType = getAuthType(resp)
 
 	if resp.StatusCode >= http.StatusOK && resp.StatusCode <= http.StatusForbidden {
 		return true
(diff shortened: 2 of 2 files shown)


errors/errors.go
```diff
@@ -174,4 +174,5 @@ var (
 	ErrInvalidSearchQuery             = errors.New("invalid search query")
 	ErrImageNotFound                  = errors.New("image not found")
 	ErrAmbiguousInput                 = errors.New("input is not specific enough")
+	ErrUnexpectedAuthHeader           = errors.New("expected bearer auth header, received basic")
```
andaaron, line 177: The string does not match the variable name exactly. How about having a more generic string here, and wrapping the error adding the specifics of 'bearer'/basic' in the code where it is used?

pkg/extensions/sync/httpclient/client.go
```diff
@@ -298,7 +284,15 @@ func (httpClient *Client) makeAndDoRequest(method, mediaType, namespace, urlStr
 		return nil, nil, err
 	}
 
-	if err := httpClient.setupAuth(req, namespace); err != nil {
+	err = httpClient.setupAuth(req, namespace)
+	if err != nil {
+		// harbor catalog requests return basicAuth by default, even if bearer is used on the rest of endpoints.
+		if errors.Is(err, zerr.ErrUnexpectedAuthHeader) &&
```
andaaron, line 290: Based on https://github.com/goharbor/harbor/blob/9e8e647b713b9e33d9b8c32adbad066f1d7b5bf6/src/server/middleware/v2auth/auth.go#L99 they expect to receive an auth header for CLIs.

Maybe instead of this weird workaround we should send an auth header when we make the /v2/_catalog request? Don't we already know what headers to use from the initial /v2/ call?

andaaron: Actually it's the other way around, if they receive and auth header on  /v2/_catalog, they return basic auth. Why would they the do this?

vooon:
Tested on Harbor 2.12.2 - works fine, both onDemand true and false.

errors/errors.go
rchincha: @eusebiu-constantin-petu-dbk ^?

andaaron: Not sure we need this PR if we merge https://github.com/project-zot/zot/pull/2903

eusebiu-constantin-petu-dbk: is it better?

andaaron (approved):

errors/errors.go
andaaron: yes

PR #2906, opened 2025-01-23
feat(sync): add tag excludeRegex filter

### LINKED ISSUE #2902: [Feat]: Sync tags rejection regex

Due to CI limitations I cannot build multiarch images directly, so i had first to upload x86_64/aarch64 images with arch suffix, then create manifest and push it back.

So as a result on "hidden" source registry i have images like:
- `base:latest`
- `base:latest-x86_64`
- `base:latest-aarch64`

On the "frontend" registry, I'd like to sync only `base:latest` (or more like `base:v1.2.3-g1234567`).
The problem that go's regex [do not have negate match operation][1], so it's impossible to write something like `.+-(?!x86_64|aarch64)$`

[1]: https://github.com/google/re2/wiki/Syntax

Add to `sync.registries.content.tags` - `reject_regex`, which will filter-out any matched tag from syncing.

_No response_

### Additional context

_No response_

vooon:
**What type of PR is this?**

feature

Fix #2902

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/config/sync/config.go
+++ b/pkg/extensions/config/sync/config.go
@@ -42,6 +42,7 @@ type Content struct {
 }
 
 type Tags struct {
-	Regex  *string
-	Semver *bool
+	Regex        *string
+	ExcludeRegex *string
+	Semver       *bool
 }
--- a/pkg/extensions/sync/content.go
+++ b/pkg/extensions/sync/content.go
@@ -63,6 +63,13 @@ func (cm ContentManager) FilterTags(repo string, tags []string) ([]string, error
 			}
 		}
 
+		if content.Tags.ExcludeRegex != nil {
+			tags, err = excludeTagsByRegex(tags, *content.Tags.ExcludeRegex, cm.log)
+			if err != nil {
+				return []string{}, err
+			}
+		}
+
 		if content.Tags.Semver != nil && *content.Tags.Semver {
 			tags = filterTagsBySemver(tags, cm.log)
 		}
@@ -233,6 +240,32 @@ func filterTagsByRegex(tags []string, regex string, log log.Logger) ([]string, e
 
 	for _, tag := range tags {
 		if tagReg.MatchString(tag) {
+			filteredTags = append(filteredTags, tag)
+		}
+	}
+
+	return filteredTags, nil
+}
+
+// excludeTagsByRegex filter-out images by tag regex given in the config.
+func excludeTagsByRegex(tags []string, regex string, log log.Logger) ([]string, error) {
+	if len(tags) == 0 || regex == "" {
+		return tags, nil
+	}
+
+	filteredTags := make([]string, 0, len(tags))
+
+	log.Info().Str("excludeRegex", regex).Msg("filtering out tags using regex")
+
+	tagReg, err := regexp.Compile(regex)
+	if err != nil {
+		log.Error().Err(err).Str("excludeRegex", regex).Msg("failed to compile regex")
+
+		return filteredTags, err
+	}
+
+	for _, tag := range tags {
+		if !tagReg.MatchString(tag) {
 			filteredTags = append(filteredTags, tag)
 		}
 	}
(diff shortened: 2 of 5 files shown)


andaaron (approved):

pkg/extensions/config/sync/config.go
```diff
@@ -42,6 +42,7 @@ type Content struct {
 }
 
 type Tags struct {
-	Regex  *string
-	Semver *bool
+	Regex       *string
+	RejectRegex *string
```
rchincha, line 46: Would call this "ExcludeRegex"

rchincha (commented):
Would also add an example under examples/


@vooon For your particular use case, are you sure there is no workaround that can avoid this PR altogether.

pkg/extensions/config/sync/config.go
rchincha: and fix symbols everywhere else accordingly.

vooon:
@rchincha well, there are workaround, negate every character of the word: `.*([^x][^8][^6][^_][^6][^4]|[^a][^a][^r][^c][^h][^6][^4])$`

But that very soon will remind me famous perl test-test-test... :)

pkg/extensions/config/sync/config.go
vooon: @rchincha done.

eusebiu-constantin-petu-dbk (approved):

rchincha:
What happens if both "regex" and "excludeRegex" are specified? and they step on each other?

vooon:
@rchincha regex first select tags, then excludeRegrex filter-out matching tags from that list.

andaaron:
Isn't it the same result if you have the following ops on sets of tags:
- All ∩ Regex - ExcludeRegex
- (All - ExcludeRegex) ∩ Regex 
?

pkg/extensions/sync/content.go
```diff
@@ -63,6 +63,13 @@ func (cm ContentManager) FilterTags(repo string, tags []string) ([]string, error
 			}
 		}
 
+		if content.Tags.ExcludeRegex != nil {
+			tags, err = excludeTagsByRegex(tags, *content.Tags.ExcludeRegex, cm.log)
+			if err != nil {
+				return []string{}, err
+			}
```
rchincha, line 70: Can we add logs here, so if something funny happens, we figure it out easily?

pkg/extensions/sync/content.go
```diff
@@ -240,6 +247,32 @@ func filterTagsByRegex(tags []string, regex string, log log.Logger) ([]string, e
...
 
+// excludeTagsByRegex filter-out images by tag regex given in the config.
+func excludeTagsByRegex(tags []string, regex string, log log.Logger) ([]string, error) {
+	if len(tags) == 0 || regex == "" {
+		return tags, nil
+	}
+
+	filteredTags := make([]string, 0, len(tags))
+
+	log.Info().Str("excludeRegex", regex).Msg("filtering out tags using regex")
```
rchincha, line 258: Make this log.Debug()

andaaron: This should be the same as the message at line 232.
If would be very confusing for users who disabled debug logs to see only that message.
Either we keen both info, or both debug.

vooon:
@andaaron yes, all of that cases gives the same result. But the code more like that:
- `(All ∩ Regex) - ((All ∩ Regex) ∩ ExcludeRegex)` (as we do not match all tags for exclude, but the ones already matched by include regex)

pkg/extensions/sync/content.go
vooon: Shouldn't i in that case also log regex matched tags?

Or even more explicit - log what tags we have on the source, what selected by regex, what excluded and the resulting set?

vooon: At the same time - i'm not sure that this really needed. Imagine report for something like `library/mariadb`, which have hundreds if tags (if not thousands already)...

andaaron: Maybe log those as debug messages. In any case we are logging the final result of the filtering in the `pkg/extensions/sync/service.go` file in function `SyncRepo`.

PR #2927, opened 2025-01-30
feat(ldap): allow to customize user filter

### LINKED ISSUE #2924: [Feat]: LDAP user filter

I want to filter disabled users, or by some other fields.
LDAP has the feature to write a filter for that, but seems Zot do not have it in the config.

Add userFilter field, so i can write something like that:

```yaml
http:
  auth:
    ldap:
      userFilter: "(&(objectClass=inetorgperson)(!(nsaccountlock=TRUE))(!(userClass=Bot))(!(sn=Notifier)))"
```

None, as filtering do not need any code (except passing string) on the client side.

### Additional context

_No response_

vooon:
**What type of PR is this?**

feature

Fix #2924

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -90,6 +90,7 @@ var (
 	LDAPBaseDN       = "ou=" + username          //nolint: gochecknoglobals
 	LDAPBindDN       = "cn=reader," + LDAPBaseDN //nolint: gochecknoglobals
 	LDAPBindPassword = "ldappass"                //nolint: gochecknoglobals
+	LDAPUserAttr     = "uid"                     //nolint: gochecknoglobals
 )
 
 func TestNew(t *testing.T) {
@@ -2974,6 +2975,13 @@ func (l *testLDAPServer) Search(boundDN string, req vldap.SearchRequest,
 		}, nil
 	}
 
+	if req.Filter == "(&(uid=locked-user)((!(nsaccountlock=TRUE))))" {
+		return vldap.ServerSearchResult{
+			Entries:    []*vldap.Entry{},
+			ResultCode: vldap.LDAPResultSuccess,
+		}, nil
+	}
+
 	check := fmt.Sprintf("(uid=%s)", username)
 
 	if check == req.Filter {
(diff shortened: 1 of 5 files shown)


pkg/api/controller_test.go
```diff
@@ -3651,27 +3652,44 @@ func TestLDAPClient(t *testing.T) {
 
 		// bad user credentials with anonymous authentication
 		lClient = &api.LDAPClient{
```
andaaron, line 3654: What are these 2 cases testing differently? Was it a redundance in our test code?

pkg/api/controller_test.go
```diff
@@ -3651,27 +3652,44 @@ func TestLDAPClient(t *testing.T) {
...
+			Port:          ldapPort,
+			BindDN:        LDAPBindDN,
+			BindPassword:  LDAPBindPassword,
+			Base:          LDAPBaseDN,
+			UserAttribute: LDAPUserAttr,
+			UserFilter:    "(!(nsaccountlock=TRUE))",
+			SkipTLS:       true,
 		}
 
 		_, _, _, err = lClient.Authenticate("fail-user-bind", "pass")
```
andaaron, line 3695: How can we check if this is a failure for the intended reason? Maybe we should pick a different user which has a good password, but has the account locked?

vooon:
Tested with `(&(objectClass=inetorgperson)(!(nsaccountlock=TRUE))(!(userClass=Bot)))` - works as expected:
- Zot's DN (bot) can't login anymore,
- Normal user - can.

pkg/api/controller_test.go
vooon: Actually, to my taste, test file is too big, so i don't really know where the server mock.

andaaron: It really is too big. We never got to refactor it. Whenever someone didn't know where to place a test, it was dropped in this file.

andaaron (approved):

andaaron:
@vooon, please sign the commit.

vooon:
@andaaron done.

Issue #2752, opened 2024-10-31
[Feat]: Ability to generate s3 presign link for blob


vanhtuan0409:
### Describe the solution you'd like

I want to get a s3 pre-sign link for image layer. This way we can tweak the client to download directly from s3 without needing to proxy through Zot

rchincha:
@vanhtuan0409 is this a CDN use case? Also, what would zot actually need to do here?

vanhtuan0409:
To be more detailed: we are using Zot as an OCI-compliant registry. During the `pull` operation, the client will actually call `GetBlob` from the standard OCI API. In turn, Zot will read remote storage (s3 in this scenario) and proxy the content back to the client. The traffic look like this `client <--> zot <--> s3`

We want to avoid the traffic passing through Zot as this can cause some heavy load.  In an ideal condition, we can have a plugin for generating s3 pre-signed link for the client to download

vanhtuan0409:
Using the 307 redirects could solve this issue. I need the client to pull directly from remote storage without causing stress on Zot server

rchincha:
https://doc.traefik.io/traefik/middlewares/http/redirectregex/
You may also want to look at solutions such as this ^ positioned in front of zot.

hown3d:
How should this be implemented outside of zot? 
How should e.g. traefik know how to generate the presigned URLs, where the blobs are stored in S3 etc.

I believe having this option in zot is necessary to reduce the heavy load as @vanhtuan0409 already stated.

> So you would have to build machinery such as once a blob is uploaded to zot, copy it somewhere else and then 30x redirect to that new location. Can this be achieved without doing this in zot?

zot, using s3 as backend, already uploads the blob to a new location. Why not just redirect there?
You wouldn't even need to implement it here directly since zot [re-uses docker-distributions s3 storage implementation](https://github.com/project-zot/zot/blob/88efa50de99d4ee510893af5d45e465966fff821/pkg/storage/s3/driver.go).
Docker-distribution has a `URLFor` function in their [`StorageDriver` interface](https://github.com/distribution/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/registry/storage/driver/storagedriver.go#L81-L85) which is leveraged in [`blobserver.ServeBlob`](https://github.com/distribution/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/registry/storage/blobserver.go#L37-L52).

[...]

PR #2965, opened 2025-02-14
feat: redirect to download urls of blobs

hown3d:
**What type of PR is this?**
feature

#2752 

redirects clients to a presigned urls for blob downloads, which significantly reduces load of the zot server, since the blobs no longer need to go through the server.

**Automation added to e2e**:

```release-note
redirect clients to S3 URLs when fetching blobs
```

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -60,6 +60,7 @@ var (
 	ErrBadBlob                        = errors.New("bad blob")
 	ErrBadBlobDigest                  = errors.New("bad blob digest")
 	ErrBlobReferenced                 = errors.New("blob referenced by manifest")
+	ErrBlobRedirectURLNotSupported    = errors.New("blob redirect url not supported")
 	ErrManifestReferenced             = errors.New("manifest referenced by index image")
 	ErrUnknownCode                    = errors.New("unknown error code")
 	ErrBadCACert                      = errors.New("invalid tls ca cert")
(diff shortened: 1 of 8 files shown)


rchincha:
https://github.com/opencontainers/distribution-spec/issues/299

@hown3d would also show your support upstream. 
zot plays a "follower" role to Open Containers specs.

zer010gic:
Is there any news on whether this PR/Feature will be implemented?

andaaron:
Assuming the spec allows for 307, what is the expected behavior for auth? The website we redirect to may have authentication.

joshbwlng:
@andaaron From what I understand, presigned URLs shouldn't have auth issues because they're presigned and only valid for a short period of time. I suppose we could potentially handle this in a download redirect proxy running in front of Zot, but it would make more sense in my opinion for it to be implemented within Zot itself - like it is for docker distribution.

hown3d:
Since the upstream spec is not moving and zot is strictly following the spec, I will close this PR.
Once upstream distribution spec is adapted to feature support for redirects, we can reopen this PR.

maboehm:
With https://github.com/opencontainers/distribution-spec/pull/607 the upstream spec has moved and now says:

> Registries MAY respond to any request with a redirect per [RFC 9110 (section 15.4)](https://www.rfc-editor.org/rfc/rfc9110#section-15.4); clients SHOULD follow such redirects, and MUST NOT forward `Authorization` headers across host boundaries unless explicitly configured to do so.

so @hown3d would you mind re-opening and polishing this PR?

hown3d:
I recreated the PR, since I had to force push to rebase and therefore cannot reopen the current PR.
 https://github.com/project-zot/zot/pull/4071

PR #2969, opened 2025-02-15
refactor: remove usage of goto in the image store

andaaron:


rchincha (approved):
lgtm

but why make this change at all?

andaaron:
Because it was very confusing for me as a developer reading the code.
I was troubleshooting an issue related to some of the calls made from this function, and it took me over an hour to notice we had a loop and those functions were called multiple times.

In any case don't merge this. I prefer to proceed with the PR also handling the locks.

PR #2996, opened 2025-02-27
Cumulative improvements for CI troubleshooting

andaaron:

### DIFF
--- a/cmd/zb/perf.go
+++ b/cmd/zb/perf.go
@@ -126,7 +126,8 @@ type statsSummary struct {
 	statusHist           map[string]int
 	rps                  float32
 	mixedSize, mixedType bool
-	errors               int
+	errorCount           int
+	errors               []error
 }
 
 func newStatsSummary(name string) statsSummary {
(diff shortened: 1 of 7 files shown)


cmd/zb/perf.go
```diff
@@ -147,11 +148,16 @@ type statsRecord struct {
...
 }
 
 func updateStats(summary *statsSummary, record statsRecord) {
 	if record.isConnFail || record.isErr {
-		summary.errors++
+		summary.errorCount++
+	}
+
+	if record.err != nil {
+		summary.errors = append(summary.errors, record.err)
```
rchincha, line 160: If every request is an error, this could get very large?

rchincha: Maybe you want buckets of things and their counts?

andaaron: Right now we have counts.

The issue is we only have counts. I was getting client side errors when the client was making requests and I did not know what they were, as status code did not get to be checked and error messages were not shown in server logs. (With this patch I found them to be EOF errors, which were probably caused by other requests deleting the file while the server was reading it for the errored request)

I think we can improve by having a map of errors and counting duplicates.

rchincha (approved):

Issue #2587, opened 2024-08-03
[Feat]: Support EC public key


giovannipapini:
### Is your feature request related to a problem? Please describe.

Currently, trying to integrate auth login with external server bearer token, I receive this error
```
{"level":"panic","error":"Key is not a valid RSA public key","goroutine":1,"caller":"zotregistry.dev/zot/pkg/api/authn.go:426","time":"2024-08-03T13:52:30.053939244Z","message":"failed to create bearer authorizer"}
```
because it seems that Zot does not support EC (not RSA) public keys.

rchincha:
@giovannipapini we use chart-museum as our library for this, which currently supports only RSA. Let us look into this.

rchincha:
https://github.com/project-zot/zot/pull/2998

has been merged

PR #2998, opened 2025-03-01
[feat]: add support for EC/ED25519 public keys for token authentication

evanebb:
**What type of PR is this?**
feature

https://github.com/project-zot/zot/issues/2587

Currently the `github.com/chartmuseum/auth` package is used for token authentication, which [only supports RSA keys](https://github.com/chartmuseum/auth/blob/2d81aa85f9fbc574ca7849526c7b6818880e269b/token.go#L93-L94). It also doesn't seem to be updated anymore, and uses a pretty [old version of the `golang-jwt` package](https://github.com/chartmuseum/auth/blob/2d81aa85f9fbc574ca7849526c7b6818880e269b/go.mod#L6) which could use a bump.
I have ripped out the important bits and re-implemented the token handling directly using `golang-jwt`, which allows it to support EC/ED25519 keys as well as RSA keys.

In addition to the added automated tests, I have done some manual testing by spinning up a local Zot instance + authorization server. I tested using an RSA (RS256), ECDSA (ES256) and ED25519 key/certificate, did some API calls with tokens signed using these keys and some logins/pushes/pulls using the Docker CLI. This all seemed to work just fine.

**Automation added to e2e**:

It shouldn't, I've tried to keep the behavior as close to what it was before as possible, but I can't guarantee it (there are some more improvements I could think of, but I don't want to break backward compatibility completely).

```release-note

```

### DIFF
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -9,7 +9,6 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"github.com/golang-jwt/jwt/v5"
 	"net"
 	"net/http"
 	"os"
@@ -18,6 +17,7 @@ import (
 	"time"
 
 	guuid "github.com/gofrs/uuid"
+	"github.com/golang-jwt/jwt/v5"
 	"github.com/google/go-github/v62/github"
 	"github.com/google/uuid"
 	"github.com/gorilla/mux"
@@ -437,6 +437,7 @@ func bearerAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 			}
 
 			var requestedAccess *resourceAction
+
 			if request.RequestURI != "/v2/" {
 				// if this is not the base route, the requested repository/action must be authorized
 				vars := mux.Vars(request)
@@ -456,7 +457,7 @@ func bearerAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 
 			err := authorizer.Authorize(header, requestedAccess)
 			if err != nil {
-				var challenge *authChallenge
+				var challenge *authChallengeError
 				if errors.As(err, &challenge) {
 					ctlr.Log.Debug().Err(challenge.err).Msg("bearer token authorization failed")
 					response.Header().Set("Content-Type", "application/json")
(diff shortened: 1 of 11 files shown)


rchincha:
@evanebb thanks for this PR. 

Pls address the CI failures.

pkg/api/authn.go
```diff
@@ -2,21 +2,21 @@ package api
...
+	"github.com/golang-jwt/jwt/v5"
 	"net"
 	"net/http"
 	"os"
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
 
-	"github.com/chartmuseum/auth"
```
rchincha, line 19: Not removed from go.mod because still used in test code.

andaaron (commented):
The code is much cleaner compared to what we had before.
Please make sure to add relevant tests and reach the coverage threshold.

pkg/api/controller_test.go
```diff
@@ -2401,6 +2405,59 @@ func TestAuthnErrors(t *testing.T) {
...
+			Cert:    tmpFile,
+		}
+
+		ctlr := makeController(conf, t.TempDir())
+
+		So(func() {
+			api.AuthHandler(ctlr)
+		}, ShouldPanic)
+
+		err = os.Chmod(tmpFile, 0o644)
```
andaaron, line 2430: Should this be in a defer statement after the file is created? If the test fails before this, wouldn't the file remain with permission 0o000?

evanebb: Seems logical yes, and other tests do it that way too.
I've moved it into a defer statement for all the tests under `TestAuthnErrors`, so that it is consistent there: https://github.com/project-zot/zot/pull/2998/commits/3480be1671414ae1a8812df4f015a775977fadd5

rchincha:
@evanebb when you are ready, move it out of draft and mark it ready for review.

evanebb:
@rchincha done, I wanted to do a bit more manual testing before doing so.

Something that _could_ cause problems is that the old code would also accept a file containing a public key for the `http.auth.bearer.cert` directive ([right here](https://github.com/chartmuseum/auth/blob/main/util.go#L56), which calls this [function under the hood](https://github.com/golang-jwt/jwt/blob/v3.2.2/rsa_utils.go#L86-L92)), whereas my changes [always assume that it is a certificate](https://github.com/evanebb/zot/blob/468d7dd004bbeb3e86b9cac0bbb552f7adaff525/pkg/api/authn.go#L916) (which makes parsing it a bit simpler).
However, this is undocumented behavior as far as I can see. The documentation only mentions [passing a certificate file through this parameter](https://zotregistry.dev/v2.1.2/articles/authn-authz/?h=#http-bearer-authentication).

(also, I noticed that the [CI will intermittently fail with a `bind: address already in use` error](https://github.com/project-zot/zot/actions/runs/13682936973/job/38259492419?pr=2998#step:7:1878). I did mess around a bit in the tests that spin up a test HTTP server, but I'm going to assume that this just happens from time-to-time and a re-run 'fixes' it?)

rchincha:
> (also, I noticed that the [CI will intermittently fail with a bind: address already in use error](https://github.com/project-zot/zot/actions/runs/13682936973/job/38259492419?pr=2998#step:7:1878). I did mess around a bit in the tests that spin up a test HTTP server, but I'm going to assume that this just happens from time-to-time and a re-run 'fixes' it?)

Some flaky tests, we just have to arbitrate free ports.

andaaron (approved):

PR #3022, opened 2025-03-11
fix: allow changing media-type when pushing an image tag

### LINKED ISSUE #3005: [Bug]: `docker build --push` results in `400 Bad Request`
### zot version

v2.1.2

## Describe the bug

**TL;DR:** Docker CLI tools seem to publish images to a registry and perform a `PUT` request that fails with `zot`.

---

With a minimal setup for `zot` (_via the [official container](https://github.com/project-zot/zot/pkgs/container/zot/338980604?tag=v2.1.2)_) and a very simple image, the following fails:

```Dockerfile
FROM alpine
RUN echo 'hello' > /tmp/hello.txt
```

```console
$ docker run --rm -d -p 127.0.0.1:5000:5000 --name zot ghcr.io/project-zot/zot
$ docker build --push --tag localhost:5000/hello/world:example .

# ...

ERROR: failed to solve: failed to push localhost:5000/hello/world:example: failed commit on ref "index-sha256:450f2a3f8d07729f8ea4db516f4a0d4d03de6e5ee6435d1d0c99289a6b3c280d": unexpected status from PUT request to http://localhost:5000/v2/hello/world/manifests/example: 400 Bad Request
```

<details>
<summary>Additional log context</summary>

```
 => [1/2] FROM docker.io/library/alpine:latest@sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c
 => => resolve docker.io/library/alpine:latest@sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c
 => CACHED [2/2] RUN echo 'hello' > /tmp/hello.txt
 => ERROR exporting to image
 => => exporting layers
 => => exporting manifest sha256:8669ac3edd2e1cba3ce21f07cef921d7687cbd13dc2a5411a56c48dfa5316998
 => => exporting config sha256:645726965a22a6f8bde99f9b456c621dee125a79b0314e16cd22c48f49c2aac6
 => => exporting attestation manifest sha256:0b318828575b92022bae7afdfee63f7ecd422024f0814c2d539b90dd23ee23e2
 => => exporting manifest list sha256:450f2a3f8d07729f8ea4db516f4a0d4d03de6e5ee6435d1d0c99289a6b3c280d
 => => naming to localhost:5000/hello/world:example
 => => unpacking to localhost:5000/hello/world:example
 => => pushing layers
 => => pushing manifest for localhost:5000/hello/world:example@sha256:450f2a3f8d07729f8ea4db516f4a0d4d03de6e5ee6435d1d0c99289a6b3c280d
------
 > exporting to image:
------
ERROR: failed to solve: failed to push localhost:5000/hello/world:example: failed commit on ref "index-sha256:450f2a3f8d07729f8ea4db516f4a0d4d03de6e5ee6435d1d0c99289a6b3c280d": unexpected status from PUT request to http://localhost:5000/v2/hello/world/manifests/example: 400 Bad Request
```

</details>

**NOTE:** `failed to solve` portion prior to `failed commit on ref` is only presented when the image exists for that tag at `zot`.

Similarly it seems that with a `docker-container` build driver, it will be successful with the first build/push provided the tag to publish is new?

```bash
docker buildx build --builder custom-builder \
  --output 'type=image,push=true,oci-mediatypes=false' \
  --platform 'linux/amd64,linux/arm64' \
  --provenance false \
  --tag localhost:5000/hello/world:example \
  .
```

Otherwise the error output is like shown earlier. Given the `PUT` request failure, the client tool here is presumably interacting with REST API to update the existing manifest for an already published tag? (_except with the `default` builder, aka the first CLI example which presumably tries to make a modification directly after_)

**NOTE:** This observation/failure was not reproducible with another client (`crane`) pushing OCI image layout exports from either of the docker CLI builds above.

### Configuration

**NOTE:** Reproduction should be quite simple with the commands in the above description. More verbose config is documented in collapsed sections below.

- Fairly standard with the zot image ([`zot:2.1.2`](https://github.com/project-zot/zot/pkgs/container/zot/338980604?tag=v2.1.2)).
- Host is a Vultr Fedora 41 VPS with [Docker installed through Docker CE repos](https://docs.docker.com/engine/install/fedora/).

<details>
<summary>Click to view</summary>

## Minimal

`compose.yaml` (_with the 1.7GB weight of trivy_):

```yaml
services:
  zot:
    image: ghcr.io/project-zot/zot:latest # 2.1.2
    ports:
      - "127.0.0.1:5000:5000"
```

**CAUTION:** As the image declares [`VOLUME /var/lib/registry/`](https://github.com/project-zot/zot/blob/d4656906309dc9b91f11158bbc26d495f54733c5/build/stacker.yaml#L75-L76) to create an implicit anonymous volume and Docker Compose can persist this, removing it will [require specific syntax](https://github.com/moby/moby/issues/49482#issuecomment-2667048431) like `docker compose down --volumes`.

or

```bash
# NOTE: Without `--rm`, each run will accumulate 1.7GB via new anonymous volume
docker run --rm -d -p 127.0.0.1:5000:5000 ghcr.io/project-zot/zot
```

---

For multi-arch image builds, you need a `docker-container` build driver like so:

```bash
docker buildx create --use \
  --name custom-builder \
  --driver docker-container \
  --driver-opt network=host \
  --platform 'linux/amd64,linux/arm64'
```

## Original reproduction config

<details>
<summary>Click to view</summary>

#### Docker host (`default`) - [`docker` build driver](https://docs.docker.com/build/builders/drivers/docker/)

`localhost` has an exception by default to trust HTTP + insecure HTTPS (no verification enforced), this doesn't apply for other private DNS names including those within `localhost` (_despite when they resolve to an [IP in the trusted loopback subnet `127.0.0.0/8`](https://docs.docker.com/reference/cli/dockerd/#insecure-registries)_).

Thus to support `zot.internal` or `zot.localhost` without a trusted CA configured edit `/etc/docker/daemon.json` to trust them over HTTP or HTTPS without verification via `insecure-registries` setting:

```json
{
  "features": {
    "containerd-snapshotter": true
  },
  "insecure-registries" : [ "zot.internal", "zot.localhost" ],
  "userland-proxy": false
}
```

The other two settings I have there appear to have no meaningful impact on this issue.

**NOTE:** Slight behaviour difference observed when a port was used (_doesn't apply to `docker-container` driver usage which has two separate BuildKit config options per registry_):
- `--tag zot.localhost/...` => `https://zot.localhost/v2/...`
- `--tag zot.localhost:5000/...` => `http://zot.localhost:5000/v2/...`

---

#### BuildKit / [`docker-container` build driver](https://docs.docker.com/build/builders/drivers/docker-container/)

When instead using `docker buildx build` with an image builder that uses the `docker-container` driver, the BuildKit config would likewise need to be adjusted (_does not recognize equivalent config at `/etc/docker/daemon.json`_).

**NOTE:** `localhost` will be trusted implicitly by default (any port), the builder instance just needs to be using the host network to reach it (_as configured below_).

```toml
# You would only need one of these options, but during troubleshooting
# I had tried with Caddy disabling HTTP => HTTPS redirects to rule HTTPS issues out

# `http=true` will switch to http:// instead of https://
# `insecure=true` will skip verification for https://
[registry."zot.internal"]
  http = true
  insecure = true

[registry."zot.localhost"]
  http = true
  insecure = true
  # Alternatively provide a path on the host to the public CA cert
  # that signed the leaf cert for this domain (will be copied on builder creation):
  #ca=["certs/ca-cert.pem"]
```

```bash
# NOTE: This option is not necessary as it's an implicit default with modern buildx:
# `--buildkitd-flags '--allow-insecure-entitlement network.host'`

# - `docker-container` driver required to build multi-arch via QEMU.
# - `network=host` opt only required if you have custom DNS on the host to use,
#   or to connect via host services such as `localhost:5000`.
# - Optional: `--config` if adding trusted insecure registries via custom BuildKit config above.
docker buildx create --use \
  --name custom-builder \
  --driver docker-container \
  --driver-opt network=host \
  --platform 'linux/amd64,linux/arm64' \
  --config buildkitd.toml
```

**NOTE:**
- If you've not created a `docker-container` builder yet on the system with multi-platform support, you'll need to have [QEMU configured](https://docs.docker.com/build/builders/drivers/docker-container/#qemu).
- Instead of the host network, if desirable you should be able to [configure for the same network](https://docs.docker.com/build/builders/drivers/docker-container/#custom-network) that the zot container is using.

---

### Zot config

**Optional:** I've only removed some of the [default image config](https://github.com/project-zot/zot/blob/d4656906309dc9b91f11158bbc26d495f54733c5/build/stacker.yaml#L18-L44), notably `extensions.search.cve` to avoid 1.7GB disk/bandwidth for Trivy (_the docs could better document opt-out for this default feature BTW_).

```json
{
  "storage":{
      "rootDirectory":"/var/lib/registry"
  },
  "http":{
      "address":"0.0.0.0",
      "port":"5000",
      "compat": ["docker2s2"]
  },
  "log":{
      "level":"debug"
  },
  "extensions": {
    "search": {
        "enable": true
    },
    "ui": {
        "enable": true
    }
  }
}
```

### Docker Compose

**`compose.yaml`:**

```yaml
name: example

services:
  zot:
    image: ghcr.io/project-zot/zot:latest # 2.1.2
    # Optional: Configures Caddy below to route traffic for `zot.localhost` and `zot.internal`
    # to this container and that traffic should go to port 5000:
    labels:
      caddy: zot.localhost, zot.internal
      caddy.reverse_proxy: "{{ upstreams 5000 }}"
    # If not using a reverse-proxy, can connect to container directly at port 5000
    ports:
      - "5000:5000"
    volumes:
      - ./zot/data/:/var/lib/registry/
      - ./zot/config.json:/etc/zot/config.json

  # Optional: Used to route HTTP/HTTPS traffic to `zot:5000` above
  reverse-proxy:
    image: lucaslorentz/caddy-docker-proxy:2.9
    # Optional: For local/private DNS route traffic for `zot.internal` through this container:
    # Host should be able to rely on `zot.localhost` when `localhost` TLD always routes to `127.0.0.1`
    networks:
      default:
        aliases:
          - zot.internal
    # For testing this will only provision locally self-signed certs instead of the default (LetsEncrypt)
    labels:
      caddy.local_certs:
    # Necessary for host with SELinux preventing container access to `docker.sock`:
    security_opt:
      - label:disable
    # Used by CDP to read labels of other containers for autoconfig:
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
    ports:
      - "80:80"
      - "443:443"
      - "443:443/udp"

# Optional: The implicit `default` bridge network.
# Change the implicit default binding IP from `0.0.0.0` to `127.0.0.1` only,
# which ensures ports are only published to the loopback interface on the host,
# without explicitly hard-coding it to each port config line above.
networks:
  default:
    driver_opts:
      com.docker.network.bridge.host_binding_ipv4: 127.0.0.1
```

**Optional:** Supporting `zot.internal` from the host (_add to above `compose.yaml`_):

```yaml
services:
  dns:
    image: ghcr.io/0xerr0r/blocky
    ports:
      - "53:53/tcp"
      - "53:53/udp"
    configs:
      - source: blocky-config
        target: /app/config.yml

configs:
  # Simple config that routes DNS queries for the `internal` TLD to
  # the containers private DNS managed by Docker at `127.0.0.11` for
  # it's bridge networks.
  # Makes resolving the configured network alias `zot.internal` available outside containers.
  # Otherwise it forwards to Cloudflare DNS.
  blocky-config:
    content: |
      upstreams:
        groups:
          default:
            - 1.1.1.1 # Cloudflare
      conditional:
        fallbackUpstream: false
        mapping:
          internal: 127.0.0.11
```

The two extra services shouldn't be necessary, but were part of the original config for the reproduction so I've documented them here.

</details>

</details>

---

### Client tool used

`docker build` as demonstrated in description above.

The original report content here is mostly redundant, thus collapsed.

<details>
<summary>Click to view original content</summary>

Always fails:

```bash
# NOTE: `docker build` is equivalent to `docker buildx build --builder default`
docker build --push --tag localhost:5000/hello/world .
```

Only seems to fail when the tag is already published at the registry and these settings change requiring a manifest update.

```bash
# NOTE: Assumes `--builder custom-builder` is the default buildx builder:
docker buildx build \
  --output 'type=image,push=true,oci-mediatypes=false' \
  --platform 'linux/amd64,linux/arm64' \
  --provenance false \
  --tag zot.internal/hello/world:latest \
  .
```

~~Failure can be avoided via these changes~~ (_**UPDATE:** Findings were inconsistent to reproduce, failure is dependent upon these settings introducing a conflict with an existing manifest for the tag_):
- `oci-mediatypes=true` in `--output`
  - Still fails when not publishing a multi-arch build (_due to omitting `--platform` or it only sets a single platform_).
- `--provenance true` (_or omission of this option as it's true by default_)
  - This appears to override `oci-mediatypes` (_forces it to true even when explicitly set to false_).
  - Also works when a single platform is set, presumably because this setting adds an `unknown/unknown` platform entry per platform image into a manifest index for the provenance attestation artifact(s).

</details>

---

### Seen error

Terminal output fails during command with:

```
ERROR: failed commit on ref "manifest-sha256:8669ac3edd2e1cba3ce21f07cef921d7687cbd13dc2a5411a56c48dfa5316998": unexpected status from PUT request to https://localhost:5000/v2/hello/world/manifests/latest: 400 Bad Request
```

## Expected behavior

No failure, image is published / updated at `zot` registry successfully.

## Additional context

I encountered this failure due to intentionally exploring these specific settings for affecting the manifest generation to compare against `ghcr.io` (_which works without this reported failure_), as I wanted to investigate `zot` with OCI Referrers API support (_which GHCR presently lacks_).

rchincha:
Fixes #3005

Previously, changing a image's media-type was disallowed. However, "docker buildx" appears to first push an image manifest and then an image index for the same image tag. So, allow this.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -10835,13 +10835,13 @@ func TestManifestImageIndex(t *testing.T) {
 					resp, err = resty.R().SetHeader("Content-Type", ispec.MediaTypeImageIndex).
 						SetBody(content).Put(baseURL + "/v2/index/manifests/test:1.0")
 					So(err, ShouldBeNil)
-					So(resp.StatusCode(), ShouldEqual, http.StatusBadRequest)
+					So(resp.StatusCode(), ShouldEqual, http.StatusCreated)
 
 					// previously an image index, try writing a manifest
 					resp, err = resty.R().SetHeader("Content-Type", ispec.MediaTypeImageManifest).
 						SetBody(m1content).Put(baseURL + "/v2/index/manifests/test:index1")
 					So(err, ShouldBeNil)
-					So(resp.StatusCode(), ShouldEqual, http.StatusBadRequest)
+					So(resp.StatusCode(), ShouldEqual, http.StatusCreated)
 				})
 			})
 		})
--- a/pkg/storage/common/common.go
+++ b/pkg/storage/common/common.go
@@ -250,15 +250,9 @@ func CheckIfIndexNeedsUpdate(index *ispec.Index, desc *ispec.Descriptor,
 
 			// changing media-type is disallowed!
 			if manifest.MediaType != desc.MediaType {
-				err := zerr.ErrBadManifest
-				log.Error().Err(err).
+				log.Info().
 					Str("old mediaType", manifest.MediaType).
-					Str("new mediaType", desc.MediaType).Msg("cannot change media-type")
-
-				reason := fmt.Sprintf("changing manifest media-type from \"%s\" to \"%s\" is disallowed",
-					manifest.MediaType, desc.MediaType)
-
-				return false, "", zerr.NewError(err).AddDetail("reason", reason)
+					Str("new mediaType", desc.MediaType).Msg("media-type changed")
 			}
 
 			oldDesc := *desc
--- a/pkg/storage/s3/s3_test.go
+++ b/pkg/storage/s3/s3_test.go
@@ -3410,11 +3410,11 @@ func TestS3ManifestImageIndex(t *testing.T) {
 					So(digest, ShouldNotBeNil)
 
 					_, _, err = imgStore.PutImageManifest("index", "test:1.0", ispec.MediaTypeImageIndex, content)
-					So(err, ShouldNotBeNil)
+					So(err, ShouldBeNil)
 
 					// previously an image index, try writing a manifest
 					_, _, err = imgStore.PutImageManifest("index", "test:index1", ispec.MediaTypeImageManifest, m1content)
-					So(err, ShouldNotBeNil)
+					So(err, ShouldBeNil)
 				})
 			})
 		})


pkg/api/controller_test.go
```diff
@@ -10835,13 +10835,13 @@ func TestManifestImageIndex(t *testing.T) {
 					resp, err = resty.R().SetHeader("Content-Type", ispec.MediaTypeImageIndex).
 						SetBody(content).Put(baseURL + "/v2/index/manifests/test:1.0")
 					So(err, ShouldBeNil)
-					So(resp.StatusCode(), ShouldEqual, http.StatusBadRequest)
+					So(resp.StatusCode(), ShouldEqual, http.StatusOK)
```
andaaron, line 10838: See GH test results. This should be 201?

rchincha: fixed

andaaron (approved):

PR #3024, opened 2025-03-12
[feat] support custom GitHub OAuth2 auth and token URLs

analytically:
Feature

Support custom Github OAuth2 auth and token URLs


andaaron (commented):
Thank you for your PR, can you also add an example configuration under https://github.com/project-zot/zot/tree/main/examples?

andaaron:
@analytically can you also add a test to verify the new configuration?

analytically:
@andaaron a test that runs against GitHub Enterprise? Or just config object validation?

andaaron:
I was thinking maybe we could mock it, similar to what we have in https://github.com/project-zot/zot/blob/c87f489a7995a06a16c967dcb8090d27b3f1cc15/pkg/test/auth/oidc.go#L13

rchincha:
> Support custom Github OAuth2 auth and token URLs

^ could you also point to pointers/docs about why this PR is needed.

@analytically thanks for taking the time for this PR.

PR #3047, opened 2025-03-21
feat: build windows binaries

rchincha:
Currently zot project doesn't build and ship Windows binaries. This PR adds that support.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/runtime_windows.go
+++ b/pkg/api/runtime_windows.go
@@ -1,0 +1,44 @@
+//go:build windows
+// +build windows
+
+package api
+
+import (
+	"runtime"
+	"syscall"
+	"unsafe"
+
+	"zotregistry.dev/zot/pkg/log"
+)
+
+func getCurrentHandleCount() (uint32, error) {
+	kernel32 := syscall.NewLazyDLL("kernel32.dll")
+
+	getProcessHandleCount := kernel32.NewProc("GetProcessHandleCount")
+	getCurrentProcess := kernel32.NewProc("GetCurrentProcess")
+	currentProcess, _, _ := getCurrentProcess.Call()
+
+	var handleCount uint32
+	ret, _, err := getProcessHandleCount.Call(
+		currentProcess,
+		uintptr(unsafe.Pointer(&handleCount)),
+	)
+
+	if ret == 0 {
+		return 0, err
+	}
+
+	return handleCount, nil
+}
+
+// DumpRuntimeParams dumps important runtime state such as file and socket limits.
+func DumpRuntimeParams(log log.Logger) {
+	evt := log.Info().Int("cpus", runtime.NumCPU()) //nolint: zerologlint
+
+	nofile, err := getCurrentHandleCount()
+	if err == nil {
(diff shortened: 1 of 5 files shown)


rchincha:
@riverar

May have beaten you to this PR. In any case, pls review or amend.

pkg/api/runtime_windows.go
```diff
@@ -0,0 +1,24 @@
...
+	"syscall"
+
+	"zotregistry.dev/zot/pkg/log"
+)
+
+// DumpRuntimeParams dumps important runtime state such as file and socket limits.
+func DumpRuntimeParams(log log.Logger) {
+	evt := log.Info().Int("cpus", runtime.NumCPU()) //nolint: zerologlint
+
+	getmaxstdio := syscall.NewLazyDLL("kernel32.dll").NewProc("_getmaxstdio")
```
riverar, line 16: This API doesn't report any useful limitation; the cap on open file handles is actually variable and based on available memory. Your best bet is to instead report the current open handle count.

Maybe something like:
```go
func getCurrentHandleCount() (uint32, error) {
	kernel32 := syscall.NewLazyDLL("kernel32.dll")

	getProcessHandleCount := kernel32.NewProc("GetProcessHandleCount")
	getCurrentProcess := kernel32.NewProc("GetCurrentProcess")
	currentProcess, _, _ := getCurrentProcess.Call()

	var handleCount uint32
	ret, _, err := getProcessHandleCount.Call(
		currentProcess,
		uintptr(unsafe.Pointer(&handleCount)),
	)

	if ret == 0 {
		return 0, err
	}
	
	return handleCount, nil
}

// ...

evt.Uint64("open handles", uint64(handleCount))
```

riverar:
Looking good, that's all I changed! (Also just threw together a workflow for my purposes.) https://github.com/riverar/zot/commits/main/

pkg/api/runtime_windows.go
riverar: (Or alternatively just not print anything on Windows since it's not very useful.)

rchincha: What we want is the max allowed file handles if possible, but if not this is the second best option.
Tested this PR as of now, and seems to work fine.

pkg/api/runtime_windows.go
```diff
@@ -0,0 +1,44 @@
...
+	return handleCount, nil
+}
+
+// DumpRuntimeParams dumps important runtime state such as file and socket limits.
+func DumpRuntimeParams(log log.Logger) {
+	evt := log.Info().Int("cpus", runtime.NumCPU()) //nolint: zerologlint
+
+	nofile, err := getCurrentHandleCount()
+	if err == nil {
+		evt = evt.Uint64("max. open files", uint64(nofile)) //nolint: unconvert // required for *BSD
```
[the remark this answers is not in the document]
rchincha: fixed

andaaron:
> Currently zot project doesn't build and ship Windows binaries. This PR adds that support.

Can we not "support"? Building it is one thing. Committing to resolve issues raised by Windows users is another.

riverar (approved):

andaaron (approved):
Looks good.

PR #3052, opened 2025-03-22
feat: handle multiple attributes

zzzhouuu:
**What type of PR is this?**
feature

#3050 

Enhance LDAP user group retrieval to handle multiple attributes

**Automation added to e2e**:

No

```release-note

```

### DIFF
--- a/pkg/api/ldap.go
+++ b/pkg/api/ldap.go
@@ -223,8 +223,9 @@ func (lc *LDAPClient) Authenticate(username, password string) (bool, map[string]
 	var userGroups []string
 
 	if lc.UserGroupAttribute != "" && len(search.Entries[0].Attributes) > 0 {
-		userAttributes := search.Entries[0].Attributes[0]
-		userGroups = userAttributes.Values
+		for _, attr := range search.Entries[0].Attributes {
+			userGroups = append(userGroups, attr.Values...)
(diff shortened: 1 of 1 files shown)


andaaron (dismissed):

andaaron:
@zzzhouuu, thank you for the PR. Do you think you can add a test with multiple ldap groups?
At the moment the existing ldap tests we have are at https://github.com/project-zot/zot/blob/main/pkg/api/controller_test.go#L3098.

zzzhouuu:
@andaaron  I also don't have a particularly good idea for writing test cases, because the existing test case runs in the first place to ensure that there are no problems with the previous use, which is not a breaking change, but just an enhancement that we don't drop any data if the LDAP server returns multiple sets of data, which is more in line with the results. Of course, if you have other ideas, you can also let me know, and I will try as much as I can.

rchincha:
@zzzhouuu can you pls take care of the ci failures?
dco specifically?

zzzhouuu:
@rchincha  It seems that there is something wrong with the processing. Do I need to close the pr and resubmit it?

zzzhouuu:
@rchincha Sorry, I've taken care of it

rchincha (approved):
lgtm

PR #3055, opened 2025-03-28
fix: remove unneeded double locks in redis implementation of metadb

andaaron:

### DIFF
--- a/pkg/meta/redis/redis.go
+++ b/pkg/meta/redis/redis.go
@@ -750,8 +750,7 @@ func (rc *RedisDB) SetRepoReference(ctx context.Context, repo string,
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/meta/redis/redis.go
```diff
@@ -750,8 +750,7 @@ func (rc *RedisDB) SetRepoReference(ctx context.Context, repo string,
 		return err
 	}
 
-	locks := []string{rc.getImageLockKey(imageMeta.Digest.String()), rc.getRepoLockKey(repo)}
-	err = rc.withRSLocks(ctx, locks, func() error {
```
rchincha, line 754: what exactly is "double locking" here?

andaaron: We take both an image lock and a repo lock at the same time for the same goroutine. Anyway, we don't need to lock at the same time.

rchincha (approved):

PR #3074, opened 2025-04-06
feat/gql-proxy: new approach for GraphQL local scale out

vrajashkr:
**What type of PR is this?**
feature

Towards #2434

Previously, only dist-spec APIs were supported for scale-out as in a shared storage environment, the metadata was shared and any instance could correctly respond to the GQL queries as all the data is available.

In a local scale-out cluster deployment, the metadata store, in addition to the file storage is isolated to each member in the cluster. Due to this, there is a need to proxy the GQL queries as well for UI and client requests to work as expected.

This change introduces a new GQL proxy + a generic fan-out handler for GQL requests.

Just manual testing with local setup for now. Real testing TODO.

No

TODO

```release-note

```


vrajashkr:
A previous attempt at this was made in #2588 

How is this different?
- Instead of a dedicated handler for each operation, in this PR, an approach is explored where there are generic and specific handlers.
- Generic handlers cater to common use cases such as fan-out proxy for GQL or single target Proxying
- Specific handlers cater to specific use cases which do not align with a generic handler
- GQL schema-merging was done manually in the previous PR, however, in this PR, a generic approach has been taken to dynamically merge the responses by treating them as maps instead of dedicated structs. This allows the code to be reusable across all types of handlers and reduces the amount of code to be maintained.

Note: Pagination is still pending
Note: The code itself is just in a Proof of concept state. Quite a bit of clean up is needed.

pkg/extensions/search/gql_proxy/gql_proxy.go
```diff
@@ -0,0 +1,105 @@
...
+// queries when running in cluster mode without shared storage (each instance has its own metadata).
+// Requests are only proxied in local cluster mode as in this mode, each instance holds only the
+// metadata for the images that it serves, however, in shared storage mode,
+// all the instances have access to all the metadata so any can respond.
+func GqlProxyRequestHandler(
+	config *config.Config,
+	log log.Logger,
+	gqlSchema *ast.Schema,
+) func(handler http.Handler) http.Handler {
+	proxyFunctionalityMap := map[string]GqlScaleOutHandlerFunc{
```
vrajashkr, line 26: @rchincha this map statically stores which GQL operation needs which kind of handler.

There will be a couple of generic handlers such as fanout and some specific handlers if any of the operations need custom behavior.

What do you think about this approach?

It's better than last time since we don't need to maintain a separate handler for each operation type, but I'm open to more ideas on making this better.

pkg/extensions/search/gql_proxy/handler_utils.go
```diff
@@ -0,0 +1,69 @@
...
+
+	_, err = response.Write(responseBody)
+	if err != nil {
+		http.Error(response, "failed to write response", http.StatusInternalServerError)
+
+		return
+	}
+}
+
+func deepMergeMaps(a, b map[string]any) map[string]any {
```
vrajashkr, line 28: @rchincha this is the new approach for aggregating the data. Since the response is a JSON, we can aggregate the data as a map type with individual logic for the embedded types - nested map, numeric types, and arrays.

This should be common across all handlers and may change when pagination comes into picture.

pkg/extensions/search/search_test.go
```diff
@@ -4508,6 +4517,118 @@ func TestGlobalSearch(t *testing.T) { //nolint: gocyclo
 	})
 }
 
+func TestGlobalSearchWithScaleOutProxyLocalStorage(t *testing.T) {
```
vrajashkr, line 4520: This seems to have failed:
```
{"level":"error","error":"address 127.0.0.135443: missing port in address","goroutine":3375,"caller":"zotregistry.dev/zot/pkg/api/controller.go:78","time":"2025-04-06T19:49:03.630787471Z","message":"failed to get member socket"}
E
Errors:

  * zotregistry.dev/zot/pkg/extensions/search/search_test.go 
  Line 65: - failed to get member socket 
  goroutine 3375 [running]:
  	github.com/smartystreets/goconvey@v1.8.1/convey/reporting/reports.go:150 +0x47
  	github.com/smartystreets/goconvey@v1.8.1/convey/reporting/reports.go:123 +0xb8
  	github.com/smartystreets/goconvey@v1.8.1/convey/context.go:253 +0xfa
  panic({0xe1e7660?, 0x11647a80?})
  	runtime/panic.go:787 +0x132
  zotregistry.dev/zot/pkg/api.NewController(0xc0015c22c0)
  	zotregistry.dev/zot/pkg/api/controller.go:79 +0xdbd
  zotregistry.dev/zot/pkg/extensions/search_test.makeController(...)
  	zotregistry.dev/zot/pkg/extensions/search/search_test.go:65
  zotregistry.dev/zot/pkg/extensions/search_test.TestGlobalSearchWithScaleOutProxyLocalStorage.func1()
  	zotregistry.dev/zot/pkg/extensions/search/search_test.go:4547 +0xbaa
  	github.com/smartystreets/goconvey@v1.8.1/convey/discovery.go:89 +0x2f
  	github.com/smartystreets/goconvey@v1.8.1/convey/context.go:279 +0x2d1
  	github.com/smartystreets/goconvey@v1.8.1/convey/context.go:112 +0x1a5
  github.com/jtolds/gls.(*ContextManager).SetValues.func1(0x0)
  	github.com/jtolds/gls@v4.20.0+incompatible/context.go:97 +0x66a
  github.com/jtolds/gls.EnsureGoroutineId.func1()
  	github.com/jtolds/gls@v4.20.0+incompatible/gid.go:24 +0x39
  github.com/jtolds/gls._m(0x0, 0xc001e31530)
  	github.com/jtolds/gls@v4.20.0+incompatible/stack_tags.go:108 +0x39
  github.com/jtolds/gls.github_com_jtolds_gls_markS(0x0, 0xc001e31530)
  	github.com/jtolds/gls@v4.20.0+incompatible/stack_tags.go:56 +0x34
  github.com/jtolds/gls.addStackTag(...)
  	github.com/jtolds/gls@v4.20.0+incompatible/stack_tags.go:49
  github.com/jtolds/gls.EnsureGoroutineId(0xc001afb9b0)
  	github.com/jtolds/gls@v4.20.0+incompatible/gid.go:24 +0x165
  github.com/jtolds/gls.(*ContextManager).SetValues(0xc000193200, 0xc001afb950, 0xc001bc1380)
  	github.com/jtolds/gls@v4.20.0+incompatible/context.go:63 +0x285
  	github.com/smartystreets/goconvey@v1.8.1/convey/context.go:107 +0x553
  	github.com/smartystreets/goconvey@v1.8.1/convey/doc.go:76 +0x9b
  zotregistry.dev/zot/pkg/extensions/search_test.TestGlobalSearchWithScaleOutProxyLocalStorage(0xc0016c0000)
  	zotregistry.dev/zot/pkg/extensions/search/search_test.go:4523 +0x10f
  testing.tRunner(0xc0016c0000, 0x107b1d98)
  	testing/testing.go:1792 +0x226
  created by testing.(*T).Run in goroutine 1
  	testing/testing.go:1851 +0x8f3
```

I'll take a look at it.

vrajashkr: Should be fixed now.

examples/scale-out-cluster-local/config-cluster-member0.json
```diff
@@ -0,0 +1,31 @@
+{
+  "distSpecVersion": "1.1.0",
+  "storage": {
+    "rootDirectory": "./workspace/zot/data/mem1",
```
rchincha, line 4: We have to figure out a scheme to append a member path.
Would like to have a single zot configuration that folks don't have to tweak.

vrajashkr:
Latest analysis of GQL queries:

**CVEListForImage**
- Proxy-once to target
- uses "repository:tag" or "repository@digest"

**CVEDiffListForImages**
- image and compared image names in "repository:tag" or "repository@digest" format.
- challenging as the images may be on different servers.

**ImageListForCVE**
- fan-out to all members

**ImageListWithCVEFixed**
- fan-out to all members

**ImageListForDigest**
- fan-out to all members

**RepoListWithNewestImage**
- fan-out to all members

**ImageList**
- use the repo param to proxy once to the target server.

**ExpandedRepoInfo**
- use the repo param to proxy once to the target server.

**GlobalSearch**
- fan-out to all members
- future optimization: if the repo names are known, these don't need to be fanned-out.

**DerivedImageList**
- uses "repository:tag" format.
- however, expected output is all images using the given one so a fan-out is needed.

**BaseImageList**
- image uses "repository:tag" format.
- as expectation is to get all images, a fan-out is needed.

**Image**
- image uses "repository:tag" format
- use repo value to proxy once to target

**Referrers**
- has a repo as an argument.
- use repo value to proxy once to target

**StarredRepos**
- fan-out to all members

**BookmarkedRepos**
- fan-out to all members

pkg/extensions/search/gql_proxy/generic_fan_out_gql_handler.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+func fanOutGqlHandler(config *config.Config, _ map[string]string, response http.ResponseWriter, request *http.Request) {
+	// Proxy to all members including self in order to get the data as calling next() won't return the
+	// aggregated data to this handler.
+	finalMap := map[string]any{}
+
+	for _, targetMember := range config.Cluster.Members {
+		proxyResponse, err := proxy.ProxyHTTPRequest(request.Context(), request, targetMember, config)
+		if err != nil {
```
rchincha, line 19: Do we want to return failure even if just one member fails?
I think if one member responds, we should return that and swallow errors maybe with some indicator somewhere. Maybe logs?  HTTP status - 206 Partial Content could work also since this is our own API.

vrajashkr: That sounds like a good idea. One thought that I had in mind is that instead of swallowing the error, perhaps we could append an error to the Errors list key in the GQL response and send it to the client so there is awareness of some error in the system.

The client can choose to ignore the error and use the valid data in the response, or ideally, show both the valid data as well as indicate that there were some errors in processing. With this approach status 206 could be the return status as you've suggested.

What do you think?

vrajashkr:
Some observations:
`DerivedImageList` and `BaseImageList` both need the metadata of the input image available to perform their logic so can't be supported with local scale-out.

Currently trying out implementing `CVEListForImage`. Working on writing tests for it which is trickier than expected since trivy needs to be running to scan the vulnerable layers. Thinking about some alternative approaches to this test case (and for future CVE related GQL test cases).

vrajashkr:
All the artifact related GQL queries are supported in the latest commit except for the operations that are explicitly marked as unsupported.

Things that need to be worked on:
- Pagination
- Standardized error handling
- Handling for Partial responses

vrajashkr:
Looks like the re-push had some more intermittent failures:
- stateless with minio and redis failed
- Additionally, an error in the extensions tests related to sync:
```
--- FAIL: TestSignatures (16.25s)
    sync_test.go:4653: 41151
    sync_test.go:5168: 33735
FAIL
```

Will force push another commit.

pkg/api/cluster_proxy.go
```diff
@@ -0,0 +1,85 @@
...
+	"net/http"
+
+	"github.com/gorilla/mux"
+
+	"zotregistry.dev/zot/pkg/api/constants"
+	"zotregistry.dev/zot/pkg/cluster"
+	"zotregistry.dev/zot/pkg/proxy"
+)
+
+// ClusterProxy wraps an http.HandlerFunc which requires proxying between zot instances to ensure
```
rchincha, line 15: This moved from pkg/api/proxy.go to here? Reason?

examples/scale-out-cluster-local/config-cluster-member0.json
rchincha: Are you able to run a `zb` benchmark to show things scaling up?

pkg/api/cluster_proxy.go
vrajashkr: This content was moved because the original api/proxy.go was split to move the common functionality that the GQL Proxy required into a new package called 'proxy' to avoid cyclic dependency.

The new file name here was `cluster_proxy.go` arbitrarily.

examples/scale-out-cluster-local/config-cluster-member0.json
vrajashkr: We did see some results when we developed scale out for the dist-spec APIs, however, can `zb` benchmark GQL queries? I believe it was only for dist-spec APIs if I recall correctly.

vrajashkr:
Couple of updates in the last push:
1. Rebased to the latest main branch
2. Part of the `TestRepoListWithNewestImageWithScaleOutProxyLocalStorage` was previously commented out due to inconsistency in tag response. After debugging in a single zot setup, I learned that the Created timestamp is used for figuring out the newest tag and not the timestamp of a tag itself being pushed - it's data in the Config that matters. This has been fixed in the test code by adding a delay to the Created timestamp and the test is now passing.

vrajashkr:
While Unit Tests are working, manual testing hit a snag:
When zot uses Auth for the UI (session and cookie are involved), the session is only available in the zot instance where the initial login was performed.
Since the load balancer sets a sticky cookie, the user will always be sent to the same instance, but internally, the GQL would be proxied to one or more instances where the session is not present and this fails Auth there.

The error seen is:
```
"error":"securecookie: the value is not valid","goroutine":729,"caller":"zotregistry.dev/zot/pkg/api/authn.go:858","time":"2025-06-19T17:35:15.281808846Z","message":"failed to decode existing session"
```
At this point, it would seem that there is a need for some manner to distribute the session data across the members as well, but this needs further discussion.

vrajashkr:
Ecosystem client tools tests appear to have failed.

I see `# {"level":"error","error":"listen tcp 0.0.0.0:39642: bind: address already in use","time":"2025-06-21T19:50:03.518264715Z","message":"failed to start controller, exiting"}`

After that message, a ton of errors show up with failed connections to the server. Will re-trigger the commit another time.

rchincha:
@vrajashkr restarted the failing test.

vrajashkr:
Previously on proxying support for GQL, there was a need for shared sessions so that other zot instances know of the login and won't return a session not found error.

Once that was supported in #3345 , the next error in the flow was to do with some user metadata that is part of the login and AuthNZ flow.

When the UI login request shows up, the following happens:
tryAuthnHandlers -> (no session header) basicAuthn function -> create new access controller -> set username, set groups, and save this data to metaDB.

Since this requires the metaDB to be shared (for the user metadata), GQL proxying is no longer required as any instance would be able to equivalently respond to GQL search queries.

For missing images, there is a thought of enabling sync to and from other members so images can be fetched. (Needs testing).

This PR can remain in this state and wait for auto-close. Once some testing and document updates are done, will consider closing the PR. I'll keep the branch around in case we want to revisit this in future.

The code where other zot members fail when the metaDB is not shared:

```
groups, err := ctlr.MetaDB.GetUserGroups(request.Context())
	if err != nil {
		ctlr.Log.Err(err).Str("identity", identity).Msg("failed to get user profile in DB")

		return false, err
	}
```

vrajashkr:
Closing this out for now as per the previous comment. We'll come back to this in future if the user metadata behavior changes or if there is an opportunity to implement this in a different manner.

PR #3128, opened 2025-04-30
Add ability to sync specific architectures.

lukaszraczylo:
**What type of PR is this?**

Feature

Adds ability to specify the images architecture for mirroring.

Instead of pulling 10s of GB of images for different architectures, you can now pull only architectures you use.

**Automation added to e2e**:

Meh.

```release-note

```


pkg/extensions/config/sync/config.go
```diff
@@ -33,7 +33,8 @@ type RegistryConfig struct {
 	RetryDelay       *time.Duration
 	OnlySigned       *bool
 	CredentialHelper string
-	PreserveDigest   bool // sync without converting
+	PreserveDigest   bool     // sync without converting
+	Architectures    []string // filter architectures during sync
```
rchincha, line 37: https://github.com/opencontainers/image-spec/blob/main/image-index.md#platform-variants

To be more accurate, the following:
```
"platform": {
        "architecture": "ppc64le",
        "os": "linux"
      }
  ```

**Comment by rchincha** [comment]
@lukaszraczylo thanks for your PR. This has been a long standing ask from the community.

**Comment by rchincha** [comment]
> @rchincha I'll work on the suggestions as well, and I'd PR it faster, but I'm using zotregistry for past two days :D

If you able to test locally, you can iterate a lot quicker.
Also do take a look at the CI and failures to understand what would matter to get this PR merged.

File: pkg/extensions/config/sync/config.go (hunk 1 above)
Comment (lukaszraczylo) [reply pkg/extensions/config/sync/config.go]: Added in https://github.com/project-zot/zot/pull/3128/commits/0ab2c7a61360502bcfca13638f5a7a8e8f60b271

**Comment by lukaszraczylo** [comment]
@rchincha sorted. I also have an idea re: resources usage improvement, but I'll leave it for the next PR to not scope creep too much :)

File: examples/config-sync-architectures.json [hunk 2]
```diff
@@ -0,0 +1,100 @@
...
+							"destination": "/repo",
+							"stripPrefix": true
+						},
+						{
+							"prefix": "/repo3/**"
+						},
+						{
+							"prefix": "/repo4/**",
+							"tags": {
+								"excludeRegex": ".*-(amd64|arm64)$"
```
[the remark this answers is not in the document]
lukaszraczylo: I'd say "meh". Platform info in the image name does not mean anything to be honest. I've seen ( and even accidentally released myself ) quite a few images tagged with one platform declared as part of the tag and image itself / binaries within being different platform.

pkg/extensions/sync/destination.go
```diff
@@ -29,26 +30,86 @@ import (
...
 
+// Platform represents an OS/architecture combination
+type Platform struct {
+	OS           string
+	Architecture string
+}
+
+// ParsePlatform parses a platform string into a Platform struct
+// The string can be in the format "os/arch" or just "arch"
+func ParsePlatform(platform string) Platform {
```
rchincha, line 41: Ok, instead of a nested map, it is a flat string with a '/' separator

pkg/extensions/config/sync/config.go
```diff
@@ -33,7 +33,9 @@ type RegistryConfig struct {
 	RetryDelay       *time.Duration
 	OnlySigned       *bool
 	CredentialHelper string
-	PreserveDigest   bool // sync without converting
+	PreserveDigest   bool     // sync without converting
+	Architectures    []string `mapstructure:",omitempty"` // filter architectures during sync (DEPRECATED: use Platforms instead)
```
rchincha, line 37: Would just drop Architectures field and keep only the Platforms field. Makes config too complex/verbose.

Realistically, this is a selector that narrows down choices.

rchincha:
> @rchincha sorted. I also have an idea re: resources usage improvement, but I'll leave it for the next PR to not scope creep too much :)

Looking forward to that also.

pkg/extensions/sync/destination.go
```diff
@@ -29,26 +30,86 @@ import (
...
+	parts := strings.Split(platform, "/")
+	if len(parts) == 2 {
+		return Platform{
+			OS:           parts[0],
+			Architecture: parts[1],
+		}
+	}
+	// If only one part, assume it's the architecture
```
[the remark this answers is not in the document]
lukaszraczylo: Sorted in https://github.com/project-zot/zot/pull/3128/commits/3eecc290ecd167becd0dec432d91c11b2e82c0cc

pkg/extensions/config/sync/config.go
lukaszraczylo: Sorted in https://github.com/project-zot/zot/pull/3128/commits/3eecc290ecd167becd0dec432d91c11b2e82c0cc

lukaszraczylo:
@rchincha I resolved all the issues / suggestions from what I see :) If anything else - I'll pick it up tomorrow

rchincha (commented):
Call this config-sync-platforms.json instead

rchincha (commented):
test/blackbox tests - just make one file right?

lukaszraczylo:
@rchincha that's done. Also rewrote the commit messages to adhere to the rules

pkg/extensions/sync/destination.go
```diff
@@ -29,26 +30,86 @@ import (
...
+type Platform struct {
+	OS           string
+	Architecture string
+}
+
+// ParsePlatform parses a platform string into a Platform struct
+// The string can be in the format "os/arch" or just "arch"
+func ParsePlatform(platform string) Platform {
+	parts := strings.Split(platform, "/")
+	if len(parts) == 2 {
```
andaaron, line 43: Note there are also cases for 2 `/` separators, meaning 3 parts, if variant is specified in the platform, for example:

```
    "Descriptor": {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "digest": "sha256:116d243c6346c44f3d458e650e8cc4e0b66ae0bcd37897e77f06054a5691c570",
      "size": 742,
      "platform": {
        "architecture": "arm",
        "os": "linux",
        "variant": "v7"
      }
    },
```

This should be taken into account.

lukaszraczylo: Added in https://github.com/project-zot/zot/pull/3128/commits/dd6a4cf955b6defac1efcb3838d35113c333cd90

pkg/extensions/sync/destination.go
```diff
@@ -29,26 +30,102 @@ import (
...
 	log             log.Logger
+	config          *syncconf.RegistryConfig // Config used for filtering architectures
 }
 
 func NewDestinationRegistry(
 	storeController storage.StoreController, // local store controller
 	tempStoreController storage.StoreController, // temp store controller
 	metaDB mTypes.MetaDB,
 	log log.Logger,
+	config ...*syncconf.RegistryConfig, // optional config for filtering
```
andaaron, line 114: Why is this optional?

pkg/extensions/sync/destination.go
```diff
@@ -254,6 +387,25 @@ func (registry *DestinationRegistry) copyManifest(repo string, desc ispec.Descri
...
+			// Update the manifest content with the filtered list
+			updatedContent, err := json.Marshal(indexManifest)
+			if err != nil {
+				registry.log.Error().Str("errorType", common.TypeOf(err)).
+					Err(err).Str("repository", repo).
+					Msg("failed to marshal updated index manifest")
+				return err
+			}
+
+			manifestContent = updatedContent
```
andaaron, line 406: Note this will break any referrer pointing to the original index digest.
Unless the same digest is kept in the filename, but a different one would be computed based on the new filename, are we doing this?

pkg/extensions/sync/service.go
```diff
@@ -447,6 +450,67 @@ func (service *BaseService) SyncRepo(ctx context.Context, repo string) error {
...
+	}
+
+	// If platform is nil, include it
+	if platform == nil || len(platformSpecs) == 0 {
+		return true
+	}
+
+	for _, spec := range platformSpecs {
+		parts := strings.Split(spec, "/")
+		if len(parts) == 2 {
```
andaaron, line 474: 1. Do we need to handle the variant here as well?
2. Maybe we should have a platform method for matching against a different platform object (including partial matches in case filtering is done just for a subset of platform attributes of course)?

pkg/extensions/sync/service.go
```diff
@@ -447,6 +450,67 @@ func (service *BaseService) SyncRepo(ctx context.Context, repo string) error {
...
+
+		// If we got here, it's a match
+		return true
+	}
+
+	return false
+}
+
+// shouldIncludeArchitecture determines if an architecture should be included in the sync
+// DEPRECATED: Use shouldIncludePlatform instead
```
andaaron, line 503: Why commit this code if it is deprecated?

pkg/extensions/sync/service.go
```diff
@@ -479,11 +554,78 @@ func (service *BaseService) syncRef(ctx context.Context, localRepo string, remot
...
+						Architecture: man.GetDescriptor().Platform.Architecture,
+						OS:           man.GetDescriptor().Platform.OS,
+						OSVersion:    man.GetDescriptor().Platform.OSVersion,
+						OSFeatures:   man.GetDescriptor().Platform.OSFeatures,
+						Variant:      man.GetDescriptor().Platform.Variant,
+					}
+					if !service.shouldIncludePlatform(platform) {
+						platformDesc := platform.Architecture
+						if platform.OS != "" {
+							platformDesc = platform.OS + "/" + platform.Architecture
```
andaaron, line 600: 1. Do we need to handle variant here?
2. Maybe we should have a Platform method to return the os/arch/variant string for that instance.

andaaron (commented):
@eusebiu-constantin-petu-dbk , please review this PR

@rchincha when sync-ing an index, do we want an exact copy, or a new index containing just a subset of the synced manifests? The image would contain invalid references in the 1st case, and digest would change in the 2nd case. There is also a 3rd option of saving the new content with the old digest, but I don't know what the implications would be, since it doesn't respect the content addressability.

rchincha:
@lukaszraczylo more folks are hitting this and complaining.

https://cloud-native.slack.com/archives/C03EGRE4QGH
^ pls join the slack channel.

lukaszraczylo:
@rchincha I may try to sit down on it this weekend although I'm a little bit busy with family and baby on the way so no promises

rchincha:
@lukaszraczylo pls do let us know if you are able to continue work on this PR.

lukaszraczylo:
@rchincha, sure - I can resume the work, although I'm disappointed with the review style and last-minute submissions of the change requests, as they've already been working on the previous comments, making it a bit of a wild goose chase.

rchincha:
> @rchincha, sure - I can resume the work, although I'm disappointed with the review style and last-minute submissions of the change requests, as they've already been working on the previous comments, making it a bit of a wild goose chase.

@lukaszraczylo feedback noted. Just so you are aware, the contributors and maintainers of this project work from diverse locations/timezones and mostly contributing their free time, hence feels a little choppy at times - overall the goal is to continue delivering a high quality project.

andaaron:
> @rchincha when sync-ing an index, do we want an exact copy, or a new index containing just a subset of the synced manifests? The image would contain invalid references in the 1st case, and digest would change in the 2nd case. There is also a 3rd option of saving the new content with the old digest, but I don't know what the implications would be, since it doesn't respect the content addressability

@rchincha let's sync on this before more effort is put in

rchincha:
For example, upstream repo has:
original_index[manifest_archA, manifest_archB, manifest_archC]

Assuming we want to sync/mirror only archA, we can either make:
1) original_index[manifest_archA, manifest_archB, manifest_archC] with missing manifest_archB and manifest_archC blobs, but this may require GC fixes potentially. Referrers etc which are downloaded will remain intact.

(OR)

2) new_index[manifest_archA], but this may have client implications because of digest mismatches potentially.

rchincha:
There appears to be a notion of "sparse index" which is option 1. We should probably implement that. We should fix any GC issues that pop up.
However, let's also ask the question, what happens if configuration is changed to reduce/increase/alter the arch pattern?

rchincha:
https://github.com/opencontainers/distribution-spec/pull/310

andaaron:
Ok, let's go for the sparse index. We'll need to make a list of all cases where we are walking repo content and check if we handle missing references nicely.

andaaron:
@lukaszraczylo I have merged https://github.com/project-zot/zot/pull/3503 today.
It should take care of all the issues sparse indexes may cause in other features after they get copied to zot storage.

apollo13:
@lukaszraczylo Are you still working on this? Would love to limit the data we pull from docker hub as well :)

Nemric:
Hi, I'm a new zot user, then I did find this issue/pr
I don't know if this is the right place to discuss this, but I try ^^
I don't really know how it works under the hood but when I perform an image pull with e.g. podman pull, I only get the right arch (amd64) image
So, as I don't have any other arch here, I would add amd64 sync filter, but if later I need another arch, I'll have to change the filter to get, e.g. arm arch and then I'll get amd64 and arm arch for all images
Isn't it possible to "filter on demand" as I think 'podman pull' will ask for the right arch, then only this one should be cached/sync in zot ?

mrudnoru:
Hi @rchincha @andaaron This pr is a complete **must have** for zot production users. We are aiming to replace our internal regirstry Harbor with Zot and without platform filtering **it would not be possible**.
Why? We run zot as a single-arch (`linux/amd64`) pull-through mirror, and syncing multi-arch images pulls several architectures we never serve, wasting storage and upstream bandwidth and slowing on-demand sync **around 3x**.

Is this PR still active? If it's stalled, we have a small focused implementation (~90 lines) that filters image-index manifests by `os/arch/variant` via a `platforms` registry config key (e.g. `["linux/amd64"]`) — happy to open it as a separate PR or help get this one over the line, whichever maintainers prefer.

Please let me know 👍

mrudnoru:
Or I could just continue this PR to apply the requests from review for it @andaaron @rchincha what do u think?

lukaszraczylo:
@mrudnoru feel free to continue with this PR :)

Nemric:
@mrudnoru I don't know how you plan to work on this PR
My best wish would be to implement something I tried to explain in this comment https://github.com/project-zot/zot/pull/3128#issuecomment-3984436156 in the case of `extensions.sync.registries.ondemand=true` when using pull-through mirror
That could avoid having to change the config file in the case of a new supported arch in the infra and prevent downloading all configured defined images arch from a remote registry when only some, e.g. `arm`, are asked
Hope being clear :smile:

ChristianCiach:
I agree with @Nemric. I think this should be possible without additional configuration. Harbor seems to support this just fine?

andaaron:
Superseded by https://github.com/project-zot/zot/pull/4410

Issue #3182, opened 2025-06-03
[Bug]: Anonymous access returns 401


rchincha:
### Discussed in https://github.com/project-zot/zot/discussions/3181

<div type='discussions-op-text'>

<sup>Originally posted by **mottetm** June  3, 2025</sup>
Hi there!

I'm trying to set up a registry with read-only anonymous access. My configuration goes as follow:

```json
{
    "storage": {
        "rootDirectory": "/var/lib/zot/data"
    },
    "http": {
        "realm": "zot",
        "address": "0.0.0.0",
        "port": "5000",
        "compat": [
            "docker2s2"
        ],
        "auth": {
            "htpasswd": {
                "path": "/var/lib/zot/htpasswd"
            }
        },
        "accessControl": {
            "metrics": {
                "users": [
                    "hello"
                ]
            },
            "repositories": {
                "**": {
                    "anonymousPolicy": [
                        "read"
                    ],
                    "defaultPolicy": [
                        "read",
                        "create"
                    ]
                }
            }
        }
    },
    "log": {
        "level": "debug"
    },
}
```

When executing an authenticated request, everything goes well:
```json
# crane manifest localhost:5000/path/to/image:v2
{"level":"info","module":"http","component":"session","clientIP":"127.0.0.1:42976","method":"GET","path":"/v2/","statusCode":401,"latency":"0s","bodySize":266,"headers":{"Accept-Encoding":["gzip"],"User-Agent":["crane/(devel) go-containerregistry/(devel)"],"X-Envoy-Expected-Rq-Timeout-Ms":["15000"],"X-Forwarded-Proto":["http"],"X-Request-Id":["e6c5f659-2c09-4d31-ba9a-a6e19e15c5d1"]},"goroutine":215,"caller":"zotregistry.dev/zot/pkg/api/session.go:137","time":"2025-06-03T09:18:54.377764572Z","message":"HTTP API"}
{"level":"info","identity":"hello","goroutine":215,"caller":"zotregistry.dev/zot/pkg/api/authn.go:141","time":"2025-06-03T09:18:54.445448425Z","message":"user profile successfully set"}
{"level":"info","identity":"hello","goroutine":215,"caller":"zotregistry.dev/zot/pkg/api/authn.go:141","time":"2025-06-03T09:18:54.509713746Z","message":"user profile successfully set"}
[...]

PR #3183, opened 2025-06-03
Fix metrics authorization middleware bleed

mottetm:
**What type of PR is this?**

bug fix

#3182 

Fixes `extension_metrics_disabled.go` to correctly isolate the authz middleware when the metrics extension is disabled.


andaaron (approved):

rchincha:
| Merging is blocked

@mottetm pls sign and sign-off your commit.

mottetm:
@rchincha I just needed to set up the key for this machine 🙈

rchincha:
https://github.com/project-zot/zot/blob/main/pkg/extensions/extension_metrics.go#L33C2-L38C3

^ why not make it identical to this?

mottetm:
In term of checking `config.IsMetricsEnabled()`? Or creating the local `extRouter`? I assumed the difference were there for a reason, I just wanted to prevent the authz middleware from bleeding out.

andaaron:
> https://github.com/project-zot/zot/blob/main/pkg/extensions/extension_metrics.go#L33C2-L38C3
> 
> ^ why not make it identical to this?

If the metrics extension is disabled at build time it makes no sense for the extensions configuration to do something. That configuration is only applicable if zot was compiled with metrics/prometheus. In the other case (including minimal), the endpoint is always accessible.

rchincha:
Agreed. I meant without the if- part.
Also, can we add tests to validate this pls?

@mottetm should be quick since we should already have most tests both for extension enabled and disabled case - will validate this path now, and protect against breakages going fwd.

mottetm:
@rchincha I added a test to the blackbox tests, that's the only place where I could think of putting it. Let me know if you had something else in mind.

rchincha (approved):
lgtm

Issue #2487, opened 2024-06-20
[Feat]: Add health endpoint


AndersBennedsgaard:
### Is your feature request related to a problem? Please describe.

I do not want to allow anonymous access to my Zot instance, and I also do not want to set the authorization header (using `htpasswd`) for my Kubernetes liveness probes, since that will expose a valid authentication header to anyone that has read access to deployments on my Kubernetes clusters.

### Describe the solution you'd like

Add a `/health` or `/healthz` endpoint to Zot, which just returns HTTP 200. This shows that the server is up and running

### Describe alternatives you've considered

Remove the liveness probes, or set a [TCP liveness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-tcp-liveness-probe) up instead, to test if the application listens on the configured port.

### Additional context

Multiple issues and other references have been created for this, but they have either been closed with no explanation or explain that it is not needed (which I think it is. It is pretty standard practice for applications running in the cloud).
Some references to existing issues/PRs include: https://github.com/project-zot/zot/issues/1767, https://github.com/project-zot/zot/issues/196, https://github.com/project-zot/zot/issues/2087, and https://github.com/project-zot/zot/pull/197

rchincha:
@AndersBennedsgaard thanks for trying out zot!

Do you want to consider a `liveness command` instead?
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command
^ this would give you more flexibility with `curl` for example.

AndersBennedsgaard:
That would definitely be a possibility, yes. That would allow for you to mount the credentials as an environment variable from a Secret, instead of having it as plain-text in the Deployment.
However, there is no `curl`, `wget`, or anything similar packaged into the Zot container at the moment since it is based on a distroless image, so if I wanted to use `exec` probes I would have to create a custom build with it.

But I would still prefer a health-check endpoint, to remove the requirement for managing extra credentials just for checking if the service is up. I could also just configure the ingress controller to expose the healthcheck path, if I didn't want users to access it

ravi-ciroos:
@AndersBennedsgaard - I am using as my liveness probe for now.

```zli status --url "$ZOT_REGISTRY" --user "$ZOT_USER_CREDS" | grep online```

asgeirn:
[...]

PR #3228, opened 2025-06-27
feat: healthz server

asgeirn:
**What type of PR is this?**

feature

#2487 

Ref #3214, adds a separate HTTP listener responsible for health endpoints to be used by Kubernetes probes or similar.  This listener has only health routes, and no authorization.

Deployed on my own instance with an updated Helm chart which produces probes pointing to this endpoint instead.

**Automation added to e2e**:

Deployments in Kubernetes not using the Helm chart must manually update the liveness, readiness and startup probes if present.

```release-note
feat: Add health endpoint for startup, readiness, and liveness probes.
```


asgeirn:
@rchincha would very much like your feedback on the best strategy of integrating this, and where it should potentially get inputs to determine the status of the probes.

For now, the controller marks the startup probe as ready once it's ready to start the HTTP server.

pkg/api/config/config.go
```diff
@@ -118,6 +118,7 @@ type HTTPConfig struct {
 	Address       string
 	ExternalURL   string `mapstructure:",omitempty"`
 	Port          string
+	HealthPort    string
```
andaaron: Why not use the same port?

zot starts responding to requests and the api becomes available after it finishes going through the images present in the storage (OCI layouts) and making sure they are available in the DB.  Not sure if that would be taken into account if we have this separate server on a different port.

asgeirn: My main thinking is to quickly start up the probes, avoiding interfering with the auth mechanisms and also limiting availability of the endpoint.  

It currently maps to `Address`:`HealthPort`, but I was considering explicitly binding to localhost as well.

If the startup is slow and the probes are not available soon enough, Zot could end up in a restart loop since the failed probes cause a restart.

If you could point me to where the image search is completed I could add a callback to the readiness state at that point.  This would ensure the instance won't get traffic until then.

andaaron: I think we're ok, `Controller.Run()` is called after `Controller.Init()` which takes care of the things I mentioned.

> If the startup is slow and the probes are not available soon enough, Zot could end up in a restart loop since the failed probes cause a restart.

Isn't this what readiness probes are for? zot shouldn't tell the world it is ready if it isn't.
In any case, it should be fine given this `Run()` is called after `Init()`

asgeirn: Yes, that's exactly what readiness probes are for.

Where it gets tricky is the relationship between this and the liveness and startup probes, as the kubelet will terminate containers which fail to respond to these.

I'll update to enable the readiness probe in controller.Run() and the startup probe in controller.Init()

rchincha: readiness - zot is ready to serve requests
liveness - are you still alive?

rchincha: We have endpoints that are different. Let's not open a separate port for this.
We should try to fix the startup delays if possible.

asgeirn: The startup delay is not the only issue. In one setup I administer we configured LDAP authentication to an Active Directory server. 

It got constantly restarted since the /v2/ endpoint regularly had over three second response times. 

I can have a look at tying it to the same HTTP listener without any interference from the authentication handlers.

andaaron: But why have the endpoint under `/v2`? 
Why not directly under `/` same as https://github.com/project-zot/zot/blob/main/pkg/api/routes.go#L217 (but before this line in order not to interfere with default routing for UI)?
`/v2` is authenticated, `/` is not

asgeirn: Yeah, I'll add the hendler at the root.  My comment describes how the existing probes work without a separate health hander.

asgeirn: I've added the health endpoints to the main listener now, and confirmed that they work properly.  A side effect of this change is that the endpoints are openly available, something I confirmed by curling my own instance.

pkg/api/constants/consts.go
```diff
@@ -14,6 +14,7 @@ const (
 	BinaryMediaType              = "application/octet-stream"
 	DefaultMetricsExtensionRoute = "/metrics"
 	AppNamespacePath             = "/zot"
+	HealthPath                   = "/health"
```
andaaron: Is this needed?
it doesn't seem to be used in the routes. Do you intend to add it to the path?

asgeirn: Ups, merge mishap, I'll fix.

pkg/api/config/config.go
rchincha: @asgeirn firstly, thanks a lot for working on this PR. Apologies in advance if we go back and forth on this, but doing the right thing is important.

| A side effect of this change is that the endpoints are openly available, something I confirmed by curling my own instance.

We need to address concerns about authenticating these endpoints.

Here are the options (thinking loud) so we can debate:

1. Open separate port that only `kubelet` can access, so these ports are not exposed via a `Service` and hence not publicly available. There is no further authn/authz requirement. As long as these endpoints are controlled via config (and not enabled by default), should still be ok for bare-metal deployments.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-tcp-liveness-probe

This is a k8s-specific solution.

Personally, I am leaning this way and sorry if you have to revert some of the changes :(

2. Keep the latest implementation

This path is broadly useful (not just k8s) since tools like `nagios` can integrate. But must now be protected via authn/authz since port is shared. Maybe create a group like we have for `metrics` (for prometheus scraping)

For k8s, we can specify `httpHeaders:` in liveness and readiness probes

andaaron: @rchincha can you please explain why authn/authz are needed for these endpoints? They don't expose any information about images/user data or the underlying system (thinking metrics).

The whole point for implementing a new endpoint in addition to /v2 was to avoid the authentication workflow on /v2.

rchincha: We definitely don't need the /v2 path prefix.
Anyone can spin-read on this now?

asgeirn: Yes, the fact that anyone can spin-read this might be a concern if the actual checks performed become computationally intensive.

Right now they - at most - check a boolean property before returning a precreated error object.

If you'd like to DDOS Zot it would be much more effective to hammer the /v2/ endpoint with a random basic auth header, as this will trigger the entire suite of auth and autz logic.

Switching the Helm chart to produce TCP probes instead of httpGet would remove the symptoms I originally experienced that got me working on this. However, this would not be useful for layer 7 load balancers or application gateways as a readiness probe. 

On my tiny Zot instance I'm not able to distinguish the two implementations I've tried since controller.Init() completes in half a second. But for larger installations the separate listener would ensure the probes respond properly during initialization. 

So which option do you prefer?

1. Separate listener
2. Shared listener
3. Modify the Helm chart

rchincha:
@asgeirn can you add an example under examples/
so we can point out usage to folks who will be interested.

rchincha:
Also, wondering if we need to protect this endpoint with ratelimiting like so:
https://github.com/project-zot/zot/blob/main/examples/config-ratelimit.json#L9

OTOH, we don't protect /v2/ endpoint in the anonymous case either.

asgeirn:
We could, but currently that would be more complex logic than just simply responding. The endpoints only output a status code and a few static characters. Not even the text from the error is returned.

rchincha (approved):

andaaron:
Thanks again @asgeirn

rchincha:
@asgeirn is there a helm chart update needed for this?

asgeirn:
Yes, I'll make a PR for the requisite changes in the Helm chart as well.  It'll be draft until a Zot release is made supporting these new probes.

PR #3230, opened 2025-06-30
fix: GetNextRepository to use a list already scanned repositories as input

andaaron:
Using just the last repository is not enough as in the case when it is deleted (either by GC or some other way), GetNextRepository returns empty string causing the generator to be marked completed without any errors.

An alternative would have been to start over from the first repository, but this can take hours if multiple repositories need to be deleted, not to mention the processing power and I/O and S3 load this could take.

### DIFF
--- a/pkg/extensions/extension_scrub.go
+++ b/pkg/extensions/extension_scrub.go
@@ -28,18 +28,26 @@ func EnableScrubExtension(config *config.Config, log log.Logger, storeController
 			log.Warn().Msg("scrub interval set to too-short interval < 2h, changing scrub duration to 2 hours and continuing.") //nolint:lll // gofumpt conflicts with lll
 		}
 
+		processedRepos := make(map[string]struct{})
+
 		generator := &taskGenerator{
-			imgStore: storeController.DefaultStore,
-			log:      log,
+			imgStore:       storeController.DefaultStore,
+			log:            log,
+			processedRepos: processedRepos,
 		}
+
 		sch.SubmitGenerator(generator, config.Extensions.Scrub.Interval, scheduler.LowPriority)
 
 		if config.Storage.SubPaths != nil {
 			for route := range config.Storage.SubPaths {
+				processedRepos := make(map[string]struct{})
+
 				generator := &taskGenerator{
-					imgStore: storeController.SubStore[route],
-					log:      log,
+					imgStore:       storeController.SubStore[route],
+					log:            log,
+					processedRepos: processedRepos,
 				}
+
 				sch.SubmitGenerator(generator, config.Extensions.Scrub.Interval, scheduler.LowPriority)
 			}
 		}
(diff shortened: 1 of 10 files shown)


andaaron:
With regards to memory usage I made a few tests:

```go
package zot

import (
	"fmt"
	"strconv"
	"strings"
	"testing"
)

var result map[string]struct{}

func createMap(prefix string, size int) map[string]struct{} {
	m := make(map[string]struct{}, size)
	for i := 0; i < size; i++ {
		key := fmt.Sprintf("prefix%d", i)
		m[key] = struct{}{}
	}
	return m
}

func BenchmarkMapMemory(b *testing.B) {
	prefix := strings.Repeat("MyRepoName", 10)
	for _, size := range []int{10, 100, 1000, 10000, 100000, 1000000} {
		b.Run(strconv.Itoa(size), func(b *testing.B) {
			var temp map[string]struct{}
			b.ReportAllocs()
			b.ResetTimer()
			for i := 0; i < b.N; i++ {
				temp = createMap(prefix, size)
			}
			result = temp
		})
	}
}

```

And run them:
```bash
go test -bench MapMemory -benchmem -benchtime 10s
goos: linux
goarch: amd64
pkg: zotregistry.dev/zot
cpu: Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz
BenchmarkMapMemory/10-8                  4309728              2816 ns/op             584 B/op         14 allocs/op
BenchmarkMapMemory/100-8                  440666             25858 ns/op            4345 B/op        104 allocs/op
BenchmarkMapMemory/1000-8                  42486            274171 ns/op           75835 B/op       1750 allocs/op
BenchmarkMapMemory/10000-8                  3846           3237897 ns/op          674397 B/op      19778 allocs/op
BenchmarkMapMemory/100000-8                  322          36492087 ns/op         5893329 B/op     200005 allocs/op
BenchmarkMapMemory/1000000-8                  26         409411538 ns/op        79919123 B/op    2003847 allocs/op
PASS
ok      zotregistry.dev/zot     80.560s
```

I think this is reasonable memory usage.
In any case, even with the implementation in https://github.com/project-zot/zot/pull/3229, just the time to go over the same repos all over again would be too much (not to mention the disk I/O and S3 calls).

pkg/extensions/extension_scrub.go
```diff
@@ -28,18 +28,26 @@ func EnableScrubExtension(config *config.Config, log log.Logger, storeController
...
 		}
 
+		processedRepos := make(map[string]struct{})
+
 		generator := &taskGenerator{
-			imgStore: storeController.DefaultStore,
-			log:      log,
+			imgStore:       storeController.DefaultStore,
+			log:            log,
+			processedRepos: processedRepos,
```
rchincha, line 36: This is the biggest resource hit at scale?
Can we not start at random offset and not store this?

andaaron: If we start on random offset we will never know when the generator should be done

andaaron: It is not that much in term of memory consumption, as you can see in test I made above.

Issue #3243, opened 2025-07-03
[Feat]: Allow Openid clientid and clientsecret to be read from file


uwej711:
### Is your feature request related to a problem? Please describe.

Similar to https://github.com/project-zot/zot/pull/1778 (LDAP credentials), I think it would be good to read openid clientid and clientsecret from a file.

### Describe the solution you'd like

Similar to https://github.com/project-zot/zot/pull/1778

### Describe alternatives you've considered

Currently we have the clientsecret in the config json, which is a configmap in Kubernetes. It would be better in my opinion to ave secrets for that.

### Additional context

I can try to do a PR similar to https://github.com/project-zot/zot/pull/1778

rchincha:
@uwej711 Sure, do give us a PR. This pattern is most useful in k8s environments where Secrets can get mounted as files.

uwej711:
fixed with https://github.com/project-zot/zot/pull/3244

PR #3244, opened 2025-07-04
Read OpenID credentials from file

uwej711:
**What type of PR is this?**

feature

https://github.com/project-zot/zot/issues/3243

This PR moves OpenID credentials (client id and client secret) to an extra file (similar to ldap config) so that you can use secrets especially in Kubernetes to configure zot.

Tested the change manually with the following config (relevant part only):

{
  "http": {
    "auth": {
      "openid": {
        "providers": {
          "github": {
            "credentialsFile": "./github.json",
            "keypath": "",
            "scopes": ["user"]
          }
        }
      }
    }
  }
}

and 

{
  "clientid": "** REDACTED **",
  "clientsecret": "** REDACTED **"
}

using an real Github OAuth2 integration

The configuration change is breaking, you have to create the extra file and replace the credentials in the config with the reference to the file.

```release-note
# OpenID credentials stored separately from configuration

The OpenID credentials are removed from zot's openid configuration and stored in (a) separate file(s). See zot's [documentation](https://zotregistry.dev/v2.1.5/articles/authn-authz/).

⚠ This OpenID configuration change is incompatible with previous zot releases. When upgrading, you must reconfigure your OpenID credentials if you use OpenID.
```

### DIFF
--- a/pkg/api/config/config.go
+++ b/pkg/api/config/config.go
@@ -94,13 +94,39 @@ type OpenIDConfig struct {
 	Providers map[string]OpenIDProviderConfig
 }
 
-type OpenIDProviderConfig struct {
-	Name         string
+type OpenIDCredentials struct {
 	ClientID     string
 	ClientSecret string
-	KeyPath      string
-	Issuer       string
-	Scopes       []string
+}
+
+type OpenIDProviderConfig struct {
+	CredentialsFile string
+	Name            string
+	clientID        string `json:"-"`
+	clientSecret    string `json:"-"`
+	KeyPath         string
+	Issuer          string
+	Scopes          []string
+}
+
+func (openIDProviderConfig *OpenIDProviderConfig) ClientID() string {
+	return openIDProviderConfig.clientID
+}
+
+func (openIDProviderConfig *OpenIDProviderConfig) SetClientID(clientID string) *OpenIDProviderConfig {
+	openIDProviderConfig.clientID = clientID
+
+	return openIDProviderConfig
+}
+
+func (openIDProviderConfig *OpenIDProviderConfig) ClientSecret() string {
+	return openIDProviderConfig.clientSecret
+}
+
+func (openIDProviderConfig *OpenIDProviderConfig) SetClientSecret(clientSecret string) *OpenIDProviderConfig {
+	openIDProviderConfig.clientSecret = clientSecret
+
+	return openIDProviderConfig
 }
 
 type MethodRatelimitConfig struct {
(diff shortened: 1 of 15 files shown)


andaaron (approved):

uwej711:
I will look at the failed checks ...

rchincha:
@uwej711 you may also need to post a PR here:
https://github.com/project-zot/helm-charts
after this PR is merged though.

rchincha:
@uwej711 pls fix ci failures.

uwej711:
> @uwej711 you may also need to post a PR here: https://github.com/project-zot/helm-charts after this PR is merged though.

I looked at the chart, but I'm don't think there are any changes needed. @rchincha do you have something in mind?

uwej711:
> @uwej711 pls fix ci failures.

I hope I caught everything now ... let's see.

pkg/api/config/config.go
```diff
@@ -94,13 +94,39 @@ type OpenIDConfig struct {
...
 	ClientSecret string
-	KeyPath      string
-	Issuer       string
-	Scopes       []string
+}
+
+type OpenIDProviderConfig struct {
+	CredentialsFile string
+	Name            string
+	clientID        string `json:"-"`
```
rchincha: This deprecates inline credentials. Not sure we want to do that because it may break existing users who have it configured like that and not everyone is on k8s.

uwej711: We could keep the existing configuration and issue a warning when the inline credentials are used. That would keep BC. The inline credentials could then be removed in a major update (as was done with the ldap credentials). Not removing them in future would be inconsistent with ldap credentials and htpassword.

rchincha: Thanks @uwej711 
Makes sense to me!

rchincha: Let's try to get this PR in for the next release.

pkg/api/config/config.go
```diff
@@ -94,13 +94,19 @@ type OpenIDConfig struct {
...
 	ClientSecret string
-	KeyPath      string
-	Issuer       string
-	Scopes       []string
+}
+
+type OpenIDProviderConfig struct {
+	CredentialsFile string
+	Name            string
+	ClientID        string
```
rchincha, line 105: Might as well "inline" these as OpenIdCredentials?

uwej711: Do you think of

```go
type OpenIDProviderConfig struct {
   ...
   OpenIDCredentials
   ...
}
```

I'm not sure, but I think that would force us to change all the fixtures in the tests and if we want to remove those fields some time in the future, the second commit can be just reverted ...

rchincha (approved):
lgtm

Issue #3246, opened 2025-07-07
[Bug]: Native FreeBSD build is disabled


dfr:
### zot version

2c7e8fd33e16e54a5085803dc60c12020d21b6be

### Describe the bug

The top-level Makefile disables building on FreeBSD, apparently due to a dependency on protoc.

### To reproduce

On a FreeBSD host (or VM):
```
$ gmake
Makefile:618: *** makefile target can't be run on freebsd.  Stop.
```

### Expected behavior

If protoc is installed on the FreeBSD host, zot should build. This is convenient for folks like me who prefer to work on FreeBSD where possible.

rchincha:
https://github.com/project-zot/zot/issues/1663

rchincha:
@dfr pls close this issue if requirements addressed.

dfr:
> [@dfr](https://github.com/dfr) pls close this issue if requirements addressed.

Right now, all the Makefile targets I need are fine. The main thing which is still blocked is `check-gh-actions` which is correct since github still doesn't support actions on FreeBSD for complicated reasons. If that situation ever changes, we can revisit.

PR #3247, opened 2025-07-07
Fix building zot natively on FreeBSD

dfr:
**What type of PR is this?**

cleanup

#3246 

Allows building zot on a FreeBSD host. Not strictly needed since FreeBSD binaries can be cross-built on a Linux host but it is convenient for developers who prefer to use FreeBSD where possible.

Run 'gmake' on a host running FreeBSD-14.3.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,7 +123,7 @@ build-metadata: $(if $(findstring ui,$(BUILD_LABELS)), ui)
 	go list $(GO_CMD_TAGS) -f '{{ join .GoFiles "\n" }}' ./... | sort -u
 
 .PHONY: gen-protobuf
-gen-protobuf: check-not-freebds $(PROTOC)
+gen-protobuf: $(PROTOC)
 	$(PROTOC) --experimental_allow_proto3_optional \
 		--proto_path=$(TOP_LEVEL)/pkg/meta/proto \
 		--go_out=$(TOP_LEVEL)/pkg/meta/proto \
@@ -610,12 +610,6 @@ ui:
 check-linux:
 ifneq ($(shell go env GOOS),linux)
 	$(error makefile target can be run only on linux)
-endif
-
-.PHONY: check-not-freebds
-check-not-freebds:
-ifeq ($(shell go env GOOS),freebsd)
-  $(error makefile target can't be run on freebsd)
 endif
 
 .PHONY: check-compatibility
--- a/pkg/api/runtime.go
+++ b/pkg/api/runtime.go
@@ -20,7 +20,7 @@ func DumpRuntimeParams(log log.Logger) {
 
 	err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rLimit)
 	if err == nil {
-		evt = evt.Uint64("max. open files", uint64(rLimit.Cur)) //nolint: unconvert // required for *BSD
+		evt = evt.Uint64("max. open files", uint64(rLimit.Cur)) //nolint: unconvert,gosec // required for *BSD
 	}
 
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -216,7 +216,7 @@ test-devmode: test-prereq
 	rm -rf /tmp/getter*; rm -rf /tmp/trivy*
 	go test -failfast -tags dev -trimpath -race -cover -coverpkg ./... -coverprofile=coverage-dev-minimal.txt -covermode=atomic ./pkg/test/... ./pkg/storage/... ./pkg/extensions/sync/... -run ^TestInject
 	rm -rf /tmp/getter*; rm -rf /tmp/trivy*
-	go test -failfast -tags stress,$(BUILD_LABELS) -trimpath -race -timeout 15m ./pkg/cli/server/stress_test.go
+	go test -failfast -tags stress,$(BUILD_LABELS) -trimpath -race -timeout 15m ./pkg/cli/server/stress_test.go ./pkg/cli/server/rlimit_linux_test.go
```
andaaron: why is this needed?

dfr: Most of this commit revolves around me trying to get 'gmake check' to run cleanly on a FreeBSD host. The problem is that on FreeBSD, `rlim_t` (the type of the two `struct rlimit` fields) is `int64` whereas on Linux, it is `uint64`. This caused lint errors on FreeBSD attempting to assign uint64 values into these fields. I attempted a workaround using conditional compilation (in rlimit_linux_test.go and rlimit_freebsd_test.go) which caused other lint failures.

This iteration passes lint checks on both platforms but to run the stress test, I needed to add the new file (rlimit_linux_test.go) to the command line. All of this is quite ugly and fragile and I would welcome any suggestions on how to make it cleaner.

andaaron: I understand, and thank you for picking it up.
The part I was not clear about is this change specifically. Doesn't ./pkg/cli/server/stress_test.go import the new type without this line change? ./pkg/cli/server/rlimit_linux_test.go shouldn't even be specified since it is specific to a hardcoded platform value.

dfr: I'm not a go expert by any means but I think that when the test source file is specified explicitly, then the compiler doesn't automatically pick up other files in the package directory. I could try moving the rlimit type declarations to another package, perhaps zotregistry.dev/zot/pkg/test/common? That might remove the need for this makefile change.

dfr: Moving the type declarations to `zotregistry.dev/zot/pkg/test/common` did avoid the Makefile change and seems to be working so far.

andaaron:
@dfr, can you please sign your commits with `-s -S` (maybe squash too)?

dfr:
I'm working on it - my gpg key expired a few weeks ago and I need to try and remember how to regenerate it...

dfr:
I squashed the commits and signed the result with an ssh key but I'm not sure if that worked right since the CI still has the 'merge is blocked...verified signatures' status. I will take another look tomorrow.

andaaron:
> I squashed the commits and signed the result with an ssh key but I'm not sure if that worked right since the CI still has the 'merge is blocked...verified signatures' status. I will take another look tomorrow.

You probably need to upload the public key to github.

rchincha:
@dfr thanks for PR. 

You may need to upload your public key to your GH account so GH knows that your commit has a verified signature.

Pls let us know how we can help.

dfr:
> @dfr thanks for PR.
> 
> You may need to upload your public key to your GH account so GH knows that your commit has a verified signature.
> 
> Pls let us know how we can help.

I figured it out, thanks!

Issue #1663, opened 2023-07-31
[Feat]: release freebsd images for zot 


rchincha:
### Is your feature request related to a problem? Please describe.

PR #1660 adds support for freebsd binaries. This issue is being filed so that we track the remaining step of releasing freebsd container images for zot once an official freebsd-minimal image is available.

cc: @dfr

rchincha:
@dfr Noticed that the images you are pushing to dockerhub are docker images and not OCI. Is this intended? For example official ubuntu images are pushed as OCI images.

https://hub.docker.com/r/dougrabson/freebsd-minimal/tags

rchincha:
@dfr There are many tools. For example,

https://github.com/google/go-containerregistry/tree/main/cmd/crane

`crane manifest <image> | jq .`

rchincha:
In comparison, ubuntu

```
$ crane manifest ubuntu:latest | jq .

{
  "manifests": [
    {
      "digest": "sha256:c9cf959fd83770dfdefd8fb42cfef0761432af36a764c077aed54bbc5bb25368",
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      
```

rchincha:
@dfr So you are pushing docker images to dockerhub. Not sure if this was intentional - from what I understand, FreeBSD is standardizing on OCI?

dfr:
This is strange - I pushed the same image to both docker.io and quay.io but the manifest type on dockerhub is different.
```
$ crane manifest docker.io/dougrabson/freebsd13.2-static:13.2p2 | jq .
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "size": 1130,
      "digest": "sha256:4378443a2023c8804997bca2f5e8c33944085e2102e192af0dcea943053b59bb",
      "platform": {
        "architecture": "amd64",
        "os": "freebsd"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "size": 1132,
      "digest": "sha256:cc62ec2301fdfab24d0e3976ff86ba16938ead54d638acdee1436d3ff5a3c1f9",
      "platform": {
        "architecture": "arm64",
        "os": "freebsd"
      }
    }
  ]
}
```

```
$ crane manifest quay.io/dougrabson/freebsd13.2-static:13.2p2 | jq .
[...]

PR #3256, opened 2025-07-08
feat(freebsd): add support native freebsd container images

rchincha:
Fixes issue #1663

freebsd is now building and releasing official freebsd OCI container images

https://hub.docker.com/r/freebsd/freebsd-runtime/tags

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


rchincha:
cc: @dfr 

Pls review and verify that freebsd images build and run correctly.

Makefile
```diff
@@ -61,8 +61,10 @@ UI_DEPENDENCIES := search,mgmt,userprefs
 # freebsd is not supported for pie builds if CGO is disabled
 # see supported platforms at https://cs.opensource.google/go/go/+/master:src/internal/platform/supported.go;l=222-231;drc=d7fcb5cf80953f1d63246f1ae9defa60c5ce2d76
 BUILDMODE_FLAGS := -buildmode=pie
+BASE_IMAGE=gcr.io/distroless/base-debian12:latest-$(ARCH)
```
andaaron, line 64: Don't we need the equivalent of this replace also in cases where the stacker build/publish GH action is used?
In this file: https://github.com/project-zot/zot/blob/main/.github/workflows/publish.yaml

dfr:
> cc: @dfr
> 
> Pls review and verify that freebsd images build and run correctly.

The Dockerfiles use a linux image (golang:1.24) as builder. It looks like this nearly works if Linux emulation is enabled on the FreeBSD host  - it gets as far as as running the go binary before giving up with an error (`failed to start telemetry sidecar: os.Executable: readlink /proc/self/exe: no such file or directory`). It might be better to have a FreeBSD-native builder image.

```
sudo gmake binary-container                                                                                                                                        2 ↵
/usr/local/bin/podman build  \
	--build-arg BASE_IMAGE=freebsd/freebsd-runtime:14.3 \
	-f build/Dockerfile -t zot-build:latest .
WARN[0000] missing "COMMIT" build argument. Try adding "--build-arg COMMIT=<VALUE>" to the command line
[1/2] STEP 1/10: FROM --platform=freebsd/amd64 ghcr.io/project-zot/golang:1.24 AS builder
Trying to pull ghcr.io/project-zot/golang:1.24...
Copying blob e9ca80243f8c skipped: already exists
Copying blob a25547fc71e9 skipped: already exists
Copying blob 900e2c02f17f skipped: already exists
Copying blob 7bbf972c6c2f skipped: already exists
Copying blob a8f67fead7e3 skipped: already exists
Copying blob 4f4fb700ef54 skipped: already exists
Copying blob d0034faf7acc skipped: already exists
Copying config aca47f4a3f done   |
WARNING: image platform (linux/amd64) does not match the expected platform (freebsd/amd64)
[1/2] STEP 2/10: ARG TARGETOS
--> Using cache 28d249185ca5428796a8feb3a1e079297be9375a43a8f97843edfbcab7bd7cff
--> 28d249185ca5
[1/2] STEP 3/10: ARG TARGETARCH
--> Using cache df8176a309a52273abebea15dc24594a981edb5a02a9b6f2b3a37d0e286cafdf
--> df8176a309a5
[1/2] STEP 4/10: ARG COMMIT
--> Using cache a36b55e5eb5809eff314ca1f3c7bd123393313dda90e5a5aeafd436e2bd17a1a
--> a36b55e5eb58
[1/2] STEP 5/10: RUN mkdir -p /go/src/github.com/project-zot/zot
--> Using cache 1e8c8186cf6cb44ef10d7984d426af0f3379adc3f9cb543443c546f37bcafcb8
--> 1e8c8186cf6c
[1/2] STEP 6/10: WORKDIR /go/src/github.com/project-zot/zot
--> Using cache ff840ca3291a20e295cb8a87a4b2165b73603535f61b9eb2bb98a2fa11e8d5b4
--> ff840ca3291a
[1/2] STEP 7/10: COPY . .
--> add460ed3bc3
[1/2] STEP 8/10: RUN make COMMIT=$COMMIT OS=$TARGETOS ARCH=$TARGETARCH clean binary
failed to start telemetry sidecar: os.Executable: readlink /proc/self/exe: no such file or directory
go: cannot find GOROOT directory: 'go' binary is trimmed and GOROOT is not set
go: cannot find GOROOT directory: 'go' binary is trimmed and GOROOT is not set
go: cannot find GOROOT directory: 'go' binary is trimmed and GOROOT is not set
go: cannot find GOROOT directory: 'go' binary is trimmed and GOROOT is not set
go: cannot find GOROOT directory: 'go' binary is trimmed and GOROOT is not set
rm -f bin/z*
rm -rf hack
rm -rf test/data/zot-test
rm -rf test/data/zot-cve-test
rm -rf test/data/zot-cve-java-test
rm -rf pkg/extensions/build
go mod tidy
go: cannot find GOROOT directory: 'go' binary is trimmed and GOROOT is not set
make: *** [Makefile:100: modtidy] Error 2
Error: building at STEP "RUN make COMMIT=$COMMIT OS=$TARGETOS ARCH=$TARGETARCH clean binary": while running runtime: exit status 2
gmake: *** [Makefile:445: binary-container] Error 2
```

dfr:
I built a suitable FreeBSD builder image using this Dockerfile:
```
FROM ghcr.io/freebsd/freebsd-runtime:14.3
RUN env ASSUME_ALWAYS_YES=yes pkg bootstrap -r FreeBSD && pkg update
RUN pkg install -y FreeBSD-utilities indexinfo
RUN pkg install -y go124 gmake git-lite
RUN pkg clean -ay && pkg delete -fy pkg && rm -rf /var/db/pkg/repos
RUN ln -sf go124 /usr/local/bin/go
```

With a slightly hacked build/Dockerfile:
```
# ---
# ---
ARG BASE_IMAGE
#FROM --platform=$BUILDPLATFORM ghcr.io/project-zot/golang:1.24 AS builder
# ---
# ---
ARG BASE_IMAGE
#FROM --platform=$BUILDPLATFORM ghcr.io/project-zot/golang:1.24 AS builder
FROM --platform=$BUILDPLATFORM localhost/golang:1.24 AS builder

ARG TARGETOS
ARG TARGETARCH
ARG COMMIT

RUN mkdir -p /go/src/github.com/project-zot/zot
WORKDIR /go/src/github.com/project-zot/zot
COPY . .
RUN gmake COMMIT=$COMMIT OS=$TARGETOS ARCH=$TARGETARCH clean binary
RUN echo '{\n\
    "storage": {\n\
        "rootDirectory": "/var/lib/registry"\n\
    },\n\
    "http": {\n\
        "address": "0.0.0.0",\n\
        "port": "5000",\n\
        "compat": ["docker2s2"]\n\
    },\n\
    "log": {\n\
        "level": "debug"\n\
    }\n\
}\n' > config.json && cat config.json

# ---
# Stage 2: Final image with nothing but certs, binary, and default config file
# ---
FROM $BASE_IMAGE AS final
ARG TARGETOS
ARG TARGETARCH
#COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
COPY --from=builder /go/src/github.com/project-zot/zot/bin/zot-$TARGETOS-$TARGETARCH /usr/bin/zot
COPY --from=builder /go/src/github.com/project-zot/zot/config.json /etc/zot/config.json
ENTRYPOINT ["/usr/bin/zot"]
EXPOSE 5000
CMD ["serve", "/etc/zot/config.json"]
```
The changes I made are:
- Change the builder image to the one I built locally
- GNU make is spelt `gmake` on FreeBSD - I could avoid this by adding a symlink in the golang builder image?
- Move the `ARG BASE_IMAGE` above the first `FROM` since Podman sets all `ARG`s declared after a `FROM` to null before executing the second `FROM`
- Remove the copy of `ca-certificates.crt` which doesn't exist on FreeBSD and is already part of the base-debian12 image on Linux.

With this, building an image nearly works on a FreeBSD host but dies after the build step:
```
time="2025-07-08T11:49:40+01:00" level=error msg="While applying layer: ApplyLayer stdout:  stderr: Lstat /go/pkg/mod/github.com/olareg/olareg@v0.1.2/testdata/layer2.txt: no such file or directory exit status 1"
Error: committing container for step {Env:[TARGETOS=freebsd TARGETARCH=amd64 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin] Command:run Args:[gmake COMMIT=$COMMIT OS=$TARGETOS ARCH=$TARGETARCH clean binary] Flags:[] Attrs:map[] Message:RUN gmake COMMIT=$COMMIT OS=$TARGETOS ARCH=$TARGETARCH clean binary Heredocs:[] Original:RUN gmake COMMIT=$COMMIT OS=$TARGETOS ARCH=$TARGETARCH clean binary}: copying layers and metadata for container "6f94dd2b1bf50465650db76bc82529dbd725832ccd79f5438fd594474ee5f53d": writing blob: adding layer with blob "sha256:03a00595f151b1268eaa059cdaa13adcc85b762a99aa16e1497a38630f48d037"/""/"sha256:03a00595f151b1268eaa059cdaa13adcc85b762a99aa16e1497a38630f48d037": ApplyLayer stdout:  stderr: Lstat /go/pkg/mod/github.com/olareg/olareg@v0.1.2/testdata/layer2.txt: no such file or directory exit status 1
```
This is while copying the result of the build step to a cache image, likely caused by the presence of a file named `.wh.layer2.txt` in the problematic go module. I worked around it by disabling cache images:
```
/usr/local/bin/podman build --squash \
	--build-arg BASE_IMAGE=freebsd/freebsd-runtime:14.3 \
	-f build/Dockerfile -t zot-build:latest .
time="2025-07-08T14:33:21+01:00" level=warning msg="missing \"COMMIT\" build argument. Try adding \"--build-arg COMMIT=<VALUE>\" to the command line"
[1/2] STEP 1/9: FROM --platform=freebsd/amd64 localhost/golang:1.24 AS builder
[1/2] STEP 2/9: ARG TARGETOS
[1/2] STEP 3/9: ARG TARGETARCH
[1/2] STEP 4/9: ARG COMMIT
[1/2] STEP 5/9: RUN mkdir -p /go/src/github.com/project-zot/zot
[1/2] STEP 6/9: WORKDIR /go/src/github.com/project-zot/zot
[1/2] STEP 7/9: COPY . .
[1/2] STEP 8/9: RUN gmake COMMIT=$COMMIT OS=$TARGETOS ARCH=$TARGETARCH clean binary
go: downloading go1.24.4 (freebsd/amd64)
rm -f bin/z*
rm -rf hack
rm -rf test/data/zot-test
rm -rf test/data/zot-cve-test
rm -rf test/data/zot-cve-java-test
rm -rf pkg/extensions/build
go mod tidy
go: downloading github.com/99designs/gqlgen v0.17.76
...
zotregistry.dev/zot/pkg/extensions/search/cve/trivy
zotregistry.dev/zot/pkg/extensions/search/cve
zotregistry.dev/zot/pkg/extensions/search/convert
zotregistry.dev/zot/pkg/extensions/search
zotregistry.dev/zot/pkg/extensions
zotregistry.dev/zot/pkg/api
zotregistry.dev/zot/pkg/cli/server
zotregistry.dev/zot/cmd/zot
[1/2] STEP 9/9: RUN echo '{\n    "storage": {\n        "rootDirectory": "/var/lib/registry"\n    },\n    "http": {\n        "address": "0.0.0.0",\n        "port": "5000",\n        "compat": ["docker2s2"]\n    },\n    "log": {\n        "level": "debug"\n    }\n}\n' > config.json && cat config.json
{\n    "storage": {\n        "rootDirectory": "/var/lib/registry"\n    },\n    "http": {\n        "address": "0.0.0.0",\n        "port": "5000",\n        "compat": ["docker2s2"]\n    },\n    "log": {\n        "level": "debug"\n    }\n}\n
[2/2] STEP 1/8: FROM freebsd/freebsd-runtime:14.3 AS final
[2/2] STEP 2/8: ARG TARGETOS
[2/2] STEP 3/8: ARG TARGETARCH
[2/2] STEP 4/8: COPY --from=builder /go/src/github.com/project-zot/zot/bin/zot-$TARGETOS-$TARGETARCH /usr/bin/zot
[2/2] STEP 5/8: COPY --from=builder /go/src/github.com/project-zot/zot/config.json /etc/zot/config.json
[2/2] STEP 6/8: ENTRYPOINT ["/usr/bin/zot"]
[2/2] STEP 7/8: EXPOSE 5000
[2/2] STEP 8/8: CMD ["serve", "/etc/zot/config.json"]
[2/2] COMMIT zot-build:latest
Copying blob sha256:0b11a04c9d7e78e922d47d6fdb7fc2a78ccff9fe43a8cff4bef079972b35cdfb
Copying blob sha256:e075dc0f6fe6ff32cc18e5671f8c818d2f538bfed6bb331f3e51112ddcd2a818
Copying config sha256:2fcb933cc0af52d3f41aace79cf88fc5591ddb741878ecfee02e0ab55be7a5bb
--> 2fcb933cc0af
Successfully tagged localhost/zot-build:latest
2fcb933cc0af52d3f41aace79cf88fc5591ddb741878ecfee02e0ab55be7a5bb
```

Makefile
rchincha: Maybe we should make this a target in the Makefile itself.

rchincha:
@dfr I think we have two issue to solve then.

1) host is linux - which is how our GH runners are set up and how we would release our images, built for freebsd target and using freebsd runtime as the base container image, which is what this PR is trying to do and afaict should work

2) host is freebsd - maybe you can give us a separate Dockerfile.freebsd under build/ plumbed with the Makefile

Makefile
andaaron: I don't recall all the details. @eusebiu-constantin-petu-dbk, why do we use the stacker GH action?

dfr:
> @dfr I think we have two issue to solve then.
> 
> 1. host is linux - which is how our GH runners are set up and how we would release our images, built for freebsd target and using freebsd runtime as the base container image, which is what this PR is trying to do and afaict should work
> 2. host is freebsd - maybe you can give us a separate Dockerfile.freebsd under build/ plumbed with the Makefile

As long as the resulting image has the right metadata (e.g. os=freebsd), then option (1) is the way to go. As far as I know, GH actions still don't support FreeBSD. For this option, we can simply remove the `COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt` line since this file is already present in the base-debian12 image (I think).

For option (2), I can probably use the existing Dockerfile (with the ca-certificates.crt line removed) if I add another ARG statement to allow the use of a different builder image for native FreeBSD builds.

rchincha:
> > @dfr I think we have two issue to solve then.
> > 
> > 1. host is linux - which is how our GH runners are set up and how we would release our images, built for freebsd target and using freebsd runtime as the base container image, which is what this PR is trying to do and afaict should work
> > 2. host is freebsd - maybe you can give us a separate Dockerfile.freebsd under build/ plumbed with the Makefile
> 
> As long as the resulting image has the right metadata (e.g. os=freebsd), then option (1) is the way to go. As far as I know, GH actions still don't support FreeBSD. For this option, we can simply remove the `COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt` line since this file is already present in the base-debian12 image (I think).
> 
> For option (2), I can probably use the existing Dockerfile (with the ca-certificates.crt line removed) if I add another ARG statement to allow the use of a different builder image for native FreeBSD builds.

@dfr I believe we need to do both options. We will do option 1 and option 2 would be needed for native FreeBSD users like yourself.

rchincha:
https://github.com/rchincha/zot/actions/runs/16211967726

$ skopeo inspect docker://freebsd/freebsd-runtime:14.3

FATA[0001] Error parsing manifest for image: Error choosing image instance: no image found in image index for architecture amd64, variant "", OS linux

rchincha:
@dfr wondering if zot should just become a part of freebsd ports and you can then release it any way you like?

dfr:
> https://github.com/rchincha/zot/actions/runs/16211967726
> 
> $ skopeo inspect docker://freebsd/freebsd-runtime:14.3
> 
> FATA[0001] Error parsing manifest for image: Error choosing image instance: no image found in image index for architecture amd64, variant "", OS linux

Try overriding the OS:
```
$ skopeo --override-os=freebsd inspect docker://freebsd/freebsd-runtime:14.3
```

dfr:
> @dfr wondering if zot should just become a part of freebsd ports and you can then release it any way you like?

I do think zot should be in the FreeBSD ports collection - this would be very easy given that it builds and works just fine already. I also still like the idea of project-zot supplied images if we can figure out how to make that work in your CI.

rchincha:
https://github.com/project-stacker/stacker/pull/711

rchincha:
We need this stacker PR first.
https://github.com/project-stacker/stacker/pull/711

rchincha:
@dfr 

Built test freebsd OCI images via CI using this PR, here:
https://github.com/rchincha/zot/pkgs/container/zot-freebsd-amd64
https://github.com/rchincha/zot/pkgs/container/zot-freebsd-arm64

multi-arch:
ghcr.io/rchincha/zot:v2.1.6-freebsd-rchincha

If you can confirm that you are able to pull and run these images, then we can iterate and fix errors.

dfr:
> @dfr
> 
> Built test freebsd OCI images via CI using this PR, here: https://github.com/rchincha/zot/pkgs/container/zot-freebsd-amd64 https://github.com/rchincha/zot/pkgs/container/zot-freebsd-arm64
> 
> multi-arch: ghcr.io/rchincha/zot:v2.1.6-freebsd-rchincha
> 
> If you can confirm that you are able to pull and run these images, then we can iterate and fix errors.

Thanks for working on this - I will take a look later today.

dfr:
The test OCI image works well under Podman on a FreeBSD 14.3 host.

Assuming that zot doesn't require a working shell, you could change the base image to `freebsd-static` (https://github.com/orgs/freebsd/packages/container/freebsd-static/432750841?tag=14.3) which contains just TLS certs and other config files which would reduce the attack surface, similar to the `base-debian12` image used for Linux. In fact, if zot on Linux is also static linked, you could change that base image to `static-debian12`.

rchincha:
Don't fully recall all the details ...

zot is built with `-buildmode=pie` and we had trouble with static base images since a minimal libc runtime was needed (again iirc)

rchincha:
With CGO disabled (as we currently do)

$ env CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -tags netgo,osusergo -o bin/zot-linux-amd64-minimal -buildmode=pie -v -trimpath ./cmd/zot

$ file bin/zot-linux-amd64-minimal 
bin/zot-linux-amd64-minimal: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=0515cb298a9fb463984e272ddf2d7de41486d60c, with debug_info, not stripped

With CGO enabled (we could consider doing this) and external linkage

$ env CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build -tags netgo,osusergo -o bin/zot-linux-amd64-minimal -buildmode=pie -v -trimpath -ldflags "-linkmode 'external' -extldflags '-static'" ./cmd/zot

$ file bin/zot-linux-amd64-minimal 
bin/zot-linux-amd64-minimal: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), statically linked, BuildID[sha1]=e359a4fbf4bdbd60147e9c10f6c974e71e5e8763, for GNU/Linux 3.2.0, with debug_info, not stripped

dfr:
> With CGO disabled (as we currently do)
> 
> $ env CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -tags netgo,osusergo -o bin/zot-linux-amd64-minimal -buildmode=pie -v -trimpath ./cmd/zot
> 
> $ file bin/zot-linux-amd64-minimal bin/zot-linux-amd64-minimal: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=0515cb298a9fb463984e272ddf2d7de41486d60c, with debug_info, not stripped
> 
> With CGO enabled (we could consider doing this) and external linkage
> 
> $ env CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build -tags netgo,osusergo -o bin/zot-linux-amd64-minimal -buildmode=pie -v -trimpath -ldflags "-linkmode 'external' -extldflags '-static'" ./cmd/zot
> 
> $ file bin/zot-linux-amd64-minimal bin/zot-linux-amd64-minimal: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), statically linked, BuildID[sha1]=e359a4fbf4bdbd60147e9c10f6c974e71e5e8763, for GNU/Linux 3.2.0, with debug_info, not stripped

Thats interesting - I looked inside the freebsd image that you built and that binary is definitely static. If at some point in the future it becomes dynamic (e.g. if the build flags are changed), then we can use https://github.com/orgs/freebsd/packages/container/freebsd-dynamic/432750723?tag=14.3 which is the same as `freebsd-static` but adds base dynamic libraries such as libc.

rchincha:
@dfr

Long-term, what is your source of truth for FreeBSD images?
ghcr or dockerhub?
https://github.com/orgs/freebsd/packages

rchincha:
ghcr.io/rchincha/zot-freebsd-amd64:v2.1.6-freebsd-static-rchincha
ghcr.io/rchincha/zot-freebsd-arm64:v2.1.6-freebsd-static-rchincha

^ with "static" base image.
Let me know if this still works, if so we can move to this.

rchincha:
Also, in future, when you bump FreeBSD versions (we are unlikely to track these), do give us a PR with changes in these two places.

https://github.com/project-zot/zot/pull/3256/commits/eac8c84e75cf53058f28da2be777b6478ded9943

dfr:
> @dfr
> 
> Long-term, what is your source of truth for FreeBSD images? ghcr or dockerhub? https://github.com/orgs/freebsd/packages

The main user-visible source of truth is https://download.freebsd.org/releases/OCI-IMAGES/ which contains oci archive format images built by the release process. The last part of our release process combines images into multi-arch manifests and pushes to both github and dockerhub - both registries should have identical images.

dfr:
> ghcr.io/rchincha/zot-freebsd-amd64:v2.1.6-freebsd-static-rchincha ghcr.io/rchincha/zot-freebsd-arm64:v2.1.6-freebsd-static-rchincha
> 
> ^ with "static" base image. Let me know if this still works, if so we can move to this.

I (lightly) tested the amd64 image and it works just fine - thanks again for this.

rchincha:
@andaaron imo, we can now merge this.

.github/workflows/publish.yaml
```diff
@@ -15,9 +15,17 @@ jobs:
       packages: write
     strategy:
       matrix:
-        os: [linux, darwin]
+        os: [linux, freebsd]
```
andaaron, line 18: @rchincha The darwin image is broken? Why are we removing it?

.github/workflows/publish.yaml
```diff
@@ -15,9 +15,17 @@ jobs:
...
         arch: [amd64, arm64]
     steps:
+      - name: Setup base image
+        run: |
+          if [ ${{ matrix.os }} = 'freebsd' ]; then
```
andaaron, line 23: Does this mean we build the same image twice, once per arch?

.github/workflows/publish.yaml
```diff
@@ -15,9 +15,17 @@ jobs:
...
+            echo "BASE_IMAGE=freebsd/freebsd-static:14.3" >> "$GITHUB_ENV"
```
andaaron, line 24: What arch is this?

.github/workflows/publish.yaml
rchincha: https://github.com/rchincha/zot/actions/runs/16696661720

OS and ARCH are specified in the stackerfile now, so the right one gets picked.
For example,
```
substitution: "OS=freebsd" was used 4 times
substitution: "ARCH=arm64" was used 4 times
```

.github/workflows/publish.yaml
rchincha: See above.

.github/workflows/publish.yaml
rchincha: We need this project to publish a base image that we can use.
https://darwin-containers.github.io/

Else we were just building with a Linux base image which wont work on darwin.

andaaron (approved):

Makefile
rchincha: Updated stackerfile with new stacker version works fine.

rchincha:
@dfr

v2.1.7 has been released with these changes.
Hopefully, it works for you.

PR #2968, opened 2025-02-15
feat(storage): enable parallel writes by using per-repo and per-digest locking

andaaron:
Should fix issues such as https://github.com/project-zot/zot/issues/2964

- lock per repo on pushes/pulls/retention, in short index operations
- lock per digest when using multiple operations affecting the cachedb and storage
(blob writes/deletes/moves/links in storage which need to be in accordance with cachedb content)

Do not lock multiple repos at the same time in the same goroutine! It will cause deadlocks.
Same applies to digests.

In separate commits:
- show more error information in zb output
- gc stress tests to save logs as artifacts

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3314, opened 2025-08-17
fix(dedupe): use a common blobs dir to simplify

rchincha:
Currently, our dedupe scheme is very complicated since master copy is kept in one of the repo dirs and if that image is deleted, the owning repo is changed, requiring multiple repo locks causing unnecessary contention.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


rchincha:
Should pair with https://github.com/project-zot/zot/pull/2968

pkg/storage/constants/constants.go
```diff
@@ -26,4 +26,5 @@ const (
 	DefaultGCInterval       = 1 * time.Hour
 	S3StorageDriverName     = "s3"
 	LocalStorageDriverName  = "local"
+	GlobalBlobsRepo         = "_blobstore"
```
andaaron, line 28: Make sure this blobs repo is per store and substore. Because substores may be on different partitions.

pkg/api/controller.go
```diff
@@ -529,7 +529,7 @@ func (c *Controller) StartBackgroundTasks() {
 			// Enable running dedupe blobs both ways (dedupe or restore deduped blobs) for subpaths
 			substore := c.StoreController.SubStore[route]
 			if substore != nil {
-				substore.RunDedupeBlobs(time.Duration(0), c.taskScheduler)
+				//substore.RunDedupeBlobs(time.Duration(0), c.taskScheduler)
```
andaaron, line 508: Is this just temporary, or we're dropping dedupe altogether?

pkg/storage/cache/boltdb.go
```diff
@@ -164,8 +149,31 @@ func (d *BoltDBDriver) PutBlob(digest godigest.Digest, path string) error {
...
 			}
+			d.log.Debug().Str("digest", digest.String()).Str("path", path).Msg("inserted in original bucket")
+
+			return nil
+		} else if origin.Get([]byte(path)) != nil { // idempotent
+			d.log.Debug().Str("digest", digest.String()).Str("path", path).Msg("inserted same key in original bucket")
+			return nil
+		}
+
+		// create nested deduped bucket where we store all the deduped blobs + original blob
```
andaaron, line 164:
```suggestion
		// create nested deduped bucket where we store all the deduped blobs while excluding the original blob
```
Correct?

pkg/storage/cache/boltdb.go
andaaron, line 1: I think the logging is too verbose for merge. It would be too much information to include logs for the code paths not returning errors.

pkg/storage/cache/boltdb.go
```diff
@@ -330,39 +345,55 @@ func (d *BoltDBDriver) DeleteBlob(digest godigest.Digest, path string) error {
...
 
 		origin := bucket.Bucket([]byte(constants.OriginalBucket))
 		if origin != nil {
-			originBlob := d.getOne(origin)
-			if originBlob != nil {
+			if origin.Get([]byte(path)) != nil {
+
+				dedupeBlob := d.getOne(deduped)
+				if dedupeBlob != nil {
+					d.log.Debug().Str("digest", digest.String()).Str("path", path).Msg("more in dedupe bucket, leaving original alone")
```
andaaron, line 375: Previous logic was replacing the old original blob with one of the duplicates, and then delete the blob.
This new logic would keep the original blob around forever? How/when would it be GCed?

pkg/storage/cache/boltdb.go
```diff
@@ -376,9 +407,11 @@ func (d *BoltDBDriver) DeleteBlob(digest godigest.Digest, path string) error {
 
 				return err
 			}
+
+			return nil
 		}
 
-		return nil
+		return zerr.ErrCacheMiss
```
andaaron, line 418: Under which scenario would this return the error?
The updated code is a bit confusing as it is returning nil in several other places and it is unclear in which scenario this line is reached.

pkg/storage/cache/dynamodb.go
andaaron, line 1: 1. Are there no other needed changes in this file? 
2. Also the redis implementation should be reviewed

pkg/storage/cache/boltdb.go
```diff
@@ -330,39 +345,55 @@ func (d *BoltDBDriver) DeleteBlob(digest godigest.Digest, path string) error {
...
 
-			return err
+				d.log.Debug().Str("digest", digest.String()).Str("path", path).Msg("deleted from dedupe bucket")
+
+				return nil
+			}
 		}
 
 		origin := bucket.Bucket([]byte(constants.OriginalBucket))
 		if origin != nil {
```
andaaron, line 370: And if origin is nil, then we should  return nil, because the blob does not exist.

It is not in the duplicates list and it is not the original, so it does not exist, so the deletion should be marked as successful, correct?

pkg/storage/gc/gc.go
```diff
@@ -366,6 +388,11 @@ func (gc GarbageCollect) removeTagsPerRetentionPolicy(ctx context.Context, repo
 		return nil
 	}
 
+	// skip the global blobs repo
+	if repo == constants.GlobalBlobsRepo {
```
andaaron, line 421: Something is fishy.
Why is this here? `constants.GlobalBlobsRepo` is not an actual repo because it has an invalid repo name. How is GC detecting it when "walking" the disk if it doesn't have a valid repo name? It shouldn't call this function at all for `constants.GlobalBlobsRepo`?

pkg/storage/imagestore/imagestore.go
```diff
@@ -1145,92 +1155,118 @@ retry:
...
-		if !is.storeDriver.SameFile(dst, dstRecord) {
-			if err := is.storeDriver.Link(dstRecord, dst); err != nil {
-				is.log.Error().Err(err).Str("blobPath", dstRecord).Str("component", "dedupe").
-					Msg("failed to link blobs")
+			return err
+		}
 
-				return err
-			}
+		goto retry
```
andaaron: Since we're refactoring this, can you pick up https://github.com/project-zot/zot/pull/2969?
Or do you want me to rebase and merge that in advance?

Is this code working right now, wouldn't it potentially execute the code multiple times?

```
		if err := is.storeDriver.Move(src, gdst); err != nil {
			is.log.Error().Err(err).Str("src", src).Str("dst", gdst).Str("component", "dedupe").
				Msg("failed to rename blob")

			return err
		}
```

rchincha:
superseded by https://github.com/project-zot/zot/pull/3906

Issue #3333, opened 2025-08-24
[Feat]: Support for remote Session storage


vrajashkr:
### Is your feature request related to a problem? Please describe.

There are a couple of use cases that require multiple zot instances to be able to have common access to web session storage.

1. In cluster mode where cache and storage are both remote and there are multiple instances such as in #3088 
2. In local scale-out mode (in development) where session information is required by other instances responding to a forwarded GQL query.

### Describe the solution you'd like

Ability to configure and use a remote session cache.

vrajashkr:
Session information is currently stored in 2 ways:
1. Local disk if the storage controller is local
2. In memory otherwise

### Proposed Solution
Create a new configuration in zot config with the following information:
1. `RemoteSessionCache` - boolean indicating whether remote session cache is enabled.
2. `SessionDriver` - a map of key value pairs that has the same information as a `CacheDriver` map.

If `RemoteSessionCache` is disabled, zot would use the session store on the local disk.
If `RemoteSessionCache` is enabled, the values from the `SessionDriver` map are used and a compatible session storage is initialized for use.

2 options could be supported (same as for remote cache drivers):
1. DynamoDB
2. Redis

vrajashkr:
zot uses https://github.com/gorilla/sessions for session storage.

There are multiple compatible interface implementations mentioned in their [repository README](https://github.com/gorilla/sessions/blob/v1.4.0/README.md).

For Redis, [this redis session store](https://github.com/rbcervilla/redisstore) seems to be a good option.

For DynamoDB, [this library](https://github.com/savaki/dynastore) seems to be written with the official AWS SDK that zot currently uses, however, there doesn't appear to be a stable release of the library yet and doesn't appear to have been updated in quite a long while.

rchincha:
> For DynamoDB, [this library](https://github.com/savaki/dynastore) seems to be written with the official AWS SDK that zot currently uses

^ this may not work currently. Probably best to port it yourself.

vrajashkr:
Work in progress

andaaron:
Shall we close this as completed in https://github.com/project-zot/zot/pull/3345?

vrajashkr:
> Shall we close this as completed in https://github.com/project-zot/zot/pull/3345?

This issue is common for redis and for dynamodb so we might want to keep it open till we implement storing sessions in dynamoDB too.

If we would like a separate issue for that, that would work just as well.

PR #3345, opened 2025-08-28
feat(sessions): add support for remote redis session store

vrajashkr:
**What type of PR is this?**
feature

Towards #3333 

- Adds support for Redis to be used as a remote storage for session information.
- Sessions are stored in Redis and get invalidated when the user logs out on the browser or after expiry.
- Allows session information to persist outside of the Zot instance for use cases where Zot is fully stateless or there are multiple Zot instances running.

A local redis 8 container is used for this testing. The container has been started and `valkey-cli` is used to access and read the entries to verify behavior.

The test config for zot is the file `examples/config-remote-session-store-redis.json`

A login is initiated through ZUI on browser. A cookie is successfully created and saved.
Screenshot from 2025-08-30 11-08-22

Inspecting on Redis indicates that it was originally empty, but later a single entry was created for the new cookie.
Screenshot from 2025-08-30 11-09-25

Post this, a logout operation was initiated from the browser and Redis was inspected again:
Screenshot from 2025-08-30 11-11-16

Additional testing after addressing review comments:
Screenshot from 2025-09-07 23-28-18

This shows that the session was successfully deleted from the Redis storage.

**Automation added to e2e**:
A new BATS test has been added for the feature.

No. Additional configuration is needed to enable it. Otherwise, the existing behavior will continue as-is.

Yes

```release-note
Zot now supports configuring a remote Redis compatible session store. Refer to examples/config-remote-session-store-redis.json for an example on how to enable it.
```

### DIFF
--- a/examples/config-remote-session-store-redis.json
+++ b/examples/config-remote-session-store-redis.json
@@ -1,0 +1,34 @@
+{
+  "distSpecVersion": "1.1.1",
+  "storage": {
+    "rootDirectory": "/tmp/zot"
+  },
+  "http": {
+    "address": "127.0.0.1",
+    "port": "8080",
+    "realm": "zot",
+    "auth": {
+      "htpasswd": {
+        "path": "/tmp/zotpasswd"
+      },
+      "sessionDriver": {
+        "name": "redis",
+        "url": "redis://localhost:6379",
+        "keyprefix": "zotsession"
+      }
+    }
+  },
+  "log": {
+    "level": "debug"
+  },
+  "extensions": {
+    "search": {
+      "cve": {
+        "updateInterval": "2h"
+      }
+    },
+    "ui": {
+      "enable": true
+    }
+  }
+}
--- a/pkg/api/cookiestore.go
+++ b/pkg/api/cookiestore.go
@@ -12,7 +12,12 @@ import (
 	"time"
 
 	"github.com/gorilla/sessions"
+	"github.com/rbcervilla/redisstore/v9"
 
+	"zotregistry.dev/zot/errors"
+	"zotregistry.dev/zot/pkg/api/config"
+	rediscfg "zotregistry.dev/zot/pkg/api/config/redis"
+	"zotregistry.dev/zot/pkg/log"
 	"zotregistry.dev/zot/pkg/scheduler"
 	"zotregistry.dev/zot/pkg/storage"
 	storageConstants "zotregistry.dev/zot/pkg/storage/constants"
@@ -36,7 +41,11 @@ func (c *CookieStore) RunSessionCleaner(sch *scheduler.Scheduler) {
 	}
 }
 
-func NewCookieStore(storeController storage.StoreController, hashKey, encryptKey []byte) (*CookieStore, error) {
+func NewCookieStore(
+	authCfg *config.AuthConfig,
+	storeController storage.StoreController,
+	log log.Logger,
+) (*CookieStore, error) {
 	// To store custom types in our cookies
 	// we must first register them using gob.Register
 	gob.Register(map[string]interface{}{})
@@ -47,10 +56,109 @@ func NewCookieStore(storeController storage.StoreController, hashKey, encryptKey
 
 	var needsCleanup bool
 
+	if authCfg.SessionDriver == nil {
+		// If the session driver is not configured, then
+		// behave in the usual way for file system cookie store and memory cookie store.
+		createdStore, returnedSessionsDir, doesStoreNeedCleanup, err := localSessionStoreInit(
+			storeController,
+			authCfg.SessionHashKey,
+			authCfg.SessionEncryptKey,
+		)
+		if err != nil {
+			return &CookieStore{}, err
+		}
+
+		store = createdStore
+		sessionsDir = returnedSessionsDir
+		needsCleanup = doesStoreNeedCleanup
+	} else {
+		switch authCfg.SessionDriver["name"] {
+		case storageConstants.RedisDriverName:
+			{
+				prefix, ok := rediscfg.GetString(authCfg.SessionDriver, "keyprefix", false, log)
+				if !ok {
+					prefix = "zotsession"
+				}
+
+				// The redisstore library code uses a colon to separate the prefix
+				// and the actual key and is expected to be part of the prefix argument.
+				// ref: https://github.com/rbcervilla/redisstore/blob/v9.0.0/redisstore.go#L44
+				// This adds a colon to the prefix only if it is not empty.
+				if prefix != "" {
+					prefix += ":"
+				}
+
+				client, err := rediscfg.GetRedisClient(authCfg.SessionDriver, log)
+				if err != nil {
+					return nil, err
+				}
+
+				redisStore, err := redisstore.NewRedisStore(context.Background(), client)
+				if err != nil {
+					return nil, err
+				}
+
+				redisStore.KeyPrefix(prefix)
+				redisStore.Options(sessions.Options{
+					MaxAge: cookiesMaxAge,
+					Path:   "/",
+				})
+
+				store = redisStore
+			}
+		case storageConstants.LocalStorageDriverName:
+			{
+				// This behaves the same as if there was no sessionDriver config.
+				// It is also the same behaviour prior to supporting this config.
+				// This allows for a backwards compatible migration path for upgrades.
+				createdStore, sessDir, cleanupReq, err := localSessionStoreInit(
+					storeController,
+					authCfg.SessionHashKey,
+					authCfg.SessionEncryptKey,
+				)
+				if err != nil {
+					return &CookieStore{}, err
+				}
+
+				store = createdStore
+				sessionsDir = sessDir
+				needsCleanup = cleanupReq
+			}
+		default:
+			return nil, fmt.Errorf(
+				"%w: sessiondriver %s not supported",
+				errors.ErrBadConfig,
+				authCfg.SessionDriver["name"],
+			)
+		}
+	}
+
+	return &CookieStore{
+		Store:        store,
+		rootDir:      sessionsDir,
+		needsCleanup: needsCleanup,
+	}, nil
+}
+
+// Handles creation and init of a local session store.
+// This can be either in memory or on the local file system.
+// Returns a session Store, root directory for the sessions if applicable,
+// a boolean indicating whether clean up is required, and an error.
+func localSessionStoreInit(
+	storeController storage.StoreController,
+	hashKey,
+	encryptKey []byte,
+) (sessions.Store, string, bool, error) {
+	var store sessions.Store
+
+	var sessionsDir string
+
+	var needsCleanup bool
+
 	if storeController.DefaultStore.Name() == storageConstants.LocalStorageDriverName {
 		sessionsDir = path.Join(storeController.DefaultStore.RootDir(), "_sessions")
 		if err := os.MkdirAll(sessionsDir, storageConstants.DefaultDirPerms); err != nil {
-			return &CookieStore{}, err
+			return &CookieStore{}, "", false, err
 		}
 
 		localStore := sessions.NewFilesystemStore(sessionsDir, hashKey, encryptKey)
@@ -67,11 +175,7 @@ func NewCookieStore(storeController storage.StoreController, hashKey, encryptKey
 		store = memStore
 	}
 
-	return &CookieStore{
-		Store:        store,
-		rootDir:      sessionsDir,
-		needsCleanup: needsCleanup,
-	}, nil
+	return store, sessionsDir, needsCleanup, nil
 }
 
 func IsExpiredSession(dirEntry fs.DirEntry) bool {
--- a/pkg/cli/server/root.go
+++ b/pkg/cli/server/root.go
@@ -332,6 +332,58 @@ func validateCacheConfig(cfg *config.Config, logger zlog.Logger) error {
 	return nil
 }
 
+func validateRemoteSessionStoreConfig(cfg *config.Config, logger zlog.Logger) error {
+	// it is okay for the session driver config to be nil
+	// this is backwards compatible for older configs
+	if cfg.HTTP.Auth.SessionDriver == nil {
+		return nil
+	}
+
+	sessionDriverName, ok := cfg.HTTP.Auth.SessionDriver["name"]
+	if !ok {
+		msg := "must provide session driver name!"
+		logger.Error().Err(zerr.ErrBadConfig).Msg(msg)
+
+		return fmt.Errorf("%w: %s", zerr.ErrBadConfig, msg)
+	}
+
+	allowedDriverNames := []string{
+		storageConstants.RedisDriverName,
+		storageConstants.LocalStorageDriverName,
+	}
+
+	isValidDriver := false
+
+	for _, allowedDriverName := range allowedDriverNames {
+		if allowedDriverName == sessionDriverName {
+			isValidDriver = true
+
+			break
+		}
+	}
+
+	if !isValidDriver {
+		msg := fmt.Sprintf("session store driver %s is not allowed!", sessionDriverName)
+		logger.Error().Err(zerr.ErrBadConfig).Msg(msg)
+
+		return fmt.Errorf("%w: %s", zerr.ErrBadConfig, msg)
+	}
+
+	// If the redis driver is being used, then session keys must not be configured
+	// as redis session store does not support these yet.
+
+	if sessionDriverName == storageConstants.RedisDriverName {
+		if cfg.HTTP.Auth.SessionKeysFile != "" {
+			msg := "session keys not supported when redis session driver is used!"
+			logger.Error().Err(zerr.ErrBadConfig).Msg(msg)
+
+			return fmt.Errorf("%w: %s", zerr.ErrBadConfig, msg)
+		}
+	}
+
+	return nil
+}
+
 func validateExtensionsConfig(cfg *config.Config, logger zlog.Logger) error {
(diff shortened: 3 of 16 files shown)


vrajashkr:
Working on manual tests

vrajashkr:
Some Redis tests seem to have failed, but the reason is unknown:
```
  echo "Redis scale-out tests failed"
  exit 1
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.11.13/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.11.13/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.13/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.13/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.13/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.11.13/x64/lib
Redis scale-out tests failed
```

examples/config-remote-session-store-redis.json
```diff
@@ -0,0 +1,35 @@
+{
+  "distSpecVersion": "1.1.1",
+  "storage": {
+    "rootDirectory": "/tmp/zot",
+    "remoteSessionCache": true,
+    "sessionDriver": {
```
andaaron: Shouldn't this be under UI as it is a UI specific setting?
@rchincha

pkg/common/common.go
```diff
@@ -246,3 +248,22 @@ func AreSocketsEqual(socketA string, socketB string) (bool, error) {
 
 	return (hostAIP.Equal(hostBIP) && (portA == portB)), nil
 }
+
+func StrValueFromMapOrDefault(configMap map[string]any, param string, defaultVal string) (string, error) {
```
andaaron: There are some similar functions in https://github.com/project-zot/zot/blob/main/pkg/api/config/redis/redis.go#L249.
To be more specific the [cast ](https://github.com/spf13/cast) library which is also used internally by the library reading the config file (https://github.com/spf13/viper): https://github.com/project-zot/zot/blob/main/pkg/api/config/redis/redis.go#L254C1-L255C30

examples/config-remote-session-store-redis.json
rchincha: yes, especially if zli is not affected by this

rchincha: also be very descriptive in your commit msg about all the config and behavior changes happening in this PR

pkg/common/common.go
vrajashkr: Thanks! I've updated the code to use the function in redis.go instead

examples/config-remote-session-store-redis.json
vrajashkr: Thanks! I've moved the config and placed it under UI config.
I've also updated the commit message with more details.

vrajashkr:
The last run seems to be hitting an unrelated error in sync:

```
Failures:

  * zotregistry.dev/zot/pkg/extensions/sync/sync_test.go 
  Line 5963:
  Expected: 1
  Actual:   2
  (Should equal)!


1382 total assertions

--- FAIL: TestOnDemandPullsOnce (0.42s)
    sync_test.go:5921: Goroutine 0: Sending request to http://127.0.0.1:38673/v2/zot-test/manifests/0.0.1
    sync_test.go:5921: Goroutine 4: Sending request to http://127.0.0.1:38673/v2/zot-test/manifests/0.0.1
    sync_test.go:5921: Goroutine 2: Sending request to http://127.0.0.1:38673/v2/zot-test/manifests/0.0.1
    sync_test.go:5921: Goroutine 1: Sending request to http://127.0.0.1:38673/v2/zot-test/manifests/0.0.1
    sync_test.go:5921: Goroutine 3: Sending request to http://127.0.0.1:38673/v2/zot-test/manifests/0.0.1
FAIL
```

@rchincha Could I get a re-run of this action?

pkg/api/cookiestore.go
```diff
@@ -47,24 +58,62 @@ func NewCookieStore(storeController storage.StoreController, hashKey, encryptKey
...
+				if err != nil {
+					return nil, err
+				}
+
+				redisStore, err := redisstore.NewRedisStore(context.Background(), client)
+				if err != nil {
+					return nil, err
+				}
+
+				redisStore.KeyPrefix(prefix)
```
andaaron, line 101:
```suggestion
				redisStore.KeyPrefix(prefix + ":")
```

Looks like in their library they expect the colon to already be part of the prefix: https://github.com/rbcervilla/redisstore/blob/master/redisstore.go#L44

pkg/cli/server/root.go
```diff
@@ -325,6 +325,45 @@ func validateCacheConfig(cfg *config.Config, log zlog.Logger) error {
 	return nil
 }
 
+func validateRemoteSessionStoreConfig(cfg *config.Config, log zlog.Logger) error {
+	if cfg.Extensions.UI.RemoteSessionStore {
```
andaaron: nitpicking, to reduce indentation levels, you can choose to skip this

```suggestion
	if !cfg.Extensions.UI.RemoteSessionStore {
	    return nil
	}
```

andaaron:
> The last run seems to be hitting an unrelated error in sync:
> 
> ```
> Failures:
> 
>   * zotregistry.dev/zot/pkg/extensions/sync/sync_test.go 
>   Line 5963:
>   Expected: 1
>   Actual:   2
>   (Should equal)!
> 
> 
> 1382 total assertions
> 
> --- FAIL: TestOnDemandPullsOnce (0.42s)
>     sync_test.go:5921: Goroutine 0: Sending request to http://127.0.0.1:38673/v2/zot-test/manifests/0.0.1
>     sync_test.go:5921: Goroutine 4: Sending request to http://127.0.0.1:38673/v2/zot-test/manifests/0.0.1
>     sync_test.go:5921: Goroutine 2: Sending request to http://127.0.0.1:38673/v2/zot-test/manifests/0.0.1
>     sync_test.go:5921: Goroutine 1: Sending request to http://127.0.0.1:38673/v2/zot-test/manifests/0.0.1
>     sync_test.go:5921: Goroutine 3: Sending request to http://127.0.0.1:38673/v2/zot-test/manifests/0.0.1
> FAIL
> ```
> 
> @rchincha Could I get a re-run of this action?

I believe I've seen this in other PRs, I think you can safely rerun.

pkg/cli/server/root.go
vrajashkr: Thanks for the suggestion! I've made this change.

pkg/api/cookiestore.go
vrajashkr: Great catch! Thank you! I've addressed this. I approached it slightly differently as in the case for an empty prefix, we wouldn't want the colon. I've added a comment for more context as well.
Let me know in case we could do it differently.

andaaron (dismissed):

examples/config-remote-session-store-redis.json
```diff
@@ -0,0 +1,35 @@
...
+  "extensions": {
+    "search": {
+      "cve": {
+        "updateInterval": "2h"
+      }
+    },
+    "ui": {
+      "enable": true,
+      "remoteSessionStore": true,
+      "remoteSessionDriver": {
```
rchincha: "sessionStore" {
    "enable": true,
     "driver": "redis"
...
  }
  
  ^ this makes more sense to me

vrajashkr: Thanks for the suggestion! I think that sounds good to me.

I do have a question about how we would want to support existing session drivers though.

For example, since it is called sessionStore, it would apply for sessions being stored on disk and in memory as well. These would be drivers that go by the name of "in-memory" and "filesystem" along with "redis" and "dynamodb" in future.

This would be a challenge for upgrade as users who upgrade, but use local sessions as they would need to update their config files.

An alternative would be to rename the parent as "remoteSessionStore" so it only applies for remote sessions, but I'm not a fan of that because we wouldn't have a config to choose between disk and memory then (in-case someone wants to use it this way).

What do you think? Are there any instances in the past where a forced configuration update was required?

rchincha: > For example, since it is called sessionStore, it would apply for sessions being stored on disk and in memory as well. These would be drivers that go by the name of "in-memory" and "filesystem" along with "redis" and "dynamodb" in future.

Keep the default behavior if this has "enable:false" or this section is missing. Only when scale-out is enabled, this needs to be enabled as well. In fact, I would throw a warning for this.

vrajashkr: I see. If enable is false, wouldn't that read as though session storage as a feature is disabled?

In my opinion, enable false would work well if the setting was called remoteSessionStorage. That way session storage has some default behaviour but remote storage is what is controlled.

If the setting is named just sessionStorage, it would ideally have to be able to configure memory and filesystem based drivers as well.

Do you think it would be suitable to do something like this:
```
{
 remoteSessionStore : {
   enabled: true,
   driver: {
      name: "redis"
   }
}
```

rchincha: Do we need to care about memory and local fs session storage at all in terms of config?

vrajashkr: If the setting is named `sessionStore`, then, for the purpose of consistency, we would need to have the name of driver as a config .

Some possible examples:
```
sessionStore: {
 driver: {
    name: "in-memory"
 }
}
```

```
sessionStore: {
  driver: {
    name: "local-storage"
  }
}
```

```
sessionStore:{
  driver:{
    name: "redis"
    ....(other driver options)
  }
}
```

If we name the config `remoteSessionStore`, then it would only be for the newer redis config and dynamo in future. Existing config remains the same.

rchincha: https://github.com/project-zot/zot/blob/main/examples/config-dynamodb.json#L17

hmm.. we should align with this as much as possible ^
We use the same pattern everywhere.

rchincha: "sessionDriver": {
  "name": "local" <- there is a constant for this
  ...
  }
  
  Would this pattern work for you?

vrajashkr: I think this works well! Thanks!

I'll make the change.

Just to make sure I have the right idea in mind, "local" would use the current zot behavior of in-memory or file system depending on the meta store, correct?

vrajashkr:
@rchincha @andaaron some of the session config seems to be split across settings.

I see we have:

```
SessionKeysFile   string
SessionHashKey    []byte `json:"-"`
SessionEncryptKey []byte `json:"-"`
```

inside the Auth config.

Should we also move the session store to be inside the Auth config since it only really applies when auth is enabled?

Another thing I came across was that the redis store library we are using doesn't appear to support the hashkey and the encrypt keys that the file system store and the memory store support. What do you think?

rchincha:
> @rchincha @andaaron some of the session config seems to be split across settings.
> 
> I see we have:
> 
> ```
> SessionKeysFile   string
> SessionHashKey    []byte `json:"-"`
> SessionEncryptKey []byte `json:"-"`
> ```
> 
> inside the Auth config.
> 
> Should we also move the session store to be inside the Auth config since it only really applies when auth is enabled?
> 
> Another thing I came across was that the redis store library we are using doesn't appear to support the hashkey and the encrypt keys that the file system store and the memory store support. What do you think?

This limits our choices then! :(

We should probably move your config here. Even for a single instance zot, it still does make sense.

examples/config-remote-session-store-redis.json
```diff
@@ -0,0 +1,34 @@
...
+  },
+  "http": {
+    "address": "127.0.0.1",
+    "port": "8080",
+    "realm": "zot",
+    "auth": {
+      "htpasswd": {
+        "path": "/tmp/zotpasswd"
+      },
+      "sessionDriver": {
```
rchincha, line 14: Also validate that "sessionFile" and this section are mutually exclusive - if you specify one, you cannot specify the other.

rchincha:
Also figure out encryption for redis sesssion cache - make this a separate PR and issue to track.

pkg/api/cookiestore.go
```diff
@@ -47,10 +56,109 @@ func NewCookieStore(storeController storage.StoreController, hashKey, encryptKey
 
 	var needsCleanup bool
 
+	if authCfg.SessionDriver == nil {
```
rchincha, line 59: As long as NewCookieStore() is called always with authCfg != nil ^

rchincha:
Also see if you can add some automated bats/nightly/stress tests to validate this PR?

examples/config-remote-session-store-redis.json
vrajashkr: Thanks for the comment! I checked on this and I think `SessionKeysFile` and `SessionDriver` are not mutually exclusive.

From my reading, the session keys file just holds the hashkey and the encryption key and is used as an alternative to raw strings in config.

Here's the reference: https://github.com/project-zot/zot/blob/main/pkg/cli/server/root.go#L897

What do you think?

vrajashkr:
> Also figure out encryption for redis sesssion cache - make this a separate PR and issue to track.

Created #3409 for tracking this

pkg/api/cookiestore.go
vrajashkr: That is correct. `NewCookieStore` is only called when `IsBasicAuthnEnabled()`
ref: https://github.com/project-zot/zot/blob/main/pkg/api/controller.go#L327

examples/config-remote-session-store-redis.json
vrajashkr: Since Redis store doesn't support the 2 keys at the moment, we'll validate this condition. Later, once there is support, the validation can be relaxed.

vrajashkr:
> Also see if you can add some automated bats/nightly/stress tests to validate this PR?

Done! I've added a new BATS test which simulates a bunch of UI requests with curl for 1000 concurrent users.

vrajashkr:
Local testing after all changes

Screenshot from 2025-09-27 20-21-30

vrajashkr:
Seeing a couple of possibly unrelated errors:

Performance regression and Ecosystem client tools

`# docker: Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint http_receiver_failure (8b56df8e038a173f478956c871d32f37e0c62886a773eb5d986301ea42f7a689): failed to bind host port for 0.0.0.0:57256:172.17.0.5:8080/tcp: address already in use`

Could I get these tests re-triggered? @rchincha @andaaron

andaaron:
> Seeing a couple of possibly unrelated errors:
> 
> Performance regression and Ecosystem client tools
> 
> `# docker: Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint http_receiver_failure (8b56df8e038a173f478956c871d32f37e0c62886a773eb5d986301ea42f7a689): failed to bind host port for 0.0.0.0:57256:172.17.0.5:8080/tcp: address already in use`
> 
> Could I get these tests re-triggered? @rchincha @andaaron

Yes.

pkg/cli/server/root.go
```diff
@@ -344,6 +396,11 @@ func validateExtensionsConfig(cfg *config.Config, log zlog.Logger) error {
 
 			return fmt.Errorf("%w: %s", zerr.ErrBadConfig, msg)
 		}
+
+		err := validateRemoteSessionStoreConfig(cfg, log)
```
andaaron: This should no longer be under the extension validation, correct?

vrajashkr: Ah, I missed this. Thanks! I've fixed this now.

vrajashkr:
I've somehow managed to half zot's performance :')

```
Warning: Performance alert! Previous value was 353.15042 and current value is 158.47853. It is 2.228380210240466x worse than previous exceeding a ratio threshold 2
Error: # :warning: **Performance Alert** :warning:

Possible performance regression was detected for benchmark.
Benchmark result of this commit is worse than the previous benchmark result exceeding threshold `2`.

| Benchmark suite | Current: 4e1171325087f4e5a9b316c654f7462488f06b28 | Previous: e6dee9f1e6495bf33291c5520ed3c8d935dad592 | Ratio |
|-|-|-|-|
| `Pull 75% and Push 25% Mixed 1MB` | `158.47853` requests per sec (`3`) | `353.15042` requests per sec (`3`) | `2.23` |
```

I've only been seeing this failure recently. Any hints on what might be causing it?

vrajashkr:
```
# Digest: sha256:2deb0891ec3f643b1d342f04cc22154e6b6a76b41044791b537093fae00b6884
# Status: Image is up to date for python:3
# 305999dd738641a2548f3fbb32a2bf7d5e846987c6bfe2a3e43e97acc1b13030
# docker: Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint http_receiver (47fd62d4dcb520327ed6f7a4c59a38b1c734c97784a024f97f7aff75b5ad370c): failed to bind host port for 0.0.0.0:57530:172.17.0.3:8080/tcp: address already in use
#
# Run 'docker run --help' for more information
```

Need a restart for ecosystem client tests :)

vrajashkr:
Thanks for restarting tests for me . I think there was a race condition in my BATS tests though.

The jobs to sign in, perform requests as users runs concurrently as they are background tasks. Due to this, if the runner is slightly slow, there is no guarantee all of them complete before the cookie count is tested.

Added some waits so that the background jobs can finish before evaluating the values.

andaaron (approved):

vrajashkr:
Depends on #3428 to hopefully address the repeated Ecosystem client tools failures.

andaaron (approved):

Issue #1707, opened 2023-08-16
[Refactor]: Migrate from zerolog to golang's `slog` package for logging


rchincha:
### Is your feature request related to a problem? Please describe.

https://pkg.go.dev/golang.org/x/exp/slog

andaaron:
Let's postpone until we can use the go 1.21 specific imports.

andaaron:
Completed in https://github.com/project-zot/zot/pull/3405

PR #3363, opened 2025-09-04
feat(sync): enable regclient logs

ljakimczuk:
**What type of PR is this?**

feature

N/A

It provides, I hope, minimal Zot logger (Zerolog logger) handler for slog and enables logging for regclient with it.

N/A

N/A

**Automation added to e2e**:

N/A

N/A

N/A

### DIFF
--- a/pkg/extensions/sync/service.go
+++ b/pkg/extensions/sync/service.go
@@ -144,9 +144,9 @@ func (service *BaseService) init() error {
 	service.clientLock.Lock()
 	defer service.clientLock.Unlock()
 
-	client, hosts, err := newClient(service.config, service.credentials)
+	client, hosts, err := newClient(service.config, service.credentials, service.log)
 	if err != nil {
-		service.log.Err(err).Msg("failed to parse sync config urls")
+		service.log.Err(err).Msg("failed to create registry client")
 
 		return err
 	}
@@ -783,7 +783,7 @@ func getTLSConfigOption(url *url.URL, tlsVerify *bool) config.TLSConf {
 	return tls
 }
 
-func newClient(opts syncconf.RegistryConfig, credentials syncconf.CredentialsFile,
+func newClient(opts syncconf.RegistryConfig, credentials syncconf.CredentialsFile, logger log.Logger,
 ) (*regclient.RegClient, []config.Host, error) {
 	urls, err := parseRegistryURLs(opts.URLs)
 	if err != nil {
@@ -875,6 +875,7 @@ func newClient(opts syncconf.RegistryConfig, credentials syncconf.CredentialsFil
 		regclient.WithDockerCreds(),
 		regclient.WithRegOpts(regOpts...),
 		regclient.WithConfigHost(hostConfigOpts...),
+		regclient.WithSlog(logger.Logger),
 	)
 
 	return client, hostConfigOpts, nil
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -308,6 +308,7 @@ func NewLoggerWithWriter(level string, writer io.Writer) Logger {
 // GoroutineID adds goroutine-id to logs to help debug concurrency issues.
 func GoroutineID() int {
 	var buf [64]byte
+
 	n := runtime.Stack(buf[:], false)
 	idField := strings.Fields(strings.TrimPrefix(string(buf[:n]), "goroutine "))[0]
 


rchincha:
Ok with the logs change.

Concurrency change - not so sure, unless it is fixing something that is breaking for you.

ljakimczuk:
@rchincha 

> Concurrency change - not so sure, unless it is fixing something that is breaking for you.

No, nothing is breaking without it, it is just seemed as a nice option to have to tweak behaviour of the regclient, though I am not sure of its usefulness alone without exposing others.

The context is, recently I have been looking at Zot behaviour and ways to speed it up, for on many simultaneous requests for new images the latencies of syncing them grow sometimes to a point clients time out. 

One side of the coin is backoff. On error, including canceled context, Zot (regclient) enters the backoff mechanism which makes all requests to a given registry subject of waiting prescribed time, which sometime grows significantly. 

But even when testing with backoff disabled, it still seems a bit slow sometimes on initial requests for new images, and I just wasn't sure it has to be that way or can be speed up, so I kept looking and then the concurrency option caught my attention. Hence I tried to test that.

I selected a group of relatively small, though not the smallest, images, see below, for which I was issuing HEAD requests to my local instance of Zot to initiate syncing. This instance each time was configured with different concurrent number of requests, default 3 and then 10, 20, 50. These images were selected mostly because they are pulled for new clusters being created and we have observed slowness for them before.

```text
kube-controller-manager:v1.33.4
kube-scheduler:v1.33.4
kube-apiserver:v1.33.4
etcd:v3.3.22
coredns:1.10.0
cert-manager-webhook:v1.18.2
snapshot-controller:v7.0.2
```

When pulling I was looking at packets/sec rates and times of completion of all HEAD requests, so basically time of the longest one, but was also looking at individual times. In general, for this set of images, better results can be achieved with higher values, though seems there a limit to it, which may suggest throttling somewhere else or maybe the natural limit of the process, like writing them to disk for example.

Here is the outcome for the default 3 concurrent requests. It took overall ~32 seconds for all HEAD requests to finish, but images with less / smaller blobs took much less, and the packets/sec graph looked like this:

obraz

Then I changed concurrency option to 10 requests, cleaned everything up, and tested this again. It took ~20 seconds for all HEAD requests to finish, again some took less then that, and the packets/sec graph looked like this:

obraz

The packets exchange rate seems visibly higher than before and the total syncing time seems significantly lower.

Then I tried concurrency of 20 requests, cleaning everything up before testing. It took again ~20 seconds for all HEAD requestes to finish and the packets/sec graph looked like this:

obraz

The initial burst of packets seems slightly higher than for 10 concurrent requests, but later on it seems to stay on a similar level. The result does not seems much different than for 10 concurrent requests.

And as the last option I tried 50 concurrent requests. It again took ~20 seconds for all HEAD requests to finish and the packets/sec graph looked like this:

obraz

The overal time is not better. Though, the initial burst was higher than before and also it seems the overall exchange have lasted for a slightly shorter period than before too.

I executed these tests couple of times always getting similar results for this set of images. I also tried mixing more images into the set. I was also getting better results for higher concurrency, though fair to be said, it wasn't always like 10 seconds difference, but ~6 seconds for example, depending on images and maybe some outside factors too, though I was trying to keep running these test under the same conditions.

Hence I concluded, it seems increasing concurrent requests from 3 to like 10 or 20 speeds up the syncing process for them, but increasing it further doesn't really change that much, at least on my laptop. It seemed enough for proposing to expose this option to configuration, but also I cannot tell these results are ultimately valid. If you do not feel convinced or feel it needs more testing, then we can skip adding it at this time and focus on logger. 

I rely on your judgment 🤝

andaaron:
The backoff is probably meant as a workaround for dockerhub rate limiting.

rchincha:
| It again took ~20 seconds for all HEAD requests

Is this with authentication enabled? basic or bearer?

ljakimczuk:
@rchincha not sure which authentication you mean. The Zot's one? Then no, I wasn't enabling it for my local instance. If you mean authentication from sync extension, towards remote registry, I was using credentials file with username and password.

ljakimczuk:
@rchincha what do you think? Should I reduce this PR to just logger part?

rchincha:
> @rchincha what do you think? Should I reduce this PR to just logger part?

@ljakimczuk sorry for the delay. Yes pls. That is a low-risk and relatively high-reward patch.

ljakimczuk:
@rchincha done, should I do something about the coverage report?

pkg/extensions/config/sync/config.go
```diff
@@ -33,7 +33,9 @@ type RegistryConfig struct {
 	RetryDelay       *time.Duration
 	OnlySigned       *bool
 	CredentialHelper string
-	PreserveDigest   bool // sync without converting
+	PreserveDigest   bool   // sync without converting
+	ClientLogs       bool   // creates regclient log/slog logger
+	ClientLogsLevel  string // defines regclient log/slog logger level
```
rchincha: Would just pass in the global logger and add the logs to that. No reason to do a separate logger which has its own lifecycle headaches

rchincha:
> @rchincha done, should I do something about the coverage report?

For this PR, pls ignore.

pkg/extensions/config/sync/config.go
ljakimczuk: @rchincha sorry for the long delay. Not sure I follow, do you mean getting rid of these two configuration options and using the global [LogConfig](https://github.com/project-zot/zot/blob/main/pkg/api/config/config.go#L208-L212) for instantiating slog.Logger for regclient?

rchincha: https://github.com/project-zot/zot/blob/main/pkg/api/controller.go#L66

^ this gets passed around. You should just use this.

ljakimczuk: @rchincha The regclient expects the [log/slog.Logger](https://pkg.go.dev/log/slog#Logger), while the logger pointed by you, which entire Zot operates on, is the [zotregistry.dev/zot/pkg/log.Logger](https://github.com/project-zot/zot/blob/main/pkg/log/log.go#L20) which is wrapper for the [github.com/rs/zerolog.Logger](https://github.com/rs/zerolog/blob/master/log.go#L229), and both are different types.

A custom logging backend can be plugged in into slog, but in order to do so your logger would have to implement the [log/slog.Handler interface](https://pkg.go.dev/log/slog#Handler), if I understand it correctly. Since we are talking about zerolog wrapper here, maybe an upstream package, like the https://github.com/samber/slog-zerolog, could simply be used for that, but I haven't tried that.

It just felt simpler to just create a new logger, but if you want I can try the slog-handler approach, probably with the  https://github.com/samber/slog-zerolog package to not re-invent the wheel. Only if it does no work, I can then try to implement the interface inside your wrapper.

Or maybe I am missing something obvious here as well, and the logger can be passed to regclient simpler. In any case, let me know what you think.

rchincha: :( https://github.com/project-zot/zot/issues/1707

Ideally, we should move to native golang's slog pkg except there were some limitations iirc.

ljakimczuk: I see, you have an issue for that, I guess though switching to slog requires some considerable amount of work, not technically hard, but one that requires laborious changes all over the place.

What we do with this PR then? Can it be merge as-is, with extra logger, which is though disabled by default? Or do you want me to try the slog-handler approach?

rchincha: >  I guess though switching to slog requires some considerable amount of work

Indeed.

> Can it be merge as-is,

Also not an option that makes a lot of sense. 

Adding another dependency in a already bloated go.mod also not a great option.

Would it be possible just to extract minimally what is needed from that package and put that under
pkg/log/slog_adapter.go?

ljakimczuk: Sure, I can give it a try 🤞

rchincha: @ljakimczuk apologies if this is turning out to be too much work, but appreciate your persistence and will help us link with any slog libs/deps in future until we get the overall slog migration done.

rchincha: @ljakimczuk 

> switching to slog requires some considerable amount of work,

Assigned the https://github.com/project-zot/zot/issues/1707 to GH copilot. Let's see what it comes up with.
https://github.com/rchincha/zot/pull/164

ljakimczuk: @rchincha nice, looks like it managed to produce something under linked PR. Feel free to close my PR if you accept the Copilot's work.

In any case, a word about my work. 

I gave it a try, and managed to create a simple handler. I took the idea from the regclient itself, they have [a handler for the Logrus there](https://github.com/regclient/regclient/tree/main/internal/sloghandle), which seems much simpler than the https://github.com/samber/slog-zerolog. I guess it makes sense, for the [slog-zerolog](https://github.com/samber/slog-zerolog) has to be universal, hence must be richer, while it looks like the Logrus handler is tailored to specific needs of the regclient. Hence it made sense to follow the same patterns. 

Tests on the other hand were inspired by the log/slog package itself and how the logger is tested there. 

I also built the binary and run it in my environment, all looked good. Though the `Ecosystem client tools / Check client tools (pull request)` job fails on this PR, not sure why. 

The `Security web scan for zot / Scan ZOT using ZAP (zot-linux-amd64) (pull_request)` job fails too, but it also fails on the `main` branch, so I guess I can ignore it.

The remaining two are codecov tests again.

ljakimczuk: @rchincha looking through the logs of the client tools CI jobs I found these pieces of information:

Easy one is, it seems there is a connectivity problem in the setup:

```text
2025-09-26T05:53:31.9260202Z # 
2025-09-26T05:53:31.9260427Z   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
2025-09-26T05:53:31.9260741Z   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
2025-09-26T05:53:31.9261128Z # curl: (7) Failed to connect to 127.0.0.1 port 48666 after 0 ms: Couldn't connect to server
2025-09-26T05:53:31.9261590Z # Warning: Problem : connection refused. Will retry in 1 seconds. 50 retries
2025-09-26T05:53:31.9261906Z # Warning: left.
```

And it seems it comes from the [test/blackbox/sync.bats](https://github.com/project-zot/zot/blob/main/test/blackbox/sync.bats#L149):

```text
2025-09-26T05:53:31.9076718Z not ok 1 setup_file failed
2025-09-26T05:53:31.9077438Z # (from function `wait_zot_reachable' in file test/blackbox/helpers_zot.bash, line 44,
2025-09-26T05:53:31.9078109Z #  from function `setup_file' in test file test/blackbox/sync.bats, line 149)
2025-09-26T05:53:31.9078833Z #   `wait_zot_reachable ${zot_port2}' failed with status 7
```

And looking further it seems, the problem is this port has already been used for something:

```text
2025-09-26T05:53:31.9231681Z # {"level":"error","error":"listen tcp 0.0.0.0:48666: bind: address already in use","time":"2025-09-26T05:52:32.047956527Z","message":"failed to start controller, exiting"}
```

ljakimczuk: I squashed the commits and force pushed again, let's see the outcome of this job when executed again.

ljakimczuk: Ok, failed again, this time on the [http_server_start](https://github.com/project-zot/zot/blob/main/test/blackbox/helpers_events.bash#L31) function, but the error is similar:

```text
docker: Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint http_receiver (5611531a51c74b87494b108303caf41ffb7389ae42a35d8aab809c3da6bf5c2c): failed to bind host port for 0.0.0.0:49478:172.17.0.3:8080/tcp: address already in use
```

@rchincha do you know what'up with these tests? I haven't touched them, hence not sure why they fail now.

ljakimczuk: Seems there is also possible data race in the API package the [`Run zot with extensions`](https://github.com/project-zot/zot/actions/runs/18036168023/job/51323524873?pr=3363#step:7:2686) test finds:

```text
WARNING: DATA RACE
2025-09-26T11:24:47.8657426Z Write at 0x00c00302d9c8 by goroutine 11777:
2025-09-26T11:24:47.8657582Z   zotregistry.dev/zot/pkg/api.(*Controller).LoadNewConfig()
2025-09-26T11:24:47.8657782Z       zotregistry.dev/zot/pkg/api/controller.go:382 +0xea
2025-09-26T11:24:47.8657959Z   zotregistry.dev/zot/pkg/cli/server.(*HotReloader).Start.func1.1()
2025-09-26T11:24:47.8658190Z       zotregistry.dev/zot/pkg/cli/server/config_reloader.go:104 +0xf52
2025-09-26T11:24:47.8658198Z 
2025-09-26T11:24:47.8658315Z Previous read at 0x00c00302d9c8 by goroutine 11914:
2025-09-26T11:24:47.8658552Z   zotregistry.dev/zot/pkg/api.(*AuthnMiddleware).basicAuthn()
2025-09-26T11:24:47.8658725Z       zotregistry.dev/zot/pkg/api/authn.go:154 +0x456
...
```

It does not always fail, not sure if you have seen it so far in other tests.

rchincha:
@ljakimczuk this PR has shaped up well imo.

But before you invest more time/effort into it, would first like to evaluate how https://github.com/project-zot/zot/pull/3405 shapes up. Overall, it has similar ideas.

rchincha:
cc: @andaaron

rchincha:
@ljakimczuk 

https://github.com/project-zot/zot/pull/3405 looks done. 
Pls take a look.

rchincha:
@ljakimczuk can you pls rebase? The adapter is no longer needed imo.

ljakimczuk:
@rchincha yup, the adapter is no longer needed, now all works fine with just passing down the logger. A very good job on your PR for integration with slog 🥇

rchincha (approved):
lgtm

PR #3411, opened 2025-09-28
chore: increase/stabilize go test coverage

andaaron:
Issues such as the ones found in:
https://app.codecov.io/gh/project-zot/zot/pull/3055/indirect-changes
https://app.codecov.io/gh/project-zot/zot/pull/3410/indirect-changes


rchincha:
"func" seems to be additional field whenever it is available, we can remove or keep it.

andaaron:
> "func" seems to be additional field whenever it is available, we can remove or keep it.

Can you elaborate? What "func" are you reffering to?

andaaron:
Looks like there is still a case I am not hitting. Will work on it.

pkg/storage/local/driver.go
```diff
@@ -359,8 +359,15 @@ func (fi fileInfo) IsDir() bool {
 	return fi.FileInfo.IsDir()
 }
 
+// fileInterface defines the interface for file operations (internal).
+type fileInterface interface {
```
rchincha: call this "fileInternal"

go.mod
```diff
@@ -22,7 +22,6 @@ require (
 	github.com/chartmuseum/auth v0.5.0
 	github.com/cloudevents/sdk-go/protocol/nats/v2 v2.16.1
 	github.com/cloudevents/sdk-go/v2 v2.16.1
-	github.com/containers/image/v5 v5.36.2
```
rchincha, line 25: We don't need this anymore?

andaaron: It was only imported for a mock, which was actually dead code.
I refactored that mock to match the current sync struct, and this line was deleted by go mod tidy.

rchincha (dismissed):

PR #3428, opened 2025-10-03
fix(ci): use fixed ranges for BATS server ports

vrajashkr:
**What type of PR is this?**
bug

In multiple CI runs, there is often a port overlap where `get_free_port` tests and returns a random port, but before a server can bind to it, another process binds to it and causes a test failure unrelated to the changes.

- Replaces the get_free_port bash function in BATS tests with get_free_port_for_service that returns a random free port in a given range for a test file and service defined in a ports.json file.
- Updates all get_free_port calls to use the new function.
- A new README file for details on the ports.json file.
- Updates some tests using fixed ports to use dynamic ports.
- Adds a ports.json file with all the allocations.
- Adds a new common helper for port fetching.

Seen in multiple CI runs

```
docker: Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint http_receiver_lint (e519fd55200c97533f0245fcca501d6621be0e89f36923775f8812d7793fe3a4): failed to bind host port for 0.0.0.0:38494:172.17.0.5:8080/tcp: address already in use
```

CI run should succeed

N/A

No user facing changes. Developers adding BATS tests must note the fixed port range allocations and make changes accordingly. README in the test directory has more details.

### DIFF
--- a/test/blackbox/annotations.bats
+++ b/test/blackbox/annotations.bats
@@ -3,6 +3,7 @@
 #       Extra tools that are not covered in Makefile target needs to be added in verify_prerequisites()
 
 load helpers_zot
+load ../port_helper
 
 function verify_prerequisites {
     if [ ! $(command -v curl) ]; then
@@ -35,7 +36,8 @@ function setup_file() {
     local zot_root_dir=${BATS_FILE_TMPDIR}/zot
     local zot_config_file=${BATS_FILE_TMPDIR}/zot_config.json
     mkdir -p ${zot_root_dir}
-    zot_port=$(get_free_port)
+
+    zot_port=$(get_free_port_for_service "zot")
     echo ${zot_port} > ${BATS_FILE_TMPDIR}/zot.port
     cat > ${zot_config_file}<<EOF
 {
--- a/test/scale-out/cloud_scale_out_redis.bats
+++ b/test/scale-out/cloud_scale_out_redis.bats
@@ -9,6 +9,7 @@ load helpers_zot
 load helpers_cloud
 load helpers_haproxy
 load helpers_redis
+load ../port_helper
 
 function verify_prerequisites() {
     if [ ! $(command -v docker) ]; then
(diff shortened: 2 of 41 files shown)


test/blackbox/annotations.bats
```diff
@@ -35,7 +35,8 @@ function setup_file() {
     local zot_root_dir=${BATS_FILE_TMPDIR}/zot
     local zot_config_file=${BATS_FILE_TMPDIR}/zot_config.json
     mkdir -p ${zot_root_dir}
-    zot_port=$(get_free_port)
+
+    zot_port=$(get_free_port_in_range 1300 1309)
```
andaaron: I suggest not to keep this in the tests as someone may copy/paste in the future and not notice. Or we may forget we are not using a range after we delete some tests.
Why not keep this data in a single file (in json format?) and read it from there (using jq?).

vrajashkr: Good idea! Thanks! I'll take a look at doing this.

rchincha: maybe in a helper file or source a script

vrajashkr: I've made this change. Currently stuck on a strange failure in cloud_only.bats though.
Looks unrelated to the port itself, but maybe something in the URL

I suspect this diff here:
https://github.com/project-zot/zot/pull/3428/files#diff-b7bcf313c3880682129ebfba248d5f93c44182935828b9f35fde6c9f8e902305L120-R124

vrajashkr: Reverted the test file to use static ports as this might have something to do with the fixed dex OIDC configuration.

This test runs alone so there is no need to use dynamic ports.

test/scale-out/cloud_scale_out_redis_scale.bats
andaaron, line 1: Can we also update the ports used by zot in these tests?

test/scale-out/cloud_scale_out_redis.bats
andaaron, line 1: Can we also update the ports used by zot in these tests?

test/scale-out/cloud_scale_out_redis_scale.bats
vrajashkr: We could. One challenge is that the ports used by zot are in series depending on the number of instances of zot that are spun up and we use this assumption in the haproxy config generation as well.

Since the tests run independently and the range starts at 10000, it is unlikely for zot to hit a conflicting port.

Redis just needed one port and could have overlapped, so it has been restricted to an allocated range.

andaaron: This test is the one failed at https://github.com/project-zot/zot/actions/runs/18194045373/job/51802106219?pr=3345, correct?
We should update this if we can.

test/scale-out/cloud_scale_out_redis.bats
vrajashkr: I've made this change - ports for zot and haproxy are also included now

test/scale-out/cloud_scale_out_redis_scale.bats
vrajashkr: I've made this change - ports for zot and haproxy are also included now

PR #3447, opened 2025-10-10
feat: the default retention delay is not the GC delay

andaaron:
Most users don't make the difference between retention deleting untagged manifests vs GC deleting other blobs.
This may cause confusion since the GC delay and the retention delay (used for untagged manifests and orphan referrers) have different defaults, and are set separately in the zot configuration.
Most users don't configure retention policies, and they still expect untagged manifests to be deleted at GC time.

With this change, if retention delay is not specified in the config file, the value used is the GC delay. If GC delay is also unspecified in the config file, the default GC delay is used for both.

### DIFF
--- a/pkg/cli/server/root.go
+++ b/pkg/cli/server/root.go
@@ -724,6 +724,10 @@ func applyDefaultValues(config *config.Config, viperInstance *viper.Viper, logge
 		if viperInstance.Get("storage::gcinterval") == nil {
 			config.Storage.GCInterval = 0
 		}
+	} else if !viperInstance.IsSet("storage::retention::delay") {
+		// if GC is enabled, retentionDelay is set to gcDelay by default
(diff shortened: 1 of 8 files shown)


pkg/cli/server/root.go
```diff
@@ -724,6 +724,10 @@ func applyDefaultValues(config *config.Config, viperInstance *viper.Viper, logge
 		if viperInstance.Get("storage::gcinterval") == nil {
 			config.Storage.GCInterval = 0
 		}
+	} else if !viperInstance.IsSet("storage::retention::delay") {
+		// if GC is enabled, retentionDelay is set to gcDelay by default
+		// it could be default gcDelay or the custom value set in the config file
+		config.Storage.Retention.Delay = config.Storage.GCDelay
```
rchincha, line 730: This is the main change. Thanks.

Issue #3445, opened 2025-10-08
[Feat]: FIPS approved crypto


TylerGillson:
### Is your feature request related to a problem? Please describe.

Deploying zot with `GODEBUG=fips140=only` produces the following panic:

```
panic: crypto/sha1: use of SHA-1 is not allowed in FIPS 140-only mode

goroutine 1 [running]:
crypto/sha1.(*digest).checkSum(0xb0?)
    crypto/sha1/sha1.go:165 +0x17c
crypto/sha1.(*digest).Sum(0x40000d62a0, {0x0, 0x0, 0x0})
    crypto/sha1/sha1.go:159 +0x70
github.com/go-redsync/redsync/v4/redis.NewScript(0x1, {0xaaaae472a43d, 0xa1})
    github.com/go-redsync/redsync/v4@v4.13.0/redis/redis.go:43 +0x104
github.com/go-redsync/redsync/v4.init()
    github.com/go-redsync/redsync/v4@v4.13.0/mutex.go:234 +0x2c
```

### Describe the solution you'd like

Explore alternatives to `github.com/go-redsync/redsync/v4` or make upstream changes to said dependency.

### Describe alternatives you've considered

Updating `github.com/go-redsync/redsync/v4` directly. If you'd prefer, I'd be happy to open an issue there and close this one.

### Additional context

https://github.com/go-redsync/redsync/issues/194

rchincha:
@TylerGillson posted a PR https://github.com/go-redsync/redsync/pull/203

rchincha:
@TylerGillson once https://github.com/project-zot/zot/pull/3451 is merged, this issue can be resolved.

rchincha:
v2.1.9 has now been released which fixes this.

PR #3451, opened 2025-10-12
fix: update go-redsync for fips-140 compatibility

rchincha:
Fixes issue #3445

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/cli/server/root.go
+++ b/pkg/cli/server/root.go
@@ -2,7 +2,6 @@ package server
 
 import (
 	"context"
-	"crypto/fips140"
 	"errors"
 	"fmt"
 	"net"
@@ -51,11 +50,6 @@ func newServeCmd(conf *config.Config) *cobra.Command {
 		Args:    cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			logger := zlog.NewLogger("info", "")
-
-			// log early about fips140
-			if fips140.Enabled() {
-				logger.Info().Msg("fips140 is currently enabled")
-			}
 
 			if len(args) > 0 {
 				if err := LoadConfiguration(conf, args[0]); err != nil {
--- a/test/blackbox/fips140.bats
+++ b/test/blackbox/fips140.bats
@@ -54,6 +54,7 @@ EOF
     git -C ${BATS_FILE_TMPDIR} clone https://github.com/project-zot/helm-charts.git
     zot_serve ${ZOT_PATH} ${zot_config_file}
     wait_zot_reachable ${zot_port}
+    log_output | jq 'contains("fips140 is currently enabled")?' | grep true
 }
 
 function teardown() {
@@ -63,6 +64,7 @@ function teardown() {
 
 function teardown_file() {
     zot_stop_all
+    unset GODEBUG
 }
 
 @test "push image" {
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"context"
+	"crypto/fips140"
 	"crypto/tls"
 	"crypto/x509"
 	goerrors "errors"
@@ -254,6 +255,11 @@ func (c *Controller) Run() error {
 }
 
 func (c *Controller) Init() error {
+	// report if fips140 mode is enabled
+	if fips140.Enabled() {
+		c.Log.Info().Msg("fips140 is currently enabled")
+	}
+
 	// print the current configuration, but strip secrets
 	c.Log.Info().Interface("params", c.Config.Sanitize()).Msg("configuration settings")
 


rchincha:
Should we just make this default in our container images?

test/blackbox/fips140.bats
```diff
@@ -0,0 +1,67 @@
...
+        "address": "0.0.0.0",
+        "port": "${zot_port}"
+    },
+    "log": {
+        "level": "debug",
+        "output": "${BATS_FILE_TMPDIR}/zot.log"
+    }
+}
+EOF
+    git -C ${BATS_FILE_TMPDIR} clone https://github.com/project-zot/helm-charts.git
```
andaaron: What is this for?

andaaron:
I suggest we also emit a log message stating fips is on, in case it is on, just to validate. And verify that in bats. Maybe use https://pkg.go.dev/crypto/fips140#Enabled

rchincha:
> I suggest we also emit a log message stating fips is on, in case it is on, just to validate. And verify that in bats. Maybe use https://pkg.go.dev/crypto/fips140#Enabled

{"time":"2025-10-12T08:46:58.596597075-07:00","level":"info","message":"fips140 is currently enabled","goroutine":1,"caller":"zotregistry.dev/zot/pkg/cli/server/root.go:57","func":"zotregistry.dev/zot/pkg/cli/server.NewServerRootCmd.newServeCmd.func2"}

test/blackbox/fips140.bats
rchincha: fixed

andaaron:
> Should we just make this default in our container images?

I dont' know if that is something we want.

Given https://go.dev/doc/security/fips140: 
1. there could be performance implications:
`Pairwise consistency tests are performed on generated cryptographic keys. Note that this can cause a slowdown of up to 2x for certain key types, which is especially relevant for ephemeral keys.`
2. there could be backwards compatibility implications for HTTPs:
`The [crypto/tls](https://go.dev/pkg/crypto/tls/) package will ignore and not negotiate any protocol version, cipher suite, signature algorithm, or key exchange mechanism that is not FIPS 140-3 approved.`
3. Is BLAKE-3 on any approved FIPS compliance list? It is not here https://csrc.nist.gov/Projects/cryptographic-module-validation-program/sp-800-140-series-supplemental-information/sp800-140c

test/blackbox/fips140.bats
rchincha: Did a full push pull test with this setting just to validate things don't crash, else not very useful.

pkg/cli/server/root.go
```diff
@@ -51,6 +52,11 @@ func newServeCmd(conf *config.Config) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			logger := zlog.NewLogger("info", "")
 
+			// log early about fips140
+			if fips140.Enabled() {
+				logger.Info().Msg("fips140 is currently enabled")
```
andaaron: Yes, this will go to stdout, I would emit after we initialize logging with the settings in the config.
Or is there a case of zot crashing earlier?

test/blackbox/fips140.bats
```diff
@@ -0,0 +1,389 @@
...
+    wait_zot_reachable ${zot_port}
+}
+
+function teardown() {
+    # conditionally printing on failure is possible from teardown but not from from teardown_file
+    cat ${BATS_FILE_TMPDIR}/zot.log
+}
+
+function teardown_file() {
+    zot_stop_all
```
andaaron, line 66: And unset GODEBUG?

test/blackbox/fips140.bats
```diff
@@ -0,0 +1,389 @@
...
+    "log": {
+        "level": "debug",
+        "output": "${BATS_FILE_TMPDIR}/zot.log"
+    }
+}
+EOF
+    export GODEBUG="fips140=only"
+    git -C ${BATS_FILE_TMPDIR} clone https://github.com/project-zot/helm-charts.git
+    zot_serve ${ZOT_PATH} ${zot_config_file}
+    wait_zot_reachable ${zot_port}
```
andaaron, line 56: We should check if zot is running in FIPS mode based on the log message.

test/blackbox/fips140.bats
```diff
@@ -0,0 +1,389 @@
...
+        "port": "${zot_port}"
+    },
+    "log": {
+        "level": "debug",
+        "output": "${BATS_FILE_TMPDIR}/zot.log"
+    }
+}
+EOF
+    export GODEBUG="fips140=only"
+    git -C ${BATS_FILE_TMPDIR} clone https://github.com/project-zot/helm-charts.git
```
andaaron, line 54: Can you remove this? Should not be needed.

rchincha: we are now doing a full push pull including helm charts right?

andaaron: Makes sense, sorry.

PR #3499, opened 2025-10-30
fix: close file handle before moving file in FullBlobUpload

andaaron:
Should fix a Windows specific issue where renaming a file fails if the handler is not closed.


rchincha:
Would ask the original poster to test this when ready.

shpatel2:
> Would ask the original poster to test this when ready.

@rchincha  I tested on windows amd64 and it works great! Thanks!

PR #3405, opened 2025-09-25
fix: migrate from github.com/rs/zerolog to golang-native log/slog

rchincha:
We have been using zerolog for a really long time. golang now has structured logging using slog.
Best to move to this in interests of long-term support.

This is a tech debt item.

**What type of PR is this?**

**Which issue does this PR fix**:

**What does this PR do / Why do we need it**:

**If an issue # is not available please add repro steps and logs showing the issue**:

**Testing done on this change**:

**Automation added to e2e**:

**Will this break upgrades or downgrades?**

**Does this PR introduce any user-facing change?**:

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3501, opened 2025-10-31
fix(log): ensure func record is correct

rchamarthy:
**What type of PR is this?**
bug

The "func" field in log output is incorrect in some cases and showing internal logging methods instead of the actual calling function. This is happening because the caller information is being captured statically (5 call frames) which can be different for different call patterns.

Changes:

* Move the caller capture to the event creation
* Use deterministic skip (3 frames) for event creation
* Add test cases to verify that the caller is captured correctly

Note: tests needed to written outside convey to avoid call insertion

Fixes the logging to log the right caller function all the time.

```
{"time":"2025-10-29T07:48:26.089019762Z","level":"info","message":"executing gc of orphaned blobs for /data/zot/images/trinetra/observer","module":"gc","goroutine":54,"caller":"zotregistry.dev/zot/pkg/storage/gc/gc.go:89","func":"zotregistry.dev/zot/pkg/log.(*Event).Str"}
{"time":"2025-10-29T07:48:26.201378659Z","level":"debug","message":"manifests without tags","module":"gc","repository":"trinetra/observer","goroutine":54,"caller":"zotregistry.dev/zot/pkg/storage/gc/gc.go:479","func":"zotregistry.dev/zot/pkg/log.(*Event).Str"}
{"time":"2025-10-29T07:48:26.210318247Z","level":"debug","message":"cleaning orphan blobs","module":"gc","repository":"trinetra/observer","goroutine":54,"caller":"zotregistry.dev/zot/pkg/storage/gc/gc.go:613","func":"zotregistry.dev/zot/pkg/log.(*Event).Str"}
{"time":"2025-10-29T07:48:26.36994592Z","level":"error","message":"failed to run GC for /data/zot/images/trinetra/observer","error":"manifest not found","module":"gc","goroutine":54,"caller":"zotregistry.dev/zot/pkg/storage/gc/gc.go:93","func":"zotregistry.dev/zot/pkg/storage/gc.GarbageCollect.CleanRepo"}
{"time":"2025-10-29T07:48:26.369991399Z","level":"info","message":"gc unsuccessfully completed for /data/zot/images/trinetra/observer","module":"gc","goroutine":54,"caller":"zotregistry.dev/zot/pkg/storage/gc/gc.go:95","func":"zotregistry.dev/zot/pkg/log.(*Event).Str"}
```

Added a test case with the PR that covers the changes.

**Automation added to e2e**:

Added a test case with the PR that covers the changes.

No

No

```release-note
improve zot logs for debuggability
```

### DIFF
--- a/pkg/cli/server/config_reloader_test.go
+++ b/pkg/cli/server/config_reloader_test.go
@@ -494,11 +494,13 @@ func TestConfigReloader(t *testing.T) {
 		// wait for config reload
 		time.Sleep(5 * time.Second)
 
+		// Wait for the async trivy download to fail and log the error
 		found, err := test.ReadLogFileAndSearchString(logFile.Name(),
 			"failed to download trivy-db to destination dir", 30*time.Second)
 		So(err, ShouldBeNil)
 		So(found, ShouldBeTrue)
 
+		// Now read the file once and check all the expected log content
 		data, err := os.ReadFile(logFile.Name())
 		So(err, ShouldBeNil)
 		t.Logf("log file: %s", data)
@@ -508,9 +510,12 @@ func TestConfigReloader(t *testing.T) {
 		So(string(data), ShouldContainSubstring, "\"UpdateInterval\":18000000000000")
 		So(string(data), ShouldContainSubstring, "\"Scrub\":null")
 		So(string(data), ShouldContainSubstring, "\"DBRepository\":\"another/unreachable/trivy/url2\"")
-		// matching log message when it errors out, test that indeed the download will try the second url
+
+		// Just verify the new URL appears in the logs to confirm config reload worked and ignore
+		// the order of json message formatting that can change independent of this functional
+		// test.
 		found, err = test.ReadLogFileAndSearchString(logFile.Name(),
-			"\"dbRepository\":\"index.docker.io/another/unreachable/trivy/url2:2\",\"goroutine", 1*time.Minute)
+			"index.docker.io/another/unreachable/trivy/url2", 1*time.Minute)
 		So(err, ShouldBeNil)
 		So(found, ShouldBeTrue)
 	})
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -15,9 +15,10 @@ import (
 )
 
 const (
-	defaultPerms         = 0o0600
-	messageKey           = "message"
-	callerSkipFrameCount = 5
+	defaultPerms = 0o0600
+	messageKey   = "message"
+	// Skip: runtime.Callers(0), newEvent(1), Info/Error/etc(2), actual caller(3).
+	callerSkipFrameCount = 3
 )
 
 // Logger extends slog's Logger with zerolog-compatible API.
@@ -27,45 +28,72 @@ type Logger struct {
 
 // Event represents a log event, mimicking zerolog.Event.
 type Event struct {
-	logger  *Logger
-	level   slog.Level
-	attrs   []slog.Attr
-	isPanic bool
+	logger   *Logger
+	level    slog.Level
+	attrs    []slog.Attr
+	isPanic  bool
+	caller   string
+	function string
+}
+
+// newEvent creates a new log event with caller information captured at the point of creation.
+func (l Logger) newEvent(level slog.Level, isPanic bool) *Event {
+	var caller, function string
+
+	// Get the program counter at the caller's location
+	pc := make([]uintptr, 1)
+	n := runtime.Callers(callerSkipFrameCount, pc)
+
+	if n > 0 {
+		frames := runtime.CallersFrames(pc)
+		frame, _ := frames.Next()
+		caller = fmt.Sprintf("%s:%d", frame.File, frame.Line)
+		function = frame.Function
+	}
+
+	return &Event{
+		logger:   &l,
+		level:    level,
+		attrs:    []slog.Attr{},
+		isPanic:  isPanic,
+		caller:   caller,
+		function: function,
+	}
 }
 
 // Info returns an event for info level logging.
 func (l Logger) Info() *Event {
-	return &Event{logger: &l, level: slog.LevelInfo, attrs: []slog.Attr{}}
+	return l.newEvent(slog.LevelInfo, false)
 }
 
 // Debug returns an event for debug level logging.
 func (l Logger) Debug() *Event {
-	return &Event{logger: &l, level: slog.LevelDebug, attrs: []slog.Attr{}}
+	return l.newEvent(slog.LevelDebug, false)
 }
 
 // Error returns an event for error level logging.
 func (l Logger) Error() *Event {
-	return &Event{logger: &l, level: slog.LevelError, attrs: []slog.Attr{}}
+	return l.newEvent(slog.LevelError, false)
 }
 
 // Warn returns an event for warn level logging.
 func (l Logger) Warn() *Event {
-	return &Event{logger: &l, level: slog.LevelWarn, attrs: []slog.Attr{}}
+	return l.newEvent(slog.LevelWarn, false)
 }
 
 // Panic returns an event for panic level logging (maps to error + panic).
 func (l Logger) Panic() *Event {
-	return &Event{logger: &l, level: slog.LevelError, attrs: []slog.Attr{}, isPanic: true}
+	return l.newEvent(slog.LevelError, true)
 }
 
 // Fatal returns an event for fatal level logging (maps to error + panic).
 func (l Logger) Fatal() *Event {
-	return &Event{logger: &l, level: slog.LevelError, attrs: []slog.Attr{}, isPanic: true}
+	return l.newEvent(slog.LevelError, true)
 }
 
 // Err logs an error directly on the logger (convenience method).
 func (l Logger) Err(err error) *Event {
-	event := l.Error()
+	event := l.newEvent(slog.LevelError, false)
 	if err != nil {
 		event.attrs = append(event.attrs, slog.String("error", err.Error()))
 	}
@@ -75,7 +103,7 @@ func (l Logger) Err(err error) *Event {
 
 // With returns a logger with additional context.
 func (l Logger) With() *Event {
-	return &Event{logger: &l, level: slog.LevelInfo, attrs: []slog.Attr{}}
+	return l.newEvent(slog.LevelInfo, false)
 }
 
 // Logger returns the logger from an event (for method chaining).
@@ -188,7 +216,20 @@ func NewTestLoggerPtr() *Logger {
 // Msgf logs the event with a formatted message.
 func (e *Event) Msgf(format string, args ...interface{}) {
 	msg := fmt.Sprintf(format, args...)
-	e.logger.LogAttrs(nil, e.level, msg, e.attrs...)
+
+	// Create a new slice to avoid modifying the original
+	attrs := make([]slog.Attr, len(e.attrs), len(e.attrs)+2)
+	copy(attrs, e.attrs)
+
+	if e.caller != "" {
+		attrs = append(attrs, slog.String("caller", e.caller))
+	}
+
+	if e.function != "" {
+		attrs = append(attrs, slog.String("func", e.function))
+	}
+
+	e.logger.LogAttrs(nil, e.level, msg, attrs...)
 
 	if e.isPanic {
 		panic(msg)
@@ -197,7 +238,20 @@ func (e *Event) Msgf(format string, args ...interface{}) {
 
 // Msg logs the event with a simple message.
 func (e *Event) Msg(msg string) {
-	e.logger.LogAttrs(nil, e.level, msg, e.attrs...)
+	// Add caller and function info to attributes if captured
+	// Create a new slice to avoid modifying the original
+	attrs := make([]slog.Attr, len(e.attrs), len(e.attrs)+2)
+	copy(attrs, e.attrs)
+
+	if e.caller != "" {
+		attrs = append(attrs, slog.String("caller", e.caller))
+	}
+
+	if e.function != "" {
+		attrs = append(attrs, slog.String("func", e.function))
+	}
+
+	e.logger.LogAttrs(nil, e.level, msg, attrs...)
 
 	if e.isPanic {
 		panic(msg)
@@ -330,18 +384,8 @@ func (h *CallerHandler) Enabled(ctx context.Context, level slog.Level) bool {
 }
 
 func (h *CallerHandler) Handle(ctx context.Context, record slog.Record) error {
-	// Add caller information
-	if pc, file, line, ok := runtime.Caller(callerSkipFrameCount); ok { // Adjust stack depth as needed
-		frame := runtime.CallersFrames([]uintptr{pc})
-		f, _ := frame.Next()
-
-		record.Add("caller", fmt.Sprintf("%s:%d", file, line))
-
-		if f.Function != "" {
-			record.Add("func", f.Function)
-		}
-	}
-
+	// Caller information is now added directly in Event.Msg/Msgf methods
+	// This handler is kept for compatibility but no longer modifies the record
 	return h.handler.Handle(ctx, record)
 }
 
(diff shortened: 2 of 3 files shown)


rchincha:
We should check all the logs carefully.

pkg/log/log.go
```diff
@@ -330,18 +378,8 @@ func (h *CallerHandler) Enabled(ctx context.Context, level slog.Level) bool {
...
-
-		record.Add("caller", fmt.Sprintf("%s:%d", file, line))
-
-		if f.Function != "" {
-			record.Add("func", f.Function)
-		}
-	}
-
+	// Caller information is now added directly in Event.Msg/Msgf methods
+	// This handler is kept for compatibility but no longer modifies the record
```
andaaron, line 388: I think we can remove the handler entirely if not used.

pkg/log/log.go
```diff
@@ -17,7 +17,7 @@ import (
 const (
 	defaultPerms         = 0o0600
 	messageKey           = "message"
-	callerSkipFrameCount = 5
+	callerSkipFrameCount = 2 // Skip: runtime.Caller and newEvent
```
andaaron: If not used. remove it or maybe we can update to 3. See my other comment.

pkg/log/log.go
```diff
@@ -27,40 +27,66 @@ type Logger struct {
...
+	function string
+}
+
+// newEvent creates a new log event with caller information captured at the point of creation.
+func (l Logger) newEvent(level slog.Level, isPanic bool) *Event {
+	var caller, function string
+	// Get the program counter at the caller's location
+	pc := make([]uintptr, 1)
+	// Skip: runtime.Callers(0), newEvent(1), Info/Error/etc(2), actual caller(3)
+	n := runtime.Callers(3, pc)
```
andaaron: Maybe
```suggestion
	n := runtime.Callers(callerSkipFrameCount, pc)
```
With callerSkipFrameCount being 3

pkg/log/log.go
```diff
@@ -188,7 +214,18 @@ func NewTestLoggerPtr() *Logger {
...
+
+	if e.caller != "" {
+		attrs = append(attrs, slog.String("caller", e.caller))
+	}
+
+	if e.function != "" {
+		attrs = append(attrs, slog.String("func", e.function))
+	}
+
+	e.logger.LogAttrs(nil, e.level, msg, attrs...)
```
andaaron, line 232: Do these changes also cover point 1 I was asking about here? https://github.com/project-zot/zot/pull/3405#issuecomment-3344341098
It would be nice if they do, but not a blocker or anything.

pkg/log/log_test.go
```diff
@@ -340,3 +340,85 @@ func TestNewAuditLogger(t *testing.T) {
 		}, ShouldPanic)
 	})
 }
+
+func TestCallerFunction(t *testing.T) {
```
andaaron, line 344: Let me see if I understand correctly:
1. The logger functioned correctly in cases such as `logger.Info().Msg("test message 1")`
2. In cases where other functions were called after event initialization, such as `logger.Info().Str("key1", "val2").Str("key2", "val2").Int("key3", 1).Msg("test message 1")`, the old code should have been counting an extra frame to be skipped for each of them in the Handle() method (in this case 3), but was not.
3. By moving the frame counting to newEvent(), the calls made between newEvent() and Handle() no longer matter.

andaaron (commented):
Thank you @rchamarthy for the PR

pkg/log/log.go
rchamarthy: agreed.

pkg/log/log.go
rchamarthy: yes adjusted as such.

pkg/log/log.go
rchamarthy: No. we are using slog. Looking at slog looks like will always put time, level and msg first and then all attrs you add. We can intercept and modify this order but that will require deeper changes. We can do that a separate PR.

rchincha: No, order not that important except for humans, even so, we can probably jq that.

rchincha:
@rchamarthy also take a look at GH copilot reviews, sometimes gets it right.

pkg/log/log.go
rchamarthy: lets keep it in case of other customizations for later.

rchamarthy:
> @rchamarthy also take a look at GH copilot reviews, sometimes gets it right.

Yes looked at it and fixed it with a minor performance penalty not too bad.

pkg/log/log_test.go
rchamarthy: The idea is to calculate the caller function and file in a deterministic call stack and that is when we are creating the event itself. The rest of the amends modifiy the same event so the first call is what matters even if we chain many times.

rchamarthy:
@rchincha  attached are the logs with the new binary:

```
loki:/zot/bin fix/log ⎈no-cluster
➜ ./zot-linux-arm64 serve config.json
{"time":"2025-10-31T11:06:42.670300417-07:00","level":"info","message":"configuration settings","params":{"distSpecVersion":"1.1.1","GoVersion":"go1.25.2":"","Audit":""},"Extensions":null,"scheduler":null,"cluster":null},"caller":"zotregistry.dev/zot/v2/pkg/api/controller.go:266","func":"zotregistry.dev/zot/v2/pkg/api.(*Controller).Init","goroutine":1}
{"time":"2025-10-31T11:06:42.670434627-07:00","level":"info","message":"runtime params","cpus":8,"max. open files":1048576,"listen backlog":"4096","max. inotify watches":"1048576","caller":"zotregistry.dev/zot/v2/pkg/api/runtime.go:19","func":"zotregistry.dev/zot/v2/pkg/api.DumpRuntimeParams","goroutine":1}
{"time":"2025-10-31T11:06:42.670490669-07:00","level":"info","message":"events disabled in configuration","caller":"zotregistry.dev/zot/v2/pkg/extensions/extension_events.go:18","func":"zotregistry.dev/zot/v2/pkg/extensions.NewEventRecorder","goroutine":1}
{"time":"2025-10-31T11:06:42.670641754-07:00","level":"debug","message":"started","caller":"zotregistry.dev/zot/v2/pkg/common/healthz.go:63","func":"zotregistry.dev/zot/v2/pkg/common.(*Healthz).Started","goroutine":1}
{"time":"2025-10-31T11:06:42.670818672-07:00","level":"info","message":"cve config not provided, skipping cve-db update","caller":"zotregistry.dev/zot/v2/pkg/extensions/extension_search.go:61","func":"zotregistry.dev/zot/v2/pkg/extensions.EnableSearchExtension","goroutine":1}
{"time":"2025-10-31T11:06:42.670904215-07:00","level":"info","message":"metrics config not provided, skipping metrics config update","caller":"zotregistry.dev/zot/v2/pkg/extensions/extension_metrics.go:21","func":"zotregistry.dev/zot/v2/pkg/extensions.EnableMetricsExtension","goroutine":1}
{"time":"2025-10-31T11:06:42.670931715-07:00","level":"debug","message":"generator is done","component":"scheduler","generator":"GCTaskGenerator","caller":"zotregistry.dev/zot/v2/pkg/scheduler/scheduler.go:480","func":"zotregistry.dev/zot/v2/pkg/scheduler.(*generator).generate","goroutine":40}
{"time":"2025-10-31T11:06:42.670960632-07:00","level":"info","message":"scrub config not provided, skipping scrub","caller":"zotregistry.dev/zot/v2/pkg/extensions/extension_scrub.go:50","func":"zotregistry.dev/zot/v2/pkg/extensions.EnableScrubExtension","goroutine":1}
{"time":"2025-10-31T11:06:42.6710573-07:00","level":"info","message":"sync config not provided or disabled, so not enabling sync","caller":"zotregistry.dev/zot/v2/pkg/extensions/extension_sync.go:82","func":"zotregistry.dev/zot/v2/pkg/extensions.EnableSyncExtension","goroutine":1}
{"time":"2025-10-31T11:06:42.671407511-07:00","level":"warn","message":"skipping enabling swagger because given zot binary doesn't include this feature, please build a binary that does so","caller":"zotregistry.dev/zot/v2/pkg/debug/swagger/swagger_disabled.go:20","func":"zotregistry.dev/zot/v2/pkg/debug/swagger.SetupSwaggerRoutes","goroutine":1}
{"time":"2025-10-31T11:06:42.67147472-07:00","level":"warn","message":"skipping enabling graphql playground extension because given zot binary doesn't include this feature, please build a binary that does so","caller":"zotregistry.dev/zot/v2/pkg/debug/gqlplayground/gqlplayground_disabled.go:17","func":"zotregistry.dev/zot/v2/pkg/debug/gqlplayground.SetupGQLPlaygroundRoutes","goroutine":1}
{"time":"2025-10-31T11:06:42.671537263-07:00","level":"info","message":"setting up metrics routes","caller":"zotregistry.dev/zot/v2/pkg/extensions/extension_metrics.go:28","func":"zotregistry.dev/zot/v2/pkg/extensions.SetupMetricsRoutes","goroutine":1}
{"time":"2025-10-31T11:06:42.67158968-07:00","level":"info","message":"skip enabling the search route as the config prerequisites are not met","caller":"zotregistry.dev/zot/v2/pkg/extensions/extension_search.go:86","func":"zotregistry.dev/zot/v2/pkg/extensions.SetupSearchRoutes","goroutine":1}
{"time":"2025-10-31T11:06:42.671633514-07:00","level":"info","message":"skip enabling the image trust routes as the config prerequisites are not met","caller":"zotregistry.dev/zot/v2/pkg/extensions/extension_image_trust.go:33","func":"zotregistry.dev/zot/v2/pkg/extensions.SetupImageTrustRoutes","goroutine":1}
{"time":"2025-10-31T11:06:42.671683306-07:00","level":"info","message":"skip enabling the mgmt route as the config prerequisites are not met","caller":"zotregistry.dev/zot/v2/pkg/extensions/extension_mgmt.go:90","func":"zotregistry.dev/zot/v2/pkg/extensions.SetupMgmtRoutes","goroutine":1}
{"time":"2025-10-31T11:06:42.67172239-07:00","level":"info","message":"skip enabling the user preferences route as the config prerequisites are not met","caller":"zotregistry.dev/zot/v2/pkg/extensions/extension_userprefs.go:34","func":"zotregistry.dev/zot/v2/pkg/extensions.SetupUserPreferencesRoutes","goroutine":1}
{"time":"2025-10-31T11:06:42.671771307-07:00","level":"info","message":"skip enabling the ui route as the config prerequisites are not met","caller":"zotregistry.dev/zot/v2/pkg/extensions/extension_ui.go:68","func":"zotregistry.dev/zot/v2/pkg/extensions.SetupUIRoutes","goroutine":1}
{"time":"2025-10-31T11:06:42.673486282-07:00","level":"debug","message":"ready","caller":"zotregistry.dev/zot/v2/pkg/common/healthz.go:58","func":"zotregistry.dev/zot/v2/pkg/common.(*Healthz).Ready","goroutine":1}
{"time":"2025-10-31T11:06:42.773454644-07:00","level":"info","message":"no repositories found in storage, finished.","component":"dedupe","caller":"zotregistry.dev/zot/v2/pkg/storage/common/common.go:905","func":"zotregistry.dev/zot/v2/pkg/storage/common.(*DedupeTaskGenerator).Next","goroutine":40}
{"time":"2025-10-31T11:06:42.773567562-07:00","level":"debug","message":"generator is done","component":"scheduler","generator":"DedupeTaskGenerator","caller":"zotregistry.dev/zot/v2/pkg/scheduler/scheduler.go:480","func":"zotregistry.dev/zot/v2/pkg/scheduler.(*generator).generate","goroutine":40}
```

andaaron (approved):

andaaron:
@rchamarthy, please check the test failure. It probably waits for a specific log message which changed and may need an update.
```
"dbRepository":"index.docker.io/another/unreachable/trivy/url2:2","caller"
```
Instead of what the test was expecting.

pkg/log/log.go
```diff
@@ -27,45 +28,72 @@ type Logger struct {
...
+	var caller, function string
+
+	// Get the program counter at the caller's location
+	pc := make([]uintptr, 1)
+	n := runtime.Callers(callerSkipFrameCount, pc)
+
+	if n > 0 {
+		frames := runtime.CallersFrames(pc)
+		frame, _ := frames.Next()
+		caller = fmt.Sprintf("%s:%d", frame.File, frame.Line)
```
rchincha, line 50: so if n == 0, then "caller" is never set?

rchincha:
Logs are much cleaner btw

pkg/log/log.go
rchamarthy: I dont think n can be 0 give the rewind we do by 3 frames.

pkg/cli/server/config_reloader_test.go
```diff
@@ -508,11 +510,11 @@ func TestConfigReloader(t *testing.T) {
 		So(string(data), ShouldContainSubstring, "\"UpdateInterval\":18000000000000")
 		So(string(data), ShouldContainSubstring, "\"Scrub\":null")
 		So(string(data), ShouldContainSubstring, "\"DBRepository\":\"another/unreachable/trivy/url2\"")
-		// matching log message when it errors out, test that indeed the download will try the second url
-		found, err = test.ReadLogFileAndSearchString(logFile.Name(),
```
andaaron, line 512: This function reads the log file in a loop until the string is found (with 1 minute timeout).
In the new code there is no guarantee the background task attempting the download is done within the initial 5 seconds.

rchamarthy: @andaaron from reading the code I thought once the condition occurs the data is already there in the log. Is this a separate condition than above? If so I will move the check back to ReadLogFileAndSearchString. The fix is actually the string we need to search.

rchamarthy: Reverting to the old test with new string.

andaaron: Apparently, "data" is used mainly to check the first log message shown when the config is loaded. 5 is supposed to be safe to assume that message shows. The other message is supposed to verify the new configuration is actually used by the trivy logic. That is a background task run by the scheduler, so a function reading the logfile in a loop was used to make sure the test waits enough.

rchamarthy: As a rule I dont like looking at logs to determine functional state, but in this case may be its legitimate :)

pkg/cli/server/config_reloader_test.go
```diff
@@ -508,9 +510,12 @@ func TestConfigReloader(t *testing.T) {
...
 		So(string(data), ShouldContainSubstring, "\"Scrub\":null")
 		So(string(data), ShouldContainSubstring, "\"DBRepository\":\"another/unreachable/trivy/url2\"")
-		// matching log message when it errors out, test that indeed the download will try the second url
+
+		// Just verify the new URL appears in the logs to confirm config reload worked and ignore
+		// the order of json message formatting that can change independent of this functional
+		// test.
 		found, err = test.ReadLogFileAndSearchString(logFile.Name(),
-			"\"dbRepository\":\"index.docker.io/another/unreachable/trivy/url2:2\",\"goroutine", 1*time.Minute)
+			"index.docker.io/another/unreachable/trivy/url2", 1*time.Minute)
```
andaaron, line 518:
```suggestion
			"\"dbRepository\":\"index.docker.io/another/unreachable/trivy/url2:2\", 1*time.Minute)
```
otherwise it also matches the other log message also checked above in `data`

andaaron: Actually this should be fine too. But it is close.
Maybe use `failed to download artifact from index.docker.io/another/unreachable/trivy/url2:2: OCI repository` should be distinct enough from the config log message

rchamarthy: I was going for the minimum unique string I need to match. but i dont understand the test in too much details. If you think I should change the matching string I can push another amend. Let me know.

andaaron: Leave it as is. The "failed to..." is already checked above. I don't know why we would have a second wait in a loop for the same thing... Your initial change should have worked too...
Maybe the test was patched a few times without a serious review/refactor.

This is good from my POV.

pkg/cli/server/config_reloader_test.go
andaaron: We don't have another way. Do you have suggestions?

PR #3503, opened 2025-11-02
fix: gracefully handle manifests missing from storage (prepare for sparse indexes)

andaaron:
GC and scrub should not stop if a manifest or index is missing from storage. Other similar changes are also included.

### DIFF
--- a/pkg/extensions/sync/destination.go
+++ b/pkg/extensions/sync/destination.go
@@ -12,6 +12,7 @@ import (
 	"path"
 	"strings"
 
+	"github.com/distribution/distribution/v3/registry/storage/driver"
 	godigest "github.com/opencontainers/go-digest"
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/regclient/regclient/types/mediatype"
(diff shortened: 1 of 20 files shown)


pkg/extensions/sync/destination.go
```diff
@@ -254,6 +271,11 @@ func (registry *DestinationRegistry) copyManifest(repo string, desc ispec.Descri
 			}
 		}
 
+		// Return error if we encountered any missing manifests
+		if firstMissingErr != nil {
+			return firstMissingErr
+		}
```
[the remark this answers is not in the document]
andaaron: This is fine for the scope of this PR. Sync is still supposed to copy all blobs, as we don't support syncing a subset of manifests with specific architectures in this PR. That is going to be a subsequent PR.

pkg/extensions/sync/destination.go
```diff
@@ -227,11 +228,27 @@ func (registry *DestinationRegistry) copyManifest(repo string, desc ispec.Descri
...
 			manifestBuf, err := tempImageStore.GetBlobContent(repo, manifest.Digest)
 			if err != nil {
+				// Handle missing manifest blobs gracefully - log warning and continue with other manifests
+				var pathNotFoundErr driver.PathNotFoundError
+				if errors.Is(err, zerr.ErrBlobNotFound) || errors.As(err, &pathNotFoundErr) {
+					if firstMissingErr == nil {
+						firstMissingErr = err
+					}
+
+					registry.log.Warn().Err(err).Str("dir", path.Join(tempImageStore.RootDir(), repo)).
```
chinchani, line 245: registry.log... ??

andaaron: Yes, what is wrong here?

PR #3513, opened 2025-11-03
fix: support custom OAuth2 URLs for GitHub Enterprise and self-hosted providers

analytically:
- Use custom authURL/tokenURL from config instead of hardcoded github.com endpoints
- Properly configure GitHub Enterprise API base URL from auth endpoints

Fixes OAuth2 authentication with GitHub Enterprise Server and other self-hosted OAuth2 providers.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -346,7 +346,7 @@ In the case of running zot with openid enabled behind a proxy/load balancer http
   "http": {
     "address": "0.0.0.0",
     "port": "8080",
-    "externalUrl: "https://zot.example.com",
+    "externalUrl": "https://zot.example.com",
     "auth": {
       "openid": {
         "providers": {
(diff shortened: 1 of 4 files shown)


examples/README.md
```diff
@@ -362,6 +362,40 @@ In the case of running zot with openid enabled behind a proxy/load balancer http
 ```
 This config value will be used by oauth2/openid clients to redirect back to zot.
 
+### OpenID/OAuth2 Social Login with Custom URLs (Self-Hosted Providers)
```
rchincha, line 365: This doc change belongs here actually.

https://github.com/project-zot/project-zot.github.io/blob/main/docs/articles/authn-authz.md#social-login-using-openidoauth2

Would also add this or a summary to the commit msg.

rchincha:
@analytically thanks for your PR btw. Without this change, zot is not able to interface with GH enterprise edition?

analytically:
> @analytically thanks for your PR btw. Without this change, zot is not able to interface with GH enterprise edition?

That's correct. GH Enterprise Server is a custom deployment (on-site) of GitHub, typically for larger enterprises. And they use custom urls. Hence the GitHub OAuth library supports these. With this PR, Zot can be configured to use these URLs.

analytically:
screentshot-2025-11-03 at 17 05 17@2x

rchincha:
@analytically is there a test we can include here https://github.com/project-zot/zot/tree/main/test and/or a mock UT, not only to test but also protect this investment?

analytically:
The GitHub authentication flow is currently not tested, so it's hard for me to add a comprehensive test suite for such a relatively small change.

Issue #2523, opened 2024-07-03
[Feat]: Allow specifying user and group scopes with OIDC authentication


AndersBennedsgaard:
### Is your feature request related to a problem? Please describe.

It should be possible to specify which claim to use for the user ID. Some users might want to use `email`, others might want to use `sub`.
Additionally, it should be possible to select a claim which is used for groups. This is already possible with LDAP: https://github.com/project-zot/zot/blob/002ff05f6e5ab3870aca75e616e8a4ed9b4a5c68/pkg/api/config/config.go#L165

### Describe the solution you'd like

Add `userclaim` and `groupsclaim` to OIDC provider config

rchincha:
@AndersBennedsgaard thanks for doing a thorough evaluation and identifying improvements for zot.
We would also respectfully urge you to consider posting PRs.

AndersBennedsgaard:
I have opened PRs in OSS projects before which have been closed with a "we do not agree with this" explanation. So I don't create PRs before I get an approval from the developers that it is something they want. If you do think it's a good improvement, and wouldn't mind contributions for it, you can say that you agree and add the https://github.com/project-zot/zot/labels/good%20first%20issue label

andaaron:
Reopening since https://github.com/project-zot/zot/pull/3999 has been opened.

PR #3518, opened 2025-11-07
feat: allow claim mapping for user name with oidc

skymoore:
**What type of PR is this?**

feature

N/A

This PR adds support for configurable OIDC claim mapping, allowing administrators to specify which claim should be used as the username during OpenID authentication.

Currently, Zot hardcodes the use of the `email` claim as the username for OIDC authentication. Many OIDC providers (like Zitadel, Keycloak, Auth0) provide a `preferred_username` claim or other custom claims that are more appropriate for use as usernames, especially in enterprise environments where email addresses may change or where usernames follow different conventions.

This change introduces a new optional `claimMapping` configuration field in the OpenID provider configuration that allows specifying which OIDC claim to use as the username, with fallback to `email` for backward compatibility.

**Before this change:**
1. Configure Zot with an OIDC provider (e.g., Zitadel) that provides `preferred_username` claim
2. Authenticate via OIDC
3. Zot uses the email address as the username, not the preferred_username

**Example configuration with this change:**
```json
{
  "http": {
    "auth": {
      "openid": {
        "providers": {
          "oidc": {
            "name": "Zitadel",
            "issuer": "https://iam.example.com",
            "credentialsFile": "/etc/zot/oidc-credentials.json",
            "scopes": ["openid", "profile", "email", "groups"],
            "claimMapping": {
              "username": "preferred_username"
            }
          }
        }
      }
    }
  }
}
```

**Supported claims:**
- `preferred_username` - The preferred username from OIDC
- `email` - Email address (default/fallback behavior)
- `sub` - Subject identifier (unique user ID)
- `name` - Full name
- Any custom claim name - Extracted from UserInfo claims

**Unit tests:**
```bash
$ go test -short -timeout 5m ./pkg/api/...
ok  	zotregistry.dev/zot/v2/pkg/api	4.140s
ok  	zotregistry.dev/zot/v2/pkg/api/config	0.190s
ok  	zotregistry.dev/zot/v2/pkg/api/config/redis	0.601s
ok  	zotregistry.dev/zot/v2/pkg/api/errors	0.711s
```

**Manual testing:**
1. Built and deployed Zot with the changes
2. Configured OIDC provider (Zitadel) with `claimMapping.username: "preferred_username"`
3. Authenticated via OIDC and verified username extraction from `preferred_username` claim
4. Verified access control policies work with the extracted username
5. Tested fallback behavior - removed `claimMapping` and confirmed email is used as username
6. Verified backward compatibility - existing configurations without `claimMapping` continue to work
7. Tested with custom claim names
8. Verified debug logging shows correct claim extraction

**Build verification:**
```bash
$ make binary
# Successful compilation with no errors
```

**Automation added to e2e**:

No - This would require setting up a mock OIDC provider in the test environment. Recommend creating a follow-up issue for adding integration tests.

No. This change is fully backward compatible:
- **Upgrades:** Existing configurations without `claimMapping` will continue to work exactly as before (using email claim)
- **Downgrades:** The new `claimMapping` field is optional and will be ignored by older versions
- **Default behavior:** When no claim mapping is configured, the email claim is used as the default/fallback

Yes - this adds a new optional configuration field.

```release-note
Add support for configurable OIDC claim mapping via the `claimMapping.username` field in OpenID provider configuration. This allows administrators to specify which OIDC claim (e.g., `preferred_username`, `sub`, `name`, or custom claims) should be used as the username instead of the default `email` claim. The change is fully backward compatible - existing configurations will continue to use the email claim as before.
```

### DIFF
--- a/test/blackbox/openid_claim_mapping.bats
+++ b/test/blackbox/openid_claim_mapping.bats
@@ -1,0 +1,430 @@
+# Note: Intended to be run as "make run-blackbox-tests" or "make run-blackbox-ci"
+#       Makefile target installs & checks all necessary tooling
+#       Extra tools that are not covered in Makefile target needs to be added in verify_prerequisites()
+
+load helpers_zot
+load helpers_wait
+load ../port_helper
+
+function verify_prerequisites {
+    if [ ! $(command -v curl) ]; then
+        echo "you need to install curl as a prerequisite to running the tests" >&3
+        return 1
+    fi
+
+    if [ ! $(command -v jq) ]; then
+        echo "you need to install jq as a prerequisite to running the tests" >&3
+        return 1
+    fi
+
+    if [ ! $(command -v docker) ]; then
+        echo "you need to install docker as a prerequisite to running the tests" >&3
+        return 1
+    fi
+
+    return 0
+}
+
+DEX_CONTAINER_NAME="dex_oidc_claim_mapping_test"
+
+function setup_file() {
+    # Verify prerequisites are available
+    if ! $(verify_prerequisites); then
+        exit 1
+    fi
+
+    # Pre-allocate all ports we'll need
+    dex_port=$(get_free_port_for_service "dex")
+    zot_port_1=$(get_free_port_for_service "zot_preferred_username")
+    zot_port_2=$(get_free_port_for_service "zot_email")
+    zot_port_3=$(get_free_port_for_service "zot_sub")
+    zot_port_4=$(get_free_port_for_service "zot_default")
+    
+    echo ${dex_port} > ${BATS_FILE_TMPDIR}/dex.port
+
+    # Setup dex OIDC provider
+    local dex_config_file=${BATS_FILE_TMPDIR}/dex_config.yaml
+
+    # Create dex config with mockCallback connector and all redirect URIs
+    cat > ${dex_config_file}<<EOF
+issuer: http://127.0.0.1:${dex_port}/dex
+
+storage:
+  type: memory
+
+web:
+  http: 0.0.0.0:${dex_port}
+
+staticClients:
+  - id: zot-client
+    redirectURIs:
+      - 'http://127.0.0.1:${zot_port_1}/zot/auth/callback/oidc'
+      - 'http://127.0.0.1:${zot_port_2}/zot/auth/callback/oidc'
+      - 'http://127.0.0.1:${zot_port_3}/zot/auth/callback/oidc'
+      - 'http://127.0.0.1:${zot_port_4}/zot/auth/callback/oidc'
+    name: 'zot'
+    secret: ZXhhbXBsZS1hcHAtc2VjcmV0
+
+connectors:
+  - type: mockCallback
+    id: mock
+    name: Example
+
+enablePasswordDB: true
+EOF
+
+    echo "Starting dex on port ${dex_port}..." >&3
+    echo "Zot ports will be: ${zot_port_1}, ${zot_port_2}, ${zot_port_3}, ${zot_port_4}" >&3
+
+    # Start dex in docker container
+    docker run -d --name ${DEX_CONTAINER_NAME} \
+        -p ${dex_port}:${dex_port} \
+        -v ${dex_config_file}:/etc/dex/config.yaml \
+        ghcr.io/dexidp/dex:v2.37.0 \
+        dex serve /etc/dex/config.yaml
+
+    # Wait for dex to be ready
+    echo "Waiting for dex to be ready on port ${dex_port}..." >&3
+    local dex_url=http://127.0.0.1:${dex_port}/dex/.well-known/openid-configuration
+    local max_attempts=60
+    local attempt=0
+    while [ $attempt -lt $max_attempts ]; do
+        if curl -f -s -o /dev/null ${dex_url} 2>&1; then
+            echo "Dex is ready!" >&3
+            break
+        fi
+        attempt=$((attempt + 1))
+        sleep 2
+    done
+    
+    if [ $attempt -eq $max_attempts ]; then
+        echo "Dex failed to start within timeout" >&3
+        docker logs ${DEX_CONTAINER_NAME} >&3
+        return 1
+    fi
+}
+
+function teardown_file() {
+    zot_stop_all
+    docker stop ${DEX_CONTAINER_NAME} 2>/dev/null || true
+    docker rm ${DEX_CONTAINER_NAME} 2>/dev/null || true
+}
+
+function teardown() {
+    # conditionally printing on failure is possible from teardown but not from teardown_file
+    if [ -f "${BATS_FILE_TMPDIR}/zot-preferred-username.log" ]; then
+        cat ${BATS_FILE_TMPDIR}/zot-preferred-username.log
+    fi
+    if [ -f "${BATS_FILE_TMPDIR}/zot-email.log" ]; then
+        cat ${BATS_FILE_TMPDIR}/zot-email.log
+    fi
+    if [ -f "${BATS_FILE_TMPDIR}/zot-sub.log" ]; then
+        cat ${BATS_FILE_TMPDIR}/zot-sub.log
+    fi
+    if [ -f "${BATS_FILE_TMPDIR}/zot-default.log" ]; then
+        cat ${BATS_FILE_TMPDIR}/zot-default.log
+    fi
+}
+
+function dex_session() {
+    local zot_port=${1}
+    local dex_port=$(cat ${BATS_FILE_TMPDIR}/dex.port)
+    
+    STATE=$(curl -L -f -s http://localhost:${zot_port}/zot/auth/login?provider=oidc | grep -m 1 -oP '(?<=state=)[^ ]*"' | cut -d \" -f1)
+    echo "STATE: $STATE" >&3
+    
+    RESPONSE=$(curl -L -f -s "http://127.0.0.1:${dex_port}/dex/auth/mock?client_id=zot-client&redirect_uri=http%3A%2F%2F127.0.0.1%3A${zot_port}%2Fzot%2Fauth%2Fcallback%2Foidc&response_type=code&scope=profile+email+groups+openid&state=$STATE")
+    echo "$RESPONSE"
+}
+
+@test "test OIDC claim mapping with preferred_username" {
+    local zot_root_dir=${BATS_FILE_TMPDIR}/zot-preferred-username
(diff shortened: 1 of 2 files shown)


rchincha:
@skymoore are there workarounds for this? without this PR I mean? The issue is whether such claims are mutable and potentially insecure.

skymoore:
> @skymoore are there workarounds for this? without this PR I mean? The issue is whether such claims are mutable and potentially insecure.

Great question!  However, this applies equally to the current email based approach, which is also mutable.  There are no workarounds without this PR, the current code is hard coded to use the email as username.  I suppose you could do some hacky stuff to your identity provider to work around this, but that's definitely not recommended.

Current:
- Emails can change (user updates email in IdP)
- Emails can be reassigned (someone leaves a company, new person gets same email)
- Result: same identity confusion and security issues

With the PR:
- Usernames can change (user updates username in IdP)
- Usernames can be reassigned (someone leaves a company, new person gets same username)
- Result: same identity confusion and security issues.

This PR does not introduce _new_ security risks.  It provides the flexibility to choose which claim to use based on the administrators IdP's guarantees.  This actually enhances security by providing administrators the ability to use the sub claim which is actually immutable.

pkg/api/routes.go
```diff
@@ -1998,17 +1998,69 @@ func (rh *RouteHandler) GithubCodeExchangeCallback() rp.CodeExchangeCallback[*oi
...
+					case "sub":
+						username = info.Subject
+					case "name":
+						username = info.Name
+					default:
+						// Try to get from custom claims in UserInfo
+						if val, ok := info.Claims[claimName].(string); ok {
+							username = val
+						}
+					}
```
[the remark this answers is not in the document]
skymoore: pretty sure that it will always be a string in the standard oidc implementation.

skymoore:
@rchincha I could enhance this PR to use sub as the internal id and just display the username claim (which is actually the proper way to handle this situation).  That completely eliminates the identity stability and security concerns, but it's a larger change and thus introduces more risk.  I can do that in this PR or in a follow up PR.  What do you think?

skymoore:
> Will review this later.
> 
> From what I remember back then we chose email for one of the reasons - 1. the name may not have been unique and we assumed the email is, and/or the characters in the email would be safe to use as keys in the DBs (not 100% sure if it was this or some other place where we were concerned about the characters). 

This may help dispel any assumptions one may have about emails.

https://e-mail.wtf/

In my opinion email addresses are in no way safer for the db than user names.

skymoore:
> Will review this later. From what I remember back then we chose email for one of the reasons - 1. the name may not have been unique and we assumed the email is, and/or the characters in the email would be safe to use as keys in the DBs (not 100% sure if it was this or some other place where we were concerned about the characters).

Also the default behavior (email as username) remains unchanged unless you explicitly configure a different claim mapping in the OIDC provider configuration.

andaaron:
> Will review this later. From what I remember back then we chose email for one of the reasons - 1. the name may not have been unique and we assumed the email is, and/or the characters in the email would be safe to use as keys in the DBs (not 100% sure if it was this or some other place where we were concerned about the characters).

OK, this may seem silly but now I remember, the use case we had in mind was: if the same person logged in with different providers (not at the same time), how do we know if it is the same returning person? So we assumed that if he reused the same email with all providers it is the same person.

skymoore:
> > Will review this later. From what I remember back then we chose email for one of the reasons - 1. the name may not have been unique and we assumed the email is, and/or the characters in the email would be safe to use as keys in the DBs (not 100% sure if it was this or some other place where we were concerned about the characters).
> 
> OK, this may seem silly but now I remember, the use case we had in mind was: if the same person logged in with different providers (not at the same time), how do we know if it is the same returning person? So we assumed that if he reused the same email with all providers it is the same person.

Ah yeah, kinda silly.  I think a better path would be to have some sort of index that tracks the sub claims when a user signs in and links different sub claims from different providers to the email addresses.  I believe it is safe to assume same email == same person, but I think fundamentally with oidc we need to track sub claims for internal identification of users.

rchincha:
@skymoore also would it be possible to add a e2e test under or separate similar to test/blackbox/cloud_only.bats

rchincha:
Look unrelated to this PR but

  * zotregistry.dev/zot/v2/pkg/extensions/search/cve/trivy/scanner_test.go 
  Line 213:
  Expected: nil
  Actual:   'open /home/ubuntu/_work/zot/zot/test/data/spring-web/blobs/sha256/506c47a6827e325a63d4b38c7ce656e07d5e98a09d748ec7ac989a45af7d6567: no such file or directory'
  
  CVE UTs are failing

skymoore:
> The code itself looks ok. Please rebase, I suspect the test failure is something transient.

done

skymoore:
> @skymoore also would it be possible to add a e2e test under or separate similar to test/blackbox/cloud_only.bats

@rchincha 
Please see my original PR description:

>Automation added to e2e:
>
>No - This would require setting up a mock OIDC provider in the test environment. Recommend creating a follow-up issue for adding integration tests.

rchincha:
| No - This would require setting up a mock OIDC provider in the test environment. Recommend creating a follow-up issue for adding integration tests.

We do have dex as an option under tests/dex. Would that be an option?

skymoore:
> | No - This would require setting up a mock OIDC provider in the test environment. Recommend creating a follow-up issue for adding integration tests.
> 
> 
> 
> We do have dex as an option under tests/dex. Would that be an option?

I will take a look.

skymoore:
> | No - This would require setting up a mock OIDC provider in the test environment. Recommend creating a follow-up issue for adding integration tests.
> 
> We do have dex as an option under tests/dex. Would that be an option?

@rchincha Added bats test:
```
bats --print-output-on-failure test/blackbox/openid_claim_mapp
ing.bats
openid_claim_mapping.bats
   fetching free port for service dex in blackbox/openid_claim_mapping.bats range 20040 to 20049
   returning free port for service dex in blackbox/openid_claim_mapping.bats => 20046
   fetching free port for service zot_preferred_username in blackbox/openid_claim_mapping.bats range 2
0060 to 20069
   returning free port for service zot_preferred_username in blackbox/openid_claim_mapping.bats => 200
69
   fetching free port for service zot_email in blackbox/openid_claim_mapping.bats range 20080 to 20089
   returning free port for service zot_email in blackbox/openid_claim_mapping.bats => 20086
   fetching free port for service zot_sub in blackbox/openid_claim_mapping.bats range 21000 to 21009
   returning free port for service zot_sub in blackbox/openid_claim_mapping.bats => 21002
   fetching free port for service zot_default in blackbox/openid_claim_mapping.bats range 21020 to 210
29
   returning free port for service zot_default in blackbox/openid_claim_mapping.bats => 21025
Starting dex on port 20046...
Zot ports will be: 20069, 20086, 21002, 21025
Waiting for dex to be ready on port 20046...
Dex is ready!
   test OIDC claim mapping with preferred_username                                                1/4
 ✓ test OIDC claim mapping with preferred_username
0060 to 20069
   returning free port for service zot_preferred_username in blackbox/openid_claim_mapping.bats => 200
69
STATE: 
 ✓ test OIDC claim mapping with email
   fetching free port for service zot_email in blackbox/openid_claim_mapping.bats range 20080 to 20089
   returning free port for service zot_email in blackbox/openid_claim_mapping.bats => 20082
STATE: 
 ✓ test OIDC claim mapping with sub
   fetching free port for service zot_sub in blackbox/openid_claim_mapping.bats range 21000 to 21009
   returning free port for service zot_sub in blackbox/openid_claim_mapping.bats => 21003
STATE: 
   test OIDC with no claim mapping (default to email)                                             4/4
 ✓ test OIDC with no claim mapping (default to email)
29
   returning free port for service zot_default in blackbox/openid_claim_mapping.bats => 21026
STATE: 

4 tests, 0 failures

```

test/blackbox/openid_claim_mapping.bats
andaaron, line 1: Please reference this file in https://github.com/project-zot/zot/blob/main/test/blackbox/ci.sh, so it runs for PRs

test/blackbox/openid_claim_mapping.bats
```diff
@@ -0,0 +1,430 @@
...
+}
+EOF
+
+    zot_serve ${ZOT_PATH} ${zot_config_file}
+    wait_zot_reachable ${zot_port}
+
+    run dex_session ${zot_port}
+    [ "$status" -eq 0 ]
+    
+    # Verify that the log contains the preferred_username claim being used
```
andaaron, line 210: How does this verification work, what is the log message to expect?

pkg/api/routes.go
```diff
@@ -1998,17 +1998,69 @@ func (rh *RouteHandler) GithubCodeExchangeCallback() rp.CodeExchangeCallback[*oi
...
+
+					rh.c.Log.Debug().
```
andaaron, line 2044: I am not convinced we should be logging a message on every successful auth, including success, but given this would be only for UI users, and after the initial success sessions are used for the following requests, I think we will be fine.

rchincha: This is at Debug level, and as long as the production default is higher at error etc, should be fine.

test/blackbox/openid_claim_mapping.bats
```diff
@@ -0,0 +1,430 @@
...
+      - 'http://127.0.0.1:${zot_port_2}/zot/auth/callback/oidc'
+      - 'http://127.0.0.1:${zot_port_3}/zot/auth/callback/oidc'
+      - 'http://127.0.0.1:${zot_port_4}/zot/auth/callback/oidc'
+    name: 'zot'
+    secret: ZXhhbXBsZS1hcHAtc2VjcmV0
+
+connectors:
+  - type: mockCallback
+    id: mock
+    name: Example
```
rchincha, line 71: You probably want to add preferredUsername: "testuser" etc under here to actually test

rchincha, line 71: You probably want to add preferredUsername: "testuser" etc under here to actually test

rchincha:
```
connectors:
  - type: mockCallback
    id: mock
    name: Example
    config:
      # The identity returned by the mockCallback connector
      # This includes the preferred_username
      preferredUsername: "testuser"
```      
like so ^?

andaaron:
This would cover https://github.com/project-zot/zot/issues/2523, correct?

rchincha:
> This would cover #2523, correct?

Perhaps the userClaim part, but groupClaim looks separate.

rchincha:
@skymoore hopefully you are able to complete the work on this PR.

andaaron:
currently being worked on in #3540

Issue #1395, opened 2023-04-30
[Feat]: for better UX, stream blobs as you fetch when on-demand sync is enabled


rchincha:
### Is your feature request related to a problem? Please describe.

Currently, we fetch blobs and store locally and then serve. It could cause some clients to error out. So basically, on-demand is mainly a trigger. A better UX for clients would be to stream to them as the blobs are fetched.

### Describe the solution you'd like

For this feature, we may have to enhance or replace the containers/image library entirely and use our own custom library.

You may need a pattern like so (or a combo):
https://pkg.go.dev/io#MultiWriter
https://pkg.go.dev/io#TeeReader

rchincha:
@liangyuanpeng ^ 
https://liangyuanpeng.com/post/new-container-image-registry-of-zot-from-cncf/

rchincha:
A blob that is sync'ed is in three possible states
1. **Found**, so not need to trigger sync on client pulls
2. **NotFound**, so need to trigger sync and add a callback (and store a http.Response ref)
3. **Partial**, sync is still in progress and a client is requesting the blob so it must be added to the list maintained in 2.

Keep in mind locking and memory usage (channels, buffers, etc cost us memory)

rchincha:
https://github.com/project-zot/zot/compare/main...rchincha:zot:issue-1395
^ just some quick pseudo-code

peusebiu:
Hello,

I encountered some issues with fsnotify package and with containers/image(used by sync) package while working on this:
so we create a temp directory (uuid.New()) where containers/image will copy image before commiting it to zot's imagestore.

1st issue with fsnotify is that it doesn't support watching dirs recursively, trying to overcome this by adding subdirs to watchlist whenever a create event comes in, will result in races (a create event could be triggered before adding a subdir to watchlist)

2nd issue with container/image is that they first write to a temp blob file in the form of "oci-put-blob3507476942" before renaming it to the actual blob path.

so I can not identify blobs being written by digest.

basically in sync: we write a blob in a temp file, then move it to a temp directory, then move it to imagestore, trying to keep all these actions synchronized would be really hard.

elee1766:
see: https://pkg.go.dev/github.com/mailgun/multibuf?utm_source=godoc#WriterOnce

rchincha:
For sync feature, add a "stream: true" to enable this ux.

rchincha:
There are pros and cons.

Pros: UX is better

Cons: blob validation etc now falls to the client

[...]

Issue #2589, opened 2024-08-04
[Feat]: Process more than one HTTP request at a time


the-sun-will-rise-tomorrow:
### Is your feature request related to a problem? Please describe.

It looks like Zot can only process one HTTP request at a time.

If a HTTP request is made to Zot while it is already processing a HTTP request, the other request will block until the first one completes.

I am seeing this at least with `HEAD /v2/$tag/blobs/$digest` requests. Each request takes about 200ms for me, and trying to perform multiple requests in parallel seems to cause them to form a queue and be processed one after another.

### Describe the solution you'd like

It would be nice if Zot could leverage Go's concurrency mechanisms to allow processing multiple requests simultaneously.

### Describe alternatives you've considered

I suppose multiple Zot instances backed by the same storage pool might work, but this seems a much less elegant solution.

### Additional context

I am experimenting with efficient uploading of images with many layers (built using Nix). In my experiments, the vast majority of the time seems to be spent on Skopeo asking Zot if a layer blob is already in the registry.

rchincha:
@the-sun-will-rise-tomorrow 

Thanks for trying out zot.
Just FYI, you can go one step further in your tests using https://zotregistry.dev/v2.1.0/articles/pprofiling/

rchincha:
@the-sun-will-rise-tomorrow can you also pls share your config (anonymized) so we can try and reproduce this.

the-sun-will-rise-tomorrow:
OK, here it is:

```json
{
    "distSpecVersion": "1.1.0-dev",
    "storage": {
        "dedupe": true,
        "gc": true,
        "gcDelay": "1h",
        "gcInterval": "6h",
        "rootDirectory": "/data/zot/"
    },
    "http": {
        "address": "0.0.0.0",
        "port": "5000",
        "realm": "zot",
        "tls": {
            "cert": "/secrets/tls_crt.pem",
            "key": "/secrets/tls_key.pem"
        },
        "auth": {
            "htpasswd": {
                "path": "/etc/zot/htpasswd"
            }
        },
        "accessControl": {
            "metrics": {
                "users": ["metrics"]
            },
            "repositories": {
                "**": {
                    "defaultPolicy": ["read", "create", "update"],
                    "anonymousPolicy": ["read"]
                }
            },
            "adminPolicy": {
                "users": ["admin"],
                "actions": ["read", "create", "update", "delete"]
            }
        }
    },
    "log": {
        "level": "warn"
    },
    "extensions": {
[...]

PR #3519, opened 2025-11-07
DRAFT: Image Stream for better UX

mariusbertram:
**What type of PR is this?**

feature
#1395 

This PR implements a better UX for on-demand Image Pulls.

This PR is in a working PoC status.

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/examples/config-stream-cache.json
+++ b/examples/config-stream-cache.json
@@ -1,0 +1,40 @@
+{
+  "distSpecVersion": "1.1.0",
+  "storage": {
+    "rootDirectory": "./tmp/zot"
+  },
+  "http": {
+    "address": "0.0.0.0",
+    "port": "8080"
+  },
+  "log": {
+    "level": "debug"
+  },
+  "extensions": {
+    "sync": {
+      "enable": true,
+      "registries": [
+        {
+          "urls": [
+            "https://registry.access.redhat.com"
+          ],
+          "onDemand": true,
+          "maxRetries": 3,
+          "retryDelay": "5m",
+          "tlsVerify": true,
+          "streamCache": {
+            "enable": true,
+            "cacheDir": "./tmp/cache/zot",
+            "maxSize": 10737418240
+          },
+          "content": [
+            {
+              "prefix": "**"
+            }
+          ]
+        }
+      ]
+    }
+  }
+}
+
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -49,6 +49,7 @@ type Controller struct {
 	EventRecorder   events.Recorder
 	CveScanner      ext.CveScanner
 	SyncOnDemand    SyncOnDemand
+	StreamProxyMgr  *StreamProxyManager
 	RelyingParties  map[string]rp.RelyingParty
 	CookieStore     *CookieStore
 	HTPasswd        *HTPasswd
@@ -285,6 +286,10 @@ func (c *Controller) Init() error {
 		return err
 	}
 
+	if err := c.InitStreamProxy(); err != nil {
+		return err
+	}
+
 	if err := c.InitMetaDB(); err != nil {
 		return err
 	}
@@ -318,6 +323,29 @@ func (c *Controller) InitImageStore() error {
 	}
 
 	c.StoreController = storeController
+
+	return nil
+}
+
+func (c *Controller) InitStreamProxy() error {
+	// Initialisiere StreamProxyManager nur wenn Extensions aktiviert sind
+	extensionsConfig := c.Config.CopyExtensionsConfig()
+	if extensionsConfig == nil || extensionsConfig.Sync == nil {
+		c.Log.Debug().Msg("sync extension not configured, skipping stream proxy initialization")
+		return nil
+	}
+
+	streamProxyMgr, err := NewStreamProxyManager(extensionsConfig.Sync, c.StoreController, c.Log)
+	if err != nil {
+		c.Log.Error().Err(err).Msg("failed to initialize stream proxy manager")
+		return err
+	}
+
+	c.StreamProxyMgr = streamProxyMgr
+
+	if streamProxyMgr != nil {
+		c.Log.Info().Msg("stream proxy manager initialized successfully")
+	}
 
 	return nil
 }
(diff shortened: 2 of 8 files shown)


examples/config-stream-cache.json
```diff
@@ -0,0 +1,40 @@
...
+      "registries": [
+        {
+          "urls": [
+            "https://registry.access.redhat.com"
+          ],
+          "onDemand": true,
+          "maxRetries": 3,
+          "retryDelay": "5m",
+          "tlsVerify": true,
+          "streamCache": {
```
rchincha, line 25: We already download blobs to a temporary dir, so no need to create a new one.
We likely just need "onDemandStreaming": enable/disable and which should depend on onDemand:true

rchincha:
@mariusbertram thanks for taking this on. This has been a long time ask.

andaaron (commented):
Hi @mariusbertram, thank you for the initiative.

The sync feature is designed to be an "extension". It is an optional feature, built using the "sync" go build tag, and the package is located under https://github.com/project-zot/zot/tree/main/pkg/extensions/sync.

This PR doesn't seem to integrate with it, except for the configuration file.

 "sync" uses regctl as a library to copy the files locally to a temporary folder. After the entire download is done, it moves the files to storage. After that is done, it starts serving them to the user.
The implementation goal of https://github.com/project-zot/zot/issues/1395 is to refactor "sync" and make the content available to the client as it becomes available.
We don't want to duplicate logic found in regctl, we tried that before, and the maintenance overhead was too large. If regctl library changes are needed, that can also be discussed.

andaaron:
Is this more of an update to serve content from storage concurrently for https://github.com/project-zot/zot/issues/2589?

mariusbertram:
> Is this more of an update to serve content from storage concurrently for #2589?

The idear is to implement something like a proxy cache for the pull through. With that zot downloads the blobs to the cache dir and can after download serve this blob to the requested client. 
It should be the priority to serve first the client and then persist the requested image. 
Because of the caching there is no need to re download the Image with regctl and just move the blobs to the final storage (s3/local directory/ etc.) 
this is a relative simple approach, while duplicating Datastreams and working with them does have some pitfalls.

I will try to implement this approach with regctl. 

Because of the different approach I did Create this PR in this early stage to get your feedback.

mariusbertram:
I did some testing. The major problem is the initial pull of the image by zot is extremely slow.
I did test different registries with the `registry.access.redhat.com/ubi10/go-toolset:latest` image.
It is a multi-arch image with multiple layers and a size of around 1 GB uncompressed and 400 MB compressed.
https://catalog.redhat.com/en/software/containers/rhel8/go-toolset/5b9c810add19c70b45cbd666

A direct pull takes about 6s 
```
time podman pull registry.access.redhat.com/ubi10/go-toolset:latest 
Trying to pull registry.access.redhat.com/ubi10/go-toolset:latest...
Checking if image destination supports signatures
Copying blob sha256:3d9bca65716dd05d5e806d53c5a82afabda97976c0f85458087d22a6df090b14
Copying blob sha256:3f3988782553a7df3a374b44628e79b412aee8c0f7b10c1f66406aa8697c8ebb
Copying blob sha256:357e652cbe64b5a370b576c03b11255daafc49e4e36a7e552c53b1886129a05a
Copying blob sha256:d368f80437f862830c17c154fa28fff5db32db910afe6c110c8314cafc36e3b8
Copying config sha256:8b179f3820c4677689a6760b296880a0dead63aa698272cc445326735c1267b2
Storing signatures
8b179f3820c4677689a6760b296880a0dead63aa698272cc445326735c1267b2
podman pull registry.access.redhat.com/ubi10/go-toolset:latest  0,02s user 0,01s system 0% cpu 6,253 total
```

with ociregistry (a very simple Pull-through oct registry )  42s
```
time podman pull --tls-verify=false 192.168.1.132:8080/registry.access.redhat.com/ubi10/go-toolset:latest
Trying to pull 192.168.1.132:8080/registry.access.redhat.com/ubi10/go-toolset:latest...
Copying blob sha256:3d9bca65716dd05d5e806d53c5a82afabda97976c0f85458087d22a6df090b14
Copying blob sha256:3f3988782553a7df3a374b44628e79b412aee8c0f7b10c1f66406aa8697c8ebb
Copying blob sha256:357e652cbe64b5a370b576c03b11255daafc49e4e36a7e552c53b1886129a05a
Copying blob sha256:d368f80437f862830c17c154fa28fff5db32db910afe6c110c8314cafc36e3b8
Copying config sha256:8b179f3820c4677689a6760b296880a0dead63aa698272cc445326735c1267b2
8b179f3820c4677689a6760b296880a0dead63aa698272cc445326735c1267b2
podman pull --tls-verify=false   0,02s user 0,01s system 0% cpu 42,167 total
```

and a current zot build with local storage 2min 32s
```
time podman pull --tls-verify=false 192.168.1.132:8080/ubi10/go-toolset:latest
Trying to pull 192.168.1.132:8080/ubi10/go-toolset:latest...
Copying blob sha256:3d9bca65716dd05d5e806d53c5a82afabda97976c0f85458087d22a6df090b14
Copying blob sha256:3f3988782553a7df3a374b44628e79b412aee8c0f7b10c1f66406aa8697c8ebb
Copying blob sha256:d368f80437f862830c17c154fa28fff5db32db910afe6c110c8314cafc36e3b8
Copying blob sha256:357e652cbe64b5a370b576c03b11255daafc49e4e36a7e552c53b1886129a05a
Copying config sha256:8b179f3820c4677689a6760b296880a0dead63aa698272cc445326735c1267b2
8b179f3820c4677689a6760b296880a0dead63aa698272cc445326735c1267b2
podman pull --tls-verify=false 192.168.1.132:8080/ubi10/go-toolset:latest  0.02s user 0.01s system 0% cpu 2:32.29 total
```

I have not found the blocking part. But I think if this problem is solved and for my environment (Internet 100mbit/s down) with an image of this size in sub 1min should be a first step to the streaming. 

BTW Redhat’s Quay has a similar approach as zot https://docs.redhat.com/en/documentation/red_hat_quay/3.7/html/use_red_hat_quay/quay-as-cache-proxy

sizowie:
I would like to see this feature in zot - as we are experiencing same performance issues with zot.

PR #3521, opened 2025-11-07
fix: re-introduce pagination

ljakimczuk:
**What type of PR is this?**

A bug / regression.

https://github.com/project-zot/zot/issues/2715.

Without it, Zot only gets first 100 repositories from the registry. This PR makes it continue building the complete list.

N/A

So far I tested this in my environment, no unit tests added yet.

**Automation added to e2e**:

N/A

It should not.

N/A

### DIFF
--- a/pkg/extensions/sync/remote.go
+++ b/pkg/extensions/sync/remote.go
@@ -51,8 +51,10 @@ func (registry *RemoteRegistry) GetHostName() string {
 func (registry *RemoteRegistry) GetRepositories(ctx context.Context) ([]string, error) {
 	var err error
 
+	var repoList []string
+
 	for _, host := range registry.hosts {
-		repoList, err := registry.getRepoList(ctx, host.Hostname)
+		repoList, err = registry.getRepoList(ctx, host.Hostname)
 		if err != nil {
 			registry.log.Error().Err(err).Str("remote", host.Name).Msg("failed to list repositories in remote registry")
 


andaaron (approved):

andaaron:
The only strange part is the coverage decrease in 
image
Maybe before this change there was some kind of error raised in case the limit was reached, and right now it isn't anymore?

ljakimczuk:
@andaaron now when I look at it, maybe it is my bad, I changed assignment to short declaration, I guess that makes `err` a new local variable inside the loop then the error return from `getRepoList` is lost, changed that back to assignment.

ljakimczuk:
@andaaron all looks good now.

ljakimczuk:
@rchincha @andaaron I saw after merging to the main branch, the codecov jobs failed there again, is it something normal?

andaaron:
The `sync` tests are ok from the point of view of functional testing, but are more `async` than `sync`, so not as good as we would want them to be for consistent test coverage on all code paths. They may or may not hit some code lines depending on timing (mostly on reaching some errors). 

I have added additional tests in the past couple of months to consistently cover some of these error branches, but it looks like there is still work to do.
Those 2 lines were covered when we merged your PR, but, as far as I remember, they were not before your last fix. One of them is in your new function, the other is not.

ljakimczuk:
Thanks for explanetion!

ljakimczuk:
@rchincha sorry for bothering you, but can you give me raw estimates on when we may expect the next Zot release?

PR #3523, opened 2025-11-09
fix: separate cipher suites and curve preferences into FIPS and non FIPS, and use them accordingly

andaaron:
See: https://github.com/project-zot/zot/actions/runs/19209741002/job/54910194536

`failed to ping registry localhost:11448: Get "https://localhost:11448/v2/": crypto/ecdh: use of X25519 is not allowed in FIPS 140-only mode`

### DIFF
--- a/.github/workflows/tls.yaml
+++ b/.github/workflows/tls.yaml
@@ -12,7 +12,15 @@ permissions: read-all
 jobs:
   tls-check:
     runs-on: ubuntu-latest
-    name: TLS check
+    strategy:
+      matrix:
+        mode: [non-fips, fips]
+        include:
+          - mode: non-fips
+            godebug: ""
+          - mode: fips
+            godebug: "fips140=only"
+    name: TLS check (${{ matrix.mode }})
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v6
@@ -25,15 +33,31 @@ jobs:
           mkdir -p test/data
           cd test/data
           ../scripts/gen_certs.sh
-      - name: Check for TLS settings
+      - name: Build binary
         run: |
           cd $GITHUB_WORKSPACE
           make binary
+      - name: Start zot server (${{ matrix.mode }})
+        run: |
+          cd $GITHUB_WORKSPACE
+          if [[ -n "${{ matrix.godebug }}" ]]; then
+            export GODEBUG="${{ matrix.godebug }}"
+          fi
           bin/zot-linux-amd64 serve examples/config-tls.json & echo $! > zot.PID
+          if [[ -n "${{ matrix.godebug }}" ]]; then
+            unset GODEBUG
+          fi
           sleep 5
           # Check if zot server is running
           cat /proc/$(cat zot.PID)/status | grep State || exit 1
           curl -k --connect-timeout 3 --max-time 5 --retry 60 --retry-delay 1 --retry-max-time 180 --retry-connrefused https://localhost:8080/v2/
-
-          # zot server is running: proceed to testing
+      - name: Run TLS tests (${{ matrix.mode }})
+        run: |
+          cd $GITHUB_WORKSPACE
           ./test/scripts/tls_scan.sh
+          ./test/scripts/tls_cipher_check.sh ${{ matrix.mode }} localhost:8080
+      - name: Cleanup
+        if: always()
+        run: |
+          cd $GITHUB_WORKSPACE
+          [[ -f zot.PID ]] && kill $(cat zot.PID) 2>/dev/null || true
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -204,21 +204,39 @@ func (c *Controller) Run() error {
 
 	tlsConfig := c.Config.CopyTLSConfig()
 	if tlsConfig != nil && tlsConfig.Key != "" && tlsConfig.Cert != "" {
-		server.TLSConfig = &tls.Config{
-			CipherSuites: []uint16{
-				tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
-				tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+		// These are the same as the cipher suites in defaultCipherSuitesFIPS for TLS 1.2
+		// see https://cs.opensource.google/go/go/+/refs/tags/go1.24.9:src/crypto/tls/defaults.go;l=123
+		// Note: Order doesn't matter - Go 1.17+ automatically orders cipher suites based on
+		// hardware capabilities and security properties. See https://go.dev/blog/tls-cipher-suites
+		cipherSuites := []uint16{
+			tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
+			tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+			tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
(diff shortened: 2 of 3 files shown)


pkg/api/controller.go
```diff
@@ -204,19 +204,33 @@ func (c *Controller) Run() error {
 
 	tlsConfig := c.Config.CopyTLSConfig()
 	if tlsConfig != nil && tlsConfig.Key != "" && tlsConfig.Cert != "" {
-		server.TLSConfig = &tls.Config{
-			CipherSuites: []uint16{
-				tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
-				tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+		// These are the same as the default cipher suites in tls.defaultCipherSuitesFIPS for TLS 1.2
```
vrajashkr: It would be nice to have a link added to point to the source for this list so that it is easier to locate.

I tried looking up `tls.defaultCipherSuitesFIPS`, but didn't get a concrete result. I was led to a Dzone article, following which I was led to https://cs.opensource.google/fuchsia/fuchsia/+/main:third_party/go/src/crypto/tls/boring.go;l=52?q=tls.defaultCipherSuitesFIPS which seems to be from an unrelated project.

pkg/api/controller.go
```diff
@@ -204,19 +204,33 @@ func (c *Controller) Run() error {
...
-				tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
-			},
-			CurvePreferences: []tls.CurveID{
-				tls.CurveP256,
-				tls.X25519,
-			},
+			}, cipherSuites...)
+		}
+
+		// These are a subset of the default curve preferences in tls.defaultCurvePreferencesFIPS for TLS 1.2
```
vrajashkr: Similar comment for this - adding the source reference in a comment would be nice.

test/scripts/tls_cipher_check.sh
```diff
@@ -0,0 +1,164 @@
...
+echo ""
+echo "=== Verification Results ==="
+
+# Summary for non-FIPS mode
+if [[ "$MODE" == "non-fips" ]]; then
+    echo "Non-FIPS Mode: Summary..."
+    echo "  TLS 1.2 FIPS: $TLS12_FIPS_PASSED/${#FIPS_TLS12_CIPHERS[@]} passed"
+    echo "  TLS 1.3 FIPS: $TLS13_FIPS_PASSED/${#FIPS_TLS13_CIPHERS[@]} passed"
+    echo ""
+    echo "Note: Non-FIPS mode may accept both FIPS and non-FIPS cipher suites"
```
vrajashkr: Shouldn't this be "should accept" instead of "may accept"?

test/scripts/tls_cipher_check.sh
```diff
@@ -0,0 +1,164 @@
...
+# Non-FIPS cipher suites (TLS 1.2)
+NON_FIPS_TLS12_CIPHERS=(
+    "ECDHE-ECDSA-CHACHA20-POLY1305"
+    "ECDHE-RSA-CHACHA20-POLY1305"
+)
+
+# Non-FIPS cipher suites (TLS 1.3)
+NON_FIPS_TLS13_CIPHERS=(
+    "TLS_CHACHA20_POLY1305_SHA256"
+)
```
vrajashkr, line 36: Similar to the comments for the source code change, it would be great to have a reference to the source of truth for these lists of ciphers.

pkg/api/controller.go
andaaron: It is a reference to the crypto/tls package in the go toolchain itself. https://cs.opensource.google/go/go/+/refs/tags/go1.24.9:src/crypto/tls/defaults.go;l=123

I will update the code to reference it

vrajashkr: Thanks! I was able to find it with your link.

Small note - it looks like in Go 1.25.3, this has been moved to a different location and renamed (hence my search failing badly :') )

https://cs.opensource.google/go/go/+/refs/tags/go1.25.3:src/crypto/tls/defaults_fips140.go;l=51

andaaron: Possibly, we're still on 1.24, and I had looked into the locally cached copy of the files used by the build.

andaaron: I updated the files.

vrajashkr (approved):
LGTM

test/scripts/tls_cipher_check.sh
andaaron: Anything (in this case a cipher suite) not explicitly in the FIPS lists, is non-FIPS.

rchincha (approved):

test/scripts/tls_cipher_check.sh
```diff
@@ -0,0 +1,166 @@
...
+        TLS13_FIPS_PASSED=$((TLS13_FIPS_PASSED + 1))
+    else
+        echo "✗ TLS 1.3 FIPS cipher '$cipher': FAILED"
+        if [[ "$MODE" == "fips" ]]; then
+            echo ""
+            echo "✗ ERROR: Required TLS 1.3 FIPS cipher '$cipher' failed"
+            echo "  All TLS 1.3 FIPS ciphers must work in FIPS mode"
+            exit 1
+        fi
+    fi
```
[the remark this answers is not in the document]
andaaron: Added a failure instead.

pkg/api/controller.go
```diff
@@ -204,19 +204,35 @@ func (c *Controller) Run() error {
...
 				tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
 				tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,
-				tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
-				tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
-			},
-			CurvePreferences: []tls.CurveID{
-				tls.CurveP256,
-				tls.X25519,
-			},
+			}, cipherSuites...)
```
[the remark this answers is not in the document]
andaaron: Done

test/scripts/tls_cipher_check.sh
```diff
@@ -0,0 +1,166 @@
...
+    else
+        echo "✗ TLS 1.2 FIPS cipher '$cipher': FAILED"
+    fi
+done
+
+# In FIPS mode, require at least one TLS 1.2 FIPS cipher to work.
+# We can't require ALL TLS 1.2 FIPS ciphers because certificate type determines which ones work:
+# - RSA certificates: only RSA-based ciphers (ECDHE-RSA-*) work
+# - ECDSA certificates: only ECDSA-based ciphers (ECDHE-ECDSA-*) work
+# If none work, it indicates a configuration issue or a certificate mismatch.
```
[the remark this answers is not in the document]
andaaron: Done

.github/workflows/tls.yaml
```diff
@@ -25,15 +33,31 @@ jobs:
...
+      - name: Run TLS tests (${{ matrix.mode }})
+        run: |
+          cd $GITHUB_WORKSPACE
           ./test/scripts/tls_scan.sh
+          ./test/scripts/tls_cipher_check.sh ${{ matrix.mode }} localhost:8080
+      - name: Cleanup
+        if: always()
+        run: |
+          cd $GITHUB_WORKSPACE
+          kill $(cat zot.PID) 2>/dev/null || true
```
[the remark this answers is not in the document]
andaaron: Done

PR #3524, opened 2025-11-09
fix: deduplicate entries in referrers responses

andaaron:
See: https://github.com/project-zot/zot/issues/2506

Note we are not losing anything functionality-wise by making this change.
Initially, we considered the tags are in the annotations present in the referrers,
but the only annotations we set on referrers are the ones inside the manifests themselves,
not the ones in the manifest descriptors, so the tags were not present anyway.


rchincha:
This would be a nice fix.

Issue #2505, opened 2024-06-29
[Feat]: Pull images with background Contexts


Namnamseo:
### Is your feature request related to a problem? Please describe.

cf. #2416 

When an on-demand sync is triggered, the `ctx` to the call to `SyncImage()` comes from the request `context.Context`.
Thus, when the user gives up, the image sync is also canceled.

Code flow:
https://github.com/project-zot/zot/blob/0d0eae502e438b419a00e8ce4816be76daf8b87a/pkg/api/routes.go#L518
https://github.com/project-zot/zot/blob/0d0eae502e438b419a00e8ce4816be76daf8b87a/pkg/api/routes.go#L1941
https://github.com/project-zot/zot/blob/0d0eae502e438b419a00e8ce4816be76daf8b87a/pkg/extensions/sync/on_demand.go#L76
https://github.com/project-zot/zot/blob/0d0eae502e438b419a00e8ce4816be76daf8b87a/pkg/extensions/sync/service.go#L311

This is unfortunate, I think, in many use cases where Zot is used as a proxy cache.
Even if we couldn't make it on time this time, it isn't that we don't want to have this image cached.

If the client has a timeout limit and a retry loop (like common setups with Kubernetes and containerd do - at least in mine), the current implementation will not be able to make any progress.
This is because, as soon as the client hits the timeout, the sync job fails (because it makes an HTTP request with a canceled Context, I guess), and when the client retries, the sync has to start over again.

It would be better if the pull keeps running in the background, so that the next time there's a request to this image, they could wait for the remaining progress and have any chance of seeing it finish.
Note that the current logic already supports the joining (to wait) to an ongoing pull in another goroutine:
https://github.com/project-zot/zot/blob/0d0eae502e438b419a00e8ce4816be76daf8b87a/pkg/extensions/sync/on_demand.go#L54-L61

### Describe the solution you'd like

Somewhere during the code flow, use a background Context instead of the request Context.
Note that we do not want to completely ignore the request Context; we need to halt the control flow and return when the request is canceled.

A good place could be here:

https://github.com/project-zot/zot/blob/0d0eae502e438b419a00e8ce4816be76daf8b87a/pkg/extensions/sync/on_demand.go#L76-L81

We could give `syncImage` a `context.Background()`, or a larger timeout (15 minutes?).
Then, below, a `select {}` with both `syncResult` and `ctx.Done()` would suffice.

### Describe alternatives you've considered

[...]

PR #3537, opened 2025-11-12
Sync images with a background context

lfrancke:
**What type of PR is this?**

feature

https://github.com/project-zot/zot/issues/2505

Images are now synced using a background context.
It is needed because the sync was tied to the incoming request that triggered the sync.
When this request was cancelled the sync is also cancelled.
In Kubernetes this caused endless loops of downloading without ever resulting in success.

I added logging statements to the code to verify the behavior.
I also ran a big test on my end which used to fail 100% of the time before this patch and now succeeded immediately.
I started a new K3d cluster, started dozens of pods with multi GB images and let them download. This used to take _hours_ before until some images more or less "randomly" downloaded in time.

Now it took ~10min to settle down. I also looked at the logs and was able to confirm the behavior.

**Automation added to e2e**:

None

I don't believe so.

```release-note
Registries can now be configured with a `syncTimeout` parameter (default: 3h) after which syncs from registries will be aborted. This used to be tied to the lifetime of the incoming request which triggered the sync. This behavior caused unwanted sync aborts. This is now fixed and configurable.
```

### DIFF
--- a/pkg/cli/server/root.go
+++ b/pkg/cli/server/root.go
@@ -663,9 +663,15 @@ func applyDefaultValues(config *config.Config, viperInstance *viper.Viper, logge
 				config.Extensions.Sync.Enable = &defaultVal
 			}
 
-			for id, regCfg := range config.Extensions.Sync.Registries {
+			defaultSyncTimeout := 3 * time.Hour
+
+			for idx, regCfg := range config.Extensions.Sync.Registries {
 				if regCfg.TLSVerify == nil {
-					config.Extensions.Sync.Registries[id].TLSVerify = &defaultVal
+					config.Extensions.Sync.Registries[idx].TLSVerify = &defaultVal
+				}
+
+				if config.Extensions.Sync.Registries[idx].SyncTimeout == 0 {
+					config.Extensions.Sync.Registries[idx].SyncTimeout = defaultSyncTimeout
 				}
 			}
 		}
--- a/pkg/extensions/config/sync/config.go
+++ b/pkg/extensions/config/sync/config.go
@@ -33,7 +33,8 @@ type RegistryConfig struct {
 	RetryDelay       *time.Duration
 	OnlySigned       *bool
 	CredentialHelper string
-	PreserveDigest   bool // sync without converting
+	PreserveDigest   bool          // sync without converting
+	SyncTimeout      time.Duration // timeout for on-demand sync operations; if zero or unset, defaults to 3 hours
 }
 
 type Content struct {
--- a/pkg/extensions/sync/on_demand.go
+++ b/pkg/extensions/sync/on_demand.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"sync"
+	"time"
 
 	zerr "zotregistry.dev/zot/v2/errors"
 	"zotregistry.dev/zot/v2/pkg/common"
@@ -64,7 +65,7 @@ func (onDemand *BaseOnDemand) SyncImage(ctx context.Context, repo, reference str
 
 	defer onDemand.requestStore.Delete(req)
 
-	go onDemand.syncImage(ctx, repo, reference, syncResult)
+	go onDemand.syncImage(repo, reference, syncResult)
 
 	err := <-syncResult
 
@@ -95,21 +96,37 @@ func (onDemand *BaseOnDemand) SyncReferrers(ctx context.Context, repo string,
 
 	defer onDemand.requestStore.Delete(req)
 
-	go onDemand.syncReferrers(ctx, repo, subjectDigestStr, referenceTypes, syncResult)
+	go onDemand.syncReferrers(repo, subjectDigestStr, referenceTypes, syncResult)
 
 	err := <-syncResult
 
 	return err
 }
 
-func (onDemand *BaseOnDemand) syncReferrers(ctx context.Context, repo, subjectDigestStr string,
+func (onDemand *BaseOnDemand) syncReferrers(repo, subjectDigestStr string,
 	referenceTypes []string, syncResult chan error,
 ) {
 	defer close(syncResult)
 
 	var err error
+
 	for serviceID, service := range onDemand.services {
-		err = service.SyncReferrers(ctx, repo, subjectDigestStr, referenceTypes)
+		timeout := service.GetSyncTimeout()
+
+		onDemand.log.Debug().
+			Str("repo", repo).
+			Str("reference", subjectDigestStr).
+			Int("serviceID", serviceID).
+			Dur("timeout", timeout).
+			Msg("starting on-demand referrer sync")
+
+		// Create a detached context with timeout to ensure sync completes even if HTTP client disconnects.
+		// This prevents Kubernetes timeout/retries from aborting in-progress referrer downloads.
+		syncCtx, cancel := context.WithTimeout(context.Background(), timeout)
+		err = service.SyncReferrers(syncCtx, repo, subjectDigestStr, referenceTypes)
+
+		cancel()
+
 		if err != nil {
 			if errors.Is(err, zerr.ErrManifestNotFound) ||
 				errors.Is(err, zerr.ErrSyncImageFilteredOut) ||
@@ -132,8 +149,10 @@ func (onDemand *BaseOnDemand) syncReferrers(ctx context.Context, repo, subjectDi
 			}
 
 			if service.CanRetryOnError() {
+				retryErr := err
+
 				// retry in background
-				go func(service Service) {
+				go func(service Service, serviceTimeout time.Duration) {
 					// remove image after syncing
 					defer func() {
 						onDemand.requestStore.Delete(req)
@@ -141,15 +160,19 @@ func (onDemand *BaseOnDemand) syncReferrers(ctx context.Context, repo, subjectDi
 							Msg("sync routine for image exited")
 					}()
 
-					onDemand.log.Info().Str("repo", repo).Str("reference", subjectDigestStr).Str("err", err.Error()).
+					onDemand.log.Info().Str("repo", repo).Str("reference", subjectDigestStr).Str("err", retryErr.Error()).
 						Msg("sync routine: starting routine to copy image, because of error")
 
-					err := service.SyncReferrers(context.Background(), repo, subjectDigestStr, referenceTypes)
+					// Use detached context with timeout for background retry
+					retryCtx, cancel := context.WithTimeout(context.Background(), serviceTimeout)
+					defer cancel()
+
+					err := service.SyncReferrers(retryCtx, repo, subjectDigestStr, referenceTypes)
 					if err != nil {
 						onDemand.log.Error().Str("errorType", common.TypeOf(err)).Str("repo", repo).Str("reference", subjectDigestStr).
 							Err(err).Msg("sync routine: starting routine to retry copy image due to error")
 					}
-				}(service)
+				}(service, timeout)
 			}
 		} else {
 			break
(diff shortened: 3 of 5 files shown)


pkg/extensions/sync/on_demand.go
```diff
@@ -38,8 +47,21 @@ func NewOnDemand(log log.Logger) *BaseOnDemand {
...
-func (onDemand *BaseOnDemand) Add(service Service) {
+func (onDemand *BaseOnDemand) Add(service Service, syncTimeout time.Duration) {
+	serviceID := len(onDemand.services)
+
+	// Use default if timeout not configured
+	if syncTimeout == 0 {
+		syncTimeout = syncOperationTimeout
+	}
+
+	onDemand.log.Info().
```
lfrancke: These were helpful to me during debugging. I am happy to remove them again if you prefer.

pkg/extensions/sync/on_demand.go
```diff
@@ -95,21 +117,38 @@ func (onDemand *BaseOnDemand) SyncReferrers(ctx context.Context, repo string,
...
 	var err error
 	for serviceID, service := range onDemand.services {
-		err = service.SyncReferrers(ctx, repo, subjectDigestStr, referenceTypes)
+		// Get timeout for this service, fallback to default if not set
+		timeout := syncOperationTimeout
+		if serviceID < len(onDemand.syncTimeouts) && onDemand.syncTimeouts[serviceID] > 0 {
+			timeout = onDemand.syncTimeouts[serviceID]
+		}
+
+		onDemand.log.Info().
```
lfrancke: As commented before: This was useful during debugging but I'm happy to remove it.

pkg/extensions/sync/on_demand.go
```diff
@@ -191,7 +253,7 @@ func (onDemand *BaseOnDemand) syncImage(ctx context.Context, repo, reference str
 				retryErr := err
 
 				// retry in background
-				go func(service Service) {
+				go func(service Service, serviceTimeout time.Duration) {
```
lfrancke, line 232: The retries were already running in the background but had no bounds at all so I thought it makes sense to also give them a timeout.

rchincha: which makes sense!

pkg/extensions/sync/on_demand.go
rchincha: If this is only for debugging, move this to Debug() instead.

pkg/extensions/sync/on_demand.go
rchincha: Same here, Debug() instead.

pkg/extensions/config/sync/config.go
```diff
@@ -33,7 +33,8 @@ type RegistryConfig struct {
 	RetryDelay       *time.Duration
 	OnlySigned       *bool
 	CredentialHelper string
-	PreserveDigest   bool // sync without converting
+	PreserveDigest   bool          // sync without converting
+	SyncTimeout      time.Duration // timeout for on-demand sync operations (default: 3 hours)
```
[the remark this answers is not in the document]
lfrancke: Dear Copilot, "lacks documentation" is quite a strong thing to say when you then go on to change _my_ documentation.

It does however give me an idea: Should it be possible to disable this and have an unlimited bound by setting this to 0?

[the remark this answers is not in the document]
lfrancke: Addressed in [`c4cdef7` (#3537)](https://github.com/project-zot/zot/pull/3537/commits/c4cdef7be08f6bc52239d92ec69a0300883e248d) 

But my previous comment is still open.

pkg/extensions/sync/on_demand.go
```diff
@@ -8,11 +8,19 @@ import (
...
+
 	zerr "zotregistry.dev/zot/v2/errors"
 	"zotregistry.dev/zot/v2/pkg/common"
 	"zotregistry.dev/zot/v2/pkg/log"
 )
 
+const (
+	// syncOperationTimeout is the maximum time allowed for a sync operation to complete.
+	// This is independent of HTTP client timeouts to ensure sync completes even if clients disconnect.
+	syncOperationTimeout = 180 * time.Minute
```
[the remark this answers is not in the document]
lfrancke: Addressed in [`c4cdef7` (#3537)](https://github.com/project-zot/zot/pull/3537/commits/c4cdef7be08f6bc52239d92ec69a0300883e248d)

pkg/extensions/sync/on_demand.go
```diff
@@ -159,13 +204,30 @@ func (onDemand *BaseOnDemand) syncReferrers(ctx context.Context, repo, subjectDi
...
+			Str("reference", reference).
+			Int("serviceID", serviceID).
+			Dur("timeout", timeout).
+			Msg("starting on-demand image sync with timeout")
+
+		// Create a detached context with timeout to ensure sync completes even if HTTP client disconnects.
+		// This prevents Kubernetes timeout/retries from aborting in-progress image downloads.
+		syncCtx, cancel := context.WithTimeout(context.Background(), timeout)
+		defer cancel()
+		err = service.SyncImage(syncCtx, repo, reference)
```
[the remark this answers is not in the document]
lfrancke: Addressed in [`c4cdef7` (#3537)](https://github.com/project-zot/zot/pull/3537/commits/c4cdef7be08f6bc52239d92ec69a0300883e248d)

pkg/extensions/sync/on_demand.go
```diff
@@ -95,21 +117,38 @@ func (onDemand *BaseOnDemand) SyncReferrers(ctx context.Context, repo string,
...
+			Str("reference", subjectDigestStr).
+			Int("serviceID", serviceID).
+			Dur("timeout", timeout).
+			Msg("starting on-demand referrer sync with timeout")
+
+		// Create a detached context with timeout to ensure sync completes even if HTTP client disconnects.
+		// This prevents Kubernetes timeout/retries from aborting in-progress referrer downloads.
+		syncCtx, cancel := context.WithTimeout(context.Background(), timeout)
+		defer cancel()
+		err = service.SyncReferrers(syncCtx, repo, subjectDigestStr, referenceTypes)
```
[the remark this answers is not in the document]
lfrancke: Good find. In practice probably not a problem but still worth fixing.

Addressed in [`c4cdef7` (#3537)](https://github.com/project-zot/zot/pull/3537/commits/c4cdef7be08f6bc52239d92ec69a0300883e248d)

pkg/extensions/sync/on_demand.go
lfrancke: Makes sense.

Addressed in [`c4cdef7` (#3537)](https://github.com/project-zot/zot/pull/3537/commits/c4cdef7be08f6bc52239d92ec69a0300883e248d)

pkg/extensions/sync/on_demand.go
lfrancke: Addressed in [`c4cdef7` (#3537)](https://github.com/project-zot/zot/pull/3537/commits/c4cdef7be08f6bc52239d92ec69a0300883e248d)

lfrancke:
I force-pushed to get the missing DCO.

lfrancke:
* I have no idea what to do with the performance regression benchmark tbh.
* Commit message style I can fix: What do you prefer? One big commit and force pushes or multiple commits on top of each other?
* The lint issues i'll look at again, I thought I fixed them all. The current golangci-lint that's used by the Makefile doesn't work for me on go 1.25 anymore. I have a half-finished PR ready to update that to a newer version but wasn't sure about the rules so aborted that.

pkg/extensions/extension_sync.go
```diff
@@ -72,7 +72,7 @@ func EnableSyncExtension(config *config.Config, metaDB mTypes.MetaDB,
 
 			if isOnDemand {
 				// onDemand services used in routes.go
-				onDemand.Add(service)
+				onDemand.Add(service, registryConfig.SyncTimeout)
```
andaaron: But doesn't the service have an attribute with the registryConfig? Why do we need to pass this per/registry configuration as an argument in these functions?

pkg/extensions/sync/on_demand.go
```diff
@@ -28,7 +35,8 @@ keep track of all parallel requests, if two requests of same image/signature/sbo
 process just the first one, also keep track of all background retrying routines.
 */
 type BaseOnDemand struct {
-	services []Service
+	services     []Service
+	syncTimeouts []time.Duration // timeout per service
```
andaaron: Why do we need this separate? Each service is aware of its own configuration.

pkg/extensions/config/sync/config.go
```diff
@@ -33,7 +33,8 @@ type RegistryConfig struct {
 	RetryDelay       *time.Duration
 	OnlySigned       *bool
 	CredentialHelper string
-	PreserveDigest   bool // sync without converting
+	PreserveDigest   bool          // sync without converting
+	SyncTimeout      time.Duration // timeout for on-demand sync operations; if zero or unset, defaults to 3 hours
```
andaaron, line 37: If we want to set a default, I suggest configuring it in pkg/cli/server/root.go, function applyDefaultValues, together with the other defaults.

andaaron (commented):
Personally, I prefer a single commit.
Let's ignore the benchmark for now, it can be flaky.
With regards to the linter, results are available here: https://github.com/project-zot/zot/actions/runs/19327762722/job/55307283725?pr=3537

pkg/extensions/config/sync/config.go
lfrancke: Thanks for all these hints. I kinda fumbled my way through the code. I don't touch Go very often and am sometimes confused by the syntax and that is on top of not knowing the Zot code very well with the exception of two files :)

I'll see if I can get your feedback incorporated.

pkg/extensions/config/sync/config.go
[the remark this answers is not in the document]
rchincha: I am not sure if it ever makes sense to have an unlimited bound.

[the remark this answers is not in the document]
lfrancke: Agreed! Thanks. People can set it to an arbitrary large number if needed.

pkg/extensions/config/sync/config.go
lfrancke: I _think_ I did that in root.go now. Not sure if I did it properly though.

lfrancke:
Please take a look again. I believe I found the service you're referring to even though it does not have a `registryConfig` but honestly I'm not 100% sure if I did the right thing here.

pkg/extensions/sync/on_demand.go
```diff
@@ -39,6 +46,19 @@ func NewOnDemand(log log.Logger) *BaseOnDemand {
 }
 
 func (onDemand *BaseOnDemand) Add(service Service) {
+	serviceID := len(onDemand.services)
```
andaaron: Why are we changing this, since we have more or less the same information in the log messages from syncReferrers and syncImage?

pkg/extensions/sync/service.go
```diff
@@ -209,6 +209,10 @@ func (service *BaseService) CanRetryOnError() bool {
 	return false
 }
 
+func (service *BaseService) GetSyncTimeout() time.Duration {
+	return service.config.SyncTimeout
```
andaaron, line 217: Why not add the `if timeout == 0` logic to return the default here, in case the value is unset?

andaaron:
Ideally, there should be a regctl library timeout to set. But until we figure that out, this should work.

Maybe it would work to initialize the [http client](https://pkg.go.dev/net/http#Client) with a timeout and use WithHTTPClient() in pkg/extensions/sync/service.go before `client := regclient.New()`. This would not take care of the request continuing in the background, but theoretically setting the `MaxRetries` sync setting `>=2` in the zot config should make sure the 2nd attempt running in the backgroud succeeds.

rchincha:
Error: pkg/cli/server/root.go:668:8: variable name 'id' is too short for the scope of its usage (varnamelen)
  			for id, regCfg := range config.Extensions.Sync.Registries {
  			


 Make this "idx"

pkg/extensions/sync/service.go
lfrancke: Yup, that makes much more sense.

pkg/extensions/sync/on_demand.go
lfrancke: I used this during debugging but you're correct. I'll remove it.

lfrancke:
Okay, thanks again for looking. I updated the patch.

- I renamed `id` to `idx` -> This patch did not create that variable, I only reused what was already there
- I moved the default timeout to GetSyncTimeout

Full disclosure: This time I asked an AI to check whether what I did looks good.


Unrelated question if you like: The golangci-lint version used by zot doesn't work on modern systems anymore, at least I get errors related to go 1.25. I can put up a PR to upgrade it to a newer versions, should I include automatted fixes to all the things it suggests as well?

andaaron:
@lfrancke, if you can put a PR for a golangci-lint version compatible with both go 1.24 and go.1.25, that would be welcomed. I am not sure about the automated fixes. What changes would it want to make?

andaaron (approved):

lfrancke:
Thanks for the review: I created a new issue for the golangci-lint issue: https://github.com/project-zot/zot/issues/3568

andaaron:
Please rebase this PR

lfrancke:
Done (automatically)

lfrancke:
Ah wel.... forgot that this doesn't work here. I'll push an updated version.

lfrancke:
Done. I hope everything passes again.

pkg/cli/server/root.go
```diff
@@ -663,9 +663,15 @@ func applyDefaultValues(config *config.Config, viperInstance *viper.Viper, logge
 				config.Extensions.Sync.Enable = &defaultVal
 			}
 
-			for id, regCfg := range config.Extensions.Sync.Registries {
+			defaultSyncTimeout := 3 * time.Hour
```
rchincha, line 666: Eventually, we need to find a better home for this. Don't like the inlined constant here.
https://github.com/project-zot/zot/blob/main/pkg/extensions/sync/constants/constants.go
^ perhaps here.

Pls don't make the change yet.

PR #3561, opened 2025-11-18
chore: update cosign from v2 to v3

andaaron:
Also, do not import cosign into zli, as it doubles the binary size
See: https://github.com/project-zot/zot/actions/runs/19506399474/job/55833719683?pr=3561


rchincha:
@andaaron thanks for taking care of this.

PR #3567, opened 2025-11-20
fix(sync): properly handle CommitAll errors in syncImage and skip failed temp sync dirs

andaaron:
- Return CommitAll errors instead of ignoring them
- Skip ErrRepoNotFound from temp sync dirs to allow other tags to sync
- Each tag uses separate temp directory, so failures are isolated

### DIFF
--- a/pkg/extensions/sync/service.go
+++ b/pkg/extensions/sync/service.go
@@ -430,8 +430,10 @@ func (service *BaseService) SyncRepo(ctx context.Context, repo string) error {
 			if errors.Is(err, zerr.ErrSyncImageNotSigned) ||
 				errors.Is(err, zerr.ErrUnauthorizedAccess) ||
 				errors.Is(err, zerr.ErrMediaTypeNotSupported) ||
-				errors.Is(err, zerr.ErrManifestNotFound) {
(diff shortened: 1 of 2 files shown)


pkg/extensions/sync/service.go
```diff
@@ -430,8 +430,10 @@ func (service *BaseService) SyncRepo(ctx context.Context, repo string) error {
 			if errors.Is(err, zerr.ErrSyncImageNotSigned) ||
 				errors.Is(err, zerr.ErrUnauthorizedAccess) ||
 				errors.Is(err, zerr.ErrMediaTypeNotSupported) ||
-				errors.Is(err, zerr.ErrManifestNotFound) {
-				// skip unsigned images or unsupported image mediatype
+				errors.Is(err, zerr.ErrManifestNotFound) ||
+				errors.Is(err, zerr.ErrRepoNotFound) {
+				// skip unsigned images, unsupported image mediatype, or temp sync dir issues
+				// ErrRepoNotFound from temp sync dir is skippable since each tag uses a different temp directory
```
andaaron, line 436: Or we could keep these conditions as they are on main, and assume having an issue with the temporary storage is a symptom of a larger problem which is not recoverable.

PR #3572, opened 2025-11-20
chore: Enable Go jsonv2 experiment and update the trivy dependency (v0.67.2)

andaaron:
It requires the encoding/json/jsontext
package which is only available when the goexperiment.jsonv2 build
tag is enabled. This was causing build constraint errors during
tests and builds.

Changes:
- Add GOEXPERIMENT=jsonv2 to Makefile export and all go build/test
  commands that use env (since env creates a fresh environment)
- Add GOEXPERIMENT=jsonv2 to GitHub workflows that use direct go
  commands (workflows using make inherit it from Makefile)

Fix other dependabot alerts.


.github/workflows/sync-3rdparty-images.yaml
```diff
@@ -18,6 +18,7 @@ jobs:
     strategy:
       matrix:
         golang_version:
+          - "1.25"
           - "1.24"
           - "1.25"
```
vrajashkr: 1.25 seems duplicated in this list

andaaron: Thanks, fixed it.

PR #3574, opened 2025-11-21
fix: Add HTTP client timeouts to prevent indefinite hangs in sync operations

andaaron:
Configure HTTP client with ResponseHeaderTimeout and overall request timeout to prevent sync operations from hanging indefinitely when upstream registries don't respond.

- Set ResponseHeaderTimeout to 30 seconds (default) to prevent hanging when servers connect but don't send response headers. This timeout is set programmatically in root.go and is not user-configurable (hidden from config files via yaml:"-" tag).
- Use SyncTimeout for overall HTTP client timeout (defaults to 3 hours)
- Move timeout constants to pkg/extensions/sync/constants/constants.go for centralized configuration
- Critical for periodic sync operations (catalog listing, SyncRepo, getTags) which don't use on-demand timeout contexts and could otherwise hang indefinitely if upstream registry doesn't respond

The timeout is applied to all regclient HTTP requests, ensuring that both on-demand and periodic sync operations have protection against unresponsive upstream registries.

### DIFF
--- a/pkg/extensions/config/sync/config.go
+++ b/pkg/extensions/config/sync/config.go
@@ -23,18 +23,19 @@ type Config struct {
 }
 
 type RegistryConfig struct {
-	URLs             []string
-	PollInterval     time.Duration
-	Content          []Content
-	TLSVerify        *bool
-	OnDemand         bool
-	CertDir          string
-	MaxRetries       *int
-	RetryDelay       *time.Duration
-	OnlySigned       *bool
-	CredentialHelper string
-	PreserveDigest   bool          // sync without converting
-	SyncTimeout      time.Duration // timeout for on-demand sync operations; if zero or unset, defaults to 3 hours
+	URLs                  []string
+	PollInterval          time.Duration
+	Content               []Content
+	TLSVerify             *bool
+	OnDemand              bool
+	CertDir               string
+	MaxRetries            *int
+	RetryDelay            *time.Duration
+	OnlySigned            *bool
+	CredentialHelper      string
+	PreserveDigest        bool          // sync without converting
+	SyncTimeout           time.Duration // overall HTTP client timeout for all sync operations
+	ResponseHeaderTimeout time.Duration `yaml:"-"` // response header timeout; set in root.go
 }
 
 type Content struct {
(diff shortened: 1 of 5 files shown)


rchincha (approved):

pkg/extensions/config/sync/config.go
```diff
@@ -23,18 +23,19 @@ type Config struct {
...
+	TLSVerify             *bool
+	OnDemand              bool
+	CertDir               string
+	MaxRetries            *int
+	RetryDelay            *time.Duration
+	OnlySigned            *bool
+	CredentialHelper      string
+	PreserveDigest        bool          // sync without converting
+	SyncTimeout           time.Duration // overall HTTP client timeout for all sync operations
+	ResponseHeaderTimeout time.Duration `yaml:"-"` // response header timeout; set in root.go
```
benoittgt, line 38: @andaaron I'm curious about why we choose to make it not serializable/configurable? Isn't 30s a lot?

andaaron: I had these considerations:
- If the headers are not returned, there is no reason to wait for the entire SyncTimeout which is theoretically enough to copy the entire image content. I wanted to timeout faster.
- The value is a lot, which should accommodate slow deployments.
- This is part of the config struct in order to enable unit tests with smaller values.
- There would have been too many knobs exposed to the user.

Are you saying you are having trouble with this large timeout in your use case?

benoittgt: Thanks for the detailed explanation, it makes sense. To clarify my thinking:

I was trying to understand the worst-case time-to-failure when the remote source registry is completely down (pull-through cache scenario). My initial (wrong) assumption was that the timeouts were additive (30s + 30s + 10s = 70s), but after reading the code more carefully I see they are component timeouts guarding different sequential stages not stacked:

- DialContext: 30s
- TLSHandshakeTimeout: 10s
- ResponseHeaderTimeout: 30s
If the remote is fully unreachable, only DialContext fires -> ~30s to report timeout (with default 0 retries). The ResponseHeaderTimeout would only matter in a different failure mode (TCP+TLS succeed but server hangs without sending headers).

To answer your question, I'm not having trouble with a specific use case right now. I was mostly asking out of curiosity because 30s felt like a long time for ResponseHeaderTimeout if you're doing on-demand pull-through and the user is waiting. But your rationale about accommodating slow deployments and avoiding too many config knobs makes sense. The DialContext (30s) is already the dominant timeout for the "registry is down" case anyway, so making ResponseHeaderTimeout configurable wouldn't help much there.

Thanks again

Issue #3085, opened 2025-04-09
[Bug]: unable to use mTLS with accessControl


frq-asgard-josi:
### zot version

v2.1.2

### Describe the bug

What I want:
- Clients can authenticate via mTLS to get read access and pull images
- I can push images via a special user that has write access (I don't really care if that user can authenticate via mTLS or a password or a combination of both)
- anonymous users can't do anything

What actually happens:
1. Scenario 1: mTLS without any accessControl options
  - unauthenticated request are blocked but authenticated requests have full access
3. Scenario 2: mTLS with accessControl options
  - Configuration cannot be validated: `invalid server config: access control config requires one of httpasswd, ldap or openid authentication or using only 'anonymousPolicy' policies`
5. Scenario 3: mTLS with accessControl options and htpasswd file
 - mTLS is completly ignored

### Scenario 1
1. Configuration
```json
{
  "distSpecVersion": "1.1.0",
  "storage": {
    "rootDirectory": "/data/zot"
  },
  "http": {
    "address": "127.0.01",
    "port": "443",
    "tls": {
      "cert": "/etc/certs/host.fullchain.cert.pem",
      "key": "/etc/certs/host.key.pem",
      "cacert": "/etc/certs/ca.crt"
    }
  }
}
```
3. Client tool used
`podman login --cert-dir <some-dir> 127.0.0.1:443/hello-world:test -u "$(hostname)" -p=""`
`podman pull --cert-dir <some-dir> 127.0.0.1:443/hello-world:test`
`podman push --cert-dir <some-dir> 127.0.0.1:443/hello-world:test`
5. Seen error
This works as expected but doesn't meet the requirement that only some users should be allowed to push.

 ### Scenario 2
1. Configuration
```json
{
  "distSpecVersion": "1.1.0",
  "storage": {
    "rootDirectory": "/data/zot"
  },
  "http": {
    "address": "127.0.01",
    "port": "443",
    "tls": {
      "cert": "/etc/certs/host.fullchain.cert.pem",
      "key": "/etc/certs/host.key.pem",
      "cacert": "/etc/certs/ca.crt"
    },
  "accessControl": {
      "repositories": {
        "**": {
          "policies": [
            {
              "users": [
                "user1"
              ],
              "actions": [
                "read",
                "create",
                "update",
                "delete"
              ]
            }
          ],
          "defaultPolicy": [
            "read"
          ]
        }
      }
    }
  }
}
```
3. Client tool used
`zot verify <config-file>`
[...]

PR #3577, opened 2025-11-23
mTLS authn and authz support both for mTLS-only and for mTLS with other auth mechanisms

Endevir:
**What type of PR is this?**

feature

https://github.com/project-zot/zot/issues/3085

This PR implements better mTLS authentication support for the Zot registry. The implementation allows mTLS to coexist with other authentication methods like basic auth and LDAP when access control is enabled, rather than requiring them to be disabled. Important to notice that this does not break existing auth flow, instead mTLS will be used as lowest-priority mechanism, then no other auth headers are present. This PR also allows to have mTLS as a single auth method with access control enabled, validating users on their CommonName. 

The PR includes tests and TLS utility functions to ensure proper mTLS validation (both single mTLS and mTLS with basic auth) and access control functionality.
These tests do fail on Main branch and pass on this PR.

**Automation added to e2e**:

Just new tests mentioned in previous section

No, this only fixes mTLS auth behavior, without introducing new params or breaking changes.
Important to notice: this PR allows enabling mTLS as only authn mechanism (by providing config.HTTP.TLS.CACert) with access control enabled, which is not supported by latest version. So if user upgrades and switches to only-mTLS config with access control enabled, then after downgrade same config will not be accepted by server

Yes.

```release-note
Implemented better mTLS handling: now mTLS-only authentication with access control is supported. 

mTLS is also correctly handled when used with other Basic authentication methods (htpasswd/LDAP/etc.) when requesting server with no credentials - then mTLS identity from CommonName will used, when any of credentials passed - then mTLS will only validate certificate, but authentication will be performed by Basic authentication mechanism 
```


pkg/api/authn_test.go
```diff
@@ -866,6 +867,326 @@ func TestAPIKeys(t *testing.T) {
...
+			client.SetTLSClientConfig(&tls.Config{
+				MinVersion:   tls.VersionTLS13,
+				Certificates: []tls.Certificate{clientCert},
+				RootCAs:      caCertPool,
+			})
+			resp, err := client.R().SetBasicAuth("httpuser", "httppass").Get(baseURL + "/v2/test-repo/tags/list")
+			So(err, ShouldBeNil)
+			So(resp.StatusCode(), ShouldEqual, http.StatusForbidden)
+
+			// Test without any auth - should fail
```
andaaron: This comment seems out of place for what the test expects.

andaaron (commented):
Overall a very nice PR. Please look into the failed checks and rebase.

As an enhancement, I am thinking CN is not always the right source for `identity` in all practical settings. Maybe we should make it configurable: CN/DN/SAN (a specific identifier under SAN).

pkg/api/authn_test.go
```diff
@@ -866,6 +867,326 @@ func TestAPIKeys(t *testing.T) {
...
+		clientCertPath := path.Join(tempDir, "client.crt")
+		clientKeyPath := path.Join(tempDir, "client.key")
+		err = tlsutils.GenerateCertWithCNToFile("testuser", caCert, caKey, clientCertPath, clientKeyPath)
+		So(err, ShouldBeNil)
+
+		// Generate self-signed client cert for "testuser" user
+		selfSignedClientCertPath := path.Join(tempDir, "client-selfsigned.crt")
+		selfSignedClientKeyPath := path.Join(tempDir, "client-selfsigned.key")
+		err = tlsutils.GenerateSelfSignedCertWithCNToFile("testuser", selfSignedClientCertPath, selfSignedClientKeyPath)
+		So(err, ShouldBeNil)
```
vrajashkr: The code here that builds the certificates looks pretty similar to the code in the previous Convey. Perhaps having a small helper function might help to reduce some repetition.

pkg/api/authn_test.go
```diff
@@ -866,6 +867,327 @@ func TestAPIKeys(t *testing.T) {
...
+				RootCAs:      caCertPool,
+			})
+
+			_, err = client.R().Get(baseURL + "/v2/test-selfsigned-repo/tags/list")
+			So(err, ShouldNotBeNil)
+			So(err.Error(), ShouldContainSubstring, "tls: certificate required")
+		})
+	})
+
+	Convey("Test mTLS with basic auth authentication", t, func() {
```
vrajashkr: According to my understanding of the Convey execution order (https://github.com/smartystreets/goconvey/wiki/Execution-order), the 3 nested Conveys under this one would re-generate the certificates for each test. I'm wondering if the certificates could be generated outside the Convey and then re-used for the tests.

pkg/api/authn_test.go
```diff
@@ -866,6 +867,327 @@ func TestAPIKeys(t *testing.T) {
...
+			client = resty.New()
+			client.SetTLSClientConfig(&tls.Config{
+				MinVersion:   tls.VersionTLS13,
+				Certificates: []tls.Certificate{clientCert},
+				RootCAs:      caCertPool,
+			})
+			resp, err = client.R().SetBasicAuth("httpuser", "httppass").Get(baseURL + "/v2/htpasswd-repo/tags/list")
+			So(err, ShouldBeNil)
+			So(resp.StatusCode(), ShouldEqual, http.StatusNotFound) // 404 meaning we successfully passed auth
+		})
```
vrajashkr, line 1089: nit: for uniformity, a newline above this convey would be nice.

Endevir:
> As an enhancement, I am thinking CN is not always the right source for `identity` in all practical settings. Maybe we should make it configurable: CN/DN/SAN (a specific identifier under SAN).

Agreed, maybe it will be better to implement that in new PR (as it will need to introduce new config option and need to write more tests)? 

As per enchancements: we can also utilize `zot/pkg/test/tls/tls.go` for making certificates instead of using bash scripts at `zot/test/scripts/gen_certs.sh` and `zot/test/scripts/gen_nameless_certs.sh`, but I suggest doing this also in a new iteration (because this PR has already grown quite big)

pkg/test/tls/tls.go
```diff
@@ -0,0 +1,329 @@
...
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"math/big"
+	"net"
+	"os"
+	"time"
+)
+
+var ErrDecondeCAPEM = errors.New("failed to decode CA certificate PEM")
```
vrajashkr: typo in variable name. Was it was meant to be `ErrDecodeCAPEM`?

pkg/api/authn_test.go
Endevir: Thanks, inaccurate copy-paste, fixed that

andaaron:
> > As an enhancement, I am thinking CN is not always the right source for `identity` in all practical settings. Maybe we should make it configurable: CN/DN/SAN (a specific identifier under SAN).
> 
> Agreed, maybe it will be better to implement that in new PR (as it will need to introduce new config option and need to write more tests)?
> 
> As per enchancements: we can also utilize `zot/pkg/test/tls/tls.go` for making certificates instead of using bash scripts at `zot/test/scripts/gen_certs.sh` and `zot/test/scripts/gen_nameless_certs.sh`, but I suggest doing this also in a new iteration (because this PR has already grown quite big)

Sounds good in both cases.

pkg/test/tls/tls.go
```diff
@@ -0,0 +1,329 @@
...
+			Country:       []string{"US"},
+			Province:      []string{""},
+			Locality:      []string{"San Francisco"},
+			StreetAddress: []string{""},
+			PostalCode:    []string{""},
+		},
+		NotBefore:             time.Now(),
+		NotAfter:              time.Now().AddDate(10, 0, 0), // Valid for 10 years
+		IsCA:                  true,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
```
vrajashkr: nit: Since the CA cert would only be used to sign the subsequent certificates, I'm wondering if it should not have any extended key usage assigned.

pkg/test/tls/tls.go
```diff
@@ -0,0 +1,329 @@
...
+
+	err = os.WriteFile(keyPath, clientKeyBytes, 0o600)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// GenerateCertWithCN generates a client certificate with a specific CommonName not signed by any CA.
```
vrajashkr: The function name in the comment appears to mismatch with the actual function name.

pkg/api/authn.go
```diff
@@ -366,57 +417,41 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
...
-				next.ServeHTTP(response, request)
+			// If error occurred during authn process - return 500 error
+			if err != nil {
+				response.WriteHeader(http.StatusInternalServerError)
 
 				return
 			}
 
-			authFail(response, request, realm, delay)
+			// even if authentication failed - serve request for anonymous calls
```
Endevir: That is an incorrect comment left from previous iteration, need to fix it before merge

pkg/api/authn_test.go
Endevir: Huge thanks for review, fixed this by moving certificates generation out of Convey and removed also duplicating code by this change

pkg/api/authn.go
```diff
@@ -525,6 +559,14 @@ func noPasswdAuth(ctlr *Controller) mux.MiddlewareFunc {
...
 							// assign identity to authz context, needed for extensions
 							userAc.SetUsername(identity)
+
+							// Get groups for the mTLS user
+							accessControl := ctlr.Config.CopyAccessControlConfig()
+							if accessControl != nil {
+								ac := NewAccessController(ctlr.Config)
+								groups := ac.getUserGroups(identity)
+								userAc.AddGroups(groups)
+							}
```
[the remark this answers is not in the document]
Endevir: > and doesn't break after finding the first certificate with a non-empty CommonName

Actually does - there're no break statements, but first certificate with non-empty CN will cause function to return

pkg/api/authn_test.go
```diff
@@ -866,6 +867,310 @@ func TestAPIKeys(t *testing.T) {
...
+					},
+				},
+			},
+		}
+		conf.Storage.RootDirectory = t.TempDir()
+
+		ctlr := api.NewController(conf)
+		cm := test.NewControllerManager(ctlr)
+
+		cm.StartAndWait(port)
```
andaaron, line 1057: Note, given this is inside a convey, it would still start and stop a server for every convey nested inside it.
I saw the certificate generation was moved outside, you may also want to move setting up the server outside the conveys (you would need separate test functions for the different server configurations in that case).

pkg/api/controller_test.go
```diff
@@ -2607,38 +2613,43 @@ func TestTLSMutualAuth(t *testing.T) {
 		So(resp, ShouldNotBeNil)
 		So(resp.StatusCode(), ShouldEqual, http.StatusBadRequest)
 
-		// without client certs and creds, should get conn error
-		_, err = resty.R().Get(secureBaseURL)
-		So(err, ShouldNotBeNil)
+		// without client certs and creds, should path auth and get expected 404 code on non-protected endpoint
```
andaaron:
```suggestion
		// without client certs and creds, should pass auth and get expected 404 code on non-protected endpoint
```

Applies to multiple comments in this file.

pkg/api/controller_test.go
```diff
@@ -2985,26 +3002,29 @@ func TestTLSMutualAndBasicAuthAllowReadAccess(t *testing.T) {
...
 
 		// with client certs but without creds, reads should succeed
-		resp, err = resty.R().Get(secureBaseURL + "/v2/")
+		resp, err = client.R().Get(secureBaseURL + "/v2/")
 		So(err, ShouldBeNil)
 		So(resp.StatusCode(), ShouldEqual, http.StatusOK)
 
-		// with only client certs, writes should fail
-		resp, err = resty.R().Post(secureBaseURL + "/v2/repo/blobs/uploads/")
+		// with only client certs, writes should fail with inefficient permissions
```
andaaron:
```suggestion
		// with only client certs, writes should fail with insufficient permissions
```

pkg/api/authn.go
andaaron, line 412: We should add a `authConfig := ctlr.Config.CopyAuthConfig()` is this function. 
`tryAuthnHandlers` is called only once when the server starts, and the configuration may have changed since then.
```suggestion
      ctlr.Config.CopyAuthConfig()
	  failDelay := authConfig.GetFailDelay()
	  realm := ctlr.Config.GetRealm()
```

pkg/api/authn.go
```diff
@@ -100,6 +100,58 @@ func (amw *AuthnMiddleware) sessionAuthn(ctlr *Controller, userAc *reqCtx.UserAc
...
+	// Check if client certificate has verified chain
+	verifiedChains := request.TLS.VerifiedChains
+	if len(verifiedChains) == 0 || len(verifiedChains[0]) == 0 {
+		ctlr.Log.Debug().Msg("mTLS authentication failed - user provided certificate not signed by CA")
+
+		return false, nil
+	}
+
+	// Extract identity from certificate
+	for _, cert := range request.TLS.PeerCertificates {
```
rchincha, line 120: "The order of certificates in the slice typically represents the certificate chain, from leaf to intermediates."

we are really interested only in the leaf certs, and just make sure it is indeed a leaf cert and it is verified? How do we ensure that?

rchincha:
When both authentication paths are enabled, emit logs about this. Also there is the question of relative strength between the two paths?

andaaron:
> When both authentication paths are enabled, emit logs about this. 

When the server starts or the config is reloaded. It's actually a good idea to show what authentication is enabled in general. Deducing from the config could be too much for the user.

pkg/api/controller_test.go
Endevir: Omg sorry for my english=)

pkg/api/authn_test.go
Endevir: Agreed, just removed unnecessary nested conveys and re-initializations, tests seem to be more simple now and do not trigger server re-init

pkg/api/controller_test.go
Endevir: Still sorry for my english=)

pkg/api/authn.go
Endevir: > we are really interested only in the leaf certs

Well at most cases we do just need a leaf cert, sure. However, there could be situations when an intermediate certificate contains identity and leaf certificate does not, so we need to extract identity from lowest certificate containing that identity.

> make sure it is verified?

We ensure, that leaf cert is verified by clientCA by checking that there is at least one trusted path for certificate chain provided by client earlier at that function:
```
	// Check if client certificate has verified chain
	verifiedChains := request.TLS.VerifiedChains
	if len(verifiedChains) == 0 || len(verifiedChains[0]) == 0 {
		ctlr.Log.Debug().Msg("mTLS authentication failed - user provided certificate not signed by CA")

		return false, nil
	}
```

Endevir:
> When both authentication paths are enabled, emit logs about this. Also there is the question of relative strength between the two paths?

Agreed, that could be really useful, but could we leave that for later enchancement iteration?
Like other improvements, such as:

- the right source for identity in all practical settings. Maybe we should make it configurable: CN/DN/SAN (a specific identifier under SAN).
- utilize zot/pkg/test/tls/tls.go for making certificates instead of using bash scripts at zot/test/scripts/gen_certs.sh and zot/test/scripts/gen_nameless_certs.sh

and also logging enabled auth methods?

andaaron:
> and also logging enabled auth methods?

If you want, I can open a separate PR, and we can merge that first.

Endevir:
> >  and also logging enabled auth methods?
>
> If you want, I can open a separate PR, and we can merge that first.

Turned out that the tests needed to be fixed anyway, so implemented that while fixing tests.

Can you check pls whether the logic is okay?

https://github.com/project-zot/zot/pull/3577/files#diff-fb1e4410b579ed6fdfb24653b76ec7aeb8e0c892ee514286f52a971695580872R122

https://github.com/project-zot/zot/pull/3577/files#diff-78b9d3397c43138d6115e314c98b5d2c06663ab6076f77e7772496cff911d095R61

andaaron:
This is what I had in mind: https://github.com/project-zot/zot/pull/3599/files#diff-672be0187484cee0bc614f3f370d1fe732434bcc65e25c9ffd89a1abf6cd43c1. Those are the places where we show the config. When the controller starts, and when the config is reloaded while the server is running.

Endevir:
> This is what I had in mind: https://github.com/project-zot/zot/pull/3599/files#diff-672be0187484cee0bc614f3f370d1fe732434bcc65e25c9ffd89a1abf6cd43c1. Those are the places where we show the config. When the controller starts, and when the config is reloaded while the server is running.

Nice, thanks, then I'll rollback my changes related to logging enabled auth methods

pkg/api/controller_test.go
```diff
@@ -2607,38 +2613,43 @@ func TestTLSMutualAuth(t *testing.T) {
 		So(resp, ShouldNotBeNil)
 		So(resp.StatusCode(), ShouldEqual, http.StatusBadRequest)
 
-		// without client certs and creds, should get conn error
-		_, err = resty.R().Get(secureBaseURL)
-		So(err, ShouldNotBeNil)
+		// without client certs and creds, should pass auth and get expected 404 code on non-protected endpoint
```
andaaron, line 2616: If MTLS is enabled, there are no other authentication methods enabled, and there is no anonymous policy, why don't we return StatusUnauthorized, isn't that what authFail() is supposed to return? Where is the code returning StatusNotFound?

pkg/api/authn_test.go
```diff
@@ -1389,7 +1652,7 @@ func TestRedisCookieStore(t *testing.T) {
 				So(cookieStore, ShouldNotBeNil)
 			} else {
 				So(err, ShouldNotBeNil)
-				So(err.Error(), ShouldEqual, testCase.expectedErrStr)
+				// So(err.Error(), ShouldEqual, testCase.expectedErrStr)
```
andaaron, line 1655: Before we merge, this needs to either be removed or uncommented.

pkg/api/authn.go
```diff
@@ -366,57 +415,54 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
...
-						authFail(response, request, realm, delay)
-					}
-
-					response.WriteHeader(http.StatusInternalServerError)
-
-					return
+				// the session header can be present also for anonymous calls - then we should pass
+				if allowAnonymous || isMgmtRequested {
+					authenticated = true
 				}
```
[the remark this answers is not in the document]
andaaron: I was thinking the same thing. But I guess the non-nil err was returned above, so the request would fail with error 500 below. 

But maybe it's better to `break` the `case` before this check in case of auth failure just to be on the safe side.

andaaron:
@Endevir, I have merged https://github.com/project-zot/zot/pull/3599

Please rebase (note the Trivy uprev was also merged today, and if you execute the go command outside the make target, you need to set `export GOEXPERIMENT=jsonv2`.

pkg/api/authn.go
rchincha: | Well at most cases we do just need a leaf cert, sure. However, there could be situations when an intermediate certificate contains identity and leaf certificate does not, so we need to extract identity from lowest certificate containing that identity.

^ this would be my main objection. Can we just enforce this? Yes some certs may not work, but then that should get fixed.

We want leaf certs with a verified chain and we will consider only the identity in the leaf cert.

rchincha:
@Endevir 

Here is a code pattern that might work. Let's see if this can be incorporated into your PR (keeping the assumption that first cert in the chain is the leaf)

```golang
package main

import (
	"crypto/tls"
	"crypto/x509"
	"fmt"
	"log"
	"net"
	"time"
)

func main() {
	// Example: Client-side connection and validation
	// In a real scenario, you'd establish a connection to a server.
	// For demonstration, we'll simulate a connection.

	// Load trusted root CAs (replace with your actual trust store)
	rootCAs, _ := x509.SystemCertPool()
	if rootCAs == nil {
		rootCAs = x509.NewCertPool()
	}

	// Assume we have a *tls.Conn from an established connection
	// For this example, we'll create a dummy one for illustration.
	// In a real application, this would come from `tls.Dial` or `tls.Server.Accept`.
	conn, err := tls.Dial("tcp", "example.com:443", &tls.Config{
		RootCAs: rootCAs,
	})
	if err != nil {
		log.Fatalf("Failed to establish TLS connection: %v", err)
	}
	defer conn.Close()

	// Get the connection state
	state := conn.ConnectionState()

	// Check if peer certificates are available
	if len(state.PeerCertificates) == 0 {
		log.Fatalf("No peer certificates received")
	}

	// The leaf certificate is typically the first one in the chain provided by the peer.
	// However, to be robust, we'll iterate and verify each one if needed,
	// focusing on finding one that is a valid leaf.
	// In most cases, `state.PeerCertificates[0]` will be the leaf.
	leafCert := state.PeerCertificates[0] // Assuming the first is the intended leaf

	// Prepare verification options
	opts := x509.VerifyOptions{
		Intermediates: x509.NewCertPool(),
		Roots:         rootCAs,
		CurrentTime:   time.Now(), // Important for checking certificate validity period
		KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth}, // Or x509.ExtKeyUsageClientAuth for client certs
	}

	// Add intermediate certificates to the pool for verification
	for i, cert := range state.PeerCertificates {
		if i == 0 { // Skip the leaf itself
			continue
		}
		opts.Intermediates.AddCert(cert)
	}

	// Verify the leaf certificate
	chains, err := leafCert.Verify(opts)
	if err != nil {
		log.Fatalf("Leaf certificate verification failed: %v", err)
	}

	fmt.Printf("Leaf certificate successfully verified. Found %d valid chains.\n", len(chains))
	// You can further inspect the 'chains' to see the full valid paths to a trusted root.
}
```

andaaron:
Picked this up, added the fixes, and more customization for certificate generation: https://github.com/project-zot/zot/pull/3624.

pkg/api/controller_test.go
andaaron: Was wrong here, 404 is correct.

rchincha:
@Endevir pls try https://github.com/project-zot/zot/pull/3624

Endevir:
@rchincha 

> @Endevir pls try #3624

Hi! Apologies for long silence and huge thanks for picking up. I'll be able to test that on my installation tomorrow

andaaron:
@Endevir, @rchincha 

There's one thing we're still having doubts about.
If a client sends both a certificate and an Authorization header, the server stack verifies the certificate is valid, but we pick the identify from the Authorization header.
I propose we either:
- reject this case with authentication failure given there are 2 different authentication methods sent by the client at the same time
or
- at least pick the identify from the certificate and ignore the Authorization header in this case.

Endevir:
@andaaron 

Since we previously supported having both mTLS and Basic Auth enabled — while still honoring the identity provided by Basic Auth — it seems more intuitive to treat Basic Auth as the higher-priority authentication method when both are enabled.

There are also valid (in my opinion) scenarios where a user may want to use mTLS primarily for host-based access control, ensuring that connections originate only from trusted hosts, while relying on standard authentication mechanisms (such as tokens or htpasswd) for actual user identity verification.

If a user wants to rely on the identity from a client certificate, they can simply omit the `Authorization` header. This leaves room for users who prefer mTLS as their identity source.

Here is the logic I would recommend going forward (still not completely sure about this=)):

* **If mTLS is enabled**, client certificates should be required for any authenticated request. Requests that do not use mTLS should be treated as anonymous, even if they include an `Authorization` header. This ensures mTLS remains a predictable access-control mechanism (mTLS validation usually prepends authorization).
* **If mTLS is enabled alongside other authentication methods**, users *may override* the identity derived from the client certificate by providing additional authentication (e.g., using Basic Auth or tokens). This allows setups like: mTLS grants read-only access by default, and users can elevate to admin by presenting credentials, without the need to generate additional certificates. 

P.S. Maybe better to move discussion into https://github.com/project-zot/zot/pull/3624 ?

Issue #3578, opened 2025-11-23
[Bug]: can't disable sync extension


mottetm:
### zot version

main

### Describe the bug

`enable: false` has no effect if any registry configuration is set in the config object.

```go
// IsSyncEnabled checks if sync is enabled in this extensions config.
func (e *ExtensionConfig) IsSyncEnabled() bool {
	if e == nil {
		return false
	}

	// Sync is enabled if either:
	// 1. Explicitly enabled (Enable == true), OR
	// 2. There are registries configured (enabled by default when registries exist)
	// This matches the behavior in root.go where Sync.Enable defaults to true when registries are present
	return e.Sync != nil && ((e.Sync.Enable != nil && *e.Sync.Enable) || len(e.Sync.Registries) > 0)
}
```

If `e.Sync.Enable` is false, the condition falls back on the last `len(e.Sync.Registries) > 0`. If this is true, then the function returns true instead of false.

### To reproduce

1. Configuration

A configuration with at least on registry configured in the sync extension and the extension disabled using `enable: false`.

2. Client tool used

`crane manifest xyz`

3. Seen error

The logs show the sync extension being triggered

### Expected behavior

The extension should be disabled.

PR #3579, opened 2025-11-23
fix: sync should be disabled when enable is false

mottetm:
**What type of PR is this?**

bug

#3578 

Instead of enabling the extension if `len(e.Sync.Registries) > 0` in the _or_ clause, only enable it if we also have `e.Sync.Enable == nil`

I updated the tests so that `Test with ExtensionConfig and Sync and Enable but disabled` uses the case with registry configuration where the current codebase incorrectly returns true when it should return false.

```
    Test IsSyncEnabled() 
      Test with nil ExtensionConfig ✔
      Test with ExtensionConfig but nil Sync ✔
      Test with ExtensionConfig and Sync but nil Enable ✔
      Test with ExtensionConfig and Sync and Enable but disabled ✘
      Test with ExtensionConfig and Sync and Enable enabled ✔
```

After adding the code change, the tests passes again as expected.

It shouldn't

No

### DIFF
--- a/.github/workflows/cluster.yaml
+++ b/.github/workflows/cluster.yaml
@@ -51,7 +51,7 @@ jobs:
                 --health-cmd "curl http://localhost:9000/minio/health/live" \
                 minio/minio:RELEASE.2024-07-16T23-46-41Z server /data
       - name: Install py minio
-        run: pip3 install minio
+        run: pip3 install minio==7.2.18
 
       - name: Wait for minio to come up
         run: |
@@ -306,7 +306,7 @@ jobs:
                 --health-cmd "curl http://localhost:9000/minio/health/live" \
                 minio/minio:RELEASE.2024-07-16T23-46-41Z server /data
       - name: Install py minio
-        run: pip3 install minio
+        run: pip3 install minio==7.2.18
 
       - name: Wait for minio to come up
         run: |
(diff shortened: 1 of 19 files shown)


pkg/extensions/config/config_test.go
```diff
@@ -57,6 +57,11 @@ func buildSyncConfig(enabled bool) *config.ExtensionConfig {
 	ext := &config.ExtensionConfig{}
 	ext.Sync = &sync.Config{
 		Enable: &enabled,
+		Registries: []sync.RegistryConfig{
```
andaaron, line 60: Why was this needed?

andaaron: Is this for the test `Test with ExtensionConfig and Sync but nil Enable ` to pass?  `Test with ExtensionConfig and Sync and Enable but disable` should have passed with the change in the config file even without this.

mottetm: No. This is to make sure that the current logic properly fails for the test `Test with ExtensionConfig and Sync and Enable but disabled` when `enable` is `false` but there are registry configurations present.

mottetm: The fix would have passed without the change, yes, but the fact that the test was passing with the flawed logic was the issue 🙂

mottetm: Also not sure about why the minio tests started to fail 🤔

andaaron: They released a new minio version 2 hours ago, and it seems like an api change slipped in: https://github.com/minio/minio-py/pull/1505#issuecomment-3569935445

andaaron:
@mottetm, I merged a fix for the monio issue: https://github.com/project-zot/zot/pull/3581
Please rebase.

andaaron:
Thanks

PR #3607, opened 2025-11-30
fix(trivy): prevent temp file accumulation in /tmp with configurable TMPDIR management

andaaron:
This commit addresses the issue where Trivy temporary directories (trivy-xxxxxx) were accumulating in /tmp/, causing disk space issues.

Changes:
- Set TMPDIR to use _trivy/tmp directory instead of system /tmp/ during scans
- Add automatic cleanup of temp directories during DB updates
- Add OverrideTmpDir configuration option to enable/disable TMPDIR management
- Default OverrideTmpDir to false to preserve old behavior (backward compatible)
- Restore original TMPDIR value after each scan completes

The cleanup removes all contents from _trivy/tmp directories for all stores (default and substores) during periodic DB updates.

Users can enable this feature by setting:
```
  "extensions": {
    "search": {
      "cve": {
        "trivy": {
          "overrideTmpDir": true
        }
      }
    }
  }
```

### DIFF
--- a/pkg/extensions/config/config.go
+++ b/pkg/extensions/config/config.go
@@ -60,6 +60,11 @@ type CVEConfig struct {
 type TrivyConfig struct {
 	DBRepository     string // default is "ghcr.io/aquasecurity/trivy-db"
 	JavaDBRepository string // default is "ghcr.io/aquasecurity/trivy-java-db"
+	// OverrideTmpDir controls whether to override TMPDIR and clean up temporary files.
+	// When true, TMPDIR is overridden to cacheDir/tmp and temp files are cleaned up during DB updates.
+	// When false, TMPDIR is not modified and temp files are not cleaned up.
+	// Default is false to preserve old behavior (backward compatibility).
+	OverrideTmpDir bool `mapstructure:",omitempty"`
 }
 
 type MetricsConfig struct {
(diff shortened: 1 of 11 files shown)


pkg/extensions/config/config.go
```diff
@@ -60,6 +60,11 @@ type CVEConfig struct {
 type TrivyConfig struct {
 	DBRepository     string // default is "ghcr.io/aquasecurity/trivy-db"
 	JavaDBRepository string // default is "ghcr.io/aquasecurity/trivy-java-db"
+	// OverrideTmpDir controls whether to override TMPDIR and clean up temporary files.
```
rchincha, line 63: Do we even need to expose this? why not set it to zot/_trivy/tmp always?

andaaron: This is a change in behavior. I am thinking it's safer this way for the users who are expecting /tmp/ to be used.

rchincha: It was already a change from v2.1.7 - we are just reverting to that.

andaaron: No, that was not the case. It was always downloaded the DB to /tmp before moving it. There may be some additional files now.

PR #3618, opened 2025-12-01
fix(trivy): cleanup Trivy temporary directory

andaaron:
This Trivy commit changed the way the cleanup is done for Trivy operations: https://github.com/aquasecurity/trivy/commit/8f5b56005a4e8752976524750089dc9ea2c91e40

We need to start calling the APIs to create and cleanup the temp dir to ensure we don't leave around content in the /tmp folder

### DIFF
--- a/pkg/extensions/search/cve/trivy/scanner.go
+++ b/pkg/extensions/search/cve/trivy/scanner.go
@@ -19,6 +19,7 @@ import (
 	"github.com/aquasecurity/trivy/pkg/flag"
 	"github.com/aquasecurity/trivy/pkg/javadb"
 	"github.com/aquasecurity/trivy/pkg/types"
+	xos "github.com/aquasecurity/trivy/pkg/x/os"
 	"github.com/google/go-containerregistry/pkg/name"
 	regTypes "github.com/google/go-containerregistry/pkg/v1/types"
 	godigest "github.com/opencontainers/go-digest"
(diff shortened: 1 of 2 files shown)


pkg/extensions/search/cve/trivy/scanner.go
```diff
@@ -193,29 +194,54 @@ func (scanner Scanner) getTrivyOptions(image string) flag.Options {
 	return opts
 }
 
+// withTempDir creates a temporary directory using xos.TempDir(), executes the provided function,
+// and then calls xos.Cleanup() to clean up Trivy's process-specific temp directory.
+func (scanner Scanner) withTempDir(wrappedFunc func() error) error {
+	// Ensure Trivy's process-specific temp directory is initialized
+	// by calling TempDir() first - this will create it if it doesn't exist
+	tempDir := xos.TempDir()
```
rchincha, line 202: Where does trivy make this? I would prefer this under zot/ root dir

andaaron: See https://github.com/aquasecurity/trivy/commit/8f5b56005a4e8752976524750089dc9ea2c91e40.
They basically use the golang os library temporary directory (`/tmp` unless `TMPDIR` is specified), and create a folder `trivy-<current_pid>`, In their code they are calling the APIs to create the folder and delete it at the end of the trivy binary execution. The creation they do only once (sync.once), which may be ok for a command line tool with short execution duration, but we want cleanup more often, so we need to create it and delete it ourselves, as we dont' want it to live for the entire zot server execution.

With regards to using the root dir I strongly disagree for 2 reasons:
1. Typically `/tmp` is tmpfs format, which is RAM-based, which is **faster** and it should be cleaned up on its own on a reboot. Note Trivy uses /tmp by default for downloading the DB before moving it to the final location, as well as for unpacking layers before scanning them.  I don't think there's any valid reason why we should keep this Trivy data in the permanent storage partition., And there's also the point of the data remaining there forever if something goes wrong and the cleanup doesn't work. As opposed to `/tmp`, which gets cleaned up on system reboot.
2. I don't think we should change the existing behavior we've had for the past few years without a solid reason to do it, and we would need to provide a way for the existing users to keep the existing behavior. As I said in my previous point it, as a zot admin I would want to keep that data in /tmp for those reasons, same as it was before.

Issue #3601, opened 2025-11-27
[Feat]: Add a upgrade test in our ci/cd pipeline


rchincha:
### Is your feature request related to a problem? Please describe.

Start up a released zot.
Do registry read, write and search operations.
Shut it down.
Start a zot with current PR and do all of the above and it MUST NOT break.

minimal and full zot, separately.

PR #3621, opened 2025-12-04
ci: add a upgrade bats test

rchincha:
Fixes https://github.com/project-zot/zot/issues/3601

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/test/blackbox/helpers_upgrade.bash
+++ b/test/blackbox/helpers_upgrade.bash
@@ -28,111 +28,149 @@ function teardown_file() {
 }
 
 # ==============================================================================
-# RELEASE TEST FUNCTIONS
-# These functions are used to test the released version of zot before upgrade
+# COMMON HELPER FUNCTIONS
+# These are the core functions used by both release and new test functions
 # ==============================================================================
 
-function test_release_push_image() {
-    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+# Helper: Get the zot port
+function get_zot_port() {
+    cat ${BATS_FILE_TMPDIR}/zot.port
+}
+
+# Helper: Push an image using skopeo
+# Args: $1 = image_name, $2 = tag
+function helper_push_image() {
+    local image_name=${1:-golang}
+    local tag=${2:-1.20}
+    local zot_port=$(get_zot_port)
     run skopeo --insecure-policy copy --dest-tls-verify=false \
-        oci:${TEST_DATA_DIR}/golang:1.20 \
-        docker://127.0.0.1:${zot_port}/golang:1.20
+        oci:${TEST_DATA_DIR}/${image_name}:${tag} \
+        docker://127.0.0.1:${zot_port}/${image_name}:${tag}
     [ "$status" -eq 0 ]
     run curl http://127.0.0.1:${zot_port}/v2/_catalog
     [ "$status" -eq 0 ]
-    [ $(echo "${lines[-1]}" | jq '.repositories[]') = '"golang"' ]
-    run curl http://127.0.0.1:${zot_port}/v2/golang/tags/list
+    [ $(echo "${lines[-1]}" | jq --arg name "${image_name}" 'any(.repositories[]; . == $name)') = true ]
+    run curl http://127.0.0.1:${zot_port}/v2/${image_name}/tags/list
     [ "$status" -eq 0 ]
-    [ $(echo "${lines[-1]}" | jq '.tags[]') = '"1.20"' ]
+    [ $(echo "${lines[-1]}" | jq --arg tag "${tag}" 'any(.tags[]; . == $tag)') = true ]
 }
 
-function test_release_pull_image() {
+# Helper: Pull an image using skopeo
+# Args: $1 = image_name, $2 = tag
+function helper_pull_image() {
+    local image_name=${1:-golang}
+    local tag=${2:-1.20}
     local oci_data_dir=${BATS_FILE_TMPDIR}/oci
-    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+    local zot_port=$(get_zot_port)
     run skopeo --insecure-policy copy --src-tls-verify=false \
-        docker://127.0.0.1:${zot_port}/golang:1.20 \
-        oci:${oci_data_dir}/golang:1.20
+        docker://127.0.0.1:${zot_port}/${image_name}:${tag} \
+        oci:${oci_data_dir}/${image_name}:${tag}
     [ "$status" -eq 0 ]
-    run cat ${BATS_FILE_TMPDIR}/oci/golang/index.json
+    run cat ${oci_data_dir}/${image_name}/index.json
     [ "$status" -eq 0 ]
-    [ $(echo "${lines[-1]}" | jq '.manifests[].annotations."org.opencontainers.image.ref.name"') = '"1.20"' ]
+    [ $(echo "${lines[-1]}" | jq --arg tag "${tag}" '.manifests[].annotations."org.opencontainers.image.ref.name" == $tag') = true ]
 }
 
-function test_release_push_image_index() {
-    # --multi-arch below pushes an image index (containing many images) instead
-    # of an image manifest (single image)
-    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+# Helper: Push an image index (multi-arch)
+# Args: $1 = source_image, $2 = dest_image_name, $3 = tag
+function helper_push_image_index() {
+    local source_image=${1:-docker://public.ecr.aws/docker/library/busybox:latest}
+    local dest_name=${2:-busybox}
+    local tag=${3:-latest}
+    local zot_port=$(get_zot_port)
     run skopeo --insecure-policy copy --format=oci --dest-tls-verify=false --multi-arch=all \
-        docker://public.ecr.aws/docker/library/busybox:latest \
-        docker://127.0.0.1:${zot_port}/busybox:latest
+        ${source_image} \
+        docker://127.0.0.1:${zot_port}/${dest_name}:${tag}
     [ "$status" -eq 0 ]
     run curl http://127.0.0.1:${zot_port}/v2/_catalog
     [ "$status" -eq 0 ]
-    [ $(echo "${lines[-1]}" | jq '.repositories[0]') = '"busybox"' ]
-    run curl http://127.0.0.1:${zot_port}/v2/busybox/tags/list
+    [ $(echo "${lines[-1]}" | jq --arg name "${dest_name}" 'any(.repositories[]; . == $name)') = true ]
+    run curl http://127.0.0.1:${zot_port}/v2/${dest_name}/tags/list
     [ "$status" -eq 0 ]
-    [ $(echo "${lines[-1]}" | jq '.tags[]') = '"latest"' ]
+    [ $(echo "${lines[-1]}" | jq --arg tag "${tag}" 'any(.tags[]; . == $tag)') = true ]
 }
 
-function test_release_pull_image_index() {
+# Helper: Pull an image index (multi-arch)
+# Args: $1 = image_name, $2 = tag
+function helper_pull_image_index() {
+    local image_name=${1:-busybox}
+    local tag=${2:-latest}
     local oci_data_dir=${BATS_FILE_TMPDIR}/oci
-    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+    local zot_port=$(get_zot_port)
     run skopeo --insecure-policy copy --src-tls-verify=false --multi-arch=all \
-        docker://127.0.0.1:${zot_port}/busybox:latest \
-        oci:${oci_data_dir}/busybox:latest
+        docker://127.0.0.1:${zot_port}/${image_name}:${tag} \
+        oci:${oci_data_dir}/${image_name}:${tag}
     [ "$status" -eq 0 ]
-    run cat ${BATS_FILE_TMPDIR}/oci/busybox/index.json
+    run cat ${oci_data_dir}/${image_name}/index.json
     [ "$status" -eq 0 ]
-    [ $(echo "${lines[-1]}" | jq '.manifests[].annotations."org.opencontainers.image.ref.name"') = '"latest"' ]
+    [ $(echo "${lines[-1]}" | jq --arg tag "${tag}" '.manifests[].annotations."org.opencontainers.image.ref.name" == $tag') = true ]
     run skopeo --insecure-policy --override-arch=arm64 --override-os=linux copy --src-tls-verify=false --multi-arch=all \
-        docker://127.0.0.1:${zot_port}/busybox:latest \
-        oci:${oci_data_dir}/busybox:latest
+        docker://127.0.0.1:${zot_port}/${image_name}:${tag} \
+        oci:${oci_data_dir}/${image_name}:${tag}
     [ "$status" -eq 0 ]
-    run cat ${BATS_FILE_TMPDIR}/oci/busybox/index.json
+    run cat ${oci_data_dir}/${image_name}/index.json
     [ "$status" -eq 0 ]
-    [ $(echo "${lines[-1]}" | jq '.manifests[].annotations."org.opencontainers.image.ref.name"') = '"latest"' ]
+    [ $(echo "${lines[-1]}" | jq --arg tag "${tag}" '.manifests[].annotations."org.opencontainers.image.ref.name" == $tag') = true ]
 }
 
-function test_release_push_oras_artifact() {
-    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+# Helper: Push an ORAS artifact
+# Args: $1 = artifact_name, $2 = tag
+function helper_push_oras_artifact() {
+    local artifact_name=${1:-hello-artifact}
+    local tag=${2:-v2}
+    local zot_port=$(get_zot_port)
     echo "{\"name\":\"foo\",\"value\":\"bar\"}" > config.json
     echo "hello world" > artifact.txt
-    run oras push --plain-http 127.0.0.1:${zot_port}/hello-artifact:v2 \
+    run oras push --plain-http 127.0.0.1:${zot_port}/${artifact_name}:${tag} \
         --config config.json:application/vnd.acme.rocket.config.v1+json artifact.txt:text/plain -d -v
     [ "$status" -eq 0 ]
     rm -f artifact.txt
     rm -f config.json
 }
 
-function test_release_pull_oras_artifact() {
-    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
-    run oras pull --plain-http 127.0.0.1:${zot_port}/hello-artifact:v2 -d -v
+# Helper: Pull an ORAS artifact
+# Args: $1 = artifact_name, $2 = tag
+function helper_pull_oras_artifact() {
+    local artifact_name=${1:-hello-artifact}
+    local tag=${2:-v2}
+    local zot_port=$(get_zot_port)
+    run oras pull --plain-http 127.0.0.1:${zot_port}/${artifact_name}:${tag} -d -v
     [ "$status" -eq 0 ]
     grep -q "hello world" artifact.txt
     rm -f artifact.txt
 }
 
-function test_release_attach_oras_artifacts() {
-    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+# Helper: Attach ORAS artifacts (signature and sbom) to an image
+# Args: $1 = image_name, $2 = tag
+function helper_attach_oras_artifacts() {
+    local image_name=${1:-golang}
+    local tag=${2:-1.20}
+    local zot_port=$(get_zot_port)
     # attach signature
     echo "{\"artifact\": \"\", \"signature\": \"pat hancock\"}" > ${BATS_FILE_TMPDIR}/signature.json
-    run oras attach --disable-path-validation --plain-http 127.0.0.1:${zot_port}/golang:1.20 --artifact-type 'signature/example' ${BATS_FILE_TMPDIR}/signature.json:application/json
+    run oras attach --disable-path-validation --plain-http 127.0.0.1:${zot_port}/${image_name}:${tag} --artifact-type 'signature/example' ${BATS_FILE_TMPDIR}/signature.json:application/json
     [ "$status" -eq 0 ]
     # attach sbom
-    echo "{\"version\": \"0.0.0.0\", \"artifact\": \"'127.0.0.1:${zot_port}/golang:1.20'\", \"contents\": \"good\"}" > ${BATS_FILE_TMPDIR}/sbom.json
-    run oras attach --disable-path-validation --plain-http 127.0.0.1:${zot_port}/golang:1.20 --artifact-type 'sbom/example' ${BATS_FILE_TMPDIR}/sbom.json:application/json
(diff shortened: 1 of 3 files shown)


rchincha:
Need this to make sure we don't break stuff after our cache refactoring.

test/blackbox/upgrade_minimal.bats
```diff
@@ -0,0 +1,720 @@
...
+    [ "$status" -eq 1 ]
+}
+
+@test "[upgrade] upgrade to new binary" {
+    zot_stop_all
+    local zot_config_file=${BATS_FILE_TMPDIR}/zot_config.json
+    local zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+    zot_serve ${ZOT_MINIMAL_PATH} ${zot_config_file}
+    wait_zot_reachable ${zot_port}
+    sleep 60
```
andaaron: Why do we need this?

test/blackbox/helpers_zot.bash
```diff
@@ -28,6 +28,46 @@ function zot_serve() {
     echo -n "$! " >> ${BATS_FILE_TMPDIR}/zot.pid
 }
 
+function zot_rel_serve() {
+    local config_file
+    if [ -n "${2}" ]; then
```
andaaron: The inded use of these arguments is confusing

test/blackbox/upgrade.bats
```diff
@@ -0,0 +1,720 @@
...
+    },
+    "http": {
+        "address": "0.0.0.0",
+        "port": "${zot_port}"
+    },
+    "log": {
+        "level": "debug",
+        "output": "${BATS_FILE_TMPDIR}/zot.log"
+    }
+}
```
andaaron, line 49: Does this add any value compared to minimal? We use the same configuration. Actually, the configuration could use more configurations. At least let's see if metadb work correctly after the upgrade, And after the upgrade we can list the images we pushed before upgrade. Using both graphql and dist spec apis.

test/blackbox/upgrade_minimal.bats
```diff
@@ -0,0 +1,720 @@
...
+    echo "{\"artifact\": \"\", \"signature\": \"pat hancock\"}" > ${BATS_FILE_TMPDIR}/signature.json
+    run oras attach --disable-path-validation --plain-http 127.0.0.1:${zot_port}/golang:1.20 --artifact-type 'signature/example' ${BATS_FILE_TMPDIR}/signature.json:application/json
+    [ "$status" -eq 0 ]
+    # attach sbom
+    echo "{\"version\": \"0.0.0.0\", \"artifact\": \"'127.0.0.1:${zot_port}/golang:1.20'\", \"contents\": \"good\"}" > ${BATS_FILE_TMPDIR}/sbom.json
+    run oras attach --disable-path-validation --plain-http 127.0.0.1:${zot_port}/golang:1.20 --artifact-type 'sbom/example' ${BATS_FILE_TMPDIR}/sbom.json:application/json
+    [ "$status" -eq 0 ]
+}
+
+@test "[new] discover oras artifacts" {
```
andaaron, line 204: Can we push new content, and verify both are available after the upgrade?
This applies in general for most of the test pushes in these files.

test/blackbox/upgrade_minimal.bats
rchincha: zot is not available immediately to serve requests. This works, so will reduce this to whatever minimum works.

andaaron: But we do have wait_zot_reachable. Also, doesn't the ready endpoint help?

test/blackbox/upgrade.bats
```diff
@@ -0,0 +1,719 @@
...
+    cat > Dockerfile <<DOCKERFILE
+    FROM ghcr.io/project-zot/test-images/busybox-docker:1.37
+    RUN echo "hello world" > /testfile
+DOCKERFILE
+    docker build -f Dockerfile . -t localhost:${zot_port}/test
+    run docker push localhost:${zot_port}/test
+    [ "$status" -eq 1 ]
+    run docker pull localhost:${zot_port}/test
+    [ "$status" -eq 1 ]
+}
```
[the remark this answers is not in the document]
rchincha: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/project-zot/zot/pull/3621#discussion_r2589890874)

test/blackbox/upgrade_minimal.bats
rchincha: that only checks /v2/ is available, zot does more.

test/blackbox/upgrade.bats
rchincha: added extensions in PR update

test/blackbox/upgrade.bats
```diff
@@ -0,0 +1,730 @@
...
+
+@test "[new] push OCI artifact references with regclient" {
+    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+    run regctl artifact put localhost:${zot_port}/manifest-ref:demo <<TXT
+test artifact
+TXT
+    [ "$status" -eq 0 ]
+    run regctl artifact list localhost:${zot_port}/manifest-ref:demo --format raw-body
+    [ "$status" -eq 0 ]
+    [ $(echo "${lines[-1]}" | jq '.manifests | length') -eq 1 ]
```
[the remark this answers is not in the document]
rchincha: because at end of the previous run, additional artifacts are added.

test/blackbox/upgrade_minimal.bats
rchincha: fixed by pushing alpine image after upgrade

test/blackbox/upgrade.bats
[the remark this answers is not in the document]
rchincha: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/project-zot/zot/pull/3621#discussion_r2589890874)

test/blackbox/helpers_zot.bash
```diff
@@ -28,6 +28,68 @@ function zot_serve() {
...
+    if [ ! -f "${zot_path}" ]; then
+        if ! curl -f -L -o "${zot_path}" https://github.com/project-zot/zot/releases/latest/download/zot-${OS}-${ARCH}; then 
+            echo "ERROR: Failed to download zot binary from GitHub." >&2
+            return 1
+        fi
+        # Download checksum file and verify integrity
+        checksum_url="https://github.com/project-zot/zot/releases/latest/download/checksums.sha256.txt"
+        checksum_file="${BATS_FILE_TMPDIR}/zot-sha256sums.txt"
+        curl -L -o "${checksum_file}" "${checksum_url}"
+        expected_sum=$(grep "zot-${OS}-${ARCH}$" "${checksum_file}" | awk '{print $1}')
```
andaaron, line 44: Nitpick: maybe this could be simplified by using something like `"$(cat ${checksum_file}) ${zot_path}" | sha256sum --check --status`

Same comment for below.

test/blackbox/upgrade.bats
```diff
@@ -0,0 +1,821 @@
...
+    wait_zot_reachable ${zot_port}
+    sleep 10    # zot does additional initialization/verification during startup
+}
+
+# After upgrading to the new binary, expect additional artifacts (a signature
+# and an sbom) that were attached
+
+@test "[new] push image" {
+    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+    run skopeo --insecure-policy copy --dest-tls-verify=false \
```
andaaron: While overwriting make sense as a test case. I think doing it first interferes with any validation we want to do on verifying the previously pushed image to the older zot version is still accessible.

rchincha: We can check existence first and then push again. We are already pushing a entirely new image.

andaaron: But we don't have tests verifying the old images already pushed to zot before upgrade are still found after the upgrade.
Or metadata introduced in metadb for those images is not wiped by the upgrade.

rchincha: fixed

test/blackbox/helpers_zot.bash
```diff
@@ -28,6 +28,70 @@ function zot_serve() {
...
+            echo "ERROR: Checksum verification failed for zot-${OS}-${ARCH}-minimal"
+            exit 1
+        fi
+        chmod +x "${zot_path}"
+    fi
+
+    ${zot_path} serve ${config_file} &
+    # zot.pid file keeps a list of zot server PIDs (in case multiple zot servers are started)
+    echo -n "$! " >> ${BATS_FILE_TMPDIR}/zot.pid
+}
```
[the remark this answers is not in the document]
vrajashkr: I agree. There is quite a bit of code that could be consolidated in these 2 functions.

test/blackbox/upgrade.bats
```diff
@@ -0,0 +1,929 @@
...
+    [ "$status" -eq 0 ]
+    [ $(echo "${lines[-1]}" | jq '.manifests[].annotations."org.opencontainers.image.ref.name"') = '"1.20"' ]
+}
+
+@test "[release] push image index" {
+    # --multi-arch below pushes an image index (containing many images) instead
+    # of an image manifest (single image)
+    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+    run skopeo --insecure-policy copy --format=oci --dest-tls-verify=false --multi-arch=all \
+        docker://public.ecr.aws/docker/library/busybox:latest \
```
vrajashkr: nit: I think it might be better if we push an internal multi-arch image instead of one fetched during the test from an different registry - that way, an outage elsewhere is less likely to affect the test runs.

test/blackbox/upgrade_minimal.bats
vrajashkr, line 1: A bunch of tests between upgrade.bats and upgrade_minimal.bats seem to be the same. Instead of a duplicate, is it possible to have a common upgrade test suite for registry tests and run the test suite against 2 binaries - full and minimal using the github actions matrix instead?

If there are any tests that are expected to only run on the full image like CVE scanning, etc. those could probably be a dedicated test suite that is modular, but is only triggered for the full binary instead.

What do you think?

test/blackbox/upgrade_minimal.bats
```diff
@@ -0,0 +1,769 @@
...
+    [ $(echo "${lines[-1]}" | jq '.manifests | length') -eq 1 ]
+    run regctl artifact list --filter-artifact-type "application/vnd.example.icecream.v1" localhost:${zot_port}/artifact-ref:demo --format raw-body
+    [ "$status" -eq 0 ]
+    [ $(echo "${lines[-1]}" | jq '.manifests | length') -eq 1 ]
+    run regctl artifact list --filter-artifact-type "application/invalid" localhost:${zot_port}/artifact-ref:demo --format raw-body
+    [ "$status" -eq 0 ]
+    [ $(echo "${lines[-1]}" | jq '.manifests | length') -eq 0 ]
+}
+
+@test "[new] push docker image" {
```
vrajashkr, line 268: Will we be adding `release - upgrade - downgrade back to release` tests later too? Doesn't have to be part of this PR, but was curious.

vrajashkr (commented):
Looks good overall. This will be a huge help for verifying upgrade scenarios. I left a couple of comments.

test/blackbox/upgrade_minimal.bats
rchincha: Downgrade out of scope

test/blackbox/upgrade.bats
andaaron: I agree. Actually, if I remember correctly, in other blackbox tests we copy images from the test/data folder.

test/blackbox/upgrade.bats
```diff
@@ -0,0 +1,929 @@
...
+    run skopeo --insecure-policy copy --src-tls-verify=false --multi-arch=all \
+        docker://127.0.0.1:${zot_port}/busybox:latest \
+        oci:${oci_data_dir}/busybox:latest
+    [ "$status" -eq 0 ]
+    run cat ${BATS_FILE_TMPDIR}/oci/busybox/index.json
+    [ "$status" -eq 0 ]
+    [ $(echo "${lines[-1]}" | jq '.manifests[].annotations."org.opencontainers.image.ref.name"') = '"latest"' ]
+    run skopeo --insecure-policy --override-arch=arm64 --override-os=linux copy --src-tls-verify=false --multi-arch=all \
+        docker://127.0.0.1:${zot_port}/busybox:latest \
+        oci:${oci_data_dir}/busybox:latest
```
andaaron: Why do we have this destination folder, shouldn't we have some place which is used per test run?
I am thinking in test/data is where we have the pristine images copied from outside sources, which we later copy to zot as part if the tests.

test/blackbox/upgrade.bats
```diff
@@ -0,0 +1,929 @@
...
+    [ "$status" -eq 0 ]
+    [ $(echo "${lines[-1]}" | jq '.tags[]') = '"latest"' ]
+}
+
+@test "[release] pull image index" {
+    local oci_data_dir=${BATS_FILE_TMPDIR}/oci
+    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+    run skopeo --insecure-policy copy --src-tls-verify=false --multi-arch=all \
+        docker://127.0.0.1:${zot_port}/busybox:latest \
+        oci:${oci_data_dir}/busybox:latest
```
andaaron: Same comment on the destination folder as below.

test/blackbox/upgrade.bats
```diff
@@ -0,0 +1,929 @@
...
+}
+
+@test "[release] discover oras artifacts" {
+    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+    run oras discover --plain-http --format json 127.0.0.1:${zot_port}/golang:1.20
+    [ "$status" -eq 0 ]
+    [ $(echo "$output" | jq -r ".manifests | length") -eq 2 ]
+}
+
+@test "[release] add and list tags using oras" {
```
andaaron, line 96: I don't think we need this test as part of the upgrade suite, at least not for the "release" its' original purpose is to verify zot pagination works on tags, not upgrade related.

test/blackbox/upgrade.bats
```diff
@@ -0,0 +1,929 @@
...
+    [ "$status" -eq 0 ]
+}
+
+@test "[release] pull image with regclient" {
+    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+    run regctl image copy localhost:${zot_port}/test-regclient ocidir://${TEST_DATA_DIR}/golang:1.20
+    [ "$status" -eq 0 ]
+}
+
+@test "[release] list repositories with regclient" {
```
andaaron, line 116: I don't think need this test before upgrade.

test/blackbox/upgrade.bats
```diff
@@ -0,0 +1,929 @@
...
+    [ "${lines[-1]}" == "golang" ]
+
+    run regctl repo ls --last busybox --limit 1 localhost:${zot_port}
+    [ "$status" -eq 0 ]
+    echo "$output"
+    [ $(echo "$output" | wc -l) -eq 1 ]
+    [ "${lines[-1]}" == "golang" ]
+}
+
+@test "[release] list image tags with regclient" {
```
andaaron, line 120: This test should not be needed before upgrade. I think it's enough to test the download in other tests.

test/blackbox/helpers_upgrade.bash
```diff
@@ -0,0 +1,732 @@
...
+    [ "$status" -eq 0 ]
+    [ $(echo "${lines[-1]}" | jq '.tags[]') = '"1.20"' ]
+}
+
+function test_release_pull_image() {
+    local oci_data_dir=${BATS_FILE_TMPDIR}/oci
+    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+    run skopeo --insecure-policy copy --src-tls-verify=false \
+        docker://127.0.0.1:${zot_port}/golang:1.20 \
+        oci:${oci_data_dir}/golang:1.20
```
andaaron: Can we keep oci:${oci_data_dir} for pristine images downloaded from upstream?
What if something goes wrong with the pull and the following test attempts to push the bad image from the same layout with the same tag?

test/blackbox/helpers_upgrade.bash
```diff
@@ -0,0 +1,732 @@
...
+    [ "$status" -eq 0 ]
+    [ $(echo "${lines[-1]}" | jq '.manifests[].annotations."org.opencontainers.image.ref.name"') = '"1.20"' ]
+}
+
+function test_release_push_image_index() {
+    # --multi-arch below pushes an image index (containing many images) instead
+    # of an image manifest (single image)
+    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+    run skopeo --insecure-policy copy --format=oci --dest-tls-verify=false --multi-arch=all \
+        docker://public.ecr.aws/docker/library/busybox:latest \
```
andaaron: Let's download to a local folder before the test code, under oci:${oci_data_dir}, and in the test copy from there.

test/blackbox/helpers_upgrade.bash
```diff
@@ -0,0 +1,732 @@
...
+    [ "$status" -eq 0 ]
+    [ $(echo "${lines[-1]}" | jq '.tags[]') = '"latest"' ]
+}
+
+function test_release_pull_image_index() {
+    local oci_data_dir=${BATS_FILE_TMPDIR}/oci
+    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+    run skopeo --insecure-policy copy --src-tls-verify=false --multi-arch=all \
+        docker://127.0.0.1:${zot_port}/busybox:latest \
+        oci:${oci_data_dir}/busybox:latest
```
andaaron: Same comment as for test_release_pull_image

test/blackbox/helpers_upgrade.bash
```diff
@@ -0,0 +1,732 @@
...
+}
+
+function test_new_existing_pull_image_index() {
+    local oci_data_dir=${BATS_FILE_TMPDIR}/oci
+    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+    run skopeo --insecure-policy copy --src-tls-verify=false --multi-arch=all \
+        docker://127.0.0.1:${zot_port}/busybox:latest \
+        oci:${oci_data_dir}/busybox:latest
+    [ "$status" -eq 0 ]
+    run cat ${BATS_FILE_TMPDIR}/oci/busybox/index.json
```
andaaron: Is this under the zot root dir?

test/blackbox/helpers_upgrade.bash
```diff
@@ -0,0 +1,732 @@
...
+    # first check existing images
+    run curl http://127.0.0.1:${zot_port}/v2/_catalog
+    [ "$status" -eq 0 ]
+    [ $(echo "${lines[-1]}" | jq 'any(.repositories[]; . == "golang")') = true ] 
+    run skopeo --insecure-policy copy --dest-tls-verify=false \
+        oci:${TEST_DATA_DIR}/golang:1.20 \
+        docker://127.0.0.1:${zot_port}/golang:1.20
+    [ "$status" -eq 0 ]
+    run skopeo --insecure-policy copy --dest-tls-verify=false \
+        docker://ghcr.io/project-zot/test-images/alpine:3.17.3 \
```
andaaron, line 570: Let's download to test/data first

test/blackbox/upgrade.bats
```diff
@@ -0,0 +1,446 @@
...
+@test "[new] existing pull image" {
+    local oci_data_dir=${BATS_FILE_TMPDIR}/oci
+    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+    # first check existing images
+    run curl http://127.0.0.1:${zot_port}/v2/_catalog
+    [ "$status" -eq 0 ]
+    [ $(echo "${lines[-1]}" | jq 'any(.repositories[]; . == "golang")') = true ] 
+    run skopeo --insecure-policy copy --src-tls-verify=false \
+        docker://127.0.0.1:${zot_port}/golang:1.20 \
+        oci:${oci_data_dir}/golang:1.20
```
andaaron: Can we not overwrite the image at  oci:${oci_data_dir}/golang:1.20?

test/blackbox/helpers_upgrade.bash
```diff
@@ -0,0 +1,732 @@
...
+    [ $(echo "${lines[-1]}" | jq '.manifests | length') -eq 1 ]
+    run regctl artifact list --filter-artifact-type "application/vnd.example.icecream.v1" localhost:${zot_port}/artifact-ref:demo --format raw-body
+    [ "$status" -eq 0 ]
+    [ $(echo "${lines[-1]}" | jq '.manifests | length') -eq 1 ]
+    run regctl artifact list --filter-artifact-type "application/invalid" localhost:${zot_port}/artifact-ref:demo --format raw-body
+    [ "$status" -eq 0 ]
+    [ $(echo "${lines[-1]}" | jq '.manifests | length') -eq 0 ]
+}
+
+function test_new_push_docker_image() {
```
vrajashkr, line 725: Thanks for making the change to dedupe a bunch of the code. I see a bit more duplication in some of the test functions. For example, the code for `test_new_existing_pull_image` and `test_new_push_docker_image` appears exactly the same to me. I'm wondering if these could be deduped as well so that it is less code to maintain if we need to make changes in future. I see a few more examples of this like - `test_new_pull_oras_artifact`, etc.

I haven't checked them all, but I think it would be useful to dedupe these identical checks.

Similarly, the pull test would be done on both release and new identically so we should probably dedupe those where required.

test/blackbox/upgrade.bats
rchincha: Well, that dep is in a lot of bats tests. We should clean that up in a separate PR.

test/blackbox/upgrade_minimal.bats
rchincha: Fixed.

test/blackbox/upgrade.bats
rchincha: local oci_data_dir=${BATS_FILE_TMPDIR}/oci

test/blackbox/upgrade.bats
rchincha: Would like to keep this test though.

test/blackbox/upgrade.bats
rchincha: Would like to keep this also.

test/blackbox/upgrade.bats
rchincha: Same.

test/blackbox/upgrade.bats
rchincha: local var and store

test/blackbox/helpers_upgrade.bash
rchincha: Fixed.

test/blackbox/helpers_upgrade.bash
```diff
@@ -0,0 +1,727 @@
...
+    run regctl registry set localhost:${zot_port} --tls disabled
+    [ "$status" -eq 0 ]
+    run regctl image copy ocidir://${TEST_DATA_DIR}/golang:1.20 localhost:${zot_port}/test-regclient
+    [ "$status" -eq 0 ]
+}
+
+# Helper: Pull image with regclient
+function helper_pull_image_with_regclient() {
+    local zot_port=$(get_zot_port)
+    run regctl image copy localhost:${zot_port}/test-regclient ocidir://${TEST_DATA_DIR}/golang:1.20
```
andaaron, line 242: Nitpick: the destination folder is not test/suite specific.

PR #3624, opened 2025-12-06
feat: support mTLS-only authn/authz with AccessControl and allow combining mTLS with other auth mechanisms

andaaron:
- Support mTLS-only authn/authz with AccessControl and allow combining mTLS with other auth mechanisms
- Reject Authorization headers Basic auth method is disabled, regardless of mTLS status
- Simplify authentication logic
- Move ErrUserDataNotFound error handling into sessionAuthn method
- TLS test certificate generation using Options pattern with CertificateOptions struct for extensibility
- Update tests to reflect new authentication behavior and certificate API

### DIFF
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -425,6 +425,10 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
 
 			// Switch authentication methods based on provided request context
 			switch {
+			// Reject requests with multiple Authorization headers as a security measure
+			case hasMultipleAuthorizationHeaders(request):
+				authenticated = false
+
 			// The authorization header presence is an explicit attempt to use basic authentication
 			case !isAuthorizationHeaderEmpty(request) && authConfig.IsBasicAuthnEnabled():
 				authenticated, err = amw.basicAuthn(ctlr, userAc, response, request)
@@ -439,7 +443,10 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
 				if err != nil {
 					break
 				}
-				// the session header can be present also for anonymous calls - then we should pass
+
+				// If session authentication fails, but anonymous or management access is allowed,
+				// treat the request as authenticated. This fallback is necessary because the session
+				// header may be present for anonymous or management requests.
 				authenticated = authenticated || allowAnonymous || isMgmtRequested
 
 			// Try mTLS authentication if client certificates are present
@@ -493,6 +500,15 @@ func bearerAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 			if request.Method == http.MethodOptions {
 				next.ServeHTTP(response, request)
 				response.WriteHeader(http.StatusNoContent)
+
+				return
+			}
+
+			// Reject requests with multiple Authorization headers as a security measure
+			if hasMultipleAuthorizationHeaders(request) {
+				ctlr.Log.Error().Msg("failed to parse Authorization header: multiple Authorization headers detected")
+				response.Header().Set("Content-Type", "application/json")
+				zcommon.WriteJSON(response, http.StatusUnauthorized, apiErr.NewError(apiErr.UNSUPPORTED))
 
 				return
 			}
@@ -736,6 +752,14 @@ func isAuthorizationHeaderEmpty(request *http.Request) bool {
 	}
 
 	return false
+}
+
+// hasMultipleAuthorizationHeaders checks if the request has multiple Authorization headers.
+// This is a security concern as it could be used to bypass authentication or cause confusion.
+func hasMultipleAuthorizationHeaders(request *http.Request) bool {
+	authHeaders := request.Header.Values("Authorization")
+
+	return len(authHeaders) > 1
 }
 
 func hasSessionHeader(request *http.Request) bool {
--- a/pkg/api/authn_test.go
+++ b/pkg/api/authn_test.go
@@ -6,7 +6,9 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/base64"
 	"encoding/json"
+	"encoding/pem"
 	"errors"
 	"io/fs"
 	"net/http"
@@ -18,6 +20,7 @@ import (
 
 	"github.com/alicebob/miniredis/v2"
 	guuid "github.com/gofrs/uuid"
+	"github.com/golang-jwt/jwt/v5"
 	godigest "github.com/opencontainers/go-digest"
 	"github.com/project-zot/mockoidc"
 	. "github.com/smartystreets/goconvey/convey"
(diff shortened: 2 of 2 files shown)


pkg/api/authn.go
```diff
@@ -366,57 +419,54 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
...
-					return
-				}
+			var err error
 
-				if authenticated {
-					next.ServeHTTP(response, request)
+			// Switch authentication methods based on provided request context
+			switch {
+			// The authorization header presence is an explicit attempt to use basic authentication
+			case !isAuthorizationHeaderEmpty(request) && authConfig.IsBasicAuthnEnabled():
```
rchincha, line 433: Per HTTP spec, we are not allowed to provide mutiple authorization headers, so we should also harden that.

| You cannot send two Authorization headers per the HTTP spec

pkg/api/authn.go
```diff
@@ -100,6 +105,57 @@ func (amw *AuthnMiddleware) sessionAuthn(ctlr *Controller, userAc *reqCtx.UserAc
...
+	if len(verifiedChains) == 0 || len(verifiedChains[0]) == 0 {
+		ctlr.Log.Debug().Msg("mTLS authentication failed - user provided certificate not signed by CA")
+
+		return false, nil
+	}
+
+	// Extract identity from certificate
+	leafCert := request.TLS.PeerCertificates[0]
+
+	identity := leafCert.Subject.CommonName
```
rchincha, line 127: You are not doing this part of the verification?
https://github.com/project-zot/zot/pull/3577#issuecomment-3607093881
May be overkill but I would rather be paranoid here.

pkg/test/tls/tls.go
```diff
@@ -0,0 +1,407 @@
...
+			Province:      []string{""},
+			Locality:      []string{"San Francisco"},
+			StreetAddress: []string{""},
+			PostalCode:    []string{""},
+		}
+		template.NotBefore = time.Now()
+		template.NotAfter = time.Now().AddDate(1, 0, 0) // 1 year for client
+	default:
+		return nil, fmt.Errorf("%w: %s", ErrInvalidCertificateType, certType)
+	}
```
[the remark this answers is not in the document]
andaaron: From https://pkg.go.dev/crypto/x509#CreateCertificate:
```
If SubjectKeyId from template is empty and the template is a CA, SubjectKeyId will be generated from the hash of the public key.

If template.SerialNumber is nil, a serial number will be generated which conforms to [RFC 5280, Section 4.1.2.2](https://rfc-editor.org/rfc/rfc5280.html#section-4.1.2.2) using entropy from rand.
```

pkg/test/tls/tls.go
```diff
@@ -0,0 +1,407 @@
...
+	// If nil, no email addresses will be included.
+	EmailAddresses []string
+
+	// Hostname is the hostname or IP address for server certificates.
+	// For server certificates, this is required and will be added to DNSNames or IPAddresses
+	// based on whether it's a valid IP address or a DNS name.
+	Hostname string
+
+	// CommonName is the CommonName (CN) for client certificates.
+	// For client certificates, this is required and will be set in the Subject.
```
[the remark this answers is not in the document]
andaaron: It is needed in most cases, but we need to test certificates without a CN, so it is not always required.

pkg/api/authn.go
andaaron: amw.basicAuthn(ctlr, userAc, response, request) verifies Authorization value is basic.
The bearer middleware verifies for bearer.

Do we want to verify the client implements the HTTP spec? That seems overkill.

pkg/api/authn.go
andaaron: AFAIK that is handled internally by the http server implementation.

We use `tls.VerifyClientCertIfGiven` in the server settings. 
https://cs.opensource.google/go/go/+/master:src/crypto/tls/common.go;l=366?q=VerifyClientCertIfGiven&sq=&ss=go%2Fgo
Which verified the certificates https://cs.opensource.google/go/go/+/master:src/crypto/tls/handshake_server.go;l=963

andaaron: I just pushed some extra tests for verifying the connection fails for unknown CA and for expired certificates. No other changes were required here.

pkg/api/authn.go
rchincha: We are likely doing  request.Header["Authorization"]
If you send multiple Authorization headers, you can get them all with request.Get("Authorization)

Multiple Authorization headers are unusual and can be a security risk if your code only checks the first one — an attacker might exploit this to bypass authentication.
Best practice: Reject requests with multiple Authorization headers unless you explicitly support multiple schemes.

Here is a piece of code that protects against this.

```golang
package main

import (
	"fmt"
	"log"
	"net/http"
)

// AuthHeaderGuard is middleware that rejects requests with multiple Authorization headers
func AuthHeaderGuard(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		authHeaders := r.Header["Authorization"]

		// If more than one Authorization header is present, reject the request
		if len(authHeaders) > 1 {
			http.Error(w, "Multiple Authorization headers are not allowed", http.StatusBadRequest)
			return
		}

		// Optionally, reject if no Authorization header is present
		// if len(authHeaders) == 0 {
		//     http.Error(w, "Authorization header required", http.StatusUnauthorized)
		//     return
		// }

		// Continue to the next handler
		next.ServeHTTP(w, r)
	})
}

// Example protected handler
func protectedHandler(w http.ResponseWriter, r *http.Request) {
	fmt.Fprintln(w, "Access granted: valid Authorization header")
}

func main() {
	mux := http.NewServeMux()
	mux.Handle("/", AuthHeaderGuard(http.HandlerFunc(protectedHandler)))

	log.Println("Server running on :8080")
	if err := http.ListenAndServe(":8080", mux); err != nil {
		log.Fatal(err)
	}
}
```

```bash
curl -H "Authorization: Bearer token123" \
     -H "Authorization: Basic dXNlcjpwYXNz" \
     http://localhost:8080
```

rchincha: While the HTTP spec allows combining same-name headers with commas, Authorization is special — most authentication schemes (like Bearer or Basic) do not define a safe way to combine multiple credentials.
In practice, servers should reject multiple Authorization headers to avoid ambiguity and security risks.

Raw HTTP Request with Multiple Authorization Headers
Authorization: Bearer token123
Authorization: Basic dXNlcjpwYXNz

Combined Form (per RFC 7230 §3.2.2)
Authorization: Bearer token123, Basic dXNlcjpwYXNz

andaaron: Pushed a new commit to fix and test this.

rchincha (dismissed):

Endevir:
Lgtm, huge thanks for picking up, I'll be able to test this on my stand tomorrow, if it is needed. Still tests are good enough, so it's ok to merge

andaaron:
@Endevir 

> @andaaron
> 
> Since we previously supported having both mTLS and Basic Auth enabled — while still honoring the identity provided by Basic Auth — it seems more intuitive to treat Basic Auth as the higher-priority authentication method when both are enabled.
> 
> There are also valid (in my opinion) scenarios where a user may want to use mTLS primarily for host-based access control, ensuring that connections originate only from trusted hosts, while relying on standard authentication mechanisms (such as tokens or htpasswd) for actual user identity verification.
> 
> If a user wants to rely on the identity from a client certificate, they can simply omit the `Authorization` header. This leaves room for users who prefer mTLS as their identity source.
> 
> Here is the logic I would recommend going forward (still not completely sure about this=)):
> 
> * **If mTLS is enabled**, client certificates should be required for any authenticated request. Requests that do not use mTLS should be treated as anonymous, even if they include an `Authorization` header. This ensures mTLS remains a predictable access-control mechanism (mTLS validation usually prepends authorization).
> * **If mTLS is enabled alongside other authentication methods**, users _may override_ the identity derived from the client certificate by providing additional authentication (e.g., using Basic Auth or tokens). This allows setups like: mTLS grants read-only access by default, and users can elevate to admin by presenting credentials, without the need to generate additional certificates.
> 
> P.S. Maybe better to move discussion into #3624 ?

That makes sense. The certificate identified the machine while the Authorization header the user. As well as the argument for keeping things as they are since this is the equivalent of what we had previous release.

> * **If mTLS is enabled**, client certificates should be required for any authenticated request. Requests that do not use mTLS should be treated as anonymous, even if they include an `Authorization` header. This ensures mTLS remains a predictable access-control mechanism (mTLS validation usually prepends authorization).

If `Authorization` is sent by the client the current the implementation in the PR right now verifies it depending on Basic auth settings. And allows access accordingly even if cert authentication is not done. This should make sure bad  `Authorization` headers are rejected, and that both machines and users can authenticates independently to the same server. I think this was the reason for the PR in the first place?

> * **If mTLS is enabled alongside other authentication methods**, users _may override_ the identity derived from the client certificate by providing additional authentication (e.g., using Basic Auth or tokens). This allows setups like: mTLS grants read-only access by default, and users can elevate to admin by presenting credentials, without the need to generate additional certificates.

This should be covered by existing implementation in the PR.

Endevir:
@andaaron 

> If Authorization is sent by the client the current the implementation in the PR right now verifies it depending on Basic auth settings. And allows access accordingly even if cert authentication is not done.

You're right, it seems that having mTLS as mandatory is too strict and will break Web UI, for example. So your approach is better in this case.

> This should make sure bad Authorization headers are rejected, and that both machines and users can authenticates independently to the same server. I think this was the reason for the PR in the first place

The primary reason for me was to enable mTLS with `accessControl` policies enabled (in current release you cannot use mTLS-only auth with non-anonymous `accessControl` policies): as end-user, I want to authorize my k8s nodes by client certs with pull-only access. 

It is the only static auth (== without imagePullSecrets) method supported by configuration in `containerd`: https://github.com/containerd/containerd/blob/main/docs/hosts.md and https://github.com/containerd/containerd/blob/main/docs/hosts.md#client-field

rchincha:
There is a configuration choice to be made based on requirements of who is going to access and how.
If k8s and UI, then have to enable both mtls and basic?
If not, pick one?

rchincha:
Not a great fan of multiple auth methods because security is only as good as the weakest one, and the rest of the methods are being done for compatibility reasons, not security.

rchincha (approved):

rchincha:
@Endevir pls check if this works for you now.

Endevir:
@rchincha @andaaron 
Just tested with my scenatios - works nicely, huge thanks! :)

andaaron:
@Endevir, I also have an enhancement to customize the property used as identity, if you're willing to try it out:
https://github.com/project-zot/zot/pull/3640

PR #3640, opened 2025-12-13
feat: add configurable mTLS identity extraction with fallback chain

andaaron:
Add support for configurable identity sources in mTLS authentication, allowing identity extraction from CommonName, Subject DN, Email SAN, URI SAN, or DNSName SAN with fallback chain support. Includes regex pattern matching for URI SANs (e.g., SPIFFE workload IDs).

- Add MTLSConfig with IdentitySources, URISANPattern, and index fields
- Implement extractMTLSIdentity with fallback chain logic
- Move the mtls tests in the api package to pkg/api/mtls_test.go


examples/README.md
```diff
@@ -202,6 +202,52 @@ authentication:
...
+      }
+    }
+}
+```
+
+**Identity Sources:**
+- `CommonName` - Extract identity from the certificate's Common Name (CN) field (default)
+- `Subject` - Extract identity from the full Subject Distinguished Name (DN)
+- `Email` - Extract identity from Email SAN (Subject Alternative Name)
+- `URL` - Extract identity from URI SAN (Subject Alternative Name)
```
vrajashkr: Should we use `URI` instead of `URL` since the entity inside the certificate is a URI?

pkg/api/authn.go
```diff
@@ -975,6 +982,104 @@ func GenerateAPIKey(uuidGenerator guuid.Generator, log log.Logger,
 	return apiKey, apiKeyID.String(), err
 }
 
+// extractIdentityFromSource attempts to extract identity from a specific source.
+func extractIdentityFromSource(cert *x509.Certificate, source string, mtlsConfig *config.MTLSConfig) (string, error) {
```
vrajashkr: nit: naming. Perhaps `extractIdentityFromCertificate` might convey the intention of the function better?

pkg/api/authn.go
```diff
@@ -975,6 +982,104 @@ func GenerateAPIKey(uuidGenerator guuid.Generator, log log.Logger,
...
+	// Normalize to lowercase for case-insensitive matching
+	normalizedSource := strings.ToLower(strings.TrimSpace(source))
+
+	switch normalizedSource {
+	case "commonname", "cn":
```
vrajashkr, line 992: The short names that are matched here don't seem to be documented. Maybe we can add these to the config README as well.

pkg/api/authn.go
```diff
@@ -975,6 +982,104 @@ func GenerateAPIKey(uuidGenerator guuid.Generator, log log.Logger,
...
+		}
+
+		return cert.Subject.CommonName, nil
+
+	case "subject", "dn":
+		return cert.Subject.String(), nil
+
+	case "url", "uri":
+		if len(cert.URIs) == 0 {
+			return "", fmt.Errorf("%w", zerr.ErrNoURISANFound)
```
vrajashkr: Since there is no other error/info here, would it be possible to directly return the error instead of wrapping it again?

Similar question for Line 1022, 1032, and 1046 below.

pkg/api/authn.go
```diff
@@ -975,6 +982,104 @@ func GenerateAPIKey(uuidGenerator guuid.Generator, log log.Logger,
...
+	}
+
+	var lastErr error
+
+	for _, source := range sources {
+		identity, err := extractIdentityFromSource(cert, source, mtlsConfig)
+		if err == nil && identity != "" {
+			return identity, nil
+		}
+		lastErr = err
```
vrajashkr: question - should there be error logs in debug for the individual errors?

pkg/test/tls/tls.go
```diff
@@ -404,6 +409,27 @@ func applyOptions(template *x509.Certificate, opts *CertificateOptions, certType
...
+		for _, uriStr := range opts.URIs {
+			uri, err := url.Parse(uriStr)
+			if err != nil {
+				// Skip invalid URIs - could log error in production
+				continue
+			}
+
+			// Validate that the URI has a valid scheme (url.Parse accepts URIs without schemes)
+			if uri.Scheme == "" {
+				// Skip URIs without a scheme - could log error in production
```
vrajashkr, line 424: Just curious - in the actual code, there is no check for a scheme. Is there a check elsewhere to verify that there is a scheme?

pkg/api/authn.go
andaaron: Actually let's wrap all errors inside the same variable and pass it along

pkg/test/tls/tls.go
andaaron: We don't validate it in the certificates coming in from the client.
I was wondering if this matters.

pkg/api/authn.go
andaaron: I kept them here as we don't have the validation on the config in this PR, and people may try multiple things until one of them works. We could implement the validation and remove these alternatives.

examples/config-mtls-spiffe.json
```diff
@@ -0,0 +1,37 @@
...
+    "port": "8080",
+    "realm": "zot",
+    "tls": {
+      "cert": "test/data/server.cert",
+      "key": "test/data/server.key",
+      "cacert": "test/data/ca.crt"
+    },
+    "auth": {
+      "mtls": {
+        "identitySources": ["URI", "CommonName"],
```
rchincha: identityAttribute? similar to LDAP. Let's not invent new language.

And this is any of the list? why not just one?

andaaron: Maybe not all your certificates are built the same way for some reason.

I can think of the scenario where you have a product with a service accessing zot, and in release A of that product you have a template for generating the certificates, and in release B you change the template, but you still want to support old certificates for backwards compatibility with release A certificates.

andaaron: Made the change to `identityAttributes`

pkg/api/authn.go
andaaron: Added the command-line validation and updated the README.

examples/config-mtls-spiffe.json
rchincha: There is also an implicit order of lookup when a list.

andaaron: The order is what the user configured

Issue #3683, opened 2026-01-08
[Bug]: Can't verify retention (server is running, while it's not)


dani:
### zot version

v2.1.13

### Describe the bug

Trying to validate my retention policy, I stopped zot (running on bare metal, so systemctl stop zot). Then
```
sudo -u zot zot verify-feature retention -i 30s /etc/zot/config.yml
```
But it fails with
```
{"time":"2026-01-08T16:58:24.752304959+01:00","level":"info","message":"checking if anonymous authorization is the only type of authorization policy configured","caller":"zotregistry.dev/zot/v2/pkg/cli/server/root.go:713","func":"zotregistry.dev/zot/v2/pkg/cli/server.validateAuthzPolicies","goroutine":1}
[...]

PR #3703, opened 2026-01-13
fix:  now attempt to bind to the zot server socket to check if the server is running

andaaron:
See https://github.com/project-zot/zot/issues/3683

### DIFF
--- a/pkg/cli/server/verify_retention.go
+++ b/pkg/cli/server/verify_retention.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"net/http"
 	"os"
 	"os/signal"
 	"syscall"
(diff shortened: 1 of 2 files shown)


pkg/cli/server/verify_retention.go
```diff
@@ -208,26 +207,25 @@ func newVerifyFeatureRetentionCmd(conf *config.Config) *cobra.Command {
...
 
 		return fmt.Errorf("%s: %w", msg, err)
 	}
 
-	response, err := http.DefaultClient.Do(req)
-	if err == nil {
-		response.Body.Close()
-		logger.Warn().Err(zerr.ErrServerIsRunning).
-			Msg("server is running, in order to perform the verify-feature retention command the server should be shut down")
+	listener, err := net.ListenTCP("tcp", addr)
```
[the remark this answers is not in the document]
andaaron: I tried but I have the linter issue

```
pkg/cli/server/verify_retention.go:212:29: net.Listen must not be called. use (*net.ListenConfig).Listen (noctx)
        listener, err := net.Listen("tcp", addr)
                                   ^
```

pkg/cli/server/verify_retention.go
```diff
@@ -208,26 +207,25 @@ func newVerifyFeatureRetentionCmd(conf *config.Config) *cobra.Command {
...
+	if err != nil {
+		msg := fmt.Sprintf("failed to bind to %s (server may be running or address unavailable)", addr.String())
+		logger.Error().Err(err).Msg(msg)
 
-		return zerr.ErrServerIsRunning
+		return fmt.Errorf("%s: %w", msg, err)
 	}
 
+	// Binding succeeded, server is not running
+	listener.Close()
```
vrajashkr: nit: the Listener intf mentions that `Close()` returns an error.
https://pkg.go.dev/net#Listener
https://pkg.go.dev/net#TCPListener.Close

Usually the linter would have mentioned that the returned error is not being checked. Is this something we would want to check and log?

vrajashkr (commented):
Looks good overall. Just 1 comment.

pkg/cli/server/verify_retention.go
andaaron: I am returning the error in the latest push. Not sure how logging it separately would improve (and I don't have a way to cover that specific error with tests)

Issue #3719, opened 2026-01-20
[Bug]: constant [low] CPU usage when idle


smemsh:
### zot version

v2.1.13

### Describe the bug

When idle (no network activity, periodic actions disabled, console log tail with debug level is silent), and no users from inception, Zot nonetheless is always using CPU, instead of blocking until there is work to do.   It's not clear why it needs to do anything except sleep.  This is not a huge total amount: about 1 minutes CPU time accumulates every 3 hours, but these are constant calls with small timeouts.  What is it doing? It's keeping my system from sleeping, etc.  Shouldn't Zot just block until there's something to do?

Here is a 5 second call count (`strace -c`) of completely idle Zot process:

```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ------------------
 52.54    0.051176         168       303        99 futex
 17.79    0.017327         173       100           nanosleep
 17.00    0.016562          32       514           epoll_pwait
 11.80    0.011490           8      1369           clock_gettime
  0.32    0.000311           9        32           write
  0.26    0.000255           7        32           read
  0.15    0.000144         144         1         1 restart_syscall
  0.05    0.000053          10         5           sched_yield
  0.05    0.000046           9         5           pread64
  0.04    0.000041           8         5           sched_getaffinity
------ ----------- ----------- --------- --------- ------------------
100.00    0.097405          41      2366       100 total
```

This is happening no matter when I sample it, about the same counts.  It's looping basically doing this repeatedly, here shown as 150ms syscall snapshot:

```
[pid 455934] 05:56:23 (+     0.000000) futex(0xc00121d158, FUTEX_WAIT_PRIVATE, 0, NULL <unfinished ...>
[pid 455933] 05:56:23 (+     0.000025) futex(0x5ba53290bbf8, FUTEX_WAIT_PRIVATE, 0, NULL <unfinished ...>
[pid 455932] 05:56:23 (+     0.000007) futex(0xc00022a158, FUTEX_WAIT_PRIVATE, 0, NULL <unfinished ...>
[pid 455931] 05:56:23 (+     0.000008) futex(0xc0000e1958, FUTEX_WAIT_PRIVATE, 0, NULL <unfinished ...>
[pid 455930] 05:56:23 (+     0.000013) futex(0xc000108158, FUTEX_WAIT_PRIVATE, 0, NULL <unfinished ...>
[...]

PR #3720, opened 2026-01-20
fix: constant [low] CPU usage when idle

andaaron:
See https://github.com/project-zot/zot/issues/3719


pkg/scheduler/scheduler.go
```diff
@@ -281,7 +282,9 @@ func (scheduler *Scheduler) RunScheduler() {
 				task := scheduler.getTask()
 
 				if task == nil {
-					<-throttle
```
rchincha, line 284: is this a non-blocking channel?

andaaron: It is basically https://cs.opensource.google/go/go/+/refs/tags/go1.25.6:src/time/tick.go;l=43

PR #3733, opened 2026-01-25
DESIGN: Streaming Sync

vrajashkr:
**What type of PR is this?**
feature design

N/A - only a design

Design for streaming sync


vrajashkr:
I've disabled a bunch of workflows for this PR since it is intended to be a design discussion PR :)
Saves on extra CI runs :)

rchincha:
1. Check if the image/blob already exists?
- 1.1 If yes, then just serve it
- 1.2 If no, then send the request to a BlobStreamer
2. BlobStreamer is a writer to a temp location and many readers (clients) - pay attention to range requests
- 2.1 - clients can disconnect
- 2.2 - upstream may timeout
3. Once the blob is fully downloaded and verified, then copy it to the actual repo

vrajashkr:
regclient feature request - https://github.com/regclient/regclient/issues/1044

PR #3750, opened 2026-01-30
[proposal] remote storage refactor proposal

andaaron:

### DIFF
--- a/pkg/storage/SHARED_BLOB_STORAGE_PROPOSAL.md
+++ b/pkg/storage/SHARED_BLOB_STORAGE_PROPOSAL.md
@@ -1,0 +1,354 @@
+# Shared Blob Storage Proposal
+
+## Executive Summary
+
+This proposal redesigns zot's S3 storage to use shared blob storage where all blobs are stored in a common location (`{rootDir}/storage/blobs/{algorithm}/{digest}`) across all repositories. This eliminates deduplication logic and reduces S3 storage overhead.
+
+**Scope:** S3 storage only. Local storage continues using per-repository blob structure with hard link deduplication.
+
+## Current Architecture
+
+### Blob Path Structure
+```
+{rootDir}/{repo}/blobs/{algorithm}/{digest}
+```
+
+### Current Issues
+1. **Storage Overhead**: Duplicate blobs stored in multiple S3 locations (full copies, not hard links)
+2. **Deduplication Complexity**: Requires cache database, periodic dedupe tasks, complex logic
+3. **S3 Limitations**: Hard links don't work on S3, requiring full blob copies
+4. **Cache Synchronization**: Cache can become out of sync with storage state
+
+## Proposed Architecture
+
+### S3 Blob Path Structure
+```
+{rootDir}/storage/blobs/{algorithm}/{digest}  # Shared across all repos
+{rootDir}/{repo}/index.json                   # Repository index (unchanged)
+```
+
+### Benefits
+- ✅ Eliminates S3 deduplication logic
+- ✅ Single source of truth per blob in S3
+- ✅ Reduces S3 storage overhead significantly
+- ✅ Simplifies cache (single path per digest vs multiple paths)
+- ✅ Local storage unchanged (no migration needed)
+
+### Trade-offs
+- **GC Complexity**: Must scan all repositories to build reference map before deleting blobs
+- **Migration Required**: Existing S3 storage needs migration to new structure
+
+## Implementation
+
+### Phase 1: Core Storage Changes
+
+#### 1.1 Modify `BlobPath()` Function
+
+```go
+func (is *ImageStore) BlobPath(repo string, digest godigest.Digest) string {
+    // S3 storage (storeDriver != nil) always uses shared blob storage
+    if is.storageDriver != nil {
+        return path.Join(is.rootDir, "storage", ispec.ImageBlobsDir, 
+            digest.Algorithm().String(), digest.Encoded())
+    }
+    // Local storage (storeDriver == nil) uses per-repo structure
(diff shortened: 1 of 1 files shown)


pkg/storage/SHARED_BLOB_STORAGE_PROPOSAL.md
```diff
@@ -0,0 +1,354 @@
...
+1. **Storage Overhead**: Duplicate blobs stored in multiple S3 locations (full copies, not hard links)
+2. **Deduplication Complexity**: Requires cache database, periodic dedupe tasks, complex logic
+3. **S3 Limitations**: Hard links don't work on S3, requiring full blob copies
+4. **Cache Synchronization**: Cache can become out of sync with storage state
+
+## Proposed Architecture
+
+### S3 Blob Path Structure
+```
+{rootDir}/storage/blobs/{algorithm}/{digest}  # Shared across all repos
```
vrajashkr, line 26: I'm curious why we are using hard links for local storage, but are fine without any kind of reference counting links in S3. i.e. if the implementation logic for blobs without links is possible, then would local storage also work fine without the hardlinks?

I'm thinking of whether there any drawbacks of not using hard links in local storage.

pkg/storage/SHARED_BLOB_STORAGE_PROPOSAL.md
```diff
@@ -0,0 +1,354 @@
...
+
+#### 1.1 Modify `BlobPath()` Function
+
+```go
+func (is *ImageStore) BlobPath(repo string, digest godigest.Digest) string {
+    // S3 storage (storeDriver != nil) always uses shared blob storage
+    if is.storageDriver != nil {
+        return path.Join(is.rootDir, "storage", ispec.ImageBlobsDir, 
+            digest.Algorithm().String(), digest.Encoded())
+    }
```
vrajashkr: In newer code, we should probably not assume that a non-empty storage driver == S3. Zot may in future support perhaps a new storage driver. Since we have the interfaces in place, we should check if it is really S3 and then return the appropriate path.

pkg/storage/SHARED_BLOB_STORAGE_PROPOSAL.md
```diff
@@ -0,0 +1,354 @@
...
+    // Local storage (storeDriver == nil) uses per-repo structure
+    return path.Join(is.rootDir, repo, ispec.ImageBlobsDir, 
+        digest.Algorithm().String(), digest.Encoded())
+}
+```
+
+**Key Point:** No configuration flag needed. S3 detection is automatic via `storeDriver != nil`.
```
vrajashkr: I'm not onboard with this - if zot supports something new in future, all this code would need to be updated, rather we should check if the storage is really S3 before proceeding.

pkg/storage/SHARED_BLOB_STORAGE_PROPOSAL.md
```diff
@@ -0,0 +1,354 @@
...
+   - Verify integrity
+   - Delete per-repo copies
+
+**Migration Steps:**
+1. **Stop zot server** (prevents concurrent writes)
+2. Run dry-run: `zot migrate-s3-storage --config config.json --dry-run`
+3. Review results
+4. Run migration: `zot migrate-s3-storage --config config.json`
+5. **Restart zot** (automatically uses shared blob storage)
+6. Verify and monitor
```
vrajashkr: Since this is a destructive operation (migrating files to the new path), will the migration logic also keep track of what was moved so that it can be rolled back if required?

Out of curiosity, what would a rollback cost on S3?

pkg/storage/SHARED_BLOB_STORAGE_PROPOSAL.md
andaaron: Using hardlinks saves space if you have multiple instances of the same blobs in different paths on the same drive. If we remove hardlinks disk usage would go up significantly.

For example: you push a base image and a derived image, and, because they have different repositories, the base image layers would take twice as much disk space for the 2 different instances, once for the base image, and once for the derived image repos.

Sure, we could implement a single instance of `{rootDir}/storage/blobs/{algorithm}/{digest}` also in case of local storage, but that would break tools reading directly from the disk and expecting each repository to be valid OCI layout of its own. One such cases is the product we developed zot for in the first place.

pkg/storage/SHARED_BLOB_STORAGE_PROPOSAL.md
andaaron: Agreed. There's also the GCS PR we should be merging soon.

pkg/storage/SHARED_BLOB_STORAGE_PROPOSAL.md
andaaron: This reads S3 because that's the only option besides local disk right now. 
I would keep the existing dedupe logic only for local.
So it makes little difference what value storeDriver has as long as it is not local driver, for which we need the dedupe.
S3 and any other possible value in the future would use `{rootDir}/storage/blobs/{algorithm}/{digest}`.

pkg/storage/SHARED_BLOB_STORAGE_PROPOSAL.md
andaaron: Those are both good questions.

I think we can keep track of 1, but unless we keep it in the "cache DB", keeping it up to date after using start using the service would be practically impossible. Even if we track it in the cache DB, It complicates maintaining it.

Maybe we should have an entry there for the original blob location same as we do now, and simply use that instead of {rootDir}/storage/blobs/{algorithm}/{digest} in case of old blobs from pre-transition. I think this could potentially help cut the cost of copying stuff back in case of a rollback.

PR #3754, opened 2026-01-31
fix(meta): fixes for LastUpdated and TaggedTimestamp

andaaron:
1. Parse repos without metadata in ParseStorage

The timestamp check in ParseStorage was skipping repos that exist in
storage but don't have metadata. When GetRepoLastUpdated returns zero
time (no metadata), we should always parse the repo to create its
metadata. Check if metaLastUpdated is zero before comparing timestamps.
If zero, always parse regardless of storageLastUpdated.

2. Change the logic of how LastUpdated is computed in RepoSummary

It is not the latest tagged timestamp from the available images or
the last updated image created timestamp, based on whichever is the
latest.

### DIFF
--- a/pkg/common/model.go
+++ b/pkg/common/model.go
@@ -33,26 +33,29 @@ type PaginatedImagesResult struct {
 }
 
 type ImageSummary struct {
-	RepoName        string                    `json:"repoName"`
-	Tag             string                    `json:"tag"`
-	Digest          string                    `json:"digest"`
-	MediaType       string                    `json:"mediaType"`
-	Manifests       []ManifestSummary         `json:"manifests"`
-	Size            string                    `json:"size"`
-	DownloadCount   int                       `json:"downloadCount"`
-	LastUpdated     time.Time                 `json:"lastUpdated"`
-	Description     string                    `json:"description"`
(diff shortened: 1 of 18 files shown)


pkg/common/model.go
```diff
@@ -33,26 +33,29 @@ type PaginatedImagesResult struct {
...
-	SignatureInfo   []SignatureSummary        `json:"signatureInfo"`
+	RepoName          string                    `json:"repoName"`
+	Tag               string                    `json:"tag"`
+	Digest            string                    `json:"digest"`
+	MediaType         string                    `json:"mediaType"`
+	Manifests         []ManifestSummary         `json:"manifests"`
+	Size              string                    `json:"size"`
+	DownloadCount     int                       `json:"downloadCount"`
+	LastUpdated       time.Time                 `json:"lastUpdated"`
+	LastPullTimestamp time.Time                 `json:"lastPullTimestamp"`
```
rchincha, line 44: are we exposing all these in zui?

andaaron: They are already exposed in the API. In UI just LastUpdated and TaggedTimestamp

Issue #3755, opened 2026-01-31
[Feat]: Fine-grained expiry for JWT Bearer access entries


matheuscscp:
### Is your feature request related to a problem? Please describe.

Hi 👋 !

I was looking at the Distribution spec: https://distribution.github.io/distribution/spec/auth/jwt/

```json
    "access": [
        {
            "type": "repository",
            "name": "samalba/my-app",
            "actions": [
                "pull",
                "push"
            ]
        }
    ]
```

It would be  awesome if Zot supported an `exp` claim per access entry, i.e.:

```json
    "access": [
        {
            "exp": 1415387315,
            "type": "repository",
            "name": "samalba/my-app",
            "actions": [
                "pull",
                "push"
            ]
        }
    ]
```

My use case is: I want to build enterprise licenses for my org's products on top of OCI JWTs. If Zot supported an `exp` claim per access entry I would be able to issue licenses containing multiple products at the same time that are mapped to OCI repositories. A customer could have a contract-long product and a free-trial one in the same license.

### Describe the solution you'd like

When access is evaluated based on the JWT claims, the access entry is discard if the expiration has passed.

PR #3761, opened 2026-02-01
feat(jwt-exp): exp claim at the access entry level

matheuscscp:
Closes: #3755

### DIFF
--- a/pkg/api/bearer.go
+++ b/pkg/api/bearer.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"regexp"
 	"slices"
+	"time"
 
 	"github.com/golang-jwt/jwt/v5"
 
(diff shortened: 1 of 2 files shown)


pkg/api/bearer.go
```diff
@@ -16,9 +17,18 @@ var bearerTokenMatch = regexp.MustCompile("(?i)bearer (.*)")
...
 // specification.
 type ResourceAccess struct {
+	// Standard claims defined in the Distribution spec:
+	// https://distribution.github.io/distribution/spec/auth/jwt/
+
 	Type    string   `json:"type"`
 	Name    string   `json:"name"`
 	Actions []string `json:"actions"`
+
+	// Zot extensions
```
andaaron: What is different from the `exp` defined here? https://distribution.github.io/distribution/spec/auth/jwt/
Maybe we don't need to call out any zot specific behavior, since this is already part of the spec.

andaaron (approved):

matheuscscp:
@andaaron Should I rebase this one?

andaaron:
> @andaaron Should I rebase this one?

yes.

matheuscscp:
On it!

PR #3792, opened 2026-02-12
feat(tls): implement dynamic TLS certificate reloading with file watching

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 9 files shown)


pkg/api/controller.go
```diff
@@ -58,6 +60,17 @@ type Controller struct {
 	Healthz         *common.Healthz
 	// runtime params
 	chosenPort int // kernel-chosen port
+	// TLS certificate management
+	TlsCertMutex   sync.RWMutex
+	TlsCert        *tls.Certificate
```
andaaron: Why not have a separate object responsible for certificate handling, same as in https://github.com/project-zot/zot/pull/3752/changes#diff-672be0187484cee0bc614f3f370d1fe732434bcc65e25c9ffd89a1abf6cd43c1R209?

rchincha: Fixed

andaaron (approved):

PR #3824, opened 2026-02-27
Skip OCI conversion for already-synced images

### LINKED ISSUE #3823: [Bug]: OCI conversion attempted causes misleading errors
### zot version

v2.1.14

### Describe the bug

When on-demand sync skips an already-synced image, `syncImage` still attempts OCI conversion on a non-existent temp directory, producing confusing error logs.

```
{"message":"skipping image because it's already synced","image":"docker.io/company-name/company-project:3f0f056..."}
{"message":"failed to convert docker image to oci","error":"unable to read oci index: oci-layout cannot be open: open /var/lib/registry/company-name/company-project/.sync/c262dae1-b723-4f03-a99f-c33cb5e49474/company-name/company-project/oci-layout: no such file or directory"}
{"message":"failed to sync image","error":"unable to read oci index: ..."}
```

But the image is served correctly (HTTP 200) because it exists in local storage. The error is logged but has no impact.

I am wondering if logs could be clearer.

I was able to remove the error by setting https://zotregistry.dev/v2.1.3/admin-guide/admin-configuration/?h=compat#linting-uploaded-images
```json
"http": {
    "address": "127.0.0.1",
    "port": "8080",
    "compat": ["docker2s2"]
}
```
And `preserveDigest: true`.

### To reproduce

1. Configuration
- Zot configured with on-demand sync to Docker Hub
- `preserveDigest` set to `false` or omitted (default)
- Remote image uses Docker manifest format not OCI
3. Client tool used
containerd
5. Seen error
```
{"message":"failed to convert docker image to oci","error":"unable to read oci index: oci-layout cannot be open: open /var/lib/registry/company-name/company-project/.sync/c262dae1-b723-4f03-a99f-c33cb5e49474/company-name/company-project/oci-layout: no such file or directory"}
```

### Expected behavior

No error or actionable error.

### Screenshots

_No response_

### Additional context

Probably dup of https://github.com/project-zot/zot/issues/3295

utafrali:
Have syncRef return a boolean flag to indicate if an image was skipped. The caller can then return early without attempting OCI conversion, which was causing misleading errors. Fixes #3823.

### DIFF
--- a/pkg/extensions/sync/service.go
+++ b/pkg/extensions/sync/service.go
@@ -622,16 +622,12 @@ func (service *BaseService) syncImage(ctx context.Context, localRepo, remoteRepo
 		return err
 	}
 
-	if skipped {
-		return nil
-	}
-
 	if withReferrers {
 		_ = service.syncReferrers(ctx, repoTags, localRepo, remoteRepo, localImageRef, remoteImageRef)
 	}
 
 	// convert image to oci if needed
-	if isConverted && !service.config.PreserveDigest {
+	if !skipped && isConverted && !service.config.PreserveDigest {
 		localImageRef, err = mod.Apply(ctx, service.rc, localImageRef,
 			mod.WithRefTgt(localImageRef),
 			mod.WithManifestToOCI(),
--- a/pkg/extensions/sync/sync_internal_test.go
+++ b/pkg/extensions/sync/sync_internal_test.go
@@ -127,7 +127,7 @@ func TestService(t *testing.T) {
 		So(panicOccurred, ShouldBeTrue)
 	})
 
-	Convey("test syncImage skips OCI conversion when image already synced", t, func() {
+	Convey("test syncImage skips OCI conversion but still commits when image already synced", t, func() {
 		conf := syncconf.RegistryConfig{
 			URLs: []string{"http://localhost"},
 		}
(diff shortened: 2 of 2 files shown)


pkg/extensions/sync/service.go
```diff
@@ -617,11 +617,15 @@ func (service *BaseService) syncImage(ctx context.Context, localRepo, remoteRepo
...
 	if err != nil {
 		return err
 	}
 
+	if skipped {
+		return nil
+	}
+
 	if withReferrers {
 		_ = service.syncReferrers(ctx, repoTags, localRepo, remoteRepo, localImageRef, remoteImageRef)
```
andaaron, line 626: What if someone pushed a new referrer to the upstream registry which was not present when the subject image was initially synced?

I am not sure about skipping syncReferrers execution (and service.destination.CommitAll(localRepo, localImageRef) afterwards).

utafrali: Good catch, you're absolutely right. New referrers can be added to the upstream registry after the initial sync, so skipping `syncReferrers` would miss them.

I've updated the PR to fix this:
- Removed the early `return nil` when `skipped` is true
- Instead, only the OCI conversion block is guarded by `!skipped` (since re-converting an already-stored image is both unnecessary and incorrect)
- `syncReferrers` and `CommitAll` now always execute, so new upstream referrers are picked up on subsequent syncs

Updated the test accordingly — `commitAllCalled` is now asserted as `true` for the skipped-image case.

andaaron (approved):

pkg/extensions/sync/service.go
andaaron: Thank you.

Issue #3839, opened 2026-03-03
[Feat]: provide flags to reduce overhead of referrer sync


jzhn:
### Is your feature request related to a problem? Please describe.

Currently, `GET /referrers` endpoint at Zot is extremely slow. 

```bash
time regctl artifact tree <upstream>/image@sha256:aaa
regctl artifact tree   0.02s user 0.03s system 3% cpu 1.359 total

time regctl artifact tree <zot>/image@sha256:aaa
regctl artifact tree   0.01s user 0.02s system 0% cpu 14.348 total
```

for a simple tree of OCI image 
```yaml
Digest: sha256:aaa
Children:
  - sha256:bbb [linux/amd64]
    Referrers:
      - sha256:ccc: sha256-bbb.att
      - sha256:ddd: sha256-bbb.sig
  - sha256:eee [linux/arm64]
    Referrers:
      - sha256:fff: sha256-eee.att
      - sha256:ggg: sha256-eee.sig
Referrers:
  - sha256:hhh: sha256-aaa.sig
```

While it’s understandable that OCI 1.1 referrers API is designed to be dynamic so it offsets the benefit of a pull through cache, looking at Zot’s code, there are a few things that seriously slows down Zot:

- It recursively syncs the tree of all referrers, and the referrers or referrers…
- It also fetches all tags in attempt to the discover old cosign compatibility tags. if matched, the tree of those are also synced. 
- there’s no parallelism when syncing multiple trees discovered from above.

### Describe the solution you'd like

adding feature flags per sync repostory, to tune down the aggressiveness of referrer sync.

- conditionally not to sync the cosign compatibility tags - those can be synced later when the client requests for the manifests. 
- conditionally not to sync recursively - only sync the root-level referrers.

### Describe alternatives you've considered

- consider the referrer API so dynamic that we just always proxy the call. this would solve the problem nicely - we still get the nice caching benefits of content-addressable objects, but leave the dynamic objects simply proxied.

Upon discussing with Zot community at Slack, it is determined that it's unsafe because there could be referrers being pushed to Zot referring a mirrored manifest/index. https://cloud-native.slack.com/archives/C03EGRE4QGH/p1772485455490499

andaaron:
Fixed in https://github.com/project-zot/zot/pull/3842

Issue #3808, opened 2026-02-21
[Feat]: Support platform filtering on index sync


jzhn:
### Is your feature request related to a problem? Please describe.

When configured with Sync extention, Zot currently always synchornize the entire tree of OCI index -> manifests -> blobs. While this behaviour is a good default, it wastes resources in enviroments where certain artifacts are not used. For example, while a typical Kubernetes environment might only have `linux/amd64` and `linux/arm64` worker nodes, but Zot would still synchronize everything at once for an Ubuntu image that comes with 6 platforms, and it takes a long time.

```
crane manifest ubuntu:noble | jq '.manifests[].platform'
{
  "architecture": "amd64",
  "os": "linux"
}
{
  "architecture": "unknown",
  "os": "unknown"
}
{
  "architecture": "arm",
  "os": "linux",
  "variant": "v7"
}
{
  "architecture": "unknown",
  "os": "unknown"
}
{
  "architecture": "arm64",
  "os": "linux",
  "variant": "v8"
}
{
  "architecture": "unknown",
  "os": "unknown"
}
{
  "architecture": "ppc64le",
  "os": "linux"
}
{
  "architecture": "unknown",
  "os": "unknown"
}
{
  "architecture": "riscv64",
  "os": "linux"
}
{
  "architecture": "unknown",
  "os": "unknown"
}
{
  "architecture": "s390x",
  "os": "linux"
}
{
  "architecture": "unknown",
  "os": "unknown"
}
```

### Describe the solution you'd like

Implement a filter at `extensions.sync.registries[].content[]` level, that optionally accepts an allow-list of platforms. For example...

```json
{
  "extensions": {
    "sync": {
      "registries": [
        {
          "content": [
            {
              "platforms": [
                "linux/amd64",
                "linux/arm64"
              ],
              "prefix": "**"
            }
          ]
        }
      ]
    }
  }
}
```

when it's configured:

- sync would filter out manifests and their blobs with mismatched platforms
- index remain unmodified - it'll still show the clients that the index contains manifests that have been filtered out, but it's safe to assume that clients in the environment will not request for manifests or blobs of the filtered platform.

### Describe alternatives you've considered

N/A. It seems that Zot currently has no workaround to reduce the overhead of unnecessary syncs.

### Additional context

- Other filters, like `mediaType`, and `artifactType`, can be considered as well. But the current pain point is platform.

rchincha:
https://github.com/project-zot/zot/pull/3128

[...]

PR #3840, opened 2026-03-03
feat(sync): allow optionally skip digest-tags and recursive referrer syncs

jzhn:
**What type of PR is this?**
feature

https://github.com/project-zot/zot/issues/3839

Add two boolean flags at `sync.RegistryConfig` to make referrer syncs lighter weight.
- `SkipTagBasedReferrerSync`: disable syncing pre-OCI 1.1 digest-tags that cosign produces
- `SkipRecursiveReferrerSync`: disable syncing referrers recursively.

I've built and deployed it internally.
- functionally I didn't notice any issue
- observed >=35% latency improvement on `GET /referrers` related operations for a typical index __on both cold and warm requests__.

```yaml
Digest: sha256:aaa
Children:
  - sha256:bbb [linux/amd64]
    Referrers:
      - sha256:ccc: sha256-bbb.att
      - sha256:ddd: sha256-bbb.sig
  - sha256:eee [linux/arm64]
    Referrers:
      - sha256:fff: sha256-eee.att
      - sha256:ggg: sha256-eee.sig
Referrers:
  - sha256:hhh: sha256-aaa.sig
```

**Automation added to e2e**:

TBD

No. Change is backwards compatible. If these two new flags aren't provided, the behaviour is same as before.

Yes, TBD

```release-note

```

### DIFF
--- a/pkg/extensions/config/sync/config.go
+++ b/pkg/extensions/config/sync/config.go
@@ -36,6 +36,17 @@ type RegistryConfig struct {
 	PreserveDigest        bool          // sync without converting
 	SyncTimeout           time.Duration // overall HTTP client timeout for all sync operations
 	ResponseHeaderTimeout time.Duration `yaml:"-"` // response header timeout; set in root.go
+	// SkipTagBasedReferrerSync disables syncing referrers that are stored as digest-encoded tags
+	// (e.g. sha256-<hex>.sig, sha256-<hex>.att, sha256-<hex>.sbom used by legacy cosign tooling).
+	// When true, only OCI-spec referrers discovered via the Referrers API are synced.
+	// Eliminates the expensive tag-listing round-trip on every on-demand referrer request.
+	// Default false to preserve backwards-compatible behaviour.
+	SkipTagBasedReferrerSync bool
+	// SkipRecursiveReferrerSync disables recursive traversal of referrer graphs.
+	// When true, only the direct referrers of the queried digest are synced; referrers-of-referrers
+	// are synced lazily when explicitly requested by a client.
+	// Default false to preserve backwards-compatible behaviour.
+	SkipRecursiveReferrerSync bool
 }
 
 type Content struct {
--- a/pkg/extensions/sync/service.go
+++ b/pkg/extensions/sync/service.go
@@ -343,12 +343,21 @@ func (service *BaseService) SyncReferrers(ctx context.Context, repo string,
 	service.log.Info().Str("remote", remoteURL).Str("repository", repo).Str("subject", subjectDigestStr).
 		Interface("reference types", referenceTypes).Msg("syncing reference for image")
 
-	tags, err := service.getTags(ctx, remoteRepo, false)
-	if err != nil {
-		service.log.Error().Str("errorType", common.TypeOf(err)).Str("repo", repo).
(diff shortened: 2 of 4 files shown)


pkg/extensions/config/sync/config.go
```diff
@@ -36,6 +36,17 @@ type RegistryConfig struct {
 	PreserveDigest        bool          // sync without converting
 	SyncTimeout           time.Duration // overall HTTP client timeout for all sync operations
 	ResponseHeaderTimeout time.Duration `yaml:"-"` // response header timeout; set in root.go
+	// SkipTagBasedReferrerSync disables syncing referrers that are stored as digest-encoded tags
+	// (e.g. sha256-<hex>.sig, sha256-<hex>.att, sha256-<hex>.sbom used by legacy cosign tooling).
+	// When true, only OCI-spec referrers discovered via the Referrers API are synced.
+	// Eliminates the expensive tag-listing round-trip on every on-demand referrer request.
+	// Default false to preserve backwards-compatible behaviour.
+	SkipTagBasedReferrerSync bool
```
rchincha, line 44: Wondering if we should just combine these two into "shallow": true

jzhn: Yea. I can merge these two into a single flag. But `shallow` might be too generic, because in the future we might want to shallow syncing other parts, like index. And there are other initiatives like "sparse index" https://github.com/project-zot/zot/issues/3808 which might be confused with `shallow`.

Maybe we combine these two into `ShallowReferrerSync`?

andaaron: IMHO we should keep them separate.

1. SkipTagBasedReferrerSync seems like an option enabled in almost all cases, since even cosign has a newer better signature format using referrers. This would be false just for people consuming legacy images.
2. SkipRecursiveReferrerSync (or more simple `SkipReferrersSync`?) would be disabled in most cases if people want to sync referrers.

andaaron:
I think I misunderstood something in the request/implementation.

There are several items we could tweak.
1. Optionally turn off referrer sync entirely (not just recursively). (The PR seems to allow the first level of referrers sync, and only others are skipped)
2. Optionally disable sync-ing tag-based signatures. (I am not convinced the PR implements this correctly, I need to check carefully)

jzhn:
The focus of this MR is to reduce the latency of `GET /referrers` call, by reducing the overhead of onDemand sync of referrers, while preserving the referrer functionatlity to the end-client.

> There are several items we could tweak.
> 
> 1. Optionally turn off referrer sync entirely (not just recursively). (The PR seems to allow the first level of referrers sync, and only others are skipped)

My intention isn't to disable referrer sync entierly because my client indeed needs the information synced from the upstream. It's Kyverno admission controller calling Zot to verify cosign signatures. Since it's in synchronous path of pod creation (admisstion controller), latency is sensitive. In my test, by disabling both legacy digest-tag and recursive referrer syncs, the `GET /referrers` endpoint latency is cut in half. 

> 2. Optionally disable sync-ing tag-based signatures. (I am not convinced the PR implements this correctly, I need to check carefully)

Please let me know which part is missing, in my implemented unit tests and manual tests at my org, it seems to be working.

jzhn:
replaced by https://github.com/project-zot/zot/pull/3842

PR #3844, opened 2026-03-06
fix(auth): prevent open redirect via callback_ui

andaaron:
Validate callback_ui and default invalid values to /. Allow absolute callback_ui only when its origin is allowlisted via http.auth.openid.callbackAllowOrigins (and externalUrl). Add/adjust unit + controller tests and update examples/docs for relative vs allowlisted absolute redirect

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -332,10 +332,26 @@ zot can be configured to use the above providers with:
   }
 ```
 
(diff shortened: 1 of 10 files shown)


andaaron:
Thanks for reporting the issue @route2shell

PR #3866, opened 2026-03-13
fix(build): update Makefile for cross-platform PIE support and static linking

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,8 @@ export GOEXPERIMENT=jsonv2
 SHELL := /bin/bash
 TOP_LEVEL=$(shell git rev-parse --show-toplevel)
 COMMIT_HASH=$(shell git describe --always --tags --long)
-RELEASE_TAG=$(shell git describe --tags --abbrev=0)
+# Repos without annotated/lightweight tags should not fail metadata resolution.
+RELEASE_TAG=$(shell git describe --tags --abbrev=0 2>/dev/null)
 GO_VERSION=$(shell go version | awk '{print $$3}')
 COMMIT ?= $(if $(shell git status --porcelain --untracked-files=no),$(COMMIT_HASH)-dirty,$(COMMIT_HASH))
 CONTAINER_RUNTIME := $(shell command -v podman 2> /dev/null || echo docker)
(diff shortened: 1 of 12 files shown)


rchincha:
@dfr your thoughts about this PR, especially the FreeBSD part?

rchincha:
@muscariello your thoughts on this PR, especially for hardening?

build/stacker-conformance.yaml
```diff
@@ -47,8 +47,7 @@ build:
   os: ${{OS}}
   arch: ${{ARCH}}
   from:
-    type: docker
-    url: docker://${{BASE_IMAGE}}
+    type: ${{BASE_IMAGE}}
```
andaaron, line 50: Having type without 'url' basically means the only working value would be 'scratch', which makes having this as a variable misleading. We should either:
a) this is refactored to accept a url as well
b) hardcode scratch

Makefile
```diff
@@ -70,13 +70,37 @@ BENCH_OUTPUT ?= stdout
 ALL_EXTENSIONS = debug,imagetrust,lint,metrics,mgmt,profile,scrub,search,sync,ui,userprefs,events
 EXTENSIONS ?= sync,search,scrub,metrics,lint,ui,mgmt,profile,userprefs,imagetrust,events
 UI_DEPENDENCIES := search,mgmt,userprefs
-# freebsd is not supported for pie builds if CGO is disabled
+# Linux: fully static PIE via Go's internal linker (CGO_ENABLED=0 + -buildmode=pie)
```
andaaron: I am not sure this is correct.

From my understanding, the choices for Linux are:
1. CGO_ENABLED=0 + -buildmode=pie -> statically linked + dynamically loaded, what we have right now, this should not work with scratch images.
2. CGO_ENABLED=1 + -buildmode=pie + -ldflags '-linkmode external -extldflags "-static -static-pie"' -> statically linked + statically loaded, what would work for scratch.

Makefile
```diff
@@ -70,13 +70,37 @@ BENCH_OUTPUT ?= stdout
...
+#         No external C compiler needed; the binary has no C lib deps so it is
+#         naturally static-PIE.  linux/amd64 and other tier-1 arches support
+#         internal-linker PIE — see InternalLinkPIESupported in platform/supported.go.
```
andaaron: If this is correct, why don't the current binaries work per report in https://github.com/project-zot/zot/discussions/3862?
Do we need to use an additional flag? Does this work for cross compilation as well?

From my understanding, the choices for Linux are:
1. CGO_ENABLED=0 + -buildmode=pie -> statically linked + dynamically loaded, what we have right now, and which is reported to not work for scratch images.
2. CGO_ENABLED=1 + -buildmode=pie + -ldflags '-linkmode external -extldflags "-static -static-pie"' -> statically linked + statically loaded, what would work for scratch.

muscariello:
@rchincha a priori scratch is emptier than the distroless currently being used so it is supposed to be better. Let me do a manual scan.

Issue #3538, opened 2025-11-13
[Bug]: docker push error unauthorized: authentication required


imaxun:
### zot version

v2.1.10

### Describe the bug

docker push error unauthorized: authentication required，Below is my configuration.

Based on the zot logs, it appears that authentication information was not included. However, pushing to the official Docker repository worked normally.

zot config：
```
http:
  address: 0.0.0.0
  port: 5000
  realm: zot
  auth: 
    htpasswd:
      path: "/etc/zot/auth/htpasswd" 
  accessControl:
    repositories:
      "**":
        anonymousPolicy: ["read"]
        policies:
          - users: ["*"]
            actions: ["read", "create", "update", "delete"]
    adminPolicy:
      users: ["admin"]
      actions: ["read", "create", "update", "delete"] 
``` 

 docker auth config：
```
.docker/config.json 

{
        "auths": {
                "zot.xxxx.com": {
                        "auth": "xxxxxxx"
                }
        }
}
``` 

zot logs
```
{
    "time": "2025-11-13T07:21:43.764554031Z",
    "level": "info",
    "message": "HTTP API",
    "module": "http",
    "component": "session",
    "clientIP": "10.44.1.1:27175",
    "method": "POST",
    "path": "/v2/nginx/blobs/uploads/",
    "statusCode": 401,
    "latency": "0s",
    "bodySize": 266,
    "headers": {
        "Accept-Encoding": [
            "gzip"
        ],
        "Content-Length": [
            "0"
        ],
        "User-Agent": [
            "docker/28.4.0 go/go1.24.7 git-commit/249d679 kernel/6.8.4-2-pve os/linux arch/amd64 UpstreamClient(Docker-Client/28.4.0 \\(linux\\))"
        ],
        "X-Forwarded-For": [
            "192.168.31.66"
        ],
        "X-Forwarded-Host": [
            "zot.xxxx.com"
        ],
        "X-Forwarded-Port": [
            "443"
        ],
        "X-Forwarded-Proto": [
            "https"
        ],
        "X-Forwarded-Scheme": [
            "https"
        ],
        "X-Real-Ip": [
            "192.168.31.66"
        ],
        "X-Request-Id": [
            "4fc3e6cccbf34bd54a97cf06e7130396"
        ],
        "X-Scheme": [
            "https"
        ]
    },
    "goroutine": 2251,
    "caller": "zotregistry.dev/zot/v2/pkg/api/session.go:137",
[...]

Issue #2928, opened 2025-01-30
[Bug]: Docker pull fails to pull from restricted path if anonymous access allowed on the other path


vooon:
### zot version

2.1.3-rc1 (master)

### Describe the bug

So, there a some base images, that i want any clients to be able to pull, but rest of the images - only for logged users.
Without login I'm able to pull `repo.tld/base/foo:bar`, but not `repo.tld/service/baz:latest`, that's expected.
Once i've done `podman login repo.tld` and `docker login repo.tld`, i can `podman pull  repo.tld/service/baz:latest`, but not the docker one!
```
# docker --debug -l debug pull repo.tld/service/baz:latest
time="2025-01-30T16:00:03Z" level=debug msg="otel error" error="1 errors occurred detecting resource:\n\t* conflicting Schema URL: https://opentelemetry.io/schemas/1.21.0 and https://opentelemetry.io/schemas/1.26.0"
Error response from daemon: unauthorized: authentication required
```

From the Zot log i see that it does:
1. `GET /v2/`, without auth, 200 OK
2.  `HEAD /v2/service/baz/manifests/latest`, 401
3. end

If there no anonymous access allowed flow:
1. `GET /v2/`, - 401
2.  `HEAD /v2/service/baz/manifests/latest`, with auth - 200 OK
3. loading layers ...

### To reproduce

1. Configuration

```yaml
http:
  address: 0.0.0.0
  port: 443
  tls:
    cert: /etc/zot/tls.crt
    key: /etc/zot/tls.key
  auth:
    accessControl:
      repositories:
        "**": {defaultPolicy: [read]}
        "base/**": {anonymousPolicy: [read]}
```

2. Client tool used: `docker/27.5.1 go/go1.22.11 git-commit/4c9b3b0`

### Expected behavior

Docker pulls protected images when there also anonymous access allowed for some other path.

andaaron:
You need a default policy for "base/**" or else the logged in users won't have access.
Access for anonymous users users doesn't affect logged in users. So if anonymous users have access it doesn't mean logged in users would, it's a separate setting.

vooon:
Oh, okay, will check tomorrow.

vooon:
@andaaron does not work. The same issue as reported:
```
$ docker logout zot.dev.tld
$ docker pull zot.dev.tld/base/almalinux9-base:latest
latest: Pulling from base/almalinux9-base
493137e25e96: Pull complete 
6dcd0140aae1: Pull complete 
[...]

PR #3867, opened 2026-03-13
fix(authn): always challenge unauthenticated requests on /v2/ ping

M0Rf30:
When any repository has anonymousPolicy set, allowAnonymous was true globally, causing the /v2/ ping endpoint to return 200 for unauthenticated clients. Docker interprets this 200 as 'no auth needed' and skips setting up auth state, so subsequent requests to protected endpoints arrive without credentials and get 401 with no retry.

Fix by excluding the /v2/ ping path from the allowAnonymous short-circuit, so it always returns 401 + WWW-Authenticate to unauthenticated clients, forcing Docker to establish proper challenge-response auth state.

Fixes: https://github.com/project-zot/zot/issues/3538

### DIFF
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -412,7 +412,20 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
 
 			// Get access control config safely
 			accessControlConfig := ctlr.Config.CopyAccessControlConfig()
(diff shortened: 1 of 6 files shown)


andaaron:
The issue with this fix is it forces other client implementations to authenticate as well.  See the failures in https://github.com/project-zot/zot/actions/runs/23061133947/job/66987724064?pr=3867.

For docker having anonymous access means having a token with anonymous access permissions. For others, it means not providing a token at all. 

This has been discussed in https://github.com/project-zot/zot/issues/2928

PR #3868, opened 2026-03-14
fix(auth): add workaround for Docker client auth with mixed anonymous policies

andaaron:
Docker client fails to authenticate to protected repositories when basic auth (htpasswd/LDAP) is used with mixed access policies (some repos anonymous, some requiring auth). This happens because Docker determines whether to send credentials based on the /v2/ response - if it returns 200, Docker assumes no auth is needed anywhere.

Add `forceDockerClientAuth` config option that, when enabled, forces 401 on /v2/ for Docker clients, triggering Docker's authentication flow.

This workaround only affects Docker clients (detected via User-Agent). Podman and other OCI-compliant clients are unaffected.

Refs:
- https://github.com/project-zot/zot/issues/3538
- https://github.com/project-zot/zot/issues/2928
- https://github.com/opencontainers/wg-auth/blob/main/docs/implementations/moby.md

The documentation for this is tracked in https://github.com/project-zot/project-zot.github.io/pull/253


pkg/api/controller_test.go
```diff
@@ -14030,3 +14030,109 @@ func TestDynamicTLSCertificateReloading(t *testing.T) {
...
+
+		htpasswdUsername, seedUser := test.GenerateRandomString()
+		htpasswdPassword, seedPass := test.GenerateRandomString()
+		htpasswdPath := test.MakeHtpasswdFileFromString(t, test.GetBcryptCredString(htpasswdUsername, htpasswdPassword))
+
+		Convey("With ForceDockerClientAuth enabled", func() {
+			conf := config.New()
+			conf.HTTP.Port = port
+			conf.HTTP.Auth = &config.AuthConfig{
+				HTPasswd: config.AuthHTPasswd{
```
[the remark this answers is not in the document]
andaaron: The initial convey executes multiple times, once per child convey, this should pick different ports per convey.

pkg/api/config/config.go
```diff
@@ -104,18 +104,19 @@ type AuthHTPasswd struct {
...
+	Bearer                *BearerConfig
+	OpenID                *OpenIDConfig
+	APIKey                bool
+	SessionKeysFile       string
+	SessionHashKey        []byte         `json:"-"`
+	SessionEncryptKey     []byte         `json:"-"`
+	SessionDriver         map[string]any `mapstructure:",omitempty"`
+	SecureSession         *bool          `json:"secureSession,omitempty"         mapstructure:"secureSession,omitempty"`
+	MTLS                  *MTLSConfig    `json:"mtls,omitempty"                  mapstructure:"mtls,omitempty"`
+	ForceDockerClientAuth bool           `json:"forceDockerClientAuth,omitempty" mapstructure:"forceDockerClientAuth,omitempty"` //nolint:lll
```
rchincha: https://github.com/project-zot/zot/blob/main/examples/config-docker-compat.json#L9

Does it make sense to just use this compat config? and not invent a new one because this already means users want docker behavior?

andaaron: No from multiple perspectives:

1. Docker client vs Docker mediatypes. 
https://github.com/project-zot/zot/blob/main/examples/config-docker-compat.json#L9 refers to allowing docker mediatypes, it is not related to a specific client. 
Enabling that configuration can be done with no additional breakage risks for every deployment. IMHO we should make it default. 

2. Breaking changes.
This changes the expectations we have for docker clients.
This does not "fix" a behavior, it changes it. And some people may not want this change.
The reason is if you want to use anonymous auth with this bool enabled, your docker client MUST login even for images you allow access to anonymously. 

With this setting you need to choose what behaviour you want for docker client:

A. Existing behavior: anonymously access images are accessible and authenticated access images are not available. Even if you previously did docker login.
B. New behavior: you need to login to docker to see both anonymously available and authenticated available images. Theoretically if you don't docker login you don't see anything, including anonymously available images.

andaaron:
There's also the alternative mentioned in https://github.com/project-zot/zot/issues/3538#issuecomment-3898958958

This is a comparison I've made with AI

## zot-docker-proxy approach

**Architecture:** External reverse proxy in front of Zot.

**Mechanism:**

1. Detects Docker clients with `strings.HasPrefix(User-Agent, "docker/")`.
2. For `GET /v2` or `GET /v2/` with no `Authorization` header, responds with **401** and  
   `WWW-Authenticate: Bearer realm="<proxy-base>/docker-token"`.
3. Docker follows the Bearer flow and calls `/docker-token`.
4. The proxy returns a short-lived, HMAC-signed token (see `internal/tokenforge`).
5. For later `/v2/...` requests carrying `Authorization: Bearer <token>`:
   - If the token verifies as a proxy-issued token, the proxy **removes** `Authorization` before forwarding (anonymous to Zot).
   - Otherwise it may rewrite `Bearer` payloads to `Basic` for upstream htpasswd-style auth.

**Path matching:** Treats both `/v2` and `/v2/` as the registry root ping.

## Comparison with native Zot (`forceDockerClientAuth`)

| Aspect | zot-docker-proxy | Native Zot (`forceDockerClientAuth`) |
|--------|------------------|--------------------------------------|
| **Where it runs** | Separate process / sidecar | Inside Zot |
| **Challenge style** | Bearer + custom token endpoint | Basic realm on `/v2` |
| **Anonymous Docker pulls** | **Preserved** (proxy token stripped upstream) | **Not preserved**; Docker must log in for `/v2` challenge path |
| **Operational cost** | Extra service, secret, URLs (`zot-url`, `my-url`) | Config flag only |
| **Docker detection** | `User-Agent` prefix `docker/` | `User-Agent` contains `Docker-Client` |
| **Root path** | `/v2` or `/v2/` | Path normalized to `/v2` (e.g. trailing slash stripped) |

## Conclusions

1. **Same underlying problem:** Docker uses the `/v2` ping to decide whether to send stored credentials; a **200** on that path with anonymous policies can prevent credentials from being sent to protected repos.

2. **Different trade-offs:** The proxy **emulates a token service** so Docker still gets a 401 + Bearer challenge while Zot can see anonymous traffic after the proxy strips the synthetic token. Native Zot instead forces a **Basic** challenge on the registry root for identified Docker clients, which fixes protected pulls but **breaks anonymous Docker** access to the registry when the option is on.

3. **Both are valid:** Use **zot-docker-proxy** when anonymous Docker access must keep working without logging in. Use **`forceDockerClientAuth`** when you want a single binary and can require `docker login` for Docker users.

4. **User-Agent nuance:** `docker/` vs `Docker-Client` are different heuristics; either can miss or misclassify unusual clients. Align expectations with your actual client population.

ravi-ciroos:
@andaaron @rchincha - tested this patch and verified that it works correctly for docker agent and does not break skopeo and other clients.

pkg/api/config/config.go
ravi-ciroos: @andaaron - we are already looking at the user agent to see if its docker client or not. I think the login behavior is now inline with the docker client expectations, so IMO we can let go of the additional configuration knob.

ravi-ciroos: How about something like this:

```
hasMixedPolicy := allowAnonymous && !accessControlConfig.ContainsOnlyAnonymousPolicy()
if isDockerClient && isV2Requested && hasMixedPolicy {
      authenticated = false
} else {
      authenticated = true
}
```

Then if its docker client, and v2 is requested path and the repo not fully anonymous we return auth required to satisfy the needs of docker client.

ravi-ciroos (commented):
Works. Verified with Anonymous, Mixed and Fully Authenticated configurations against docker client && skopeo.

pkg/api/config/config.go
andaaron: Ok, that makes sense. I will update the PR.

andaaron:
@ravi-ciroos, I made the change you suggested. Do you want to test again before we merge?

pkg/api/controller_test.go
```diff
@@ -14202,3 +14202,112 @@ func TestDynamicTLSCertificateReloading(t *testing.T) {
...
+			defer cm.StopServer()
+
+			// Docker client without credentials should get 401
+			resp, err := resty.R().
+				SetHeader("User-Agent", "docker/26.1.3 go/go1.22.2 UpstreamClient(Docker-Client/26.1.3 (linux))").
+				Get(baseURL + "/v2/")
+			So(err, ShouldBeNil)
+			So(resp, ShouldNotBeNil)
+			So(resp.StatusCode(), ShouldEqual, http.StatusUnauthorized)
+			So(resp.Header().Get("WWW-Authenticate"), ShouldContainSubstring, "Basic realm=")
```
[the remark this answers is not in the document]
andaaron: Not relevant

pkg/api/authn.go
```diff
@@ -466,7 +468,16 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
...
+				// Docker workaround: force 401 on /v2/ when anonymous policies coexist with
+				// authenticated-only policies. Otherwise Docker treats 200 on /v2/ as "no auth"
+				// and will not send stored credentials for protected repositories.
+				// See: https://github.com/opencontainers/wg-auth/blob/main/docs/implementations/moby.md
+				hasMixedPolicy := allowAnonymous && !accessControlConfig.ContainsOnlyAnonymousPolicy()
+				if isDockerClient && isV2Requested && hasMixedPolicy {
+					authenticated = false
+				} else {
+					authenticated = true
+				}
```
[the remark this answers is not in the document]
andaaron: not relevant

pkg/api/authn.go
```diff
@@ -466,7 +468,16 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
 
 			// If no credentials provided - check for anonymous / mgmt requests
 			case allowAnonymous || isMgmtRequested:
-				authenticated = true
+				// Docker workaround: force 401 on /v2/ when anonymous policies coexist with
+				// authenticated-only policies. Otherwise Docker treats 200 on /v2/ as "no auth"
+				// and will not send stored credentials for protected repositories.
+				// See: https://github.com/opencontainers/wg-auth/blob/main/docs/implementations/moby.md
+				hasMixedPolicy := allowAnonymous && !accessControlConfig.ContainsOnlyAnonymousPolicy()
+				if isDockerClient && isV2Requested && hasMixedPolicy {
```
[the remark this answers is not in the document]
andaaron: Fixed this.

PR #3885, opened 2026-03-21
feat: support pushing multiple tags for a single manifest

andaaron:
See https://github.com/opencontainers/distribution-spec/pull/600

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -266,7 +266,7 @@ func getUIHeadersHandler(config *config.Config, allowedMethods ...string) func(h
 // @Router  /v2/ [get]
 // @Accept  json
 // @Produce json
-// @Success 200 {string} string "ok".
+// @Success 200 {string} string "ok"
 func (rh *RouteHandler) CheckVersionSupport(response http.ResponseWriter, request *http.Request) {
 	if request.Method == http.MethodOptions {
 		return
@@ -304,7 +304,7 @@ func (rh *RouteHandler) CheckVersionSupport(response http.ResponseWriter, reques
 // @Param   last  query  string   true   "last tag value for pagination"
 // @Success 200 {object}     common.ImageTags
 // @Failure 404 {string}     string                 "not found"
-// @Failure 400 {string}     string                 "bad request".
+// @Failure 400 {string}     string                 "bad request"
 func (rh *RouteHandler) ListTags(response http.ResponseWriter, request *http.Request) {
 	if request.Method == http.MethodOptions {
 		return
@@ -436,9 +436,9 @@ func (rh *RouteHandler) ListTags(response http.ResponseWriter, request *http.Req
 // @Param   name          path    string     true        "repository name"
 // @Param   reference     path    string     true        "image reference or digest"
 // @Success 200 {string} string "ok"
-// @Header  200 {object} constants.DistContentDigestKey
+// @Header  200 {string} Docker-Content-Digest "Manifest digest of the content"
 // @Failure 404 {string} string "not found"
-// @Failure 500 {string} string "internal server error".
+// @Failure 500 {string} string "internal server error"
 func (rh *RouteHandler) CheckManifest(response http.ResponseWriter, request *http.Request) {
 	if request.Method == http.MethodOptions {
 		return
@@ -509,7 +509,7 @@ type ExtensionList struct {
 // @Param   name         path    string     true        "repository name"
 // @Param   reference    path    string     true        "image reference or digest"
 // @Success 200 {object} api.ImageManifest
-// @Header  200 {object} constants.DistContentDigestKey
+// @Header  200 {string} Docker-Content-Digest "Manifest digest of the content"
 // @Failure 404 {string} string "not found"
 // @Failure 500 {string} string "internal server error"
 // @Router /v2/{name}/manifests/{reference} [get].
@@ -683,11 +683,12 @@ func (rh *RouteHandler) GetReferrers(response http.ResponseWriter, request *http
 // @Param   name         path    string     true        "repository name"
 // @Param   reference    path    string     true        "image reference or digest"
 // @Param   tag          query   []string   false       "additional tag(s) for digest pushes" collectionFormat(multi)
-// @Success 201 {string} string "created"
-// @Header  201 {object} constants.DistContentDigestKey
-// @Header  201 {string} constants.OCITagResponseKey "echoed tag= value (one header per query tag)"
+// @Success 201 "created"
+// @Header  201 {string} Docker-Content-Digest "Manifest digest of the uploaded content"
+// @Header  201 {string} OCI-Tag "Echoed tag= value; this header is repeatable (one field per tag= query parameter)"
 // @Failure 400 {string} string "bad request"
 // @Failure 404 {string} string "not found"
+// @Failure 414 {string} string "too many tag query parameters"
 // @Failure 500 {string} string "internal server error"
 // @Router /v2/{name}/manifests/{reference} [put].
 func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *http.Request) {
@@ -873,7 +874,12 @@ func normalizeManifestExtraTags(raw []string) ([]string, error) {
 // @Produce json
 // @Param   name          path    string     true        "repository name"
 // @Param   reference     path    string     true        "image reference or digest"
-// @Success 200 {string} string "ok"
+// @Success 202 "accepted"
+// @Failure 400 {string} string "bad request"
+// @Failure 404 {string} string "not found"
+// @Failure 405 {string} string "method not allowed"
+// @Failure 409 {string} string "conflict"
+// @Failure 500 {string} string "internal server error"
 // @Router /v2/{name}/manifests/{reference} [delete].
 func (rh *RouteHandler) DeleteManifest(response http.ResponseWriter, request *http.Request) {
 	vars := mux.Vars(request)
@@ -1012,7 +1018,7 @@ func canMount(userAc *reqCtx.UserAccessControl, imgStore storageTypes.ImageStore
 // @Param   name     path    string     true        "repository name"
 // @Param   digest   path    string     true        "blob/layer digest"
 // @Success 200 {object} api.ImageManifest
-// @Header  200 {object} constants.DistContentDigestKey
+// @Header  200 {string} Docker-Content-Digest "Manifest digest of the content"
 // @Router /v2/{name}/blobs/{digest} [head].
 func (rh *RouteHandler) CheckBlob(response http.ResponseWriter, request *http.Request) {
 	vars := mux.Vars(request)
@@ -1155,7 +1161,7 @@ func parseRangeHeader(contentRange string) (int64, int64, error) {
 // @Produce application/vnd.oci.image.layer.v1.tar+gzip
 // @Param   name     path    string     true        "repository name"
 // @Param   digest   path    string     true        "blob/layer digest"
-// @Header  200 {object} constants.DistContentDigestKey
+// @Header  200 {string} Docker-Content-Digest "Manifest digest of the content"
 // @Success 200 {object} api.ImageManifest
 // @Router /v2/{name}/blobs/{digest} [get].
 func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Request) {
@@ -1266,7 +1272,7 @@ func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Requ
 // @Produce json
 // @Param   name      path    string     true        "repository name"
 // @Param   digest    path    string     true        "blob/layer digest"
-// @Success 202 {string} string "accepted"
+// @Success 202 "accepted"
 // @Router /v2/{name}/blobs/{digest} [delete].
 func (rh *RouteHandler) DeleteBlob(response http.ResponseWriter, request *http.Request) {
 	vars := mux.Vars(request)
@@ -1325,7 +1331,7 @@ func (rh *RouteHandler) DeleteBlob(response http.ResponseWriter, request *http.R
 // @Accept  json
 // @Produce json
 // @Param   name    path    string     true        "repository name"
-// @Success 202 {string} string "accepted"
+// @Success 202 "accepted"
 // @Header  202 {string} Location "/v2/{name}/blobs/uploads/{session_id}"
 // @Header  202 {string} Range "0-0"
 // @Failure 401 {string} string "unauthorized"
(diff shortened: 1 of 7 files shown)


pkg/api/constants/consts.go
```diff
@@ -3,12 +3,19 @@ package constants
...
+	DistAPIVersion       = "Docker-Distribution-API-Version"
+	DistContentDigestKey = "Docker-Content-Digest"
+	// OCITagResponseKey is returned on digest manifest pushes that include tag query
+	// parameters (distribution-spec PR #600).
+	OCITagResponseKey = "OCI-Tag"
+	SubjectDigestKey  = "OCI-Subject"
+	// MaxManifestDigestQueryTags is the maximum number of tag query parameters accepted on
+	// PUT .../manifests/<digest>?tag=... (draft OCI distribution-spec: registries MUST
+	// support at least this many and MAY respond with 414 beyond it).
+	MaxManifestDigestQueryTags   = 10
```
vrajashkr: Should we fix a numerical limit for the max number of tags or should we keep a max URL length and let the number of tags be dynamic based on that?

pkg/api/routes.go
```diff
@@ -727,7 +727,34 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
...
+		zcommon.WriteJSON(response, http.StatusBadRequest, apiErr.NewErrorList(err))
+
+		return
+	}
+
+	if len(digestQueryTags) > constants.MaxManifestDigestQueryTags {
+		e := apiErr.NewError(apiErr.MANIFEST_INVALID).AddDetail(map[string]string{
+			"reason": fmt.Sprintf("too many tag query parameters (max %d)", constants.MaxManifestDigestQueryTags),
+		})
+		zcommon.WriteJSON(response, http.StatusRequestURITooLong, apiErr.NewErrorList(e))
```
vrajashkr: The URL could be too long even if the number of tags are less than 10. Should we also validate by setting a max URL length which we can validate early on and return status 400?

pkg/api/routes.go
```diff
@@ -769,12 +796,33 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
...
-			return
+			for _, tag := range digestQueryTags {
+				mErr := meta.SetImageMetaFromInput(request.Context(), name, tag, mediaType,
+					digest, body, imgStore, rh.c.MetaDB, rh.c.Log)
+				if mErr != nil {
+					rh.c.Log.Error().Err(mErr).Str("repository", name).Str("tag", tag).
+						Msg("multi-tag digest push: failed to update meta for tag")
+
+					metaUpdateFailed = true
+				}
```
vrajashkr: If writing a particular tag fails, then the meta is left in a partial state:
- Meta is updated for some tags
- Meta is not updated for tags after the failed tag

Is there a way to reconcile the meta state later or remove the image from store / untag the other tags. I guess we get to pick the behavior here, but it should be recoverable.

pkg/api/routes.go
```diff
@@ -784,9 +832,45 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
...
+	}
+
 	response.WriteHeader(http.StatusCreated)
 }
 
+// normalizeManifestExtraTags deduplicates tag query values in order and rejects empty tag components.
+func normalizeManifestExtraTags(tagQueryPresent bool, raw []string) ([]string, error) {
+	if !tagQueryPresent {
+		return nil, nil
+	}
```
vrajashkr: nit: IMO, we could probably check this outside the function and conditionally call it only if there are tag query params.

pkg/api/routes.go
```diff
@@ -784,9 +832,45 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
...
+
+	seen := map[string]struct{}{}
+
+	out := make([]string, 0, len(raw))
+
+	for _, rawTag := range raw {
+		rawTag = strings.TrimSpace(rawTag)
```
vrajashkr: nit: it might be helpful to rename `rawTag` on the left side as `cleanedTag`.

pkg/api/routes.go
```diff
@@ -784,9 +832,45 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
...
+			continue
+		}
+
+		seen[rawTag] = struct{}{}
+		out = append(out, rawTag)
+	}
+
+	if len(out) == 0 {
+		return nil, zerr.ErrEmptyManifestTagQuery
+	}
```
vrajashkr: We may not need this check - if there were tags query params present, then the output list must have something in it. If any were empty, the function would have already returned an error on line 856.

pkg/storage/imagestore/imagestore.go
```diff
@@ -647,27 +647,105 @@ func (is *ImageStore) PutImageManifest(repo, reference, mediaType string, //noli
...
-		_, ok := manifest.Annotations[ispec.AnnotationRefName]
-		if !ok && manifest.Digest.String() == desc.Digest.String() {
-			// matching descriptor does not have a tag, we need to remove it and add the new descriptor
-			index.Manifests = append(index.Manifests[:midx], index.Manifests[midx+1:]...)
+	if len(extraTags) > 0 {
+		for midx := 0; midx < len(index.Manifests); {
+			manifest := index.Manifests[midx]
+			_, hasTag := manifest.Annotations[ispec.AnnotationRefName]
+			if !hasTag && manifest.Digest.String() == mDigest.String() {
+				index.Manifests = append(index.Manifests[:midx], index.Manifests[midx+1:]...)
```
vrajashkr, line 660: Wouldn't this exceed the bounds on the last iteration?
`midx` would be equal to `len(index.Manifests) - 1` so accessing + 1 here goes out of bounds.

pkg/storage/imagestore/imagestore.go
```diff
@@ -647,27 +647,105 @@ func (is *ImageStore) PutImageManifest(repo, reference, mediaType string, //noli
...
+				Size:      desc.Size,
+				Digest:    mDigest,
+				Annotations: map[string]string{
+					ispec.AnnotationRefName: tag,
+				},
+			}
+
+			updateIndex, oldDgst, cerr := common.CheckIfIndexNeedsUpdate(&index, &descLocal, is.log)
+			if cerr != nil {
+				err = cerr
```
vrajashkr: Is this line needed since the next line is a return?

pkg/storage/imagestore/imagestore.go
```diff
@@ -647,27 +647,105 @@ func (is *ImageStore) PutImageManifest(repo, reference, mediaType string, //noli
...
+			return mDigest, subjectDigest, nil
+		}
+
+		lintDesc = ispec.Descriptor{
+			MediaType:    mediaType,
+			Size:         desc.Size,
+			Digest:       mDigest,
+			ArtifactType: artifactType,
+			Annotations: map[string]string{
+				ispec.AnnotationRefName: extraTags[0],
```
vrajashkr: Does lint not require the other tags too?

pkg/storage/imagestore/imagestore.go
```diff
@@ -647,27 +647,105 @@ func (is *ImageStore) PutImageManifest(repo, reference, mediaType string, //noli
...
+		if err != nil {
+			return "", "", err
+		}
+
+		// now update "index.json"
+		for midx, manifest := range index.Manifests {
+			_, ok := manifest.Annotations[ispec.AnnotationRefName]
+			if !ok && manifest.Digest.String() == desc.Digest.String() {
+				// matching descriptor does not have a tag, we need to remove it and add the new descriptor
+				index.Manifests = append(index.Manifests[:midx], index.Manifests[midx+1:]...)
```
vrajashkr, line 742: Same question here about the possibility of bounds of `index.Manifests` being exceeded on the last iteration.

pkg/storage/storage_test.go
```diff
@@ -189,6 +189,76 @@ func TestStorageNew(t *testing.T) {
 	})
 }
 
+func TestDigestPushMultiTagQuery(t *testing.T) {
```
vrajashkr: Should we also have tests for:
1. Existing manifest pushed again, but with a different set of tags (maybe a subset)
2. Non-digest push with tags resulting in an error.

pkg/api/constants/consts.go
andaaron: The number is based on the discussions in https://github.com/opencontainers/distribution-spec/pull/600

pkg/api/routes.go
andaaron: Initially I put a revert for the tags which succeeded. I guess we need to decide one way of the other.

pkg/storage/imagestore/imagestore.go
andaaron: I don't know if anyone uses this lint feature.

pkg/storage/imagestore/imagestore.go
andaaron: No, this is fine https://go.dev/wiki/SliceTricks#delete

pkg/storage/imagestore/imagestore.go
andaaron: Yes, because there's a deferred function reading the err. I refactored the code for clarity.

rchincha:
we may need to upgrade swagger, etc also

rchincha:
just fyi, cc:@sudo-bmitch

pkg/api/routes.go
andaaron: Fixed

rchincha:
@sudo-bmitch does regclient support this today? so we can add it to our ci/cd.

sudo-bmitch:
> @sudo-bmitch does regclient support this today? so we can add it to our ci/cd.

It's in edge/main: https://github.com/regclient/regclient/pull/1062

The place you see it is when pushing a non-canonical digest. I've got an outstanding TODO to add pushing multiple tags via the library.

pkg/api/constants/consts.go
```diff
@@ -3,12 +3,23 @@ package constants
...
+	OCITagResponseKey = "OCI-Tag"
+	SubjectDigestKey  = "OCI-Subject"
+	// MaxManifestDigestQueryTags is the maximum number of raw `tag=` query parameters accepted on
+	// PUT .../manifests/<digest>?tag=... (draft OCI distribution-spec: registries MUST support at
+	// least 10 and MAY respond with 414 beyond this limit). It uses the OCI tag max length (128;
+	// must match pkg/regexp.TagMaxLen) and an ~8KiB request-target budget, reserving 2048 bytes
+	// for path and digest:
+	//
+	//	(8192 - 2048) / (len("tag=") + 128 + 1) == 46
+	MaxManifestDigestQueryTags   = (8192 - 2048) / (len("tag=") + 128 + 1)
```
vrajashkr, line 22: nit: I think we might have wanted to use the constant `zreg.TagMaxLen` here.

pkg/api/constants/consts_test.go
```diff
@@ -0,0 +1,18 @@
...
+	zreg "zotregistry.dev/zot/v2/pkg/regexp"
+)
+
+func TestMaxManifestDigestQueryTagsDerived(t *testing.T) {
+	t.Parallel()
+
+	want := (8192 - 2048) / (len("tag=") + zreg.TagMaxLen + 1)
+
+	if constants.MaxManifestDigestQueryTags != want {
+		t.Fatalf("MaxManifestDigestQueryTags = %d, want %d", constants.MaxManifestDigestQueryTags, want)
```
vrajashkr, line 16: nit: since we've used `github.com/stretchr/testify/assert`, it would be ideal to use that here too for consistency.

pkg/api/controller_test.go
```diff
@@ -7808,6 +7809,142 @@ func TestManifestValidation(t *testing.T) {
...
+		baseURL := test.GetBaseURL(port)
+
+		conf := config.New()
+		conf.HTTP.Port = port
+
+		dir := t.TempDir()
+		ctlr := makeController(conf, dir)
+		cm := test.NewControllerManager(ctlr)
+		cm.StartServer()
+		time.Sleep(1000 * time.Millisecond)
```
vrajashkr, line 7824: If I recall correctly, we had a way to test for server readiness so we could potentially wait only as long as needed. (Not sure if I'm getting this mixed up with the BATS tests).

pkg/api/routes.go
```diff
@@ -769,12 +806,22 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
 	}
 
 	if rh.c.MetaDB != nil {
-		err := meta.OnUpdateManifest(request.Context(), name, reference, mediaType,
-			digest, body, rh.c.StoreController, rh.c.MetaDB, rh.c.Log)
-		if err != nil {
-			response.WriteHeader(http.StatusInternalServerError)
+		if len(digestQueryTags) > 0 {
+			err := meta.OnUpdateManifestDigestTags(request.Context(), name, digestQueryTags, mediaType,
```
vrajashkr, line 810: q: do we log this error inside the functions for all error cases?

pkg/api/constants/consts.go
andaaron: I know, but I didn't want to add the extra package dependency. We can update if you think it is worth it.

pkg/api/controller_test.go
andaaron: I don't think we really needed this here. Must be a leftover.

pkg/api/constants/consts.go
vrajashkr: Gotcha, that makes sense. I don't expect the value to change so I think it is okay as-is.

If we see any changes in future, then we could have the constant available in a different package that regex and api consts can import without a cycle.

pkg/api/constants/consts_test.go
andaaron: I think we use it in a single file in an older test, I haven't noticed it until now. We mosty use goconvey which has its own assertions. Here we could have used github.com/stretchr/testify/assert since goconvey doesn't work well with t.Parallel() as far as I understand.

pkg/api/routes.go
andaaron: Looks like I missed one, commented.

pkg/meta/hooks.go
```diff
@@ -43,6 +171,43 @@ func OnUpdateManifest(ctx context.Context, repo, reference, mediaType string, di
...
+	digest godigest.Digest, body []byte,
+	storeController storage.StoreController, metaDB mTypes.MetaDB, log log.Logger,
+) error {
+	if len(tags) == 0 {
+		return nil
+	}
+
+	priorTagManifests, err := priorTagManifestsFromMetaDB(ctx, metaDB, repo, tags)
+	if err != nil {
+		return err
```
andaaron, line 189: Need to add a log here.

pkg/api/constants/consts_test.go
vrajashkr: Cool. Sounds good. I'm fine with either the standard testing library way of doing it or testify assert. It would be good to standardize on one though to help with consistency across tests.

PR #3887, opened 2026-03-21
fix(cve): fall back to vendor severity when primary is unknown

### LINKED ISSUE #3886: CVE scanner shows 'Unknown' severity for all vulnerabilities — missing VendorSeverity fallback
## Description

Zot's built-in Trivy CVE scanner reports all detected vulnerabilities as "Unknown" severity, even when standalone Trivy correctly assigns severity levels (HIGH, MEDIUM, etc.) for the same image.

For example, scanning an Alpine 3.23-based Node.js app image:
- **Zot UI**: 18 CVEs, all "Unknown" severity (C:0, H:0, M:0, L:0, U:18)
- **Standalone Trivy** (`trivy image`): same CVEs with proper severity ratings from Alpine/vendor sources

## Environment

- **Zot version**: v2.1.15
- **Platform**: linux/arm64 (Docker Desktop on macOS)
- **Image base**: node:22-alpine (Alpine 3.23.3)
- **Trivy DB**: default ghcr.io/aquasecurity/trivy-db

## Root cause

In \`pkg/extensions/search/cve/trivy/scanner.go\`, the \`convertSeverity\` function reads only \`vulnerability.Severity\` (the primary NVD severity string). It does not consult the \`VendorSeverity\` map that Trivy populates with distro-specific severity data (Alpine, Red Hat, Ubuntu, etc.).

Due to the ongoing NVD analysis backlog (25,000+ unscored CVEs since 2024), many CVEs have no CVSS score in NVD, so the primary \`Severity\` field is empty/"UNKNOWN". Standalone Trivy handles this via \`--vuln-severity-source\` which falls back to vendor-specific data. Zot's embedded usage does not configure this fallback.

## Proposed fix

When the primary severity is "UNKNOWN" and \`VendorSeverity\` has entries, use the highest vendor-reported severity. This matches standalone Trivy's behavior.

I have a PR ready with the fix and tests: will link shortly.

## Similar issues in other projects

- [goharbor/harbor#11210](https://github.com/goharbor/harbor/issues/11210) — Harbor had the same issue with their Trivy integration.

jpgrace:
## Summary

- When the primary `Severity` field on a CVE is empty or "UNKNOWN", fall back to the highest severity from the `VendorSeverity` map (Alpine, Red Hat, Ubuntu, etc.)
- This matches standalone Trivy's behavior and resolves #3886

## Changes

- **`pkg/extensions/search/cve/trivy/scanner.go`**: Updated `convertSeverity` to accept `VendorSeverity` and use the highest vendor-reported severity when the primary is unknown
- **`pkg/extensions/search/cve/trivy/scanner_internal_test.go`**: Added 6 test cases covering primary severity, vendor fallback, highest-wins logic, and edge cases

## Context

Due to the NVD analysis backlog (25,000+ unscored CVEs since 2024), many CVEs have no CVSS score in NVD. Standalone Trivy handles this via `--vuln-severity-source` which falls back to vendor data. Zot's embedded Trivy usage was only reading the primary `Severity` string, causing all such CVEs to show as "Unknown" in the UI.

## Test plan

- [x] Added unit tests for `convertSeverity` with vendor severity fallback
- [ ] Verify with a real image that previously showed all "Unknown" severities
- [ ] Confirm existing CVE tests still pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/extensions/search/cve/trivy/scanner.go
+++ b/pkg/extensions/search/cve/trivy/scanner.go
@@ -501,7 +501,7 @@ func (scanner Scanner) scanManifest(ctx context.Context, repo, digest string) (m
 					Title:       vulnerability.Title,
 					Description: vulnerability.Description,
 					Reference:   getCVEReference(vulnerability.PrimaryURL, vulnerability.References),
-					Severity:    convertSeverity(vulnerability.Severity),
+					Severity:    convertSeverity(vulnerability.Severity, vulnerability.VendorSeverity),
 					PackageList: newPkgList,
 				}
 			}
@@ -701,15 +701,32 @@ func findMediaTypeForDigest(metaDB mTypes.MetaDB, digest godigest.Digest) (bool,
 	return false, ""
 }
 
-func convertSeverity(detectedSeverity string) string {
-	trivySeverity, _ := dbTypes.NewSeverity(detectedSeverity)
-
+func convertSeverity(detectedSeverity string, vendorSeverity dbTypes.VendorSeverity) string {
 	sevMap := map[dbTypes.Severity]string{
 		dbTypes.SeverityUnknown:  cvemodel.SeverityUnknown,
 		dbTypes.SeverityLow:      cvemodel.SeverityLow,
 		dbTypes.SeverityMedium:   cvemodel.SeverityMedium,
 		dbTypes.SeverityHigh:     cvemodel.SeverityHigh,
 		dbTypes.SeverityCritical: cvemodel.SeverityCritical,
+	}
+
+	trivySeverity, _ := dbTypes.NewSeverity(detectedSeverity)
+
+	// When the primary severity is unknown (common due to NVD analysis backlog),
+	// fall back to the highest severity reported by any vendor-specific source
+	// (e.g. Alpine, Red Hat, Ubuntu). This matches standalone Trivy's behavior
+	// when using --vuln-severity-source.
+	if trivySeverity == dbTypes.SeverityUnknown && len(vendorSeverity) > 0 {
+		highest := dbTypes.SeverityUnknown
+		for _, sev := range vendorSeverity {
+			if sev > highest {
+				highest = sev
+			}
+		}
+
+		if highest > dbTypes.SeverityUnknown {
+			return sevMap[highest]
+		}
 	}
 
 	return sevMap[trivySeverity]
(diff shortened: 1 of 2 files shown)


pkg/extensions/search/cve/trivy/scanner.go
```diff
@@ -712,5 +710,24 @@ func convertSeverity(detectedSeverity string) string {
...
+
+	// When the primary severity is unknown (common due to NVD analysis backlog),
+	// fall back to the highest severity reported by any vendor-specific source
+	// (e.g. Alpine, Red Hat, Ubuntu). This matches standalone Trivy's behavior
+	// when using --vuln-severity-source.
+	if trivySeverity == dbTypes.SeverityUnknown && len(vendorSeverity) > 0 {
+		highest := dbTypes.SeverityUnknown
+		for _, sev := range vendorSeverity {
+			if sev > highest {
+				highest = sev
```
andaaron, line 723: I don't agree with this approach. For example each disto has its own severity depending on how that vulnerability affects that specific distro. Some distros may decide the impact is minimal and give a low score. So in this specific case we should match the severity for the specific distro the image is based on. But I don't think that should be zot's role, that should be done by trivy internally.

Picking the highest severity would just raise alarm bells for everyone even if that should not be the case.

pkg/extensions/search/cve/trivy/scanner.go
```diff
@@ -712,5 +710,24 @@ func convertSeverity(detectedSeverity string) string {
 		dbTypes.SeverityCritical: cvemodel.SeverityCritical,
 	}
 
+	trivySeverity, _ := dbTypes.NewSeverity(detectedSeverity)
+
+	// When the primary severity is unknown (common due to NVD analysis backlog),
+	// fall back to the highest severity reported by any vendor-specific source
+	// (e.g. Alpine, Red Hat, Ubuntu). This matches standalone Trivy's behavior
+	// when using --vuln-severity-source.
```
andaaron, line 718: In this case why not use the equivalent of `--vuln-severity-source`  as a parameter in the trivy library.
I see https://github.com/aquasecurity/trivy/issues/8180 was added after our initial implementation, and we could probably leverage that.

I see they reason:
```
Some users prefer to prioritize NVD severities
Others might want to https://github.com/aquasecurity/trivy/discussions/8042 if they consider them problematic
Different organizations might have different preferences for severity sources
```

Given the above I think we should add a zot configuration to be able to switch from `auto` to something else. That way zot wouldn't be making assumptions about what the user would be interested to see.

jpgrace:
Closing — Aqua Security / Trivy supply chain compromise makes this less relevant right now. May revisit later.

PR #3906, opened 2026-03-28
feat(storage): add a common blobstore to store all blobs

rchincha:
Currently, zot uses one of the existing repos as the master copy for a blob to achieve dedupe, which complicates dedupe tracking logic. Furthermore, we have a global storage lock which is becoming a bottleneck. In order to move to a per-repo lock, we first need to simplify this logic.

Now use a single hidden global repo (_blobstore/) as a blob store instead.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -89,6 +89,8 @@ var (
 	ErrInvalidRoute                     = errors.New("invalid route prefix")
 	ErrImgStoreNotFound                 = errors.New("image store not found corresponding to given route")
 	ErrLocalImgStoreNotFound            = errors.New("local image store not found corresponding to given route")
+	ErrDefaultImgStoreCreate            = errors.New("failed to create image store for default config")
+	ErrSubpathImgStoreCreate            = errors.New("failed to create image store for subpath")
 	ErrEmptyValue                       = errors.New("empty cache value")
 	ErrEmptyRepoList                    = errors.New("no repository found")
 	ErrCVESearchDisabled                = errors.New("cve search is disabled")
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -649,14 +649,26 @@ func TestObjectStorageController(t *testing.T) {
 		endpoint := os.Getenv("S3MOCK_ENDPOINT")
 		tmp := t.TempDir()
 
+		// create s3 bucket
+		resp, err := resty.R().Put("http://" + endpoint + "/" + bucket)
+		if err != nil {
+			panic(err)
+		}
+		if resp.StatusCode() != http.StatusOK {
+			panic(fmt.Sprintf("failed to create bucket: %d %s", resp.StatusCode(), resp.String()))
+		}
+
 		storageDriverParams := map[string]any{
 			"rootdirectory":  tmp,
 			"name":           storageConstants.S3StorageDriverName,
 			"region":         "us-east-2",
 			"bucket":         bucket,
 			"regionendpoint": endpoint,
+			"accesskey":      "minioadmin",
+			"secretkey":      "minioadmin",
 			"secure":         false,
 			"skipverify":     false,
+			"forcepathstyle": true,
 		}
 
 		conf.Storage.StorageDriver = storageDriverParams
@@ -682,8 +694,11 @@ func TestObjectStorageController(t *testing.T) {
 			"region":         "us-east-2",
 			"bucket":         bucket,
 			"regionendpoint": endpoint,
+			"accesskey":      "minioadmin",
+			"secretkey":      "minioadmin",
 			"secure":         false,
 			"skipverify":     false,
+			"forcepathstyle": true,
 		}
 		conf.Storage.RemoteCache = true
 		conf.Storage.StorageDriver = storageDriverParams
@@ -730,9 +745,12 @@ func TestObjectStorageController(t *testing.T) {
 		}
 
 		// create s3 bucket
-		_, err = resty.R().Put("http://" + os.Getenv("S3MOCK_ENDPOINT") + "/" + bucket)
+		resp, err := resty.R().Put("http://" + os.Getenv("S3MOCK_ENDPOINT") + "/" + bucket)
 		if err != nil {
 			panic(err)
+		}
+		if resp.StatusCode() != http.StatusOK {
+			panic(fmt.Sprintf("failed to create bucket: %d %s", resp.StatusCode(), resp.String()))
 		}
 
 		ctlr := makeController(conf, "/")
@@ -758,14 +776,26 @@ func TestObjectStorageControllerSubPaths(t *testing.T) {
 		endpoint := os.Getenv("S3MOCK_ENDPOINT")
 		tmp := t.TempDir()
 
+		// create s3 bucket
+		resp, err := resty.R().Put("http://" + endpoint + "/" + bucket)
+		if err != nil {
+			panic(err)
+		}
+		if resp.StatusCode() != http.StatusOK {
+			panic(fmt.Sprintf("failed to create bucket: %d %s", resp.StatusCode(), resp.String()))
+		}
+
 		storageDriverParams := map[string]any{
 			"rootdirectory":  tmp,
 			"name":           storageConstants.S3StorageDriverName,
 			"region":         "us-east-2",
 			"bucket":         bucket,
 			"regionendpoint": endpoint,
+			"accesskey":      "minioadmin",
+			"secretkey":      "minioadmin",
 			"secure":         false,
 			"skipverify":     false,
+			"forcepathstyle": true,
 		}
 		conf.Storage.StorageDriver = storageDriverParams
 		ctlr := makeController(conf, tmp)
--- a/pkg/cli/server/verify_retention_test.go
+++ b/pkg/cli/server/verify_retention_test.go
@@ -476,7 +476,7 @@ func TestRetentionCheckWithRetentionEnabledAndRedisDriver(t *testing.T) {
 
 		defer ctrlManager.StopServer()
 
-		os.Args = []string{"cli_test", "verify-feature", "retention", "-l", logFile, "-t", "2s", configFile}
+		os.Args = []string{"cli_test", "verify-feature", "retention", "-l", logFile, "-t", "10s", configFile}
 		err = cli.NewServerRootCmd().Execute()
 		So(err, ShouldBeNil)
 
--- a/pkg/extensions/sync/destination.go
+++ b/pkg/extensions/sync/destination.go
@@ -90,6 +90,12 @@ func (registry *DestinationRegistry) GetImageReference(repo, reference string) (
 func (registry *DestinationRegistry) CommitAll(repo string, imageReference ref.Ref) error {
 	tempImageStore := getImageStoreFromImageReference(repo, imageReference, registry.log)
 
+	if tempImageStore == nil {
+		registry.log.Error().Str("repo", repo).Msg("failed to get temp image store for sync commit")
+
+		return zerr.ErrLocalImgStoreNotFound
+	}
+
 	defer os.RemoveAll(tempImageStore.RootDir())
 
 	repoDir := path.Join(tempImageStore.RootDir(), repo)
--- a/pkg/extensions/sync/sync_test.go
+++ b/pkg/extensions/sync/sync_test.go
@@ -1720,7 +1720,16 @@ func TestDockerImagesAreSkipped(t *testing.T) {
 
 				// trigger config blob upstream error
 				// remove synced image
-				err = os.RemoveAll(path.Join(destDir, indexRepoName))
+				dstRepoPath := path.Join(destDir, indexRepoName)
+				for range 5 {
+					err = os.RemoveAll(dstRepoPath)
+					if err == nil {
+						break
+					}
+
+					time.Sleep(100 * time.Millisecond)
+				}
+
 				So(err, ShouldBeNil)
 
 				configBlobPath := path.Join(srcDir, indexRepoName, "blobs/sha256", configBlobDigest.Encoded())
--- a/pkg/storage/cache/boltdb.go
+++ b/pkg/storage/cache/boltdb.go
@@ -114,6 +114,10 @@ func (d *BoltDBDriver) PutBlob(digest godigest.Digest, path string) error {
 		}
 	}
 
+	if len(path) == 0 {
+		return zerr.ErrEmptyValue
+	}
+
 	if err := d.db.Update(func(tx *bbolt.Tx) error {
 		root := tx.Bucket([]byte(constants.BlobsCache))
 		if root == nil {
@@ -128,21 +132,6 @@ func (d *BoltDBDriver) PutBlob(digest godigest.Digest, path string) error {
 		if err != nil {
 			// this is a serious failure
 			d.log.Error().Err(err).Str("bucket", digest.String()).Msg("failed to create a bucket")
-
-			return err
-		}
-
-		// create nested deduped bucket where we store all the deduped blobs + original blob
-		deduped, err := bucket.CreateBucketIfNotExists([]byte(constants.DuplicatesBucket))
-		if err != nil {
-			// this is a serious failure
-			d.log.Error().Err(err).Str("bucket", constants.DuplicatesBucket).Msg("failed to create a bucket")
-
-			return err
-		}
-
-		if err := deduped.Put([]byte(path), nil); err != nil {
-			d.log.Error().Err(err).Str("bucket", constants.DuplicatesBucket).Str("value", path).Msg("failed to put record")
 
 			return err
 		}
@@ -164,6 +153,28 @@ func (d *BoltDBDriver) PutBlob(digest godigest.Digest, path string) error {
 
 				return err
 			}
+
+			return nil
+		}
+
+		// original bucket exists; if path is the same as the original, this is idempotent
+		if origin.Get([]byte(path)) != nil {
+			return nil
+		}
+
+		// otherwise, this is a duplicate blob - add to the duplicates bucket
+		deduped, err := bucket.CreateBucketIfNotExists([]byte(constants.DuplicatesBucket))
+		if err != nil {
+			// this is a serious failure
+			d.log.Error().Err(err).Str("bucket", constants.DuplicatesBucket).Msg("failed to create a bucket")
+
+			return err
+		}
+
+		if err := deduped.Put([]byte(path), nil); err != nil {
+			d.log.Error().Err(err).Str("bucket", constants.DuplicatesBucket).Str("value", path).Msg("failed to put record")
+
+			return err
 		}
 
 		return nil
@@ -275,18 +286,20 @@ func (d *BoltDBDriver) HasBlob(digest godigest.Digest, blob string) bool {
 			return zerr.ErrCacheMiss
 		}
 
-		deduped := bucket.Bucket([]byte(constants.DuplicatesBucket))
-		if deduped == nil {
-			return zerr.ErrCacheMiss
+		// check original bucket first
+		if origin.Get([]byte(blob)) != nil {
+			return nil
 		}
 
-		if origin.Get([]byte(blob)) == nil {
-			if deduped.Get([]byte(blob)) == nil {
-				return zerr.ErrCacheMiss
+		// check duplicates bucket
+		deduped := bucket.Bucket([]byte(constants.DuplicatesBucket))
+		if deduped != nil {
+			if deduped.Get([]byte(blob)) != nil {
+				return nil
 			}
 		}
 
-		return nil
+		return zerr.ErrCacheMiss
 	}); err != nil {
 		return false
 	}
@@ -330,22 +343,33 @@ func (d *BoltDBDriver) DeleteBlob(digest godigest.Digest, path string) error {
 			return zerr.ErrCacheMiss
 		}
 
+		// check duplicates bucket first
 		deduped := bucket.Bucket([]byte(constants.DuplicatesBucket))
-		if deduped == nil {
-			return zerr.ErrCacheMiss
+		if deduped != nil {
+			if deduped.Get([]byte(path)) != nil {
+				if err := deduped.Delete([]byte(path)); err != nil {
+					d.log.Error().Err(err).Str("digest", digest.String()).Str("bucket", constants.DuplicatesBucket).
+						Str("path", path).Msg("failed to delete")
+
+					return err
+				}
+
+				return nil
+			}
 		}
 
-		if err := deduped.Delete([]byte(path)); err != nil {
-			d.log.Error().Err(err).Str("digest", digest.String()).Str("bucket", constants.DuplicatesBucket).
-				Str("path", path).Msg("failed to delete")
-
-			return err
-		}
-
+		// check original bucket
 		origin := bucket.Bucket([]byte(constants.OriginalBucket))
+		deleted := false
+
 		if origin != nil {
-			originBlob := d.getOne(origin)
-			if originBlob != nil {
+			if origin.Get([]byte(path)) != nil {
+				// if duplicates still exist, keep the original (global blobstore file stays)
+				if deduped != nil && d.getOne(deduped) != nil {
+					return nil
+				}
+
+				// no more duplicates, safe to remove the original
 				if err := origin.Delete([]byte(path)); err != nil {
 					d.log.Error().Err(err).Str("digest", digest.String()).Str("bucket", constants.OriginalBucket).
 						Str("path", path).Msg("failed to delete")
@@ -353,17 +377,12 @@ func (d *BoltDBDriver) DeleteBlob(digest godigest.Digest, path string) error {
 					return err
 				}
 
-				// move next candidate to origin bucket, next GetKey will return this one and storage will move the content here
-				dedupedBlob := d.getOne(deduped)
-				if dedupedBlob != nil {
-					if err := origin.Put(dedupedBlob, nil); err != nil {
-						d.log.Error().Err(err).Str("digest", digest.String()).Str("bucket", constants.OriginalBucket).Str("path", path).
-							Msg("failed to put")
-
-						return err
-					}
-				}
+				deleted = true
 			}
+		}
+
+		if !deleted {
+			return zerr.ErrCacheMiss
 		}
 
 		// if no key in origin bucket then digest bucket is empty, remove it
@@ -376,9 +395,11 @@ func (d *BoltDBDriver) DeleteBlob(digest godigest.Digest, path string) error {
 
 				return err
 			}
+
+			return nil
 		}
 
-		return nil
+		return zerr.ErrCacheMiss
 	}); err != nil {
 		return err
 	}
--- a/pkg/storage/cache/boltdb_test.go
+++ b/pkg/storage/cache/boltdb_test.go
@@ -55,7 +55,7 @@ func TestBoltDBCache(t *testing.T) {
 		So(err, ShouldEqual, errors.ErrCacheMiss)
 
 		err = cacheDriver.DeleteBlob("key", "bogusValue")
-		So(err, ShouldBeNil)
+		So(err, ShouldEqual, errors.ErrCacheMiss)
 
 		// try to insert empty path
 		err = cacheDriver.PutBlob("key", "")
@@ -85,14 +85,21 @@ func TestBoltDBCache(t *testing.T) {
 		err = cacheDriver.PutBlob("key1", "duplicateBlobPath")
 		So(err, ShouldBeNil)
 
+		// deleting original when duplicates exist should keep the original (no promotion)
 		err = cacheDriver.DeleteBlob("key1", "originalBlobPath")
 		So(err, ShouldBeNil)
 
+		// original should still be "originalBlobPath" since duplicates exist
 		val, err = cacheDriver.GetBlob("key1")
-		So(val, ShouldEqual, "duplicateBlobPath")
+		So(val, ShouldEqual, "originalBlobPath")
 		So(err, ShouldBeNil)
 
+		// now delete the duplicate
 		err = cacheDriver.DeleteBlob("key1", "duplicateBlobPath")
+		So(err, ShouldBeNil)
+
+		// now delete the original (no more duplicates, should clean up)
+		err = cacheDriver.DeleteBlob("key1", "originalBlobPath")
 		So(err, ShouldBeNil)
 
 		// should be empty
@@ -147,15 +154,17 @@ func TestBoltDBCache(t *testing.T) {
 		blobs, err := cacheDriver.GetAllBlobs("digest")
 		So(err, ShouldBeNil)
 
+		// "first" is the original, "second" and "third" are duplicates
 		So(blobs, ShouldResemble, []string{"first", "second", "third"})
 
+		// deleting "first" (original) should keep it since duplicates exist
 		err = cacheDriver.DeleteBlob("digest", "first")
 		So(err, ShouldBeNil)
 
 		blobs, err = cacheDriver.GetAllBlobs("digest")
 		So(err, ShouldBeNil)
 
-		So(blobs, ShouldResemble, []string{"second", "third"})
+		So(blobs, ShouldResemble, []string{"first", "second", "third"})
 
 		err = cacheDriver.DeleteBlob("digest", "third")
 		So(err, ShouldBeNil)
@@ -163,6 +172,6 @@ func TestBoltDBCache(t *testing.T) {
 		blobs, err = cacheDriver.GetAllBlobs("digest")
 		So(err, ShouldBeNil)
 
-		So(blobs, ShouldResemble, []string{"second"})
+		So(blobs, ShouldResemble, []string{"first", "second"})
 	})
 }
--- a/pkg/storage/cache/dynamodb.go
+++ b/pkg/storage/cache/dynamodb.go
@@ -187,13 +187,22 @@ func (d *DynamoDBDriver) PutBlob(digest godigest.Digest, path string) error {
 		return zerr.ErrEmptyValue
 	}
 
-	if originBlob, _ := d.GetBlob(digest); originBlob == "" {
-		// first entry, so add original blob
+	originBlob, _ := d.GetBlob(digest)
+	if originBlob == "" {
+		// first entry, so add original blob only
 		if err := d.putOriginBlob(digest, path); err != nil {
 			return err
 		}
+
+		return nil
 	}
 
+	// if same as original, this is idempotent
+	if originBlob == path {
+		return nil
+	}
+
+	// add as duplicate
 	expression := "ADD DuplicateBlobPath :i"
 	attrPath := types.AttributeValueMemberSS{Value: []string{path}}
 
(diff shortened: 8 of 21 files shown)


rchincha:
The following seems to be getting stuck!

```
@test "sync image on demand from registry.gitlab.com" {
    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
    run skopeo copy docker://127.0.0.1:${zot_port}/gitlab-org/public-image-archive/gitlab-ee:15.11.6-ee.0 oci:${TEST_DATA_DIR} --src-tls-verify=false
    [ "$status" -eq 0 ]

    run curl http://127.0.0.1:${zot_port}/v2/_catalog
    [ "$status" -eq 0 ]
    [ $(echo "${lines[-1]}"| jq '.repositories | map(select(. == "gitlab-org/public-image-archive/gitlab-ee"))' | jq '.[]') = '"gitlab-org/public-image-archive/gitlab-ee"' ]
    run curl http://127.0.0.1:${zot_port}/v2/gitlab-org/public-image-archive/gitlab-ee/tags/list
    [ "$status" -eq 0 ]
    [ $(echo "${lines[-1]}" | jq '.tags[]') = '"15.11.6-ee.0"' ]
}
```

pkg/extensions/search/cve/trivy/scanner.go
```diff
@@ -435,11 +442,8 @@ func (scanner Scanner) scanManifest(ctx context.Context, repo, digest string) (m
 	cveidMap := map[string]cvemodel.CVE{}
 	image := repo + "@" + digest
 
-	scanner.dbLock.Lock()
```
andaaron, line 609: Why have these been changed?
There's trivy global variables we need to protect. Unless they refactored those out of the code.

pkg/extensions/search/cve/trivy/scanner.go
```diff
@@ -575,12 +579,6 @@ func (scanner Scanner) scanIndex(ctx context.Context, repo, digest string) (map[
 
 // UpdateDB downloads the Trivy DB / Cache under the store root directory.
 func (scanner Scanner) UpdateDB(ctx context.Context) error {
-	// We need a lock as using multiple substores each with its own DB
-	// can result in a DATARACE because some varibles in trivy-db are global
-	// https://github.com/project-zot/trivy-db/blob/main/pkg/db/db.go#L23
-	scanner.dbLock.Lock()
```
andaaron, line 919: This should stay, see the deleted comment

pkg/extensions/search/cve/trivy/scanner.go
andaaron, line 431: This should not be locked. We want only the scan part to be blocking.

pkg/extensions/search/cve/trivy/scanner.go
rchincha: reverted

pkg/extensions/search/cve/trivy/scanner.go
rchincha: reverted

pkg/extensions/search/cve/trivy/scanner.go
rchincha: reverted

rchincha:
Don't merge this yet until we discuss the remote storage strategy first.

andaaron:
AI evaluation on curent state and what to do next

# Evaluation: PR #3906 — `feat(storage): add a common blobstore to store all blobs`

- **PR:** https://github.com/project-zot/zot/pull/3906
- **Merged into:** `refactor/storage` (base `273b1536`, head `3b71bcb`) — **not** main
- **Size:** 21 files, +1,100 / −426, patch coverage ~70%
- **Author note on the PR:** *"Don't merge this yet until we discuss the remote storage strategy first."*

---

## What the PR does (summary)

The PR changes zot's deduplication model from **"use one arbitrary repo as the master copy of a blob"** to **"a single hidden global blob store (`_blobstore/`) holds the master copy."** Stated goal: simplify dedupe tracking so they can later move from a global storage lock to a per-repo lock.

### Core change (`pkg/storage/imagestore/imagestore.go`)
- New constant `GlobalBlobsRepo = "_blobstore"` (leading underscore so it can't collide with a valid OCI repo name).
- `NewImageStore(...)` now calls `initRepo("_blobstore")` when dedupe is enabled, creating the global store at startup.
- `DedupeBlob` now moves the first copy of a blob into `_blobstore` and records *that* path as the cache "original"; per-repo paths become hard links (local FS) or 0-byte marker files (S3/GCS) recorded as duplicates.
- New `upgradeToGlobalBlobstore()` — a one-time migration that walks all existing repos and links/copies their blobs into `_blobstore`, run lazily the first time `_blobstore` is initialized.
- `deleteBlob` rewritten to reference-count: remove the per-repo path, and only delete the `_blobstore` master file/cache entry when the last reference goes away.
- Repo-name validation moved out of internal `initRepo` into exported `InitRepo` (so `_blobstore` can be created internally).

### Cache drivers (`boltdb.go`, `dynamodb.go`, `redis.go`) — semantics rewritten (bucket/key *structure* unchanged)
- First `PutBlob` for a digest goes **only** to the "original" bucket (previously it was also added to duplicates).
- `PutBlob` is now idempotent for the same path.
- `DeleteBlob` now **keeps the original while any duplicates remain**, and returns `ErrCacheMiss` when the path isn't found (previously returned `nil`).
- `HasBlob` checks original first, then duplicates.

### GC (`pkg/storage/gc/gc.go`)
- Skip retention on `_blobstore`.
- Treat a missing manifest blob as garbage-collectable (remove stale index entry) instead of erroring.

### Other
- `storage.go` now errors out instead of silently using a `nil` store.
- sync `destination.go` nil-guards the temp store.
- New errors in `errors/errors.go`.
- The rest (≈ the bulk of the +1,100) is test changes for boltdb/dynamodb/redis/s3/gcs/local plus bats tweaks.

---

## Bugs / correctness concerns

### P1. `DedupeBlob` can spin in an infinite loop if `_blobstore` content is missing but cache entries remain (most serious)
`DedupeBlob` keeps a `for {}` self-heal loop: when it can't `Stat` the cache's "original" path, it calls `cache.DeleteBlob(...)` and `continue`s. But the new cache semantics **refuse to delete the original while duplicates exist** (`"duplicates still exist, keep the original"`). So if the `_blobstore` master file is gone while per-repo duplicate references still exist in the cache, `GetBlob` keeps returning the same dead path, `Stat` keeps failing, `DeleteBlob` keeps being a no-op, and the loop never makes progress. The author worked around the *symptom* in `test/blackbox/sync_docker.bats` (the comment literally says wiping the root *"results in an infinite retry loop"*), but the underlying footgun in production code remains. There is no max-retry bound.

### P2. The migration is not resumable on partial failure
`upgradeToGlobalBlobstore` decides whether to run by checking `if len(globalBlobs) > 0 { return nil }`. If the upgrade fails partway (it `return`s the error, which `initRepo` only *logs*), the next restart sees a non-empty `_blobstore` and **skips the rest of the migration permanently**. Blobs not yet linked/copied are never migrated and never get their per-repo duplicate cache entries.

### P3. The migration assumes a fresh dedupe cache, but real upgrades keep it
The upgrade does `PutBlob(digest, globalPath)` first, expecting it to become the "original." But on a real upgrade the existing `cache.db` / DynamoDB / Redis already has `original = <some per-repo path>`. With the new `PutBlob` logic, since an original already exists and differs, **the `_blobstore` path is recorded as a duplicate, not the master.** Net effect: the very thing the PR set out to fix (a per-repo path acting as master) persists for all pre-existing blobs, and because the new `DeleteBlob` keeps the original-while-duplicates-exist, deleting that master repo can leave the cache pointing at a deleted file → feeds directly into P1. The new test `TestUpgradeToGlobalBlobstore` only ever uses a brand-new cache, so this realistic path is untested.

### P4. Object-storage migration loads whole blobs into memory
For S3/GCS the upgrade uses `ReadFile` then `WriteFile` per blob. Copilot flagged this and suggested streaming via `Reader`/`Writer` + `io.Copy`; **the suggestion was not applied** — the merged code still buffers entire layers in memory. Multi-GB layers can OOM the process during startup migration.

### P5. Migration runs synchronously at startup and walks the entire store
On a large registry this makes the first post-upgrade boot very slow; on S3/GCS it's a full server-side copy of every blob (temporarily ~doubling stored bytes) plus the memory issue above.

### P6. Minor: `_blobstore` leaks into mount-permission candidates
`GetAllDedupeReposCandidates` now returns `_blobstore`, and `canMount` (`pkg/api/routes.go`) iterates candidates checking `userAc.Can(ReadPermission, repo)`. A user won't normally have an explicit policy for `_blobstore`, but a wildcard/default read policy would match it. Low risk (such a user would already have access to the real repos), but it's an unintended entry in the authz surface.

### Verified non-issues
- `_blobstore` is correctly hidden from `/v2/_catalog`, metaDB indexing, and per-repo GC, because `ValidateRepo` rejects names that fail `FullNameRegexp` (a leading underscore is invalid), and the upgrade walk also explicitly skips it.
- Moving name validation out of internal `initRepo` is safe: the only internal caller (`copyBlob`) passes already-validated repo names, and exported `InitRepo` still validates.

---

## Upgrade risk from an existing main-based deployment

### Cache "schema": structurally compatible, semantically not
No bucket/table/key renames, so boltdb/DynamoDB/Redis won't fail to open. But the *meaning* of "original" vs "duplicate" changes, and (per P3) the migration doesn't reconcile a pre-existing populated cache. Expect bookkeeping that doesn't match the intended model on any registry that carries its dedupe cache across the upgrade.

### On-disk layout changes and (critically) downgrade safety
- **Local FS:** per-repo blobs become hard links to `_blobstore` files — content stays accessible, and a downgrade still works *as long as `_blobstore/` is not deleted*. Reasonable.
- **S3 / GCS:** the master content moves to `_blobstore` and per-repo blobs are 0-byte markers that rely on the cache to redirect reads. **Rolling back to a pre-blobstore zot would serve empty/0-byte layers — i.e. data corruption.** Downgrade is effectively one-way on object storage. The author's own "discuss the remote storage strategy first" comment is about exactly this area.

### `DeleteBlob` return-value change is a behavioral contract change
(`nil` → `ErrCacheMiss` for not-found). The in-tree callers were updated to tolerate `ErrCacheMiss`, but any out-of-tree integration or future caller relying on the old "delete is always nil" behavior would break.

### Net assessment
- For **local-filesystem, single-instance** deployments the upgrade is plausibly OK (with a backup), but you still inherit P1–P3.
- For **S3/GCS/remote-cache** deployments treat this as **not upgrade-safe as-is**: irreversible layout change, memory-heavy non-resumable migration, and the unhandled "existing cache" path.

---

## Cache driver bugs (boltdb / redis / dynamodb)

Reviewed the merged drivers at the PR head against the new contract: `original` = master (`_blobstore`) path, `duplicates` = the rest; first `PutBlob` → original only; `DeleteBlob` keeps the original while any duplicate remains; missing path → `ErrCacheMiss`.

### C1. BoltDB `GetAllBlobs` returns `ErrCacheMiss` when a digest has an original but no duplicates bucket (real regression)

```go
bucket := root.Bucket([]byte(digest.String()))
if bucket != nil {
    origin := bucket.Bucket([]byte(constants.OriginalBucket))
    blobPath.Write(d.getOne(origin))
    originBlob := blobPath.String()
    blobPaths = append(blobPaths, originBlob)          // origin collected...

    deduped := bucket.Bucket([]byte(constants.DuplicatesBucket))
    if deduped != nil {
        // ... append duplicates ...
        return nil                                     // only success path
    }
}

return zerr.ErrCacheMiss                                // <-- reached if deduped == nil
```

If the digest has an original but the duplicates sub-bucket was never created, the function discards the already-collected original and returns `(nil, ErrCacheMiss)`. This state is newly reachable because the PR's `PutBlob` writes the first entry to the origin **only** (previously the first entry also went into the duplicates bucket, so it was never nil). Redis and DynamoDB both correctly return `[original]` in this case, so BoltDB is now inconsistent. Impact: the imagestore `deleteBlob` `len(paths) == 1` cleanup is skipped, orphaning the global cache entry/blob. **Fix:** move `return nil` out to after the `if deduped != nil` block.

### C2. DynamoDB `DeleteBlob`/`PutBlob` are not atomic and have no locking

Redis serializes writes with a per-digest redsync mutex; BoltDB uses a single `db.Update` transaction. DynamoDB has neither, and the new `DeleteBlob` is now a 4–5 round-trip check-then-act (`GetDuplicateBlob` → `GetItem` → `GetBlob` → `GetDuplicateBlob` → `DeleteItem`) with no `ConditionExpression`. A concurrent `PutBlob` adding a duplicate can interleave so the item (the master entry) is deleted right after a new repo reference was registered — losing the master while a repo still points at it. **Fix:** guard the final `DeleteItem`/`putOriginBlob` with conditional expressions (e.g. delete only if `DuplicateBlobPath` absent).

### C3. "Keep original while duplicates exist" makes a stale original un-evictable (root of the DedupeBlob infinite loop)

All three drivers now return `nil` without changing state when asked to delete the original while duplicates remain (`getOne(deduped) != nil` / `dupes > 0` / `remainingDuplicate != ""`). Correct for the model, but it removes the only mechanism `DedupeBlob` had to evict a stale origin: when the `_blobstore` master *file* is missing but duplicate refs remain, `DedupeBlob`'s self-heal (`Stat` fails → `cache.DeleteBlob(original)` → `continue`) can never make progress → unbounded loop (the same one worked around in `sync_docker.bats`). The root cause is the cache `DeleteBlob` semantics, not just the loop.

### C4. DynamoDB `PutBlob` swallows the `GetBlob` error and can overwrite a valid original

```go
originBlob, _ := d.GetBlob(digest)   // error ignored
if originBlob == "" {
    if err := d.putOriginBlob(digest, path); err != nil { ... }  // SET OriginalBlobPath
    return nil
}
```

A transient `GetItem` failure yields `originBlob == ""`, treated as "first entry", and **overwrites** `OriginalBlobPath` with the new path — silently re-pointing the master. Pre-existing, but more dangerous now that the origin is the load-bearing `_blobstore` master. **Fix:** distinguish `ErrCacheMiss` from a real error.

### C5. Minor / style
- BoltDB `DeleteBlob` uses `root.DeleteBucket([]byte(digest))` while everything else uses `[]byte(digest.String())` (functionally identical, inconsistent).
- BoltDB `GetAllBlobs`/`GetBlob` append an empty string when `origin` is nil/empty instead of treating it as a miss.
- Redis `PutBlob` issues a direct `d.db.HGet(...)` inside the `TxPipelined` callback (immediate round-trip, not pipelined) — harmless under the lock, but defeats the pipeline.

---

## Design evaluation: removing the original from the duplicates list

**Change:** previously the first `PutBlob` wrote the path into *both* the origin and the duplicates set, so `duplicates` held the full set of paths (including the master). Now the first entry goes to the origin **only**, and `duplicates` holds *references other than the master*.

### Is it reasonable? — Yes, it's the cleaner model for the blobstore architecture.

- The blobstore design makes the master an **immutable** canonical location (`_blobstore/blobs/...`). Keeping it out of the duplicates set turns the data model into a clean "1 master + N references" shape, where `len(duplicates)` is exactly the number of repos still referencing the blob.
- It removes the old, fragile "promote a duplicate to origin and move the content there" dance that the previous `DeleteBlob` performed whenever the master repo was deleted. With an immutable master that logic — and a class of content-move races — simply goes away.
- It makes idempotency and "first entry" detection trivial.

### Impact on GC — net positive, but it shifts correctness onto cache/disk consistency

- GC of the master now keys off `GetAllBlobs(digest) == [origin]` (i.e. `len(duplicates) == 0`): the imagestore `deleteBlob` removes the repo's duplicate, and when only the origin remains it deletes the master cache entry and the `_blobstore` file. With the original excluded from duplicates, this is straightforward reference counting; under the old scheme the master was itself one of the repo paths, so counting references was muddier.
- The **risk** is that GC correctness now depends entirely on the duplicates set staying in lock-step with on-disk references:
  - **Under-count** (a real on-disk reference missing from the cache) → GC deletes the master while a repo still points at it → on S3 that repo's 0-byte marker now references deleted content = **data loss**.
  - **Over-count** (a duplicate left in the cache for a repo that no longer references it) → master never collected = **storage leak**.
- Concretely, bug **C1** pushes toward the *leak* direction (the cleanup is skipped), while bug **C2** (DynamoDB races) and the "existing cache" upgrade problem can push toward the *data-loss* direction. So the design is sound, but its safety margin is thinner than the old model where the master was a normal content file that survived a stale cache.

### Impact on dedupe — net positive

- `DedupeBlob` now always links new repo references against a single stable master (`GetBlob` → origin → `_blobstore`), instead of chasing a master that could move between repos. Simpler and race-free for the common path.
- But the same "immutable master / keep-original-while-duplicates-exist" rule (C3) means the old self-heal path in `DedupeBlob` (evict stale origin, retry) no longer works. The design decision is fine; the PR is incomplete because it didn't update the dependent retry loop, hence the infinite-loop hazard.

### Verdict
Removing the original from the duplicates list is a **reasonable and arguably better** design for the blobstore model — cleaner reference counting for GC and simpler, race-free dedupe. The problems are not with the decision itself but with **incomplete follow-through**: an invariant that other code relied on changed, and not all dependents were updated (BoltDB `GetAllBlobs`, the `DedupeBlob` retry loop), plus GC now leans harder on cache/disk consistency that isn't strongly guaranteed (especially DynamoDB, which has no locking, and the upgrade path that doesn't reconcile an existing cache).

---

## Architecture: would splitting storage (1 local / 2 remote) make dedupe + GC easier?

### What the structure looks like today
There are already two layers:
- **`pkg/storage/imagestore/imagestore.go`** — one ~2,300-line `ImageStore` implementing *all* OCI semantics (manifests, index.json, referrers, tags, uploads, validation, **dedupe + GC**), parameterized by a `driver.StorageDriver`.
- **`pkg/storage/{local,s3,gcs}/`** — thin `NewImageStore` constructors (~1 KB each) plus the real `driver.go` implementing the storage primitives (`Read/Write/Move/Stat/Walk/Delete/Link`).

So the backend seam is the **driver interface**, and there is exactly **one** ImageStore shared by all backends. The only `Name() == local` branches left in the ImageStore are cosmetic (metrics `SetStorageUsage`, the empty-dir check in `ValidateRepo`). The real local-vs-remote dedupe difference is hidden inside `StorageDriver.Link`: hard link on local, **0-byte marker + cache redirect** on s3/gcs.

### Splitting the *whole* ImageStore into local vs remote — not worth it
~80% of the file (manifests, index, referrers, tags, uploads, OCI layout) is backend-agnostic and identical; forking it causes duplication/drift and doubles the test surface. It also wrongly assumes s3 and gcs are interchangeable — they aren't (gcs has its own root-dir normalization and consistency quirks; `gcs/driver.go` already differs meaningfully from `s3/driver.go`). "2 remote" belongs at the **driver** level, not as a second ImageStore.

### Splitting just the dedupe + blob-lifecycle/GC strategy — yes, this is the real win
This is aligned with the PR's stated goal of simplifying dedupe tracking. Extract a small strategy interface (e.g. `BlobDedupe` / `BlobLifecycle`) with two implementations:

- **Local = hard-link / filesystem-refcounted.** The filesystem *already* does content reference counting: content GC becomes "unlink the repo blob; the OS frees the inode when the last link goes." The cache degrades to a pure dedupe *index*, and all the `size == 0` / `SameFile` / cache-redirect special cases (which exist only to emulate object storage) disappear. It also removes local exposure to the "under-count → delete master while a ref exists → data loss" risk, because the FS won't free an inode that still has links.
- **Remote = marker + explicit refcount (the genuinely hard part).** Object stores have no links, weaker consistency, no atomic rename in some cases, and per-object/listing cost. Isolating this makes the `_blobstore` canonical-object + marker + reference-counting model explicit and first-class (and properly testable for races) instead of implicit special cases in a shared method. This is exactly where this PR's bugs concentrate (DynamoDB races, marker/redirect edges, the infinite-loop self-heal).

Note: the `_blobstore` master concept is *most* valuable for remote (a stable canonical object you can't get from links); for local it is largely redundant with what hard links already provide.

### Recommendation (decided direction)
**Keep the single `ImageStore` and keep `_blobstore` for both backends** (do *not* fork into local/remote ImageStores). The shared deterministic `_blobstore` location is actually what makes the single-store design easier, because it unifies the master location across backends; only two things differ per backend, behind a thin capability/strategy seam:
1. Route dedupe/GC through a **capability/strategy seam** — e.g. a driver capability (`SupportsHardLinks()` / `LinkOrCopy` / a `LinkCount`/`Dereference` primitive) or a small `BlobLifecycle` strategy with `localHardlink` and `remoteMarker` implementations.
2. **Local = hard links + filesystem refcounting.** Because `_blobstore` is at a deterministic path, local dedupe needs no cache bookkeeping for content location (Stat the master path; link if present, else write+link), and **content GC is driven by `st_nlink`** rather than our own ref counting — delete the master only when its link count shows no repo references remain. This removes the C1/C3 cache hazards and the under-count data-loss risk *for local*. The cache becomes a pure dedupe index (still used for the mount-candidate query).
3. **Remote = converge toward the deterministic shared path (#3750).** Reuse `_blobstore` as the canonical object, but make remote reads target it deterministically (with per-repo fallback for un-migrated blobs) instead of via a cache redirect, and make GC/mount-candidate queries authoritative index scans. Keep s3-vs-gcs differences at the **driver** level, not as separate ImageStores.

Net: the helpful split is **"isolate the dedupe/GC strategy, share everything else,"** not "two whole ImageStores." It delivers the manageability win, lets local become much simpler (FS-refcounted), and confines the hard remote-storage problems (which is also what the author flagged with "discuss the remote storage strategy first").

### Hard-link support on macOS / Windows (and graceful degradation)
The hard-link approach is **already probed at runtime**, so it is portable:
- `pkg/storage/storage.go` calls `local.ValidateHardLink(rootDir)` whenever `Dedupe` is enabled on a local store; it writes a temp file and attempts `os.Link`. If the filesystem does **not** support hard links it logs a warning and **auto-disables dedupe** (`config.Storage.Dedupe = false`).
- With #3906, `_blobstore` is created only `if dedupe`, so on a no-hardlink filesystem there is no `_blobstore` and blobs are stored as full per-repo copies (existing non-dedupe behavior). Graceful degradation, no special handling needed.
- macOS (APFS/HFS+) and Windows (NTFS) support hard links, so the probe passes and dedupe works. On FAT/exFAT or some network mounts it fails → dedupe off.
- Caveat for the `st_nlink` GC idea: reading link count is trivial on Unix (`syscall.Stat_t.Nlink`) but Go's `os.FileInfo.Sys()` does **not** expose it on Windows — you'd need `GetFileInformationByHandle` (`nNumberOfLinks`) in a `driver_windows.go`. The package already uses a `driver_unix.go` / `driver_windows.go` build-tag split, so there's precedent; otherwise keep the cache/index refcount as the portable fallback when link count isn't available.

---

## Cross-check vs proposal PR #3750 (`SHARED_BLOB_STORAGE_PROPOSAL.md`)

PR #3750 is the design *proposal doc* (by `andaaron`); #3906 is an implementation that took a **different route**. Key differences:

| Aspect | #3750 proposal | #3906 implementation |
|---|---|---|
| Shared location | Deterministic path `{rootDir}/storage/blobs/{algo}/{digest}` | Hidden `_blobstore/` repo (OCI-layout dir) |
| Scope | **Remote only** (S3/GCS); local untouched | **All backends**, incl. local |
| Dedupe machinery | **Removed** for remote (deterministic path, no redirect) | **Kept** (DedupeBlob, cache, markers, candidates) |
| Cache for remote | One path per digest, no redirect | Original + duplicates + redirect |
| GC for remote | New global ref-tracking via index scan | Cache-based refcount via `deleteBlob` |
| Migration | Explicit `migrate-remote-storage` (dry-run, integrity verify, idempotent) | Automatic lazy startup walk |
| Rollback | Explicit `rollback-remote-storage` | **None** |
| Mount authz | `GetReposReferencingBlob` (index scan) | `GetAllDedupeReposCandidates` (cache) + adds `_blobstore` |

### Status of #3750's stated concerns in #3906
- **Eliminate dedupe complexity for remote** → **not done** (kept and arguably increased; this is where C1–C4 and the infinite loop live).
- **"Local unchanged, no migration"** → **not honored** (#3906 migrates local too via hard-linking into `_blobstore`). *Accepted as the decided direction* (see Architecture): keeping `_blobstore` for both backends is fine and enables FS-refcounted local dedupe — but the `#3750` proposal doc is now stale and should be reconciled to "`_blobstore` for both backends."
- **"Each repo a valid OCI layout on disk"** (andaaron's stated requirement) → **preserved for local** (hard links keep full content per repo); **broken for remote** (0-byte markers), same as the proposal accepted for remote.
- **vrajashkr: don't hardcode S3 / support future drivers** → **addressed** (branches on `Name() == LocalStorageDriverName`, i.e. local-vs-not, not S3-specific; GCS works).
- **vrajashkr: destructive migration needs rollback tracking / S3 rollback cost** → **decided out of scope** (see D): downgrade is not supported and no rollback tooling will be built. Migration is still made resumable + integrity-verified (B/C), but it is intentionally one-way.

### Decided plan — fix within the cherry-pick PR (as separate commits)
These are **not** deferred to a future PR; each lands as its own commit on top of the cherry-pick commit in the same PR.

- **A. Architecture decision:** keep single `ImageStore` + `_blobstore` for both backends; local degrades gracefully on no-hardlink filesystems (dedupe auto-disabled via `ValidateHardLink`). **No separate migration CLI tool is needed** — the in-process automatic migration is kept, but made robust via a success-marker file (see B). The `#3750` proposal doc should be reconciled to this decision: `_blobstore` for both backends, automatic migration guarded by a marker, and **no migrate/rollback CLI commands**.
- **B. Migration robustness + remote storage doubling — fix in this PR.** Replace the brittle `len(globalBlobs) > 0` completion heuristic (the source of the not-resumable P2) with an explicit **success-marker file** written in the storage itself (e.g. a sentinel object/file under `{rootDir}/_blobstore/`, so it travels with local *and* remote data). Semantics: if the marker is absent, run (or re-run) the migration idempotently; write the marker **only after the whole migration succeeds**, so a crash mid-migration simply re-runs on next start. Because downgrade is out of scope (see D), migration can be **destructive on remote**: after a **verified** copy into `_blobstore`, convert the per-repo blob to a 0-byte marker so there is no storage doubling. (Local uses hard links, so no doubling and nothing to convert.) Migration should also re-point the cache `original` to the `_blobstore` path to reconcile a pre-populated cache (P3). Remaining caveat: this remote walk is heavy — stream the copy (P4) and prefer running it off the synchronous startup path (P5).
- **C. Integrity verification on migration copy — fix in this PR (separate commit). Doc: no further update needed beyond this note.** Because remote migration now destructively converts the per-repo blob to a marker, the `_blobstore` copy **must** be digest-verified before the source content is replaced. Local hard links need no verification (same inode).
- **D. Downgrade/rollback — out of scope (removed).** We do **not** support downgrading to a pre-`_blobstore` zot, and will not build rollback tooling. Document downgrade as unsupported for remote (one-way once migrated). (For information only: local happens to remain readable by older zot because per-repo hard links hold full content — but this is not a supported/maintained guarantee.)
- **E. GC robustness — fix in this PR.** Make the cache **advisory, never authoritative** for deletion, plus a drift-proof backstop: **local** reclaims masters via `st_nlink` (Windows needs `GetFileInformationByHandle`); **remote** uses an authoritative index/manifest scan (per #3750 §2.2) to delete unreferenced `_blobstore` objects regardless of cache state.
- **F. Deterministic remote read path on top of #3906 — fix in this PR.** Reuse `_blobstore` as the canonical location, but for remote: (1) reads target `_blobstore` deterministically with per-repo fallback for not-yet-migrated blobs (kills the infinite loop and C1 ambiguity); (2) stop recording remote per-repo paths as cache duplicates for content location; (3) candidates + GC become index scans (E). End state: one ImageStore; **local = hard links + `Nlink`**, **remote = markers + deterministic read + index-scan GC**; cache is an index/optimization, not the GC source of truth.

---

## Interaction with PR #4089 (`fix(storage): release global write lock during blob restore I/O`)

PR #4089 (by `shcherbak`, open, milestone v2.1.18) is a **narrow, near-merge** fix to the *background restore/rebuild* path — distinct from the inline `DedupeBlob`, `deleteBlob`, and GC that #3906 rewrites. It touches `RunDedupeForDigest`, `restoreDedupedBlobs`, `dedupeBlobs` (`imagestore.go:2168-2306`) and the `DedupeTaskGenerator` in `common.go`:

- **Lock narrowing:** `RunDedupeForDigest` holds the global write lock only for the `dedupe=true` (`dedupeBlobs`) path; the restore (`dedupe=false`) path no longer holds it for the whole task.
- **Restore I/O off-lock:** `restoreDedupedBlobs` runs the slow S3 `Stat`/`ReadFile` outside the lock, locking only briefly around `WriteFile` (plus buffer reuse per pass).
- **`_restore_complete` marker:** the `DedupeTaskGenerator` writes a sentinel after a fully successful restore so subsequent `dedupe=false` startups skip the scan entirely (atomic counter + `sync.Once`).

### Conflict assessment

- **vs the dedupe/GC strategy seam (Phase 3 step 10) — low.** Different functions; overlap only if the seam is extended to own `restoreDedupedBlobs`/`dedupeBlobs` (it arguably should). No design conflict — #4089's lock-narrowing is behavior the strategy seam should preserve.
- **vs #3906 — moderate–high.** Three collisions:
  1. **Textual (minor–moderate):** both edit the `imagestore.go:2168-2306` dedupe region and `common.go`'s `DedupeTaskGenerator`. #3906 (marked "don't merge yet") rebases onto #4089, not the reverse.
  2. **Semantic (moderate–high):** `restoreDedupedBlobs`/`dedupeBlobs` are built on the current `binfo.Size() == 0` marker + `getOriginalBlob`. #3906 moves the "original" into the `_blobstore` master and changes the marker meaning, so "restore" becomes "copy from `_blobstore` back into each per-repo path" — the function must be reworked regardless of #4089, and #4089's buffer/lock optimizations must be re-applied to the post-`_blobstore` restore.
  3. **Marker proliferation (most important):** there are now **three** overlapping "skip the expensive startup walk" mechanisms — `#4026`'s metaDB fast-restart stamp (already merged, the base `273b1536`), `#4089`'s `_restore_complete` file marker, and this plan's item **B** success-marker under `{rootDir}/_blobstore/`. `rchincha` already raised this on #4089 ("I am wondering if this belongs in the db we maintain rather than in a file"). These should be **reconciled into one mechanism** before #3906's migration marker lands.
- **vs #2968 — complementary, not conflicting.** #4089 is a tactical narrowing of the global write lock on one path; #2968 replaces the global lock wholesale. #4089 is evidence *for* the per-digest direction (the contention it fixes is exactly what #2968 targets).

### Decision

- **Order:** `#4089` → `#3906`. Merge the small, milestoned #4089 first and rebase the larger work onto it.
- **Reconcile the markers:** pick one completion/skip mechanism (metaDB stamp vs file marker) before #3906's migration marker (item B) lands. Do not ship three sentinels with subtly different semantics.
- **Expect to rewrite restore under `_blobstore`:** treat #4089's `restoreDedupedBlobs` optimizations as something to re-express inside the eventual `remoteMarker` strategy (F), not as code that survives #3906 untouched.

---

## Feasibility of follow-on per-repo / per-digest locking (#2968)

PR #2968 (`feat(storage): enable parallel writes by using per-repo and per-digest locking`, by `andaaron`, open/stalled) is the end goal #3906 was created to unblock — #3906's description states the global storage lock is a bottleneck and *"in order to move to a per-repo lock, we first need to simplify this [dedupe] logic."*

### Verdict: the decided changes make #2968 substantially *more* feasible — they remove its primary blocker
The global `is.Lock()` couldn't be replaced by per-repo locks before because of **cross-repo coupling in dedupe**: the master copy lived in "one of the existing repos," so an operation on repoB's blob could require touching repoA's master — which a per-repo lock can't express without grabbing two repo locks (the deadlock #2968 explicitly warns against: *"Do not lock multiple repos at the same time… Same applies to digests"*).

The decided architecture dissolves that coupling:
- **`_blobstore` master keyed by digest** turns the shared write target into a **per-digest** resource, independent of any repo. That maps directly onto #2968's two-tier scheme: **per-repo lock for index.json operations, per-digest lock for blob/master/cache operations.**
- **Local = hard links + `st_nlink`, cache advisory** is the safest model for concurrency: the FS atomically refuses to free an inode that still has links, so the classic **GC-vs-concurrent-push race cannot lose data** on local.
- **Remote = deterministic path + index-scan GC** makes blob writes idempotent per-digest (no cache-redirect race), so a per-digest lock fully covers master + marker creation.

### Still needs deliberate design (feasible, but not free)
1. **Strict lock ordering.** Most write paths need both a repo lock (index) and a digest lock (blob). Enforce one hierarchy — **always repo-before-digest; never hold two repo locks or two digest locks at once** — or the deadlock returns. `_blobstore` makes this easy because the digest lock is global-per-digest, not nested under a repo.
2. **GC-vs-write race on the shared master (remote).** Index-scan GC has a TOCTOU: it builds the reference set, then deletes an "unreferenced" master while a concurrent push references it. Mitigate by taking the per-digest lock and **re-checking references at delete time**, plus the existing GC `delay` window. Local sidesteps this via `nlink`.
3. **Distributed locks for scale-out.** Multiple zot instances over shared S3 + shared remote cache need the per-digest lock to be **distributed** (Redis redsync, which the cache already provides), not an in-process mutex. This makes the **DynamoDB C2 fix a prerequisite** — without conditional writes/locking, per-digest concurrency on DynamoDB is unsafe.
4. **Migration interaction.** The marker-conversion migration (B) must participate in per-digest locking if it runs while serving (P5 / background migration); simplest if it runs before serving.

### Net
#2968 goes from "blocked on dedupe complexity" to "a natural next step." Framed as phases: **Phase 1 = #3906 + our fixes (simplify shared state into per-digest `_blobstore`, cache advisory); Phase 2 = #2968 (parallelize)**. The remaining hard work is lock-ordering discipline and the remote GC-vs-write race (largely free on local thanks to `nlink`), and it depends on two items already on our list: the **DynamoDB locking fix (C2)** and the **authoritative GC backstop (E)**.

---

## Cherry-pick mechanics onto main

The PR's base (`refactor/storage` @ `273b1536`) is already an ancestor of `main` (it's the `metadb: ... fast restart` commit). So the diff applies relative to a point in main's history. Main has only 4 newer commits on top. Expect **minor conflicts**, most likely in `pkg/storage/storage_test.go` / the `GetAllDedupeReposCandidates` tests, which main's `a675ac96` ("treat dedupe-candidate cache miss as no candidates") also touched. The core `imagestore.go` changes should apply cleanly. CI on the PR was green (41 checks), but patch coverage was only ~70% — and notably the migration's realistic paths (existing cache, partial-failure resume, large S3 blobs) are the under-tested parts.

---

## Recommendation

**Don't cherry-pick #3906 to main as a single commit and ship it.** Instead, cherry-pick it as commit 1 of a PR and land the fixes below as **separate commits in the same PR** (this is the agreed plan; nothing is deferred to a later PR). Architecture is decided: keep the single `ImageStore` with `_blobstore` for both backends; local = hard links + `st_nlink`; remote converges to the deterministic shared path + index-scan GC.

Suggested commit sequence on top of the cherry-pick:

1. **Cache-driver + loop bug fixes:** BoltDB `GetAllBlobs` no-duplicates fall-through (C1); conditional-delete/lock for DynamoDB `DeleteBlob`/`PutBlob` (C2, C4); bounded/fail-fast `DedupeBlob` retry to kill the infinite loop (P1 / C3).
2. **Migration hardening (B/C):** replace the `len(globalBlobs) > 0` heuristic with a **success-marker file** under `{rootDir}/_blobstore/` (resumable, idempotent); re-point a pre-populated cache `original` to `_blobstore` (P3); **integrity-verify** remote copies then convert per-repo blobs to markers (no doubling); stream the copy (P4) and prefer running off the synchronous startup path (P5). No separate migration CLI tool.
3. **GC robustness (E):** cache advisory only; `st_nlink` backstop for local (Windows via `GetFileInformationByHandle`), authoritative index-scan sweep for remote.
4. **Deterministic remote read path (F):** read `_blobstore` directly with per-repo fallback; stop using cache redirect for content location on remote.
5. **Docs (A/D):** reconcile the stale `#3750` proposal doc to the "`_blobstore` for both backends, marker-guarded automatic migration, no migrate/rollback CLI" decision; document remote downgrade as **unsupported** (out of scope).

Independent of all the above: test the realistic paths that #3906 under-covers — existing populated cache, partial-failure resume, large S3 blobs, and a no-hardlink filesystem.

> The detailed, ordered commit-by-commit plan is in **Execution plan** below.

---

## Execution plan (chronological, commit-by-commit)

Ordering principle: land the cherry-pick first, then **correctness bug fixes** (so the base is safe to run), then **migration robustness**, then the **architecture convergence** (deterministic reads → drop doubling → GC backstop), then **docs/tests**, and finally the **follow-on locking work** as a separate PR. Each bug is its own commit (per preference); a few tightly-coupled ones are grouped where splitting would be artificial. Run targeted tests (`go test -tags=... ./pkg/storage/...`, relevant blackbox) after each commit.

### Phase 0 — Prep
1. **Branch + resolve the cherry-pick.** Branch from `main`; `git cherry-pick 3b71bcb`. Resolve the expected conflicts (mainly `pkg/storage/storage_test.go` / `GetAllDedupeReposCandidates` tests vs main's `a675ac96`). This is **commit 1** (the unmodified-as-possible cherry-pick). Validate: build + `pkg/storage` unit tests (local; s3/gcs where mocks allow).

### Phase 1 — Correctness bug fixes (safe to run base) — one commit each
2. **C1 — BoltDB `GetAllBlobs` no-duplicates fall-through.** Move `return nil` out of the `if deduped != nil` block so a digest with only an original returns `[original]` (consistent with Redis/DynamoDB). *Smallest, highest-value fix.*
3. **P1 / C3 — bounded `DedupeBlob` retry.** Replace the unbounded `for {}` self-heal with a capped retry (or fail-fast) so a missing `_blobstore` master with live duplicate refs can't loop forever.
4. **C2 — DynamoDB atomicity.** Add conditional expressions / a per-digest guard to `DeleteBlob`/`PutBlob` so the master item can't be deleted concurrently with a new duplicate. *(Also a prerequisite for #2968 distributed locking.)*
5. **C4 — DynamoDB `PutBlob` swallows `GetBlob` error.** Distinguish `ErrCacheMiss` from a real read error so a transient failure can't overwrite a valid `original`. (Fold C5 style nits into the commit that touches each file.)

### Phase 2 — Migration robustness (non-destructive) — separate commits
6. **P2 — success-marker file.** Replace the `len(globalBlobs) > 0` completion heuristic with a sentinel under `{rootDir}/_blobstore/` written **only after full success**; absent marker ⇒ run/re-run idempotently. Makes migration resumable after a crash.
7. **P3 — reconcile a pre-populated cache.** During migration, re-point the cache `original` to the `_blobstore` path (don't rely on `PutBlob`, which keeps an existing original). Test against an **existing** populated cache.
8. **P4 — stream the remote copy** (`Reader`+`Writer`/`io.Copy`) to bound memory on large layers; **add integrity verification (C)** of the `_blobstore` copy (digest check) before it is trusted. Local stays hard-link (no copy/verify).
9. **P5 — move migration off the synchronous startup path** (background/explicit), so first boot of a large store isn't blocked. (May be combined with step 6 if simpler.)

### Phase 3 — Architecture convergence
10. **Seam — introduce the backend capability/strategy seam.** Add a thin `BlobLifecycle` strategy (or a driver capability: `SupportsHardLinks()` / `LinkOrCopy` / a `LinkCount`/`Dereference` primitive) with `localHardlink` and `remoteMarker` implementations, and route `DedupeBlob`/`deleteBlob`/GC/mount-candidate calls through it. **Pure non-behavioral refactor** — local still hard-links, remote still uses today's path; this just creates the seam that steps 11–13 fill in. Keep s3-vs-gcs differences at the **driver** level (one `ImageStore`, no fork). Land it *after* the Phase 1 bug fixes (so their small diffs aren't tangled in this refactor) and *before* F. Validate: full `pkg/storage` suite green with no behavior change.
11. **F — deterministic remote read path.** Reads target `_blobstore/<digest>` directly with per-repo fallback for not-yet-migrated blobs; stop using the cache redirect for content location on remote. Subsumes **P6** (`canMount`/candidates move to index scans). Fills in the `remoteMarker` read side of the seam. *Must land before the destructive marker conversion in step 12.*
12. **B — remove remote storage doubling.** Now that reads are deterministic, have migration convert each per-repo remote blob to a 0-byte marker after the verified copy (no double storage). One-way by design (downgrade is out of scope, see step 14).
13. **E — GC robustness.** Make the cache **advisory, not authoritative** for deletion; add the drift-proof backstop via the seam: **local** reclaims masters via `st_nlink` (Windows via `GetFileInformationByHandle`); **remote** uses an authoritative index/manifest scan to delete unreferenced `_blobstore` objects.

### Phase 4 — Docs & tests
14. **A/D — docs.** Reconcile the `#3750` proposal doc to the decided design (`_blobstore` for both backends, marker-guarded automatic migration, **no** migrate/rollback CLI); document remote downgrade as **unsupported**.
15. **Tests.** Cover the under-tested paths: existing populated cache, partial-failure/resume migration, large S3 blob (streaming), and a no-hardlink filesystem (dedupe auto-disabled). Keep runs package-scoped per repo testing rules.

### Phase 5 — Follow-on (separate PR)
16. **#2968 — per-repo + per-digest locking.** Replace the global `is.Lock()` with per-repo locks (index ops) + per-digest locks (blob/master/cache ops). Enforce **repo-before-digest** ordering, never two of a kind. For scale-out use the cache's distributed (redsync) per-digest lock — depends on **C2 (step 4)** and the **GC backstop (step 13)**. This is Phase 2 of the larger effort and should not be mixed into the cherry-pick PR.

PR #3954, opened 2026-04-09
Pin actions and tighten workflow permissions

benoittgt:
**What type of PR is this?**

cleanup

N/A. Proactive security hardening based on zizmor static analysis.

We still have area of improvement. See https://github.com/zizmorcore/zizmor/issues/1775#issue-4112001655

You can review commit by commit.

- Addresses all fixable findings from [zizmor](https://docs.zizmor.sh), a static analysis tool for GitHub Actions security, and tightens Dependabot configuration.
- `artipacked` added `persist-credentials: false` to every `actions/checkout` call across 22 workflow files. Without it, the GitHub token is written to `.git/config` and accessible to all subsequent steps and uploaded artifacts.
- `template-injection` moved `${{ github.* }}` expressions out of `run:` blocks and yq `cmd:` inputs into `env:` variables. Direct interpolation lets a crafted value (e.g. a release tag containing shell metacharacters) execute arbitrary code.
- `unpinned-uses` pinned all 24 unique action references to their full 40-char commit SHA using `pinact`, with the version tag preserved as an inline comment. Branch-based refs (`@main`, `@master`, partial SHAs) resolved via GitHub API.

Dependabot grouped Go and GitHub Actions updates into one PR per ecosystem, 14-day cooldown on action updates, fixed Monday schedule. What do you think ? I find it much more digest to handle.

```
zizmor .
346 findings: 172 high, 9 medium, 45 low
```

After this PR: 7 findings remain (5 `excessive-permissions`, 2 `secrets-outside-env`) both require out-of-band repo configuration (GitHub Environments, per-job permission auditing) and are tracked separately.

Re-ran `zizmor .` after each commit to verify findings resolved. 

**Automation added to e2e**:

No, zizmor can be added to CI as a follow-up. 🤷 

No

No

### DIFF
--- a/.github/actions/setup-localstack/action.yaml
+++ b/.github/actions/setup-localstack/action.yaml
@@ -13,12 +13,12 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - uses: docker/login-action@v4
+    - uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
       with:
         registry: ghcr.io
         username: ${{ inputs.username }}
         password: ${{ inputs.password }}
-    - uses: actions/setup-python@v6
+    - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
       with:
         python-version: '3.11'
     - shell: bash
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,8 +10,12 @@ updates:
     open-pull-requests-limit: 10
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
   - package-ecosystem: "github-actions" # See documentation for possible values
     directory: "/" # Location of package manifests
     open-pull-requests-limit: 10
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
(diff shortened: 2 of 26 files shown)


rchincha:
@benoittgt thanks for this PR, we have been meaning to get to this.
However, we also have to be careful with the PR review.

benoittgt:
@rchincha of course. This is not perfect but it is going to the right direction.

rchincha:
jmgilman/actions-generate-checksum: Update from the old v1 branch (2021) to the v1.0.1 release SHA 3ea6dc9bf8eecf28e2ecc982fab683484a1a8561.

mikefarah/yq: Pin to a versioned v4.x.y release commit rather than an arbitrary master HEAD

rchincha:
The PR itself lgtm

However,
https://docs.github.com/en/code-security/concepts/supply-chain-security/about-dependabot-alerts

| Dependabot alerts have some limitations:
| For GitHub Actions, alerts are only generated for actions that use semantic versioning, not SHA versioning.

which means going forward we have to rely on zizmor instead of GH dependabot (at least for now)

and which conflicts with the following guidance

https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/#enforce-sha-pinning

andaaron (approved):

benoittgt:
@rchincha those are good points !

PR #3959, opened 2026-04-10
feat(events): include actor and request metadata in webhook payloads

### LINKED ISSUE #3859: [Feat]: Include actor and request metadata in Webhook Events

I am building a SaaS platform / multi-tenant system on top of Zot and heavily rely on the `events` extension (specifically the `http` sink) to react to pushes, calculate billing, and write audit logs.

While the usage of the CloudEvents standard is fantastic, the event payload itself is missing critical context. Currently, events like `zotregistry.image.updated` only include `name`, `reference`, `digest`, `mediaType`, and `manifest`.

Crucially, **it is impossible to know WHO triggered the event**.
To maintain a proper audit trail or correlate actions to specific users/teams, downstream systems need to know the identity of the authenticated user (`actor`) and ideally metadata about the HTTP request (IP address, user-agent), similar to [how `distribution/distribution` handles its webhook payloads](https://distribution.github.io/distribution/about/notifications/).

Looking at the codebase, this omission seems to be a side effect of where events are fired. The `eventRecorder` is tightly coupled to the storage layer (passed via `storage.New(...)`).
When a push happens, the api layer knows the actor via `reqCtx.UserAcFromContext(request.Context())` in UpdateManifest ([pkg/api/routes.go](https://github.com/project-zot/zot/blob/main/pkg/api/routes.go#L690)). However, this context is not passed down to `imgStore.PutImageManifest(...)`. Thus, when the storage layer finally triggers `ImageUpdated`, the HTTP and Authentication context has been completely lost.

Interestingly, Zot already recognizes the importance of tracking the actor! Right after the storage operation completes in `UpdateManifest`, Zot calls `meta.OnUpdateManifest(request.Context(), ...)`. By passing the HTTP context to the `meta` extension, the database layers (e.g., boltdb.go) are able to extract the user and successfully populate the `PushedBy` field in `DescriptorStatistics`. This data is just currently siloed in the MetaDB and never reaches the event system.

I would like the CloudEvents data payload to include information about the actor and the request. For example:
```json
{
  "name": "space/my-image",
  "reference": "latest",
  "actor": {
    "name": "username_or_token_subject"
  },
  "request": {
    "addr": "192.168.0.1",
    "method": "PUT",
    "useragent": "docker/24.0.5"
  }
}
```

**Querying the GraphQL API / MetaDB**: Knowing that Zot already tracks the actor internally via the `meta` extension (populating the `PushedBy` field in `DescriptorStatistics` within the database), the natural alternative was to simply let the webhook trigger a synchronous GraphQL query back to Zot to ask for the actor. However, this is currently impossible, because the `PushedBy` field is not exposed in the `ImageSummary` GraphQL schema ([pkg/extensions/search/schema.graphql](https://github.com/project-zot/zot/blob/ace12e2a12d0e104f6fb0aecf162b29e8c67b8a9/pkg/extensions/search/schema.graphql#L178)). Therefore, I have no way to retrieve this internal data without parsing raw DB files.

**Auth Token Correlation**: Since my SaaS acts as the identity provider (issuing JWTs for Zot), I could temporarily cache whenever a user requests a push token for a specific repository. When the Zot webhook arrives a few seconds later, I correlate the webhook's name (repo) with the cached token intent. While this works as a band-aid, it adds stateful caching complexity, is susceptible to race conditions, and completely leaves out users relying on standard htpasswd or LDAP authentication.

### Additional context

_No response_

cainydev:
## Summary

Closes #3859

- Adds `actor` (username) and `request` (addr, method, useragent) fields to CloudEvents payloads for `ImageUpdated`, `ImageDeleted`, `ImageLintFailed`, and `RepositoryCreated` events
- Introduces `EventContext`, `ActorInfo`, and `RequestInfo` types in `pkg/extensions/events/common.go` (no build tag — always compiled so the storage layer can reference them unconditionally)
- The API layer (`UpdateManifest`, `DeleteManifest`) builds an `EventContext` from the live HTTP request and attaches it to the context via `events.WithEventContext` before passing it down to storage
- The storage layer (`imagestore`) extracts the context at event-fire time via `events.EventContextFromContext`; nil context (internal/GC paths) cleanly omits both fields from the payload
- Fields are omitted entirely when not present (unauthenticated requests, internal triggers), matching the issue's requested JSON shape

Example payload:
```json
{
  "name": "space/my-image",
  "reference": "latest",
  "digest": "sha256:abc...",
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "actor": { "name": "john" },
  "request": { "addr": "192.168.0.1:54321", "method": "PUT", "useragent": "docker/24.0.5" }
}
```

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -42,6 +42,7 @@ import (
 	"zotregistry.dev/zot/v2/pkg/debug/pprof"
 	debug "zotregistry.dev/zot/v2/pkg/debug/swagger"
 	ext "zotregistry.dev/zot/v2/pkg/extensions"
+	"zotregistry.dev/zot/v2/pkg/extensions/events"
 	"zotregistry.dev/zot/v2/pkg/log"
 	"zotregistry.dev/zot/v2/pkg/meta"
 	mTypes "zotregistry.dev/zot/v2/pkg/meta/types"
@@ -756,6 +757,8 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
 		return
 	}
 
+	ctx := events.WithEventContext(request.Context(), eventContextFromRequest(request))
+
 	if len(digestQueryTags) > 0 && !zcommon.IsDigest(reference) {
 		err := apiErr.NewError(apiErr.MANIFEST_INVALID).AddDetail(map[string]string{
 			"reason": "tag query parameters are only valid when pushing a manifest by digest",
@@ -765,7 +768,7 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
 		return
 	}
 
-	digest, subjectDigest, err := imgStore.PutImageManifest(name, reference, mediaType, body, digestQueryTags)
+	digest, subjectDigest, err := imgStore.PutImageManifest(ctx, name, reference, mediaType, body, digestQueryTags)
 	if err != nil {
 		details := zerr.GetDetails(err)
 		if errors.Is(err, zerr.ErrRepoNotFound) { //nolint:gocritic // errorslint conflicts with gocritic:IfElseChain
@@ -792,7 +795,7 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
 			// could be syscall.EMFILE (Err:0x18 too many opened files), etc
 			rh.c.Log.Error().Err(err).Msg("unexpected error, performing cleanup")
 
-			if err = imgStore.DeleteImageManifest(name, reference, false); err != nil {
+			if err = imgStore.DeleteImageManifest(ctx, name, reference, false); err != nil {
 				// deletion of image manifest is important, but not critical for image repo consistency
 				// in the worst scenario a partial manifest file written to disk will not affect the repo because
 				// the new manifest was not added to "index.json" file (it is possible that GC will take care of it)
@@ -937,7 +940,9 @@ func (rh *RouteHandler) DeleteManifest(response http.ResponseWriter, request *ht
 		return
 	}
 
-	err = imgStore.DeleteImageManifest(name, reference, detectCollision)
+	ctx := events.WithEventContext(request.Context(), eventContextFromRequest(request))
+
+	err = imgStore.DeleteImageManifest(ctx, name, reference, detectCollision)
 	if err != nil { //nolint: dupl
 		details := zerr.GetDetails(err)
 		if errors.Is(err, zerr.ErrRepoNotFound) { //nolint:gocritic // errorslint conflicts with gocritic:IfElseChain
@@ -2499,4 +2504,24 @@ func isSyncOnDemandEnabled(ctlr *Controller) bool {
 	}
 
 	return false
+}
+
+func eventContextFromRequest(r *http.Request) *events.EventContext {
+	ectx := &events.EventContext{
+		Request: &events.RequestInfo{
+			Addr:      r.RemoteAddr,
+			Method:    r.Method,
+			UserAgent: r.UserAgent(),
+		},
+	}
+
+	userAc, err := reqCtx.UserAcFromContext(r.Context())
+	if err == nil && userAc != nil {
+		username := userAc.GetUsername()
+		if username != "" {
+			ectx.Actor = &events.ActorInfo{Name: username}
+		}
+	}
+
+	return ectx
 }
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -1112,7 +1112,8 @@ func (rh *RouteHandler) CheckBlob(response http.ResponseWriter, request *http.Re
 	var blen int64
 
 	if userCanMount {
-		ok, blen, err = imgStore.CheckBlob(name, digest)
+		ctx := events.WithEventContext(request.Context(), eventContextFromRequest(request))
+		ok, blen, err = imgStore.CheckBlob(ctx, name, digest)
 	} else {
 		var lockLatency time.Time
 
@@ -1536,7 +1537,8 @@ func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Requ
 	}
 
 	if rangeHeaderPresent {
-		ok, bsize, err := imgStore.CheckBlob(name, digest)
+		ctx := events.WithEventContext(request.Context(), eventContextFromRequest(request))
+		ok, bsize, err := imgStore.CheckBlob(ctx, name, digest)
 		if err != nil {
 			writeBlobError(err)
 
@@ -1724,6 +1726,8 @@ func (rh *RouteHandler) CreateBlobUpload(response http.ResponseWriter, request *
 
 	imgStore := rh.getImageStore(name)
 
+	ctx := events.WithEventContext(request.Context(), eventContextFromRequest(request))
+
 	// currently zot does not support cross-repository mounting, following dist-spec and returning 202
 	if mountDigests, ok := request.URL.Query()["mount"]; ok {
 		if len(mountDigests) != 1 {
@@ -1755,11 +1759,11 @@ func (rh *RouteHandler) CreateBlobUpload(response http.ResponseWriter, request *
 		// check blob looks for actual path (name+mountDigests[0]) first then look for cache and
 		// if found in cache, will do hard link and if fails we will start new upload.
 		if userCanMount {
-			_, _, err = imgStore.CheckBlob(name, mountDigest)
+			_, _, err = imgStore.CheckBlob(ctx, name, mountDigest)
 		}
 
 		if err != nil || !userCanMount {
-			upload, err := imgStore.NewBlobUpload(name)
+			upload, err := imgStore.NewBlobUpload(ctx, name)
 			if err != nil {
 				details := zerr.GetDetails(err)
 				if errors.Is(err, zerr.ErrRepoNotFound) {
@@ -1833,7 +1837,7 @@ func (rh *RouteHandler) CreateBlobUpload(response http.ResponseWriter, request *
 			return
 		}
 
-		sessionID, size, err := imgStore.FullBlobUpload(name, request.Body, digest)
+		sessionID, size, err := imgStore.FullBlobUpload(ctx, name, request.Body, digest)
 		if err != nil {
 			rh.c.Log.Error().Err(err).Int64("actual", size).Int64("expected", contentLength).
 				Msg("failed to full blob upload")
@@ -1856,7 +1860,7 @@ func (rh *RouteHandler) CreateBlobUpload(response http.ResponseWriter, request *
 		return
 	}
 
-	upload, err := imgStore.NewBlobUpload(name)
+	upload, err := imgStore.NewBlobUpload(ctx, name)
 	if err != nil {
 		details := zerr.GetDetails(err)
 		if errors.Is(err, zerr.ErrRepoNotFound) {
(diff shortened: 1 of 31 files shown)


rchincha:
https://github.com/project-zot/zot/pull/3962

Take a look at this as well.

cainydev:
yes i saw it i will work on this PR a bit. I had it all figured out and then broke it on the rebase. copilot had a similar solution as this one.

rchincha:
@cainydev are you planning on working on this?

cainydev:
> @cainydev are you planning on working on this?

yes - i just couldn't figure out why the ci failed tbh. maybe flaky?

rchincha:
The fix should target the failing blackbox test behind this workflow step:

Job definition: [.github/workflows/ecosystem-tools.yaml](https://github.com/project-zot/zot/blob/1e81b7d7c309b4c83537e72257501c64f883d472/.github/workflows/ecosystem-tools.yaml#L74-L80)
Failing step: make run-blackbox-ci
The failure itself is in the regclient image pull path, not in the workflow YAML. The relevant log lines are:

PullImage from image service failed
rpc error: code = Canceled desc = copying config: context canceled
level=fatal msg="pulling image: rpc error: code = Canceled desc = copying config: context canceled"
This happened immediately after the test:

ok 20 [release] push OCI artifact with regclient
So the likely broken case is the next regclient-related test that pulls an image through CRI-O/containerd after push.

Solution
Make the regclient blackbox test wait until the pushed manifest and its config blob are fully readable before calling the runtime pull.

The logs show many blob uploads completing with PUT, then the runtime pull fails while copying config. That points to a race: the test proceeds as soon as push returns, but CRI pulls before the config blob is consistently available to the image service.

Recommended code change
In the failing regclient test, add a readiness loop after push and before the runtime pull:

bash
manifest_digest="$(regctl manifest head localhost:${PORT}/test-regclient --format '{{.Digest}}')"

# wait for manifest to be readable
for i in $(seq 1 30); do
  if regctl manifest get "localhost:${PORT}/test-regclient@${manifest_digest}" >/dev/null 2>&1; then
    break
  fi
  sleep 1
done

# resolve and wait for config blob
config_digest="$(regctl manifest get "localhost:${PORT}/test-regclient@${manifest_digest}" --format '{{ .Config.Digest }}')"

for i in $(seq 1 30); do
  if curl -fsI "http://localhost:${PORT}/v2/test-regclient/blobs/${config_digest}" >/dev/null 2>&1; then
    break
  fi
  sleep 1
done

sudo crictl pull "localhost:${PORT}/test-regclient@${manifest_digest}"
If the test uses a tag instead of digest, prefer switching the runtime pull to the digest form after push. That removes any tag propagation timing issue.

Why this should fix it
The error is specifically:

copying config: context canceled
That usually means the runtime started pulling, requested the image config blob, and the request got canceled because the content was not yet stably available. Since the push had just completed, the most likely cause is eventual visibility/race in the local registry storage path rather than a bad image.

Additional hardening suggestion
If the test currently does something like:

bash
regctl image copy ...
crictl pull localhost:${PORT}/test-regclient
change it to:

bash
digest="$(regctl image digest localhost:${PORT}/test-regclient:latest)"
crictl pull "localhost:${PORT}/test-regclient@${digest}"
and keep the wait loop above.

Summary
The practical fix is:

Locate the failing regclient pull test in the blackbox suite run by make run-blackbox-ci.
After pushing the image/artifact, poll until:
the manifest is readable, and
the config blob returns success on HEAD.
Pull by digest instead of tag.
That directly addresses the race shown by the logs and is the most likely way to make this job stable.

cainydev:
@rchincha Pushed a rebase and two follow-up commits.

The first addresses Copilot's review on NewBlobUpload. I threaded the request context through the blob-upload entry points (NewBlobUpload, FullBlobUpload, PutBlobChunk, PutBlobChunkStreamed, CheckBlob) so RepositoryCreated events fired on first push or cross-repo mount carry actor and request metadata.

On the CI failure: looking at the logs, zot serves every blob with 200 OK and the cancellation happens ~14s later inside cri-o itself. The same failure reproduces on main, so it's a pre-existing runner flake. The second commit wraps crictl pull in a bounded retry so the test stops blocking PRs.

cainydev:
@rchincha Can you maybe start Actions for this one?

rchincha:
> @rchincha Can you maybe start Actions for this one?

Done. Let's try to target this for next release window.

pkg/extensions/sync/destination.go
```diff
@@ -217,7 +217,7 @@ func (registry *DestinationRegistry) copyManifest(repo string, desc ispec.Descri
 			return err
 		}
 
-		digest, _, err := imageStore.PutImageManifest(repo, reference,
+		digest, _, err := imageStore.PutImageManifest(context.Background(), repo, reference,
```
vrajashkr, line 220: The functions in this file's diff are carried out by the system during a sync. I think that it would be great to have a "system" user actor to track automated operations.

@rchincha @andaaron what do you think about a "system" actor?

That said, that need not be part of this PR and can be taken up separately if it may be useful.

test/blackbox/helpers_upgrade.bash
```diff
@@ -311,7 +311,7 @@ function helper_pull_manifest_with_docker_client() {
 # Helper: Pull manifest with crictl
 function helper_pull_manifest_with_crictl() {
     local zot_port=$(get_zot_port)
-    run crictl pull localhost:${zot_port}/test-regclient
+    run crictl_pull_with_retry localhost:${zot_port}/test-regclient
```
vrajashkr: While I like the idea of having a retry, I'm worried about whether there is a possibility that having retries here might accidentally mask future regressions where functionality is restored by retrying a pull.
That said, more often that not a retry of the action would fix it so it is probably fair to have a retry here.

cc: @rchincha

vrajashkr (commented):
@cainydev thanks for the PR!

The changes look good to me overall. Left a couple of comments

I think there is a file where the context hasn't been added in yet:
```
Error: pkg/extensions/monitoring/monitoring_test.go:539:87: not enough arguments in call to imgStore.FullBlobUpload
```

pkg/extensions/events/common.go
```diff
@@ -22,11 +23,43 @@ func (e EventType) String() string {
...
+
+// RequestInfo describes the HTTP request that triggered an event.
+type RequestInfo struct {
+	Addr      string `json:"addr"`
+	Method    string `json:"method"`
+	UserAgent string `json:"useragent"`
+}
+
+// EventContext carries actor and request metadata for events.
+type EventContext struct {
```
andaaron, line 39: Is this structure (to be more specific the field names and types) something expected by 3rd party libraries we integrate with are receiving the events?

pkg/storage/imagestore/imagestore.go
```diff
@@ -762,7 +759,14 @@ func (is *ImageStore) PutImageManifest(repo, reference, mediaType string, //noli
 			Msg("linter didn't pass")
 
 		if is.events != nil {
-			is.events.ImageLintFailed(repo, reference, mDigest.String(), mediaType, string(body))
+			evCtx := events.EventContextFromContext(ctx)
+			if len(changedTags) > 0 {
+				for _, tag := range changedTags {
+					is.events.ImageLintFailed(repo, tag, mDigest.String(), mediaType, string(body), evCtx)
```
andaaron: Do we really need all these events?
The previous implementation only emitted once, since the tag itself is not subject to linting.
Now we send an event for every tag with the same result, instead of a single event per manifest.

Does the user care the lint failed on a specific manifest (regardless of tag) or a specific image (in which case we should send for all tags)?

pkg/storage/imagestore/imagestore.go
```diff
@@ -753,7 +751,6 @@ func (is *ImageStore) PutImageManifest(repo, reference, mediaType string, //noli
 		desc.ArtifactType = artifactType
 
 		lintDesc = desc
```
andaaron, line 754:
```suggestion
		lintDesc = desc
		changedTags = []string{reference}
```

This way you can later use simply
```
		for _, ref := range changedTags {
			is.events.ImageUpdated(repo, ref, mDigest.String(), mediaType, string(body), evCtx)
		}
```
instead of
```
		if len(changedTags) > 0 {
			for _, tag := range changedTags {
				is.events.ImageUpdated(repo, tag, mDigest.String(), mediaType, string(body), evCtx)
			}
		} else {
			is.events.ImageUpdated(repo, reference, mDigest.String(), mediaType, string(body), evCtx)
		}
```

pkg/extensions/sync/destination.go
andaaron: Makes sense to me

cainydev:
Rebased onto latest main and pushed updates addressing the review feedback and the CI failure.

CI failure (lint + extensions tests). This was a build break, not flakiness. After main merged the trivy SBOM artifact feature and a new GC-metrics test, both call ImageStore methods whose signatures this PR changes to take context.

@andaaron
- lint events: Good point. lint is a property of the manifest, not the tags. Reverted to a single ImageLintFailed per manifest instead of one per changed tag, and updated the test accordingly.
- event-emission refactor: changedTags is now set to the path reference in the non-extra-tags path, so ImageUpdated emission collapses to a single loop over changedTags instead of the if/else fallback.
- actor/request field shape: No third-party contract dictates these. The shape is our own, modeled on the events the CNCF distribution registry emits. Its not a 1:1 but close.

@vrajashkr
- "system" actor for sync: Agree it'd be useful for tracking automated operations. Keeping it out of scope here as you suggested. can be a follow-up.

pkg/storage/imagestore/imagestore.go
```diff
@@ -716,8 +716,6 @@ func (is *ImageStore) PutImageManifest(repo, reference, mediaType string, //noli
 				ispec.AnnotationRefName: changedTags[0],
```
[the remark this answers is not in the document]
andaaron: This is guarded by anyIndexChange

pkg/storage/imagestore/imagestore.go
```diff
@@ -762,7 +760,10 @@ func (is *ImageStore) PutImageManifest(repo, reference, mediaType string, //noli
 			Msg("linter didn't pass")
 
 		if is.events != nil {
-			is.events.ImageLintFailed(repo, reference, mDigest.String(), mediaType, string(body))
+			// lint is a property of the manifest, not of the tag(s) it is applied under,
+			// so a single event is emitted per manifest regardless of how many tags changed.
+			is.events.ImageLintFailed(repo, changedTags[0], mDigest.String(), mediaType, string(body),
+				events.EventContextFromContext(ctx))
```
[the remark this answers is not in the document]
andaaron: This is not the case here

pkg/extensions/events/common.go
```diff
@@ -22,11 +23,43 @@ func (e EventType) String() string {
...
+func WithEventContext(ctx context.Context, ec *EventContext) context.Context {
+	return context.WithValue(ctx, eventContextKey{}, ec)
+}
+
+// EventContextFromContext retrieves the EventContext from a context.Context.
+func EventContextFromContext(ctx context.Context) *EventContext {
+	ec, _ := ctx.Value(eventContextKey{}).(*EventContext)
+
+	return ec
+}
```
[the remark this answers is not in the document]
andaaron: I would include this fix

pkg/api/routes.go
```diff
@@ -2882,3 +2895,23 @@ func isSyncOnDemandEnabled(ctlr *Controller) bool {
 
 	return false
 }
+
+func eventContextFromRequest(r *http.Request) *events.EventContext {
+	ectx := &events.EventContext{
+		Request: &events.RequestInfo{
+			Addr:      r.RemoteAddr,
```
andaaron, line 2959: `request.addr` is populated from `r.RemoteAddr`. For deployments behind a reverse proxy or ingress, this will usually be the proxy address, not the end-client IP.

I think this is acceptable in this specific case, @cainydev?

andaaron:
Also please rebase.

pkg/api/routes.go
cainydev: Yes, acceptable for this PR. r.RemoteAddr is the immediate peer, so behind a proxy/ingress it's the proxy address.

andaaron (dismissed):

andaaron (approved):

PR #3963, opened 2026-04-12
fix: Updating a repository should not result in a corrupted index.json file if disk is full

andaaron:
See https://github.com/project-zot/zot/issues/3924

### DIFF
--- a/pkg/storage/local/driver_unix.go
+++ b/pkg/storage/local/driver_unix.go
@@ -1,0 +1,11 @@
+//go:build !windows
+
+package local
+
+import "os"
+
+// renameReplace moves src to dst, replacing dst if it already exists (POSIX rename).
+// commit is ignored on Unix; durability for committed writers is handled via fsync in fileWriter.Close.
+func renameReplace(src, dst string, _ bool) error {
+	return os.Rename(src, dst)
+}
--- a/pkg/storage/storage_test.go
+++ b/pkg/storage/storage_test.go
@@ -10,9 +10,11 @@ import (
 	"io"
 	"os"
 	"path"
+	"path/filepath"
 	"slices"
 	"strings"
 	"sync"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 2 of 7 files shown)


rchincha:
Did we root cause this bug? Or is this a speculative fix?

rchincha:
Also is it not possible to figure out if we are doing to run out of disk space before modifying the storage per the request?

andaaron:
> Also is it not possible to figure out if we are doing to run out of disk space before modifying the storage per the request?

Zot is not a tool to monitor disk usage. How would you even monitor that, run `df`/`du` every time before writing files?

> Did we root cause this bug? Or is this a speculative fix?

If you mean to ask if I debugged the issue on the reporter's setup or reproduced it locally, then the answer is no.
But there is nothing speculative about the error path below, see `pkg/storage/local/driver.go` (`Driver.Writer` and `Driver.WriteFile`).

1. **Overwrite begins**
   - https://github.com/project-zot/zot/pull/3963/changes#diff-431cd64995137c385aa1a7a3497fd441e9f217fa55d206f63ce83790e802511dL1775 calls `driver.WriteFile(filepath, newBytes)` intending to write/replace the file’s contents.

2. **A writer is opened with `append=false`**
   - `WriteFile` calls `driver.Writer(filepath, false)`.

3. **The target file is opened for write (create if missing)**
   - `Writer` creates the parent directory (`os.MkdirAll`), then opens the file with `os.OpenFile(filepath, os.O_WRONLY|os.O_CREATE, ...)`
   - If the file already exists, it is opened for writing.

4. **Existing content is destroyed immediately (truncate)**
   - Because `append` is `false`, `Writer` truncates the file using `file.Truncate(0)`
   - At this point, the **original index content is no longer present as file content** (the file length becomes 0). There is no “old version” preserved by this code path.

5. **New content starts being written**
   - `WriteFile` then copies data into the writer using `io.Copy(writer, bytes.NewReader(content))`
   - The local `fileWriter` uses a buffered writer (`bufio.Writer`), so some bytes may be written before an error is surfaced.

6. **Disk fills during the write**
   - If the filesystem runs out of space while writing or flushing, the OS returns `ENOSPC` (“no space left on device”)
   - This error can be reported during `io.Copy(...)` (when the buffer flushes) or later at close/commit time depending on timing.

7. **On copy error, the write is cancelled**
   - If `io.Copy` returns an error, `WriteFile` executes `writer.Cancel(context.Background())`

8. **Cancel removes the file**
   - `Cancel()` closes the file and then calls `os.Remove(fw.file.Name())`.

9. **Net result**
   - Because truncation happened before new content was safely written, an `ENOSPC` during the overwrite means:
     - the original contents have already been destroyed (truncate-to-zero)
     - and the error path removes the file entirely (Cancel → Remove)


The safe pattern for replacing critical metadata is:
- write new content to a **separate staging file**, any failure, including disk space errors, at this step leaves the original index file intact
- then **rename/replace into place** as a final step

This preserves the original file unless the new version is fully written and the final rename succeeds.

rchincha:
In a disk-full situation, a copy of .uploads is likely to fail also. So the protection is the _original_ index.json is unaffected.

pkg/storage/local/driver_unix.go
```diff
@@ -0,0 +1,11 @@
+//go:build !windows
+
+package local
+
+import "os"
+
+// renameReplace moves src to dst, replacing dst if it already exists (POSIX rename).
+// commit is ignored on Unix; durability for committed writers is handled via fsync in fileWriter.Close.
+func renameReplace(_ bool, src, dst string) error {
```
vrajashkr: nit: it would be nice if the params order was `src, dst string, commit bool` instead. Since the bool is more of a modifier for the behavior rather than a primary rename input.

This works just fine either ways though so just a nit.

pkg/storage/storage_test.go
```diff
@@ -3901,3 +3937,135 @@ func DumpKeys(t *testing.T, redisURL string) {
...
+		if err, ok := h.moveHook(src, dst); ok {
+			return err
+		}
+	}
+
+	return h.Driver.Move(src, dst)
+}
+
+func TestPutIndexContent_atomicReplace(t *testing.T) {
+	Convey("PutIndexContent stages under .uploads then renames (atomic replace, GitHub #3924)", t, func() {
```
vrajashkr: nit: I don't think the Github issue is required in the code base. The PR tracking will take care of that :)

vrajashkr (commented):
LGTM overall. Left a couple of nits.

andaaron:
> In a disk-full situation, a copy of .uploads is likely to fail also. So the protection is the _original_ index.json is unaffected.

Exactly

rchincha (approved):

PR #518, opened 2022-05-03
dependabot alert: fix CVE-2022-29810

rchincha:
https://github.com/project-zot/zot/security/dependabot/14

Signed-off-by: Ramkumar Chinchani <rchincha@cisco.com>

**What type of PR is this?**

bug

**Which issue does this PR fix**:

https://github.com/project-zot/zot/security/dependabot/14

**What does this PR do / Why do we need it**:

Fixes a medium CVE

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3975, opened 2026-04-17
feat(auth): support OIDC RP-Initiated Logout

krjakbrjak:
Implements the server-side half of [OpenID Connect RP-Initiated Logout 1.0](https://openid.net/specs/openid-connect-rpinitiated-1_0.html). When a session was established via an OIDC provider whose discovery document advertises an `end_session_endpoint`, `POST /zot/auth/logout` now returns a JSON body containing an `end_session_url` the client can navigate to in order to terminate the session at the IdP in addition to clearing zot's local cookie.

Before this change, "logout" only invalidated zot's own session cookie. The IdP's SSO session stayed alive, so clicking "Login with OIDC" again immediately re-authenticated the user without any credential prompt — effectively not a logout.

This PR is similar to the [one merged in gitea](https://github.com/go-gitea/gitea/pull/36724).

The frontend changes for this PR are implemented in [zui PR](https://github.com/project-zot/zui/pull/518).

**What type of PR is this?**

* Configure OIDC
* Sign in via OIDC provider
* Logout
* Sign in again -> The user will be logged in immediately without asking credentials

**Automation added to e2e**:

No

```release-note

```

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -208,4 +208,5 @@ var (
 	ErrOIDCAudienceMismatch             = errors.New("token audience does not match any of the expected audiences")
 	ErrCertificateNotLoaded             = errors.New("tls certificate not yet loaded")
 	ErrCertificateWatcherAlreadyRunning = errors.New("certificate watcher is already running")
+	ErrInvalidEndSessionEndpoint        = errors.New("end_session_endpoint must be an absolute http(s) URL")
 )
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -204,7 +204,7 @@ func (amw *AuthnMiddleware) basicAuthn(ctlr *Controller, userAc *reqCtx.UserAcce
 		// saved logged session only if the request comes from web (has UI session header value)
 		if hasSessionHeader(request) {
 			secure := ctlr.Config.UseSecureSession()
-			if err := saveUserLoggedSession(cookieStore, response, request, identity, secure, ctlr.Log); err != nil {
+			if err := saveUserLoggedSession(cookieStore, response, request, identity, "", secure, ctlr.Log); err != nil {
 				return false, err
 			}
 		}
@@ -243,7 +243,7 @@ func (amw *AuthnMiddleware) basicAuthn(ctlr *Controller, userAc *reqCtx.UserAcce
 			// saved logged session only if the request comes from web (has UI session header value)
 			if hasSessionHeader(request) {
 				secure := ctlr.Config.UseSecureSession()
-				if err := saveUserLoggedSession(cookieStore, response, request, identity, secure, ctlr.Log); err != nil {
+				if err := saveUserLoggedSession(cookieStore, response, request, identity, "", secure, ctlr.Log); err != nil {
 					return false, err
 				}
 			}
@@ -901,6 +901,24 @@ func NewRelyingPartyGithub(config *config.Config, provider string, hashKey, encr
 	return relyingParty
 }
 
+// originFromConfig returns the server's base URL (scheme + host[:port], no trailing
+// slash) used as the origin for OIDC redirect URIs. It prefers ExternalURL; otherwise
+// it derives the origin from cfg.HTTP.Address, cfg.HTTP.Port, and cfg.HTTP.TLS. Using
+// a single source for both the login redirect_uri and the logout
+// post_logout_redirect_uri ensures the IdP sees matching origins.
+func originFromConfig(cfg *config.Config) string {
+	if trimmed := strings.TrimSuffix(cfg.HTTP.ExternalURL, "/"); trimmed != "" {
+		return trimmed
+	}
+
+	scheme := constants.SchemeHTTP
+	if cfg.HTTP.TLS != nil {
+		scheme = constants.SchemeHTTPS
+	}
+
+	return scheme + "://" + net.JoinHostPort(cfg.HTTP.Address, cfg.HTTP.Port)
+}
+
 func getRelyingPartyArgs(cfg *config.Config, provider string, hashKey, encryptKey []byte, log log.Logger) (
 	string, string, string, string, []string, []rp.Option,
 ) {
@@ -918,26 +936,11 @@ func getRelyingPartyArgs(cfg *config.Config, provider string, hashKey, encryptKe
 		scopes = append([]string{oidc.ScopeOpenID}, scopes...)
 	}
 
-	port := cfg.HTTP.Port
 	issuer := providerConfig.Issuer
 	keyPath := providerConfig.KeyPath
-	baseURL := net.JoinHostPort(cfg.HTTP.Address, port)
 
 	callback := constants.CallbackBasePath + "/" + provider
-
-	var redirectURI string
-
-	if cfg.HTTP.ExternalURL != "" {
-		externalURL := strings.TrimSuffix(cfg.HTTP.ExternalURL, "/")
-		redirectURI = fmt.Sprintf("%s%s", externalURL, callback)
-	} else {
-		scheme := constants.SchemeHTTP
-		if cfg.HTTP.TLS != nil {
-			scheme = constants.SchemeHTTPS
-		}
-
-		redirectURI = fmt.Sprintf("%s://%s%s", scheme, baseURL, callback)
-	}
+	redirectURI := originFromConfig(cfg) + callback
 
 	options := []rp.Option{
 		rp.WithVerifierOpts(rp.WithIssuedAtOffset(issuedAtOffset)),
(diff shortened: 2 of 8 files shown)


krjakbrjak:
@rchincha @andaaron latest push addresses the Copilot feedback. Thanks!

pkg/api/routes.go
```diff
@@ -2213,6 +2323,22 @@ func (rh *RouteHandler) OpenIDCodeExchangeCallbackWithProvider(providerName stri
...
+// save, keeping its Secure/HttpOnly/SameSite attributes on the second Save.
+func recordSessionProvider(ctlr *Controller, w http.ResponseWriter, r *http.Request, provider string) error {
+	session, err := ctlr.CookieStore.Get(r, "session")
+	if err != nil {
+		return err
+	}
+
+	session.Values["provider"] = provider
+
+	return session.Save(r, w)
```
andaaron: saveUserLoggedSession already calls session.Save(r, w) during login, and then recordSessionProvider does another session.Save(r, w) on the same session to write provider. This emits two Set-Cookie: session=... headers per OIDC login. Can we thread provider into saveUserLoggedSession (or a sibling helper) so the login path does a single session.Save? That also removes the in-request registry-caching assumption that recordSessionProvider's comment currently relies on.

pkg/api/routes.go
```diff
@@ -2038,20 +2025,38 @@ func (rh *RouteHandler) ListExtensions(w http.ResponseWriter, r *http.Request) {
...
-// @Success 200 {string} string "ok"
+// @Success 200 {object} api.LogoutResponse
 // @Failure 500 {string} string "internal server error"
 func (rh *RouteHandler) Logout(response http.ResponseWriter, request *http.Request) {
 	if request.Method == http.MethodOptions {
 		return
 	}
 
 	session, _ := rh.c.CookieStore.Get(request, "session")
+	provider, _ := session.Values["provider"].(string)
```
andaaron, line 2083: Existing code discards this error too, but now we also read session.Values["provider"] from the returned session. If the cookie fails to decrypt (key rotation, corrupted cookie), we get a fresh session with provider == "" and silently degrade to local-only logout. Please log at debug/warn like GetAuthUserFromRequestSession does — operators will thank us when diagnosing "my RP-Initiated Logout stopped working after I rotated sessionHashKey."

pkg/api/logout_internal_test.go
```diff
@@ -0,0 +1,558 @@
...
+		t.Parallel()
+
+		rh := newTestRouteHandler(t)
+		req := httptest.NewRequest(http.MethodOptions, "/zot/auth/logout", nil)
+		rec := httptest.NewRecorder()
+
+		rh.Logout(rec, req)
+
+		// The handler returns before touching the response, so status stays at the
+		// recorder default (200) and body is empty.
```
andaaron, line 270: Please add assert.Equal(t, http.StatusOK, rec.Result().StatusCode) so a regression that writes an error status before the early OPTIONS return is caught.

pkg/api/routes.go
```diff
@@ -2061,7 +2066,100 @@ func (rh *RouteHandler) Logout(response http.ResponseWriter, request *http.Reque
...
+		return trimmed + "/login"
+	}
+
+	host := request.Host
+	if host == "" {
+		return ""
+	}
+
+	scheme := "http"
+	if request.TLS != nil || strings.EqualFold(request.Header.Get("X-Forwarded-Proto"), "https") {
```
andaaron: When http.externalUrl is unset, postLogoutRedirectURI derives the origin from request.Host + X-Forwarded-Proto, but the matching login-side redirect_uri in getRelyingPartyArgs (pkg/api/authn.go:~922) uses cfg.HTTP.Address:Port + cfg.HTTP.TLS. In the same deployment these two rules can produce different origins, and both need to be pre-registered on the IdP. Can we align the fallbacks — either both use request-derived values, or (preferred) both use the configured Address:Port/TLS? A single shared helper would make this impossible to drift.

andaaron: The fallback honors X-Forwarded-Proto but reads request.Host directly, ignoring X-Forwarded-Host. Behind a reverse proxy that's half-proxy-aware: we'd trust the proxy for the scheme but not the host. Either also honor X-Forwarded-Host, or drop the X-Forwarded-Proto special case and tell users that proxied deployments must set http.externalUrl. Right now it's easy to end up with https://internal-lb:5000/login sent to the IdP, which won't match the registered URI.

pkg/api/routes.go
```diff
@@ -2038,20 +2025,38 @@ func (rh *RouteHandler) ListExtensions(w http.ResponseWriter, r *http.Request) {
 
 // The following routes are specific to zot and NOT part of the OCI dist-spec
 
+// errInvalidEndSessionEndpoint is returned by composeEndSessionURL when the URL read
+// from the IdP's discovery document is not a usable absolute http(s) URL — see
+// https://openid.net/specs/openid-connect-rpinitiated-1_0.html#OPMetadata.
+var errInvalidEndSessionEndpoint = errors.New("end_session_endpoint must be an absolute http(s) URL")
```
krjakbrjak: changed in b644d6579c53d5f4b98b12fbc841730d5ae43529

pkg/api/routes.go
krjakbrjak: changed in 6990a6b865b225296f3f04270104ec18d4efd2e3

krjakbrjak:
Thanks for the review @andaaron. I pushed the fixups.

pkg/api/routes.go
```diff
@@ -2080,7 +2162,7 @@ func (rh *RouteHandler) GithubCodeExchangeCallback() rp.CodeExchangeCallback[*oi
 			return
 		}
 
-		callbackUI, err := OAuth2Callback(rh.c, w, r, state, email, groups) //nolint: contextcheck
+		callbackUI, err := OAuth2Callback(rh.c, w, r, state, email, "", groups) //nolint: contextcheck
 		if err != nil {
 			if errors.Is(err, zerr.ErrInvalidStateCookie) {
 				w.WriteHeader(http.StatusUnauthorized)
```
andaaron, line 2192: These branches need the returns as well. same as lines 2284/2289, correct?

errors/errors.go
```diff
@@ -208,4 +208,5 @@ var (
 	ErrOIDCAudienceMismatch             = errors.New("token audience does not match any of the expected audiences")
 	ErrCertificateNotLoaded             = errors.New("tls certificate not yet loaded")
 	ErrCertificateWatcherAlreadyRunning = errors.New("certificate watcher is already running")
+	ErrInvalidEndSessionEndpoint        = errors.New("endSessionEndpoint must be an absolute http(s) URL")
```
andaaron: The OIDC Discovery field is end_session_endpoint (snake_case). Using the same spelling in the error message makes it easier for operators to correlate with the IdP's discovery document when they see this log line.

andaaron:
Please squash the commits afterwards, and make sure the commit msg matches the mandatory style in the GH check.

krjakbrjak:
> Please squash the commits afterwards, and make sure the commit msg matches the mandatory style in the GH check.

Done

pkg/api/routes.go
krjakbrjak: right

andaaron (dismissed):

andaaron:
@krjakbrjak can you rebase this PR?

krjakbrjak:
> @krjakbrjak can you rebase this PR?

Done

pkg/api/routes.go
```diff
@@ -2072,7 +2090,84 @@ func (rh *RouteHandler) Logout(response http.ResponseWriter, request *http.Reque
 		return
 	}
 
-	response.WriteHeader(http.StatusOK)
+	zcommon.WriteJSON(response, http.StatusOK, LogoutResponse{
+		EndSessionURL: rh.buildEndSessionURL(provider),
+	})
```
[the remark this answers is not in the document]
andaaron: The previous response body was empty (response.WriteHeader(http.StatusOK) with no body write), not the string "ok" — the Swagger annotation was stale. An empty body → {} transition on a JSON-producing endpoint is not a semantic breaking change, and the sole client (zui PR #518) is being updated alongside this PR.

pkg/api/authn.go
```diff
@@ -1071,6 +1074,10 @@ func saveUserLoggedSession(cookieStore sessions.Store, response http.ResponseWri
 	session.Values["authStatus"] = true
 	session.Values["user"] = identity
 
+	if provider != "" {
+		session.Values["provider"] = provider
```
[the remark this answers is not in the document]
andaaron: This seems to be an actual bug @krjakbrjak please check this out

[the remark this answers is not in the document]
krjakbrjak: Fixed: saveUserLoggedSession now deletes the provider key when none is passed, so a stale OIDC value can't survive a non-OIDC re-login. Added a regression test.

andaaron (approved):

PR #3977, opened 2026-04-17
fix(security): limit manifest PUT body to 4 MiB (INPUT-1)

rchincha:
Wrap request.Body with http.MaxBytesReader before io.ReadAll in UpdateManifest. Bodies exceeding MaxManifestBodySize (4 MiB) now return HTTP 413 with a MANIFEST_INVALID error body instead of buffering unlimited data into memory.

Add the MaxManifestBodySize constant and a unit test that sends an oversized body and asserts the 413 status.

Agent-Logs-Url: https://github.com/project-zot/zot/sessions/5eca86eb-9749-4cf8-9fb8-7b9ace2ba87f

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -689,6 +689,7 @@ func (rh *RouteHandler) GetReferrers(response http.ResponseWriter, request *http
 // @Header  201 {string} OCI-Tag "Echoed tag= value; this header is repeatable (one field per tag= query parameter)"
 // @Failure 400 {string} string "bad request"
 // @Failure 404 {string} string "not found"
+// @Failure 413 {string} string "request entity too large"
 // @Failure 414 {string} string "too many tag query parameters"
 // @Failure 500 {string} string "internal server error"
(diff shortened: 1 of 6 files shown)


pkg/api/routes.go
```diff
@@ -765,6 +756,24 @@ func (rh *RouteHandler) UpdateManifest(response http.ResponseWriter, request *ht
 		return
 	}
 
+	body, err := io.ReadAll(http.MaxBytesReader(response, request.Body, constants.MaxManifestBodySize))
+	// hard to reach test case, injected error (simulates an interrupted image manifest upload)
+	// err could be io.ErrUnexpectedEOF or *http.MaxBytesError
+	if err := inject.Error(err); err != nil {
```
andaaron, line 762: Let's not use the inject anymore, please.

andaaron: Actually this was already there. We can skip last note.

andaaron (approved):

PR #3978, opened 2026-04-17
fix(security): limit API key creation body to 4 KiB (INPUT-2)

rchincha:
Wrap req.Body with http.MaxBytesReader before io.ReadAll in CreateAPIKey. Requests with bodies larger than MaxAPIKeyBodySize (4 KiB) now return HTTP 413 instead of buffering arbitrary data.

Add the MaxAPIKeyBodySize constant, update the Swagger @Failure annotation to document 413, and add a unit test.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/constants/consts.go
+++ b/pkg/api/constants/consts.go
@@ -22,7 +22,9 @@ const (
 	MaxManifestDigestQueryTags = (8192 - 2048) / (len("tag=") + 128 + 1)
 	// MaxManifestBodySize is the maximum number of bytes accepted for a manifest PUT request body.
 	// OCI manifest JSON is always small metadata; 4 MiB is well above any realistic manifest.
-	MaxManifestBodySize          = 4 * 1024 * 1024
+	MaxManifestBodySize = 4 * 1024 * 1024
+	// MaxAPIKeyBodySize is the maximum number of bytes accepted for an API-key creation request body.
+	MaxAPIKeyBodySize            = 4 * 1024
 	BlobUploadUUID               = "Blob-Upload-UUID"
 	DefaultMediaType             = "application/json"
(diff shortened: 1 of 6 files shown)


andaaron (commented):
Swagger doc needs an update

pkg/api/constants/consts.go
```diff
@@ -22,7 +22,9 @@ const (
 	MaxManifestDigestQueryTags = (8192 - 2048) / (len("tag=") + 128 + 1)
 	// MaxManifestBodySize is the maximum number of bytes accepted for a manifest PUT request body.
 	// OCI manifest JSON is always small metadata; 4 MiB is well above any realistic manifest.
-	MaxManifestBodySize          = 4 * 1024 * 1024
+	MaxManifestBodySize = 4 * 1024 * 1024
+	// MaxAPIKeyBodySize is the maximum number of bytes accepted for an API-key creation request body.
+	MaxAPIKeyBodySize            = 4 * 1024
```
vrajashkr, line 27: Is there a justification on why 4KiB exactly? Could it be smaller or should it be larger?
IMO, 4KiB is pretty large for API key payload.

vrajashkr (commented):
Code changes look good overall. Had a question about the payload size.

pkg/api/constants/consts.go
rchincha: It is, but then it could be any value - large enough to fit and known upper limit.

vrajashkr: Makes sense. We can stick with this value for now and adjust later if required. Current value seems sufficiently large.

vrajashkr (approved):
LGTM

andaaron (approved):

PR #3982, opened 2026-04-18
fix(security): remove InsecureSkipVerify from metrics client (TLS-1)

rchincha:
Replace the unconditional InsecureSkipVerify: true TLS config in newHTTPMetricsClient with the system cert pool (+ TLS 1.2 minimum).

Add an optional CACert field to MetricsConfig and to the exporter ServerConfig so operators running zot with a self-signed or private CA can point the exporter at the correct CA file instead of disabling certificate verification entirely.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/monitoring/minimal_client_test.go
+++ b/pkg/extensions/monitoring/minimal_client_test.go
@@ -1,6 +1,6 @@
 //go:build !metrics
 
-package monitoring
+package monitoring_test
 
 import (
 	"context"
--- a/pkg/extensions/monitoring/minimal_client_test.go
+++ b/pkg/extensions/monitoring/minimal_client_test.go
@@ -1,0 +1,189 @@
+//go:build !metrics
+
+package monitoring
+
+import (
+	"crypto/rand"
+	"crypto/rsa"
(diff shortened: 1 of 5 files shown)


pkg/extensions/monitoring/minimal_client_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+	if resp.StatusCode != http.StatusOK {
+		t.Fatalf("expected status %d, got %d", http.StatusOK, resp.StatusCode)
+	}
+}
+
+func TestNewMetricsClientFallbackKeepsTLSHardening(t *testing.T) {
+	t.Parallel()
+
+	cfg := &MetricsConfig{Address: "https://127.0.0.1:8443", CACert: filepath.Join(t.TempDir(), "missing-ca.pem")}
```
andaaron, line 129: Let's not hardcode ips, that may result in test failures if tests are run locally and that ip is used for something else. Or may fail in gh actions if multiple prs run in parallel

andaaron (approved):

PR #3984, opened 2026-04-19
fix(security): enhance timeout configurations and body size limits fo…

rchincha:
…r HTTP requests

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


pkg/api/proxy.go
```diff
@@ -171,42 +180,9 @@ func proxyHTTPRequest(ctx context.Context, req *http.Request,
...
-
-	// after closing the original body, substitute it with a new reader
-	// using the buffer that was just created.
-	// this buffer should be closed later by the consumer of the response.
-	resp.Body = io.NopCloser(bytes.NewReader(clonedRespBody.Bytes()))
-
 	return resp, nil
 }
 
-func cloneRequestBody(src *http.Request) io.Reader {
```
andaaron, line 189: Can you explain why these are no longer needed?

rchincha: Yes. The short reason is that ownership of the request body changed when the proxy stopped cloning and buffering it.

In the current implementation, [proxy.go:133-153](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) passes the incoming [req.Body](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) straight into the forwarded request instead of reading it into a separate buffer first. Once you do that, there is no extra temporary request-body copy to clean up. The HTTP transport consumes that stream when [httpClient.Do(fwdRequest)](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) runs, and the original inbound request body is still owned by the server-side [net/http](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) request lifecycle, not by [proxyHTTPRequest](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

That helper only made sense in the old model, where proxying eagerly read the entire request body, duplicated it, and rebuilt replacement readers so both the original request and the forwarded request could still be read. In that design, there were extra intermediate bodies/readers, so explicit cleanup had value. After the refactor to streaming:

there is no cloned request-body buffer anymore
there is no second synthetic request body that needs explicit disposal
manually closing the inbound request body inside [proxyHTTPRequest](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) would be the wrong ownership boundary
The response side is different. The proxied response body is now returned to the caller and closed by the caller in [proxy.go:68-77](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), which is the correct place because that code is the consumer of the response stream.

So the removal is not “we forgot to close something”; it is “the proxy no longer creates an extra resource that it owns and therefore no longer needs a close helper for it.”

pkg/api/config/config.go
```diff
@@ -1117,6 +1129,70 @@ func (c *Config) GetHTTPPort() string {
...
+
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	if c.HTTP.ReadTimeout == nil {
+		return 0, false
+	}
+
+	return *c.HTTP.ReadTimeout, true
+}
```
vrajashkr: `GetHTTPReadTimeout` above already returns a 0 when the value is not set. Since a 0 timeout is not a realistic thing, we could assume that regardless of it being explicitly set or not, a value of 0 is no timeout. With that, in my opinion, this function is not needed.

pkg/api/config/config.go
```diff
@@ -1117,6 +1129,70 @@ func (c *Config) GetHTTPPort() string {
...
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	if c.HTTP.WriteTimeout == nil {
+		return 0, false
+	}
+
+	return *c.HTTP.WriteTimeout, true
+}
+
```
vrajashkr: `GetHTTPWriteTimeout` above already returns a 0 when the value is not set. Since a 0 timeout is not a realistic thing, we could assume that regardless of it being explicitly set or not, a value of 0 is no timeout. With that, in my opinion, this function is not needed.

pkg/api/controller.go
```diff
@@ -34,8 +34,10 @@ import (
 )
 
 const (
-	idleTimeout       = 120 * time.Second
-	readHeaderTimeout = 5 * time.Second
+	idleTimeout         = 120 * time.Second
+	readHeaderTimeout   = 5 * time.Second
+	defaultReadTimeout  = 30 * time.Second
+	defaultWriteTimeout = 30 * time.Second
```
vrajashkr: These timeouts seem very restrictive. Blobs being pushed and pulled will very easily require more than 30 seconds right once the size goes beyond a certain point.

IMO, requiring all zot configs to have timeouts set seems to make the simple zot config more complex. Instead, I feel that users can set the timeouts if they want and by default, there are no timeouts for the zot server. This way, I feel users get a choice and are not forced to override the timeouts for every deployment.

To encourage setting timeouts, we can include it in the examples.

pkg/exporter/api/exporter.go
```diff
@@ -169,10 +171,23 @@ func GetCollector(c *Controller) *Collector {
...
+		return *configured
+	}
+
+	return defaultReadTimeout
+}
+
 func runExporter(c *Controller) {
 	exporterAddr := ":" + c.Config.Exporter.Port
+	readTimeout := selectedTimeout(c.Config.Exporter.ReadTimeout)
+	writeTimeout := selectedTimeout(c.Config.Exporter.WriteTimeout)
```
vrajashkr, line 184: `selectedTimeout` always returns the `defaultReadTimeout` when the value is not configured. This causes the default write timeout to always be the same as read.

This should probably be fixed to fetch the correct default for write timeout.

pkg/api/constants/consts.go
```diff
@@ -24,7 +24,9 @@ const (
 	// OCI manifest JSON is always small metadata; 4 MiB is well above any realistic manifest.
 	MaxManifestBodySize = 4 * 1024 * 1024
 	// MaxAPIKeyBodySize is the maximum number of bytes accepted for an API-key creation request body.
-	MaxAPIKeyBodySize            = 4 * 1024
+	MaxAPIKeyBodySize = 4 * 1024
+	// MaxImageTrustBodySize is the maximum number of bytes accepted for image-trust key/certificate uploads.
+	MaxImageTrustBodySize        = 4 * 1024 * 1024
```
vrajashkr: Do we have a reference we could add here for the max size of the image trust body size?
Is it part of the spec or a reasonable maximum?

pkg/api/proxy.go
```diff
@@ -131,15 +130,28 @@ func proxyHTTPRequest(ctx context.Context, req *http.Request,
 	cloneURL.Scheme = proxyQueryScheme
 	cloneURL.Host = targetMember
 
-	clonedBody := cloneRequestBody(req)
```
vrajashkr, line 134: It's been a while so I don't remember too well, but there was some reason for which we wanted to avoid consuming the original request's body - hence the cloning thing we did.

vrajashkr (commented):
Left some comments on the changes.

pkg/api/config/http_timeout_accessors_test.go
andaaron, line 1: Why not have these in pkg/api/config/config_test.go? They don't seem to stand out from other configuration methods.

pkg/api/controller.go
andaaron: This makes sense also in case of existing deployments which may need larger timeouts.

pkg/api/config/config.go
rchincha: Difference is whether you can distinguish unset from explicitly set-to-zero.

pkg/api/controller.go
rchincha: Leaving the default (which is infinite) is insecure to start with? We can make this a little bigger.

At a theoretical speed of 100 Mbps (megabits per second), a 100 MiB (mebibyte) file takes approximately 8 to 10 seconds to download

pkg/api/constants/consts.go
rchincha: For specific implementations, such as those using AWS Signer, the signed payload has a maximum length of 4096 bytes. 

https://docs.aws.amazon.com/signer/latest/api/API_SignPayload.html#API_SignPayload_RequestBody

rchincha: We could make this a little larger

pkg/api/config/config.go
rchincha: readTimeout, readTimeoutSet := c.Config.GetHTTPReadTimeoutWithSet()
	if !readTimeoutSet {
		readTimeout = defaultReadTimeout
	}

rchincha: So basically, we don't want a zero value by default.

pkg/api/proxy.go
rchincha: the proxy was simplified to single-pass streaming plus explicit ContentLength handling, and because the original request body no longer needs to be preserved for reuse.

rchincha: The new design still forwards the body itself, but it forwards it as a stream from req.Body instead of cloning/buffering it first.

rchincha:
reviews have been addressed, left them open for any further rebuttal.

pkg/api/controller.go
vrajashkr: Fair point. I think the 60 second time out that is currently used as default is reasonable.

pkg/api/config/config.go
vrajashkr: Fair point. In that case, do we need `GetHTTPReadTimeout` and `GetHTTPWriteTimeout`? Could we just use the `withSet` functions instead (and ignore the boolean if not required)?

I think the code is mostly a duplicate of the other corresponding getter and the only additional thing is the boolean.

pkg/api/constants/consts.go
andaaron: So let's make them larger. 4k could be too small. How about double the largest certificate we know of? 
We want this to be larger that any reasonable certificate, so users are not impacted in any way (remember this is not configurable by the zot admin), and small enough to avoid attacks attempting to trigger OOM issues.

pkg/api/proxy.go
andaaron: Well yes, I get it that we stopped cloning the body so that's why this is not longer needed, but.

Why did we start cloning in the first place instead of passing the request body?
Why did we decide to stop cloning and pass the request body?

What are the pros and cons of cloning vs not cloning.
I assume the cloning was there for a reason, so if we remove that without knowing why it was initially included, we will end up with some failure down the line.

pkg/api/controller.go
andaaron: How about 5 minutes? Don't we want this to be seemless for existing deployments? Do we want the admins to update existing zot configs after upgrade?

pkg/api/proxy.go
rchincha: I think it was a store and forward earlier because Content-Length was unknown or we didn't know how to calculate it. But it takes a large amount of memory depending on the blob size being proxied.

@vrajashkr would you remember anything about this?

pkg/api/config/config.go

pkg/api/config/config.go

pkg/api/constants/consts.go
rchincha: fixed

pkg/api/proxy.go
rchincha: is there a bats test that covers this?

pkg/api/config/config.go
andaaron: The "pattern" we used for other settings is we check if the field is nil or not in root.go, where we set defaults for the configuration.

According to that pattern, given the config fed to the controller and the other components already has the defaults set by the time the config is passed to them, it is not relevant for the controller to know if the timeout was set by the user or not. It is outside the controllers' responsibility to decide what default to use. By the time the controller calls GetHTTPWriteTimeout, the default had already been set in root.go.

I am also fine with setting some defaults in the controller.go to guard against someone forgetting to set everything in root.go, but this new `WithSet` pattern opens the gate for any component to decide it's own default configuration which I don't agree with.

The config should be complete and valid by the time GetHTTPWriteTimeout() is called.

pkg/api/proxy.go
andaaron: I think there are some tests, @vrajashkr ?

pkg/api/config/config.go
andaaron: We'd also have to look everywhere in the codebase to figure out what defaults are used where, instead of just root.go. It's bad enough that we need to do that for extensions and metrics.

pkg/api/proxy.go
vrajashkr: > @vrajashkr would you remember anything about this?

I don't recall unfortunately. I did take a look at the old PR as well and there doesn't appear to be any discussion related to that.

One use case I can think of for not consuming the request's body was for fan-out where the same request is sent to multiple target servers. The same request could be used again and again since the body is always cloned.

IMO, it should be fine to remove the cloning. We don't have any fan-out use cases for the dist-spec APIs and I think we've decided not to implement the GQL scale out functionality so there shouldn't be any need.

> is there a bats test that covers this?
> I think there are some tests, @vrajashkr ?

Yes, we have a bunch of bats tests covering scale out proxying:
https://github.com/project-zot/zot/tree/main/test/scale-out

vrajashkr:
Changes LGTM other than Andrei's comments above.

pkg/api/proxy.go
andaaron: I'll resolve this.

pkg/api/config/config.go

andaaron (approved):

andaaron (approved):

PR #3995, opened 2026-04-25
Support multipart range blob pulls

### LINKED ISSUE #3623: [Feat]: Multi-range requests not supported
### zot version

v2.1.11

### Describe the bug

A partial pull using podman v5 (5.7.0) fails with an HTTP 416 error:
```bash
podman pull --tls-verify=false localhost:5000/alpine:zstd-chunked
Trying to pull localhost:5000/alpine:zstd-chunked...
Error: unable to copy from source docker://localhost:5000/alpine:zstd-chunked: partial pull of blob sha256:c364852357d6a2fd3557daa7f1fda6d22621c012e2d13f2ba77feefec0115025: read zstd:chunked manifest: fetching partial blob: StatusCode: 416, ""
```

I've enabled partial pulls locally with this in `/etc/containers/storage.conf`:
```
[storage.options.pull_options]
enable_partial_images = "true"
```

The `zot` instance outputs this log indicating is unable the request included was a multi-byte range header:
> {"time":"2025-12-06T21:19:20.753427237Z","level":"info","message":"HTTP API","module":"http","component":"session","clientIP":"10.89.0.2:41478","method":"GET","path":"/v2/alpine/blobs/sha256:c364852357d6a2fd3557daa7f1fda6d22621c012e2d13f2ba77feefec0115025","statusCode":416,"latency":"0s","bodySize":0,"headers":{"Docker-Distribution-Api-Version":["registry/2.0"],"Range":["bytes=3832641-3852897,3852906-3869099"],"User-Agent":["containers/5.38.0 (github.com/containers/image)"]},"caller":"zotregistry.dev/zot/v2/pkg/api/session.go:93","func":"zotregistry.dev/zot/v2/pkg/api.SessionLogger.func1.1","goroutine":227}

In this case: `"Range":["bytes=3832641-3852897,3852906-3869099"]`, which is a valid format referenced in [RFC 7233](https://datatracker.ietf.org/doc/html/rfc7233#section-2.1)

After looking at the code base, it seems that the current pattern used to parse the range header does not take multi-byte ranges into consideration: https://github.com/project-zot/zot/blob/9dfa7c3ae621c4c98b7027a1283db80de4e7677b/pkg/api/routes.go#L1027-L1028

### To reproduce

Run local `zot` instance via docker compose:
```yaml
services:
  zot:
    image: ghcr.io/project-zot/zot-linux-amd64:v2.1.11
    ports:
      - "5000:5000"
    volumes:
      - ./zot-config.json:/etc/zot/config.json:ro
```

Using this `zot-config.json`:
```json
{
  "distSpecVersion": "1.1.0",
  "storage": {
    "rootDirectory": "/var/lib/zot"
  },
  "http": {
    "address": "0.0.0.0",
    "port": "5000"
  },
  "log": {
    "level": "info"
  },
  "extensions": {
    "search": {
      "enable": true
    },
    "ui": {
      "enable": true
    }
  }
}
```

Prepare test image:
```bash
podman pull docker.io/library/alpine:latest
podman tag docker.io/library/alpine:latest localhost:5000/alpine:zstd-chunked
```

Push with `zstd:chunked` compression format to local zot instance without issue:
```bash
podman push --tls-verify=false --compression-format=zstd:chunked localhost:5000/alpine:zstd-chunked
Copying blob 5aa68bbbc67e done   | 
Copying config 7acffee03f done   | 
```

Remove local images and pull zstd-chunked image from `zot`:
```bash
podman rmi docker.io/library/alpine:latest
podman rmi localhost:5000/alpine:zstd-chunked

podman pull --tls-verify=false localhost:5000/alpine:zstd-chunked
Trying to pull localhost:5000/alpine:zstd-chunked...
Error: unable to copy from source docker://localhost:5000/alpine:zstd-chunked: partial pull of blob sha256:c364852357d6a2fd3557daa7f1fda6d22621c012e2d13f2ba77feefec0115025: read zstd:chunked manifest: fetching partial blob: StatusCode: 416, ""
```

Confirm with `curl` that while "Range: bytes=0-100" works, "Range: bytes=-10" does not:
```bash
curl -v -H "Range: bytes=0-100" http://localhost:5000/v2/alpine/blobs/sha256:c364852357d6a2fd3557daa7f1fda6d22621c012e2d13f2ba77feefec0115025
...
< HTTP/1.1 206 Partial Content
< Content-Length: 101
< Content-Range: bytes 0-100/3869172
...


curl -v -H "Range: bytes=0-100,101-200" http://localhost:5000/v2/alpine/blobs/sha256:c364852357d6a2fd3557daa7f1fda6d22621c012e2d13f2ba77feefec0115025
...
< HTTP/1.1 416 Requested Range Not Satisfiable
...
```

### Expected behavior

Should be able to partial pull images with podman from zot without issue

### Screenshots

_No response_

### Additional context

_No response_

AkashKumar7902:
## What
- add size-aware blob Range parsing for suffix and multi-range requests
- return multipart/byteranges responses for multi-range blob GETs
- cover suffix and multi-range blob pulls in TestPullRange

## Validation
- env GOEXPERIMENT=jsonv2 go test -tags "sync scrub metrics search lint mgmt imagetrust ui userprefs" ./pkg/api -run TestPullRange -count=1
- env GOEXPERIMENT=jsonv2 go test -tags "sync scrub metrics search lint mgmt imagetrust ui userprefs" -race ./pkg/api -run TestPullRange -count=1
- env GOEXPERIMENT=jsonv2 go vet -tags "sync scrub metrics search lint mgmt imagetrust ui userprefs" ./pkg/api
- git diff --check

Fixes #3623

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -1250,7 +1250,7 @@ func coalesceRanges(ranges []httpRange) []httpRange {
 	return coalesced
 }
 
-func writeMultipartRanges(response http.ResponseWriter, mediaType string, ranges []blobRangeReader, bsize int64,
+func writeMultipartRanges(response http.ResponseWriter, ranges []blobRangeReader, bsize int64,
 	logger log.Logger,
 ) {
 	writer := multipart.NewWriter(response)
@@ -1266,10 +1266,6 @@ func writeMultipartRanges(response http.ResponseWriter, mediaType string, ranges
 	for _, rangeReader := range ranges {
 		partHeader := textproto.MIMEHeader{}
 		partHeader.Set("Content-Range", fmt.Sprintf("bytes %d-%d/%d", rangeReader.start, rangeReader.end, bsize))
-
-		if mediaType != "" {
-			partHeader.Set("Content-Type", mediaType)
-		}
 
 		part, err := writer.CreatePart(partHeader)
 		if err != nil {
@@ -1392,10 +1388,10 @@ func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Requ
 			rangeReaders = append(rangeReaders, blobRangeReader{httpRange: httpRange, reader: repo})
 		}
 
-		defer closeRangeReaders(rangeReaders)
+		response.Header().Set(constants.DistContentDigestKey, digest.String())
 
 		if len(rangeReaders) > 1 {
-			writeMultipartRanges(response, mediaType, rangeReaders, bsize, rh.c.Log)
+			writeMultipartRanges(response, rangeReaders, bsize, rh.c.Log)
 
 			return
 		}
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -13,10 +13,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"mime/multipart"
 	"net/http"
+	"net/textproto"
 	"net/url"
 	"path"
-	"regexp"
 	"slices"
 	"sort"
 	"strconv"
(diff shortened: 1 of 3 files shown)


pkg/api/routes.go
```diff
@@ -1121,51 +1122,129 @@ func (rh *RouteHandler) CheckBlob(response http.ResponseWriter, request *http.Re
...
+/* parseRangeHeader validates the "Range" HTTP header and returns normalized byte ranges. */
+func parseRangeHeader(contentRange string, size int64) ([]httpRange, error) {
+	if size <= 0 || !strings.HasPrefix(contentRange, "bytes=") {
+		return nil, zerr.ErrParsingHTTPHeader
+	}
 
-	paramsMap := make(map[string]string)
+	rangeSet := strings.TrimPrefix(contentRange, "bytes=")
+	rangeSpecs := strings.Split(rangeSet, ",")
+	ranges := make([]httpRange, 0, len(rangeSpecs))
```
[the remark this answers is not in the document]
andaaron: Good catch

pkg/api/routes.go
```diff
@@ -1257,26 +1308,91 @@ func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Requ
...
+
+		if !ok {
+			e := apiErr.NewError(apiErr.BLOB_UNKNOWN).AddDetail(map[string]string{"digest": digest.String()})
+			zcommon.WriteJSON(response, http.StatusNotFound, apiErr.NewErrorList(e))
+
+			return
+		}
+
+		ranges, err := parseRangeHeader(contentRange, bsize)
+		if err != nil {
```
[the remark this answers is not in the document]
andaaron: Good catch

pkg/api/routes.go
```diff
@@ -1257,26 +1308,91 @@ func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Requ
...
+			response.WriteHeader(http.StatusRequestedRangeNotSatisfiable)
+
+			return
+		}
+
+		rangeReaders := make([]blobRangeReader, 0, len(ranges))
```
andaaron, line 1365:
```suggestion
		rangeReaders := make([]blobRangeReader, 0, len(ranges))
        defer func() { closeRangeReaders(rangeReaders) }()
```

pkg/api/routes.go
```diff
@@ -1257,26 +1308,91 @@ func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Requ
...
+
+		for _, httpRange := range ranges {
+			repo, blen, _, err := imgStore.GetBlobPartial(name, digest, mediaType, httpRange.start, httpRange.end)
+			if err != nil {
+				closeRangeReaders(rangeReaders)
+				writeBlobError(err)
+
+				return
+			}
+
```
andaaron, line 1375: rangeReaders = append(rangeReaders, blobRangeReader{httpRange: httpRange, reader: repo})

So we close all repos, including the current one if we include the suggestion I added at line 1135

pkg/api/routes.go
```diff
@@ -1257,26 +1308,91 @@ func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Requ
...
+		if err != nil {
+			response.WriteHeader(http.StatusRequestedRangeNotSatisfiable)
+
+			return
+		}
+
+		rangeReaders := make([]blobRangeReader, 0, len(ranges))
+
+		for _, httpRange := range ranges {
+			repo, blen, _, err := imgStore.GetBlobPartial(name, digest, mediaType, httpRange.start, httpRange.end)
```
andaaron, line 1369: For each range the loop calls `imgStore.GetBlobPartial`, which does:

```text
digest validation → RLock → originalBlobInfo (Stat + cache lookup)
→ storeDriver.Reader(path, from) → newBlobStream
```

For an N-range request that is N opens, N stats, N lock acquires. Plus
the PR's added `CheckBlob` adds one more `Stat` up front. With S3
backend that's a fresh `GetObject` per range — exactly the opposite of
what `multipart/byteranges` is supposed to optimize.

Options, in increasing order of effort:

1. Cache `binfo` per request and pass it through (one `Stat` instead
   of N+1).
2. Open the blob once for the whole request (`io.ReadSeekCloser` /
   `io.ReaderAt`) and `Seek` + `io.LimitReader` per range. Storage
   already takes an offset, so this is a small extension.
3. Add `imgStore.GetBlobRanges(name, digest, mediaType, []httpRange)
   ([]io.ReadCloser, int64, error)` so the storage layer can decide
   how to fan out (one S3 GetObject + multiple LimitReaders, etc).

pkg/api/routes.go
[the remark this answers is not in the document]
AkashKumar7902: Thanks, good catch. I capped the number of requested range specs at 16 and also coalesce overlapping/adjacent ranges before opening blob readers, so large Range headers get rejected instead of fanning out unbounded reads.

pkg/api/routes.go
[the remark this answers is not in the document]
AkashKumar7902: Done. Invalid or unsatisfiable ranges now return 416 with `Content-Range: bytes */<size>` after we have the blob size from `CheckBlob`. Added coverage for this as well.

pkg/api/routes.go
```diff
@@ -1121,51 +1122,129 @@ func (rh *RouteHandler) CheckBlob(response http.ResponseWriter, request *http.Re
...
+			return
 		}
 
-		if to < from {
-			return -1, -1, zerr.ErrParsingHTTPHeader
+		if _, err := io.Copy(part, rangeReader.reader); err != nil {
+			logger.Error().Err(err).Msg("failed to copy range into multipart response")
+
+			return
 		}
```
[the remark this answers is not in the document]
AkashKumar7902: Done. `writeMultipartRanges` now defers closing the multipart writer and logs close errors, so even early returns still finalize the response boundary as cleanly as possible.

pkg/api/routes.go
AkashKumar7902: Applied this cleanup. The range readers are now closed through a deferred cleanup right after the slice is created, which keeps the error paths simpler and consistent.

pkg/api/routes.go
AkashKumar7902: Yep, that path is covered now. Once a reader is appended to `rangeReaders`, the deferred cleanup closes it; if the current reader fails validation before append, it is closed directly.

pkg/api/routes.go
AkashKumar7902: Thanks for the detailed breakdown. I kept this patch scoped by adding the range cap and coalescing adjacent/overlapping ranges, which reduces unnecessary opens and prevents unbounded fan-out. I did not add a new storage-layer API here; that feels like a larger follow-up than this fix should carry.

pkg/api/routes.go
```diff
@@ -1257,26 +1341,92 @@ func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Requ
...
+					Msg("unexpected partial blob length")
+				response.WriteHeader(http.StatusInternalServerError)
+
+				return
+			}
+
+			rangeReaders = append(rangeReaders, blobRangeReader{httpRange: httpRange, reader: repo})
+		}
+
+		defer closeRangeReaders(rangeReaders)
```
andaaron: This duplicates line 1370, resulting in trying to close() twice.

pkg/api/routes_internal_test.go
```diff
@@ -0,0 +1,106 @@
+//go:build sync && scrub && metrics && search && lint && userprefs && mgmt && imagetrust && ui
```
andaaron: This sis restrictive for no reason.

pkg/api/routes.go
```diff
@@ -1257,26 +1341,92 @@ func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Requ
...
+
+		if len(rangeReaders) > 1 {
+			writeMultipartRanges(response, mediaType, rangeReaders, bsize, rh.c.Log)
+
+			return
+		}
+
+		rangeReader := rangeReaders[0]
+		response.Header().Set("Content-Range", fmt.Sprintf("bytes %d-%d/%d", rangeReader.start, rangeReader.end, bsize))
```
andaaron, line 1400: can you add `response.Header().Set(constants.DistContentDigestKey, digest.String())` also for this case?

pkg/api/routes.go
```diff
@@ -1257,26 +1341,92 @@ func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Requ
...
+				return
+			}
+
+			rangeReaders = append(rangeReaders, blobRangeReader{httpRange: httpRange, reader: repo})
+		}
+
+		defer closeRangeReaders(rangeReaders)
+
+		if len(rangeReaders) > 1 {
+			writeMultipartRanges(response, mediaType, rangeReaders, bsize, rh.c.Log)
```
andaaron: Can you also include `response.Header().Set(constants.DistContentDigestKey, digest.String())` also for this case?

pkg/api/routes.go
andaaron: OK, will follow up on that separately

pkg/api/routes.go
```diff
@@ -1257,26 +1341,92 @@ func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Requ
...
+				response.WriteHeader(http.StatusInternalServerError)
+
+				return
+			}
+
+			rangeReaders = append(rangeReaders, blobRangeReader{httpRange: httpRange, reader: repo})
+		}
+
+		defer closeRangeReaders(rangeReaders)
+
```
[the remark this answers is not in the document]
AkashKumar7902: Removed — only the early defer remains so the readers are closed exactly once on every path.

pkg/api/routes.go
```diff
@@ -1121,51 +1122,168 @@ func (rh *RouteHandler) CheckBlob(response http.ResponseWriter, request *http.Re
...
+	response.WriteHeader(http.StatusPartialContent)
+
+	for _, rangeReader := range ranges {
+		partHeader := textproto.MIMEHeader{}
+		partHeader.Set("Content-Range", fmt.Sprintf("bytes %d-%d/%d", rangeReader.start, rangeReader.end, bsize))
+
+		if mediaType != "" {
+			partHeader.Set("Content-Type", mediaType)
+		}
+
```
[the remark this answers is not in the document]
AkashKumar7902: Dropped the per-part Content-Type entirely. Echoing the request `Accept` header in there could embed a media range (`application/octet-stream,*/*`) which is not valid for Content-Type. RFC 9110 allows omitting it for byteranges parts.

pkg/api/routes.go
AkashKumar7902: Right, removed the duplicate. The defer registered when `rangeReaders` is created handles all the paths.

pkg/api/routes_internal_test.go
AkashKumar7902: Dropped the build tag — the parser only exercises plain `api` symbols, no extension code involved.

pkg/api/routes.go
AkashKumar7902: Done — `Docker-Content-Digest` is now set once on the partial-content response before either branch (single range or multipart) writes its body.

pkg/api/routes.go
AkashKumar7902: Same fix covers this case — the digest header is set on the response right after we collect the readers, so multipart responses include it too.

andaaron (approved):

PR #3996, opened 2026-04-25
feat(cli): support default zli config

### LINKED ISSUE #129: cli: add a default config if not specified
Right now, we need to pass a config alias with each command. Instead we can make it optional and make user able to choose a config as default.

For example
```console
$ zot images remote-zot
```
Instead of this we could have a default config and then the command will look like
```console
$ zot images
```

AkashKumar7902:
**What type of PR is this?**

feature

Fixes #129

Adds support for marking a saved zli config as the default so commands can omit `--config` when neither `--url` nor `--config` is provided.

This adds `zli config <config-name> --set-default`, keeps the `default` config key single-default aware, and reuses the default config for URL, spinner, and TLS settings. Explicit `--url` still takes precedence and does not read options from the default config.

N/A

- `GOEXPERIMENT=jsonv2 go test -tags search ./pkg/cli/client -run 'TestConfigCmdMain|TestUtils'`
- `GOEXPERIMENT=jsonv2 go test -tags search ./pkg/cli/client -run 'TestConfigCmdBasics|TestConfigCmdMain|TestUtils|TestCVECommandErrors|TestServerStatusCommandErrors|TestSearchCLIErrors|TestReferrersCLIErrors'`
- `git diff --check`

I also ran `GOEXPERIMENT=jsonv2 go test -json -tags search ./pkg/cli/client`; it fails locally because the checkout is missing `test/data/alpine/blobs/sha256/f56be85fc22e46face30e2c3de3f7fe7c15f8fd7c4e5add29d7f64b87abdaa09`. The failing tests are `TestImagesCommandGQL`, `TestImageCommandREST`, and `TestNegativeServerResponse`.

**Automation added to e2e**:

No. Focused CLI unit coverage was added for the default config behavior.

No.

Yes.

```release-note
zli can now use a saved default registry configuration when --config is omitted. Mark one with `zli config <config-name> --set-default`.
```

### DIFF
--- a/pkg/cli/client/config_cmd.go
+++ b/pkg/cli/client/config_cmd.go
@@ -28,6 +28,8 @@ func NewConfigCommand() *cobra.Command {
 
 	var isReset bool
 
+	var isSetDefault bool
+
 	configCmd := &cobra.Command{
 		Use:     "config <config-name> [variable] [value]",
 		Example: examples,
@@ -42,8 +44,16 @@ func NewConfigCommand() *cobra.Command {
 
 			configPath := path.Join(home, "/.zot")
 
+			if isSetDefault && (isListing || isReset) {
+				return zerr.ErrInvalidArgs
+			}
+
 			switch len(args) {
 			case noArgs:
+				if isSetDefault {
+					return zerr.ErrInvalidArgs
+				}
+
 				if isListing { // zot config -l
 					res, err := getConfigNames(configPath)
 					if err != nil {
@@ -57,6 +67,10 @@ func NewConfigCommand() *cobra.Command {
 
 				return zerr.ErrInvalidArgs
 			case oneArg:
+				if isSetDefault { // zot config <name> --set-default
+					return setDefaultConfig(configPath, args[0])
+				}
+
 				// zot config <name> -l
 				if isListing {
 					res, err := getAllConfig(configPath, args[0])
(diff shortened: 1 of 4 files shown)


pkg/cli/client/config_cmd.go
```diff
@@ -313,6 +336,95 @@ func addDefaultConfigs(config map[string]any) {
...
+		return "", err
+	}
+
+	for _, val := range configs {
+		configMap, ok := val.(map[string]any)
+		if !ok {
+			return "", zerr.ErrBadConfig
+		}
+
+		defaultValue, ok := configMap[defaultConfigKey]
```
andaaron, line 405: I am not convinced this is the right approach, especially if there is a change where we can end up in a state with multiple configs with the defautl value set.
How about having a second key besides "configs", such as "defaultConfig", with value a reference to one of the configs in the "configs" list.

andaaron:
@AkashKumar7902, if you are still interested in resolving this, I recently refactored the config commands and the code handling the zli config file.
I have updated the description in https://github.com/project-zot/zot/issues/129 to match.
You would probably need to open a fresh PR.

AkashKumar7902:
I opened a fresh replacement from current main with the updated issue shape (`defaultConfigName` plus set/clear-default commands): #4094. Closing this older conflicted PR so review can continue on the clean branch.

PR #3999, opened 2026-04-25
feat(auth): map OpenID groups claim

### LINKED ISSUE #2523: [Feat]: Allow specifying user and group scopes with OIDC authentication

AkashKumar7902:
## What

Adds `claimMapping.groups` for OpenID providers so deployments can map group membership from a custom OIDC claim. The existing `claimMapping.username` behavior remains unchanged, and group mapping defaults to `groups` when unset.

Updates the OpenID claim-mapping example and README snippet.

Fixes #2523

## Validation

- `GOEXPERIMENT=jsonv2 go test -tags 'sync scrub metrics search lint mgmt' ./pkg/api -run TestRoutes -count=1`\n- `GOEXPERIMENT=jsonv2 go test -tags 'sync scrub metrics search lint mgmt' ./pkg/api -run 'TestRoutes|TestAuthn' -count=1`\n- `GOEXPERIMENT=jsonv2 go test ./pkg/api/config -run 'TestConfig|TestUpdateReloadableConfig|TestSanitize' -count=1`\n- `GOEXPERIMENT=jsonv2 go test ./pkg/cli/server -run 'TestSchema|TestVerify|TestLoadConfig' -count=1`\n- `jq empty examples/config-openid-claim-mapping.json`\n- `git diff --check`

### DIFF
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -1106,6 +1106,161 @@ func saveUserLoggedSession(cookieStore sessions.Store, response http.ResponseWri
 	return nil
 }
 
+const (
+	defaultUsernameClaim = "email"
+	defaultGroupsClaim   = "groups"
+)
+
+// getOpenIDClaimMapping resolves the username and groups claim names for a given provider.
+// The third return value reports whether the username claim was explicitly configured
+// (false means the default "email" claim is being used as a fallback).
+func getOpenIDClaimMapping(authConfig *config.AuthConfig, providerName string) (string, string, bool) {
+	usernameClaim := defaultUsernameClaim
+	groupsClaim := defaultGroupsClaim
+	usernameConfigured := false
+
+	if authConfig == nil || authConfig.OpenID == nil || providerName == "" {
+		return usernameClaim, groupsClaim, usernameConfigured
+	}
+
+	providerConfig, ok := authConfig.OpenID.Providers[providerName]
+	if !ok || providerConfig.ClaimMapping == nil {
+		return usernameClaim, groupsClaim, usernameConfigured
+	}
+
+	if providerConfig.ClaimMapping.Username != "" {
+		usernameClaim = providerConfig.ClaimMapping.Username
+		usernameConfigured = true
+	}
+
+	if providerConfig.ClaimMapping.Groups != "" {
+		groupsClaim = providerConfig.ClaimMapping.Groups
+	}
+
+	return usernameClaim, groupsClaim, usernameConfigured
+}
+
+func getOpenIDUsername(info *oidc.UserInfo, claimName string) string {
+	if info == nil {
+		return ""
+	}
+
+	switch claimName {
+	case "preferred_username":
+		return info.PreferredUsername
+	case defaultUsernameClaim:
+		return info.UserInfoEmail.Email
+	case "sub":
+		return info.Subject
+	case "name":
+		return info.Name
+	default:
+		if val, ok := info.Claims[claimName].(string); ok {
+			return val
+		}
+	}
+
+	return ""
+}
+
+func appendOpenIDGroups(groups []string, claims map[string]any, claimName string) ([]string, bool) {
+	switch val := claims[claimName].(type) {
+	case []any:
+		for _, group := range val {
+			if group == nil {
+				continue
+			}
+
+			if str := fmt.Sprint(group); str != "" {
+				groups = append(groups, str)
+			}
+		}
+
+		return groups, true
+	case []string:
+		for _, group := range val {
+			if group != "" {
+				groups = append(groups, group)
+			}
+		}
+
+		return groups, true
+	case string:
+		if val != "" {
+			groups = append(groups, val)
+		}
+
+		return groups, true
+	}
+
+	return groups, false
+}
+
+// extractOpenIDIdentity resolves the username and groups for an OIDC callback
+// based on the provider's configured claim mapping. It returns the resolved
+// username, the deduplicated/sorted groups, and a boolean reporting whether
+// the username could be resolved at all (false means callers should reject).
+func extractOpenIDIdentity(logger log.Logger, authConfig *config.AuthConfig, providerName string,
+	info *oidc.UserInfo, idTokenClaims map[string]any,
+) (string, []string, bool) {
+	usernameClaim, groupsClaim, usernameConfigured := getOpenIDClaimMapping(authConfig, providerName)
+
+	username := getOpenIDUsername(info, usernameClaim)
+
+	if username == "" && usernameConfigured {
+		configuredClaim := usernameClaim
+		usernameClaim = defaultUsernameClaim
+		usernameConfigured = false
+		username = getOpenIDUsername(info, usernameClaim)
+
+		logger.Warn().
+			Str("provider", providerName).
+			Str("claim", configuredClaim).
+			Msg("configured username claim missing or empty, falling back to email")
+	}
+
+	if username == "" {
+		return "", nil, false
+	}
+
+	if usernameConfigured {
+		logger.Debug().
+			Str("provider", providerName).
+			Str("claim", usernameClaim).
+			Str("username", username).
+			Msg("extracted username from configured claim")
+	} else {
+		logger.Debug().
+			Str("provider", providerName).
+			Str("username", username).
+			Msg("using email as username (fallback)")
+	}
+
+	var (
+		groups      []string
+		groupsFound bool
+	)
+
+	if info != nil {
+		groups, groupsFound = appendOpenIDGroups(groups, info.Claims, groupsClaim)
+		if !groupsFound {
+			logger.Info().Msgf("failed to find any %q claim for user %s in UserInfo", groupsClaim, username)
+		}
+	}
+
+	if idTokenClaims != nil {
+		groups, groupsFound = appendOpenIDGroups(groups, idTokenClaims, groupsClaim)
+		if !groupsFound {
+			logger.Info().Msgf("failed to find any %q claim for user %s in IDTokenClaimsToken", groupsClaim, username)
+		}
+	}
+
+	slices.Sort(groups)
+	groups = slices.Compact(groups)
+
+	return username, groups, true
+}
+
 // OAuth2Callback is the callback logic where openid/oauth2 will redirect back to our app.
 func OAuth2Callback(ctlr *Controller, w http.ResponseWriter, r *http.Request, state, email, provider string,
 	groups []string,
--- a/pkg/api/authn_internal_test.go
+++ b/pkg/api/authn_internal_test.go
@@ -1,5 +1,3 @@
-//go:build sync && scrub && metrics && search && lint && mgmt
-
 package api
 
 import (
@@ -191,8 +189,22 @@ func TestAppendOpenIDGroups(t *testing.T) {
 			expectedFound: true,
 		},
 		{
+			name:          "skips nil and empty entries in any slice",
+			claims:        map[string]any{"roles": []any{"dev", nil, ""}},
+			claim:         "roles",
+			expected:      []string{"dev"},
+			expectedFound: true,
+		},
+		{
 			name:          "appends string slice",
 			claims:        map[string]any{"roles": []string{"admin", "ops"}},
+			claim:         "roles",
+			expected:      []string{"admin", "ops"},
+			expectedFound: true,
+		},
+		{
+			name:          "skips empty entries in string slice",
+			claims:        map[string]any{"roles": []string{"admin", "", "ops"}},
 			claim:         "roles",
 			expected:      []string{"admin", "ops"},
 			expectedFound: true,
--- a/pkg/api/config/config.go
+++ b/pkg/api/config/config.go
@@ -322,6 +322,10 @@ type ClaimMapping struct {
 	// Acceptable values include "preferred_username", "email", "sub", "name", or any custom claim name.
 	// If not configured, the default is "email".
 	Username string `mapstructure:"username,omitempty"`
+
+	// Groups specifies which OpenID claim to use as the groups for the authenticated user.
+	// If not configured, the default is "groups".
+	Groups string `mapstructure:"groups,omitempty"`
 }
 
 // CELClaimValidationAndMapping specifies Common Expression Language (CEL) expressions
--- a/pkg/api/openid_claim_mapping_internal_test.go
+++ b/pkg/api/openid_claim_mapping_internal_test.go
@@ -175,42 +175,69 @@ func TestAppendOpenIDGroups(t *testing.T) {
 	t.Parallel()
 
 	tests := []struct {
-		name     string
-		groups   []string
-		claims   map[string]any
-		claim    string
-		expected []string
+		name          string
+		groups        []string
+		claims        map[string]any
+		claim         string
+		expected      []string
+		expectedFound bool
 	}{
 		{
-			name:     "appends any slice",
-			groups:   []string{"existing"},
-			claims:   map[string]any{"roles": []any{"dev", 7}},
-			claim:    "roles",
-			expected: []string{"existing", "dev", "7"},
+			name:          "appends any slice",
+			groups:        []string{"existing"},
+			claims:        map[string]any{"roles": []any{"dev", 7}},
+			claim:         "roles",
+			expected:      []string{"existing", "dev", "7"},
+			expectedFound: true,
 		},
 		{
-			name:     "appends string slice",
-			claims:   map[string]any{"roles": []string{"admin", "ops"}},
-			claim:    "roles",
-			expected: []string{"admin", "ops"},
+			name:          "appends string slice",
+			claims:        map[string]any{"roles": []string{"admin", "ops"}},
+			claim:         "roles",
+			expected:      []string{"admin", "ops"},
+			expectedFound: true,
 		},
 		{
-			name:     "appends non-empty string",
-			claims:   map[string]any{"roles": "admin"},
-			claim:    "roles",
-			expected: []string{"admin"},
+			name:          "appends non-empty string",
+			claims:        map[string]any{"roles": "admin"},
+			claim:         "roles",
+			expected:      []string{"admin"},
+			expectedFound: true,
 		},
 		{
-			name:     "ignores empty string",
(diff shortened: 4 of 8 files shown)


pkg/api/routes.go
```diff
@@ -2129,50 +2193,21 @@ func (rh *RouteHandler) OpenIDCodeExchangeCallbackWithProvider(providerName stri
...
+		if username != "" {
 			rh.c.Log.Debug().
 				Str("provider", providerName).
+				Str("claim", usernameClaim).
 				Str("username", username).
-				Msg("using email as username (fallback)")
+				Msg("extracted username from configured claim")
 		}
 
 		if username == "" {
```
andaaron: Let's use an if/else here. The username value is also checked above.

pkg/api/routes.go
```diff
@@ -2184,22 +2219,16 @@ func (rh *RouteHandler) OpenIDCodeExchangeCallbackWithProvider(providerName stri
...
 		var groups []string
 
-		val, ok := info.Claims["groups"].([]any)
-		if !ok {
-			rh.c.Log.Info().Msgf("failed to find any 'groups' claim for user %s in UserInfo", username)
-		}
-
-		for _, group := range val {
-			groups = append(groups, fmt.Sprint(group))
+		userInfoGroupCount := len(groups)
```
andaaron: But groups is always 0 here, correct?

pkg/api/routes.go
```diff
@@ -2184,22 +2219,16 @@ func (rh *RouteHandler) OpenIDCodeExchangeCallbackWithProvider(providerName stri
...
+		groups = appendOpenIDGroups(groups, info.Claims, groupsClaim)
+		if len(groups) == userInfoGroupCount {
```
andaaron:
```suggestion
		if len(groups) == 0 {
```

pkg/api/routes.go
AkashKumar7902: Done — collapsed into a nested if so the empty-username check happens in one place.

pkg/api/routes.go
AkashKumar7902: Right — replaced the length-based detection with a bool from `appendOpenIDGroups` in the followup commit, so we no longer rely on the saved count.

pkg/api/routes.go
AkashKumar7902: Switched to having `appendOpenIDGroups` return a found bool, so the length comparison is gone entirely.

pkg/api/routes.go
```diff
@@ -2281,24 +2319,30 @@ func (rh *RouteHandler) OpenIDCodeExchangeCallbackWithProvider(providerName stri
...
+			rh.c.Log.Info().Msgf("failed to find any %q claim for user %s in UserInfo", groupsClaim, username)
 		}
 
-		for _, group := range val {
-			groups = append(groups, fmt.Sprint(group))
+		idTokenGroupCount := len(groups)
+		groups = appendOpenIDGroups(groups, tokens.IDTokenClaims.Claims, groupsClaim)
+		if len(groups) == idTokenGroupCount {
+			rh.c.Log.Info().Msgf("failed to find any %q claim for user %s in IDTokenClaimsToken", groupsClaim, username)
 		}
```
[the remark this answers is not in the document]
AkashKumar7902: Good catch — `appendOpenIDGroups` now returns a found bool that distinguishes "claim missing/wrong type" from "claim present but empty", and the log uses that instead of length.

pkg/api/routes.go
```diff
@@ -2281,24 +2319,30 @@ func (rh *RouteHandler) OpenIDCodeExchangeCallbackWithProvider(providerName stri
...
-		if !ok {
-			rh.c.Log.Info().Msgf("failed to find any 'groups' claim for user %s in UserInfo", username)
+		if usernameConfigured {
+			rh.c.Log.Debug().
+				Str("provider", providerName).
+				Str("claim", usernameClaim).
+				Str("username", username).
+				Msg("extracted username from configured claim")
+		} else {
+			rh.c.Log.Info().
```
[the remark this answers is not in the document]
AkashKumar7902: Agreed, dropped the fallback log back to Debug.

pkg/api/routes.go
```diff
@@ -2217,6 +2217,86 @@ func (rh *RouteHandler) OpenIDCodeExchangeCallback() rp.CodeExchangeUserinfoCall
...
 
+const (
+	defaultUsernameClaim = "email"
+	defaultGroupsClaim   = "groups"
+)
+
+// getOpenIDClaimMapping resolves the username and groups claim names for a given provider.
+// The third return value reports whether the username claim was explicitly configured
+// (false means the default "email" claim is being used as a fallback).
+func getOpenIDClaimMapping(authConfig *config.AuthConfig, providerName string) (string, string, bool) {
```
andaaron: Maybe we should move this to authn.go. Its only relevant for authentication and doesn't depend on anything else here.

pkg/api/routes.go
```diff
@@ -2217,6 +2217,86 @@ func (rh *RouteHandler) OpenIDCodeExchangeCallback() rp.CodeExchangeUserinfoCall
...
+	}
+
+	if providerConfig.ClaimMapping.Groups != "" {
+		groupsClaim = providerConfig.ClaimMapping.Groups
+	}
+
+	return usernameClaim, groupsClaim, usernameConfigured
+}
+
+func getOpenIDUsername(info *oidc.UserInfo, claimName string) string {
```
andaaron: Maybe we should move this to authn.go. Its only relevant for authentication and doesn't depend on anything else here.

pkg/api/routes.go
```diff
@@ -2228,77 +2308,50 @@ func (rh *RouteHandler) OpenIDCodeExchangeCallbackWithProvider(providerName stri
 		relyingParty rp.RelyingParty, info *oidc.UserInfo,
 	) {
 		// Extract username based on claim mapping configuration
-		var username string
 		authConfig := rh.c.Config.CopyAuthConfig()
```
andaaron, line 2528: Maybe we should move this whole block up until line 2358 into a new function in authn.go

pkg/api/routes.go
```diff
@@ -2217,6 +2217,86 @@ func (rh *RouteHandler) OpenIDCodeExchangeCallback() rp.CodeExchangeUserinfoCall
...
+	default:
+		if val, ok := info.Claims[claimName].(string); ok {
+			return val
+		}
+	}
+
+	return ""
+}
+
+func appendOpenIDGroups(groups []string, claims map[string]any, claimName string) ([]string, bool) {
```
andaaron: Maybe we should move this to authn.go. Its only relevant for authentication and doesn't depend on anything else here.

pkg/api/openid_claim_mapping_internal_test.go
```diff
@@ -0,0 +1,266 @@
+//go:build sync && scrub && metrics && search && lint && mgmt
```
andaaron: These labels shouldn't be needed.

pkg/api/routes.go
```diff
@@ -2228,77 +2308,50 @@ func (rh *RouteHandler) OpenIDCodeExchangeCallbackWithProvider(providerName stri
...
-							Str("claim", claimName).
-							Str("username", username).
-							Msg("extracted username from configured claim")
-					}
-				}
+		username := getOpenIDUsername(info, usernameClaim)
+		if username == "" {
+			if usernameClaim != defaultUsernameClaim {
+				usernameClaim = defaultUsernameClaim
+				usernameConfigured = false
```
andaaron: When an operator configures claimMapping.username = "preferred_username" (or any custom claim) and the IdP returns it empty/absent, we silently reset usernameConfigured = false and log "using email as username (fallback)" at Debug — losing the signal that the configured claim was tried and missing.

pkg/api/routes.go
```diff
@@ -2217,6 +2217,86 @@ func (rh *RouteHandler) OpenIDCodeExchangeCallback() rp.CodeExchangeUserinfoCall
...
+			return val
+		}
+	}
+
+	return ""
+}
+
+func appendOpenIDGroups(groups []string, claims map[string]any, claimName string) ([]string, bool) {
+	switch val := claims[claimName].(type) {
+	case []any:
```
andaaron: The string branch filters empty values, but []string and []any don't. With []any{"dev", nil, ""} you'd end up with "<nil>" and "" group entries (and they'd survive slices.Compact). Three options:

Filter empties / skip nil entries consistently in all three branches.
Or document the intent (e.g., "we trust the IdP to send only non-empty entries").
Adding a test case like {"roles": []any{nil, ""}} would also pin the chosen behavior down.

pkg/api/authn_internal_test.go
andaaron, line 1: Rename to authn_internal_test.go

andaaron:
Please rebase this PR

pkg/api/routes.go
AkashKumar7902: Done — moved `getOpenIDClaimMapping` (along with the `defaultUsernameClaim`/`defaultGroupsClaim` constants) into authn.go in 30160a72.

pkg/api/routes.go
AkashKumar7902: Moved `getOpenIDUsername` over to authn.go as well.

pkg/api/routes.go
AkashKumar7902: Same — `appendOpenIDGroups` is now in authn.go.

pkg/api/routes.go
AkashKumar7902: Good idea — pulled the whole claim/group extraction block into a new `extractOpenIDIdentity` helper in authn.go. The HTTP callback in routes.go is now just `extractOpenIDIdentity(...) -> OAuth2Callback(...)` plus error handling, which keeps it readable and lets the helper be unit-tested directly.

pkg/api/openid_claim_mapping_internal_test.go
AkashKumar7902: Removed — the file no longer has any build tags.

pkg/api/authn_internal_test.go
AkashKumar7902: Done — renamed to `authn_internal_test.go` (via `git mv` so the history is preserved).

pkg/api/routes.go
AkashKumar7902: Good point — silently resetting and only logging at Debug hid the signal. Now `extractOpenIDIdentity` logs a Warn (`configured username claim missing or empty, falling back to email`) including the configured claim name before falling back, so an operator can see exactly what they configured and why it didn't match.

pkg/api/routes.go
AkashKumar7902: Went with option 1 — `appendOpenIDGroups` now filters `nil` and empty entries consistently across all three branches (`[]any`, `[]string`, and `string`). Added two test cases pinning the behavior down: `[]any{"dev", nil, ""}` → `["dev"]` and `[]string{"admin","","ops"}` → `["admin","ops"]`.

AkashKumar7902:
Rebased onto main in 30160a72. The rebase commit also folds in the round of changes from your latest review (helpers and the extraction block moved into authn.go behind a single `extractOpenIDIdentity` helper, build tags dropped, test file renamed to `authn_internal_test.go`, group filter consistent across all branches, and a Warn log when a configured username claim is missing/empty).

pkg/api/config/config.go
andaaron, line 319: Either remove this line or update it to specify the user groups

pkg/api/authn.go
```diff
@@ -1106,6 +1106,161 @@ func saveUserLoggedSession(cookieStore sessions.Store, response http.ResponseWri
...
+
+	if idTokenClaims != nil {
+		groups, groupsFound = appendOpenIDGroups(groups, idTokenClaims, groupsClaim)
+		if !groupsFound {
+			logger.Info().Msgf("failed to find any %q claim for user %s in IDTokenClaimsToken", groupsClaim, username)
+		}
+	}
+
+	slices.Sort(groups)
+	groups = slices.Compact(groups)
```
andaaron, line 1259: Since it is likely groups in info and idTokenClaims aren't both populated for every request, this would probably result in one of the messages at line 1247 and 1254 always be emitted, resulting in a lot of log messages.

I suggest we only log if the groups slice is empty here, a debug log level message.

pkg/api/authn.go
```diff
@@ -1106,6 +1106,161 @@ func saveUserLoggedSession(cookieStore sessions.Store, response http.ResponseWri
...
+	if username == "" && usernameConfigured {
+		configuredClaim := usernameClaim
+		usernameClaim = defaultUsernameClaim
+		usernameConfigured = false
+		username = getOpenIDUsername(info, usernameClaim)
+
+		logger.Warn().
+			Str("provider", providerName).
+			Str("claim", configuredClaim).
+			Msg("configured username claim missing or empty, falling back to email")
```
andaaron, line 1219: If an operator explicitly sets claimMapping.username = "email" (matching the default) and the email is missing, this branch fires (usernameConfigured is true) and re-queries the same defaultUsernameClaim, so the second getOpenIDUsername call returns the same empty string and the message reads "configured username claim 'email' missing or empty, falling back to email". Either:

Skip the warn/refallback when configuredClaim == defaultUsernameClaim (the re-query is a no-op anyway), or
Rephrase the message to "configured username claim missing or empty" without the "falling back to email" tail when the configured claim already is email.

pkg/api/authn.go
```diff
@@ -1106,6 +1106,161 @@ func saveUserLoggedSession(cookieStore sessions.Store, response http.ResponseWri
...
+	}
+
+	slices.Sort(groups)
+	groups = slices.Compact(groups)
+
+	return username, groups, true
+}
+
 // OAuth2Callback is the callback logic where openid/oauth2 will redirect back to our app.
 func OAuth2Callback(ctlr *Controller, w http.ResponseWriter, r *http.Request, state, email, provider string,
```
andaaron, line 1265: Can we rename "email" to "username" for consistency?

andaaron (approved):
Will approve this for now and take care of the nits in a separate PR.

andaaron:
> Will approve this for now and take care of the nits in a separate PR.

https://github.com/project-zot/zot/pull/4028

pkg/api/authn.go
```diff
@@ -1106,6 +1106,161 @@ func saveUserLoggedSession(cookieStore sessions.Store, response http.ResponseWri
...
+func extractOpenIDIdentity(logger log.Logger, authConfig *config.AuthConfig, providerName string,
+	info *oidc.UserInfo, idTokenClaims map[string]any,
+) (string, []string, bool) {
+	usernameClaim, groupsClaim, usernameConfigured := getOpenIDClaimMapping(authConfig, providerName)
+
+	username := getOpenIDUsername(info, usernameClaim)
+
+	if username == "" && usernameConfigured {
+		configuredClaim := usernameClaim
+		usernameClaim = defaultUsernameClaim
```
vrajashkr, line 1212: IMO, if the user has explicitly configured a claimMapping, we probably shouldn't try to fall back to email during claim evaluation.

Using email as a default claim is fine if the user has not configured anything.

pkg/api/authn.go
```diff
@@ -1106,6 +1106,161 @@ func saveUserLoggedSession(cookieStore sessions.Store, response http.ResponseWri
...
+			Str("provider", providerName).
+			Str("claim", usernameClaim).
+			Str("username", username).
+			Msg("extracted username from configured claim")
+	} else {
+		logger.Debug().
+			Str("provider", providerName).
+			Str("username", username).
+			Msg("using email as username (fallback)")
+	}
```
vrajashkr, line 1237: These debug logs seem to be based on `usernameConfigured` which if I'm not mistaken is deterministic based on the config that the user has provided. As such, it may not be a helpful log message.

Instead, a generic log mentioning the provider, claim, and username in all cases might be better. That said, there is also a question as to whether usernames should be logged.

pkg/api/authn.go
```diff
@@ -1106,6 +1106,161 @@ func saveUserLoggedSession(cookieStore sessions.Store, response http.ResponseWri
...
+
+	var (
+		groups      []string
+		groupsFound bool
+	)
+
+	if info != nil {
+		groups, groupsFound = appendOpenIDGroups(groups, info.Claims, groupsClaim)
+		if !groupsFound {
+			logger.Info().Msgf("failed to find any %q claim for user %s in UserInfo", groupsClaim, username)
```
vrajashkr, line 1247: Having this in info would make the logs quite a bit chatty.
Would recommend moving it to debug.

Also, "failed" is often used in context of errors which this is not. Would instead suggest something like below:


```suggestion
			logger.Info().Msgf("no groups found for user %s in UserInfo", groupsClaim, username)
```

pkg/api/authn.go
```diff
@@ -1106,6 +1106,161 @@ func saveUserLoggedSession(cookieStore sessions.Store, response http.ResponseWri
...
+		}
+	}
+
+	if idTokenClaims != nil {
+		groups, groupsFound = appendOpenIDGroups(groups, idTokenClaims, groupsClaim)
+		if !groupsFound {
+			logger.Info().Msgf("failed to find any %q claim for user %s in IDTokenClaimsToken", groupsClaim, username)
```
vrajashkr, line 1254: Similar comment as above.

pkg/api/authn.go
andaaron: This was also the case before this PR https://github.com/project-zot/zot/pull/3999/changes#diff-20343ca31da2491d1160ba5ed434f5dfa0698341d7ed516306872ebe03758b49L2569

Let's not change behavior.

andaaron:
@vrajashkr I opened https://github.com/project-zot/zot/pull/4028 to tackle most of these aspects. Please check it out

Issue #2966, opened 2025-02-14
[Feat]: Consolidate AWS DynamoDB table names in configuration to a table name prefix


CajuM:
### Is your feature request related to a problem? Please describe.

Presently a configuration file for zot that uses DynamoDB as the cache driver has the following format:
```json
        "cacheDriver": {
            "name": "dynamodb",
            "endpoint": "http://localhost:4566",
            "region": "us-east-2",
            "cacheTablename": "BlobTable",
            "repoMetaTablename": "RepoMetadataTable",
            "imageMetaTablename": "ImageMetaTable",
            "repoBlobsInfoTablename": "RepoBlobsInfoTable",
            "apiKeyTablename": "ApiKeyDataTable",
            "userDataTablename": "UserDataTable",
            "versionTablename": "VersionTable"
        }
```

### Describe the solution you'd like

I believe it would be easier for the administrator if we could could configure a table name prefix instead of each table name individually.
For example:
```json
        "cacheDriver": {
            "name": "dynamodb",
            "endpoint": "http://localhost:4566",
            "region": "us-east-2",
            "tablenamePrefix": "Zot"
        }
```

mdaffad:
Hi, can i take this issue? Willing to try to take a good start for this project

andaaron:
> Hi, can i take this issue? Would like to try and take a good start for this project

Sure, @mdaffad, thanks for the help.

PR #4016, opened 2026-04-26
feat(storage): support DynamoDB tablenamePrefix for cache driver

AkashKumar7902:
**What type of PR is this?**

feature

**Which issue does this PR fix**: #2966

Adds a new `tablenamePrefix` option to the `dynamodb` cache driver so administrators can configure all DynamoDB table names with a single prefix instead of declaring each table individually.

When `tablenamePrefix` is set, zot derives the following table names automatically:

| Purpose                | Derived name (with `"tablenamePrefix": "Zot"`) |
| ---------------------- | --------------------------------------------- |
| Dedupe blob cache      | `ZotBlobTable`                                |
| Repo metadata          | `ZotRepoMetadataTable`                        |
| Repo blobs info        | `ZotRepoBlobsInfoTable`                       |
| Image metadata         | `ZotImageMetaTable`                           |
| User data              | `ZotUserDataTable`                            |
| API key                | `ZotApiKeyDataTable`                          |
| Schema version         | `ZotVersionTable`                             |

Explicit per-table options (`cacheTablename`, `repoMetaTablename`, etc.) continue to take precedence, so existing configurations keep working unchanged.

Example new config:

```json
"cacheDriver": {
    "name": "dynamodb",
    "endpoint": "http://localhost:4566",
    "region": "us-east-2",
    "tablenamePrefix": "Zot"
}
```

N/A — issue #2966 is open with the feature request.

- Added unit tests in `pkg/storage/cache_internal_test.go` covering: derivation from prefix, explicit name precedence, missing/empty/non-string prefix, and missing/empty/non-string explicit table name.
- Added unit tests in `pkg/meta/meta_internal_test.go` covering the equivalent cases for the metadata tables.
- `go test ./pkg/storage/ ./pkg/meta/` passes.
- `golangci-lint run ./pkg/storage/... ./pkg/meta/...` reports 0 issues.
- `examples/config-dynamodb.json` was simplified to use the new `tablenamePrefix` and validated via the unit tests.

**Automation added to e2e**:

No new e2e cases. The existing search/auth e2e suite still exercises explicit per-table configuration, which the new code path preserves verbatim.

No. Existing configurations that use explicit table names continue to work because explicit names always take precedence over the derived prefix. Older zot versions ignore the new `tablenamePrefix` field.

Yes — administrators can configure DynamoDB tables via a single `tablenamePrefix` instead of seven individual table-name fields.

```release-note
DynamoDB cache driver now supports a `tablenamePrefix` option that derives all required table names (`<prefix>BlobTable`, `<prefix>RepoMetadataTable`, `<prefix>RepoBlobsInfoTable`, `<prefix>ImageMetaTable`, `<prefix>UserDataTable`, `<prefix>ApiKeyDataTable`, `<prefix>VersionTable`) from a single prefix. Explicit per-table options still take precedence.
```

### DIFF
--- a/pkg/storage/constants/constants.go
+++ b/pkg/storage/constants/constants.go
@@ -20,10 +20,27 @@ const (
 	BoltdbName              = "cache"
 	DynamoDBDriverName      = "dynamodb"
 	RedisDriverName         = "redis"
-	RedisLocksBucket        = "locks"
-	DefaultGCDelay          = 1 * time.Hour
-	DefaultGCInterval       = 1 * time.Hour
-	S3StorageDriverName     = "s3"
-	GCSStorageDriverName    = "gcs"
-	LocalStorageDriverName  = "local"
+
+	DynamoDBTableNamePrefix = "tablenamePrefix"
+	DynamoDBCacheTableName  = "cachetablename"
+	DynamoDBRepoMetaTable   = "repometatablename"
+	DynamoDBRepoBlobsTable  = "repoblobsinfotablename"
+	DynamoDBImageMetaTable  = "imagemetatablename"
+	DynamoDBUserDataTable   = "userdatatablename"
+	DynamoDBAPIKeyTable     = "apikeytablename"
+	DynamoDBVersionTable    = "versiontablename"
+
+	DynamoDBCacheTableSuffix = "BlobTable"
+	DynamoDBRepoMetaSuffix   = "RepoMetadataTable"
+	DynamoDBRepoBlobsSuffix  = "RepoBlobsInfoTable"
+	DynamoDBImageMetaSuffix  = "ImageMetaTable"
+	DynamoDBUserDataSuffix   = "UserDataTable"
+	DynamoDBAPIKeySuffix     = "ApiKeyDataTable"
+	DynamoDBVersionSuffix    = "VersionTable"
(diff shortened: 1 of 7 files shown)


rchincha:
I like this change for better ux but concern is whether this is a breaking change.

rchincha:
I wouldn't drop support for earlier config. I would add this new behavior as an addendum. It is always useful to have the older way to override defaults.

pkg/storage/constants/constants.go
```diff
@@ -20,10 +20,27 @@ const (
...
+	DynamoDBVersionTable    = "versiontablename"
+
+	DynamoDBCacheTableSuffix = "BlobTable"
+	DynamoDBRepoMetaSuffix   = "RepoMetadataTable"
+	DynamoDBRepoBlobsSuffix  = "RepoBlobsInfoTable"
+	DynamoDBImageMetaSuffix  = "ImageMetaTable"
+	DynamoDBUserDataSuffix   = "UserDataTable"
+	DynamoDBAPIKeySuffix     = "ApiKeyDataTable"
+	DynamoDBVersionSuffix    = "VersionTable"
+	RedisLocksBucket         = "locks"
```
rchincha, line 40: Restore this before the new entries. Will make the patch a lot cleaner.

PR #4017, opened 2026-04-26
feat(gc): add gcTimeWindow to limit periodic GC to a local time range

### LINKED ISSUE #3001: [Feat]: Implement gcTimeWindow in order to reduce storage locking during active usage time
this relates to problem described in: https://github.com/project-zot/zot/issues/2999

GC leads to storage locks which is not always good on high loaded setups

Image
the graph shows dependency between gc and storage latency. The gap at apx.13-40 starts after disabling gc

May be it is possible to set gc time window? f/e:
```
    "storage": {
        "gc": true,
        "gcTimeWindow": "01.00 - 08.00"
```
and implement simple time check against config to start gc task or not

_No response_

### Additional context

_No response_

AkashKumar7902:
## Summary

- Adds a new optional `gcTimeWindow` storage setting that confines scheduled garbage collection to a local time range (e.g. `01.00 - 08.00`), supporting both `.` and `:` separators and windows that cross midnight.
- Honours the window for the default store and each subpath, validates the value at config load, and gates the GC task generator's readiness on the window.
- Updates `examples/README.md` and `examples/config-gc-periodic.json` so operators can discover and copy the new option.

Closes #3001

## Implementation notes

- `pkg/storage/gc/time_window.go` exposes `ParseTimeWindow` / `TimeWindow.Contains`. The window is empty by default (matches today's behaviour).
- `pkg/api/controller.go` plumbs the parsed window through `gc.Options` for both the default store and subpaths; an invalid value logs and skips scheduled GC for that store rather than failing the whole server.
- `pkg/cli/server/root.go` rejects malformed values up-front and warns when a window is set without `gc: true`.
- `UpdateReloadableConfig` carries the new field across reloads so live config edits take effect without a restart.

## Test plan

- [x] `go test ./pkg/storage/gc/... ./pkg/api/config/... ./pkg/cli/server/...`
- [x] `golangci-lint run` on the touched packages
- [x] Verified parsing of same-day window, midnight-crossing window, and invalid inputs via unit tests
- [x] Verified GC task generator readiness honours the window via unit test

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -91,6 +91,14 @@ Orphan blobs are removed if they are older than gcDelay.
         "gcDelay": "2h"
 ```
 
+Periodic garbage collection can also be limited to a local time window to avoid
+running during active usage hours. The window accepts `HH.MM - HH.MM` or
+`HH:MM - HH:MM` and can cross midnight.
+
+```
+        "gcTimeWindow": "01.00 - 08.00"
+```
+
 To limit the maximum number of repositories that can be created, set:
 
 ```
--- a/examples/config-gc-periodic.json
+++ b/examples/config-gc-periodic.json
@@ -5,12 +5,14 @@
         "gc": true,
         "gcDelay": "1h",
         "gcInterval": "1h",
+        "gcTimeWindow": "01.00 - 08.00",
         "subPaths": {
             "/a": {
                 "rootDirectory": "/tmp/zot1",
                 "gc": true,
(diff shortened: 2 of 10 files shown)


rchincha:
@AkashKumar7902 pls fix merge conflicts and ci failures

andaaron:
Closing in favor of https://github.com/project-zot/zot/pull/4223

PR #4019, opened 2026-04-26
fix(exporter): forward Authorization header from zxp to zot metrics endpoint

### LINKED ISSUE #3263: [Bug]: zxp doesn't export zot metrics with turned on authentication
### zot version

v2.1.5-9-g552242f5

### Describe the bug

With user authentication (at least htpasswd) turned on in `zot-minimal` config `zxp` exports only some general metrics like `go_*`, `process_*`, but not the `zot_*`

And it doesn't matter, whether you specify user:password in request to `zxp` or not.

zot log:
`
{"level":"info","module":"http","component":"session","clientIP":"127.0.0.1:33456","method":"GET","path":"/metrics","statusCode":401,"latency":"0s","bodySize":266,"headers":{"Accept-Encoding":["gzip"],"User-Agent":["Go-http-client/1.1"]},"goroutine":129,"caller":"zotregistry.dev/zot/pkg/api/session.go:137","time":"2025-07-16T09:35:18.51698461Z","message":"HTTP API"}
`

It looks like `zxp` ignores the authorization header it gets from user and sends an unauthorized request to `zot`

### To reproduce

```
zot-linux-amd64-minimal serve /tmp/zot_test/zot_config2.json &
zxp-linux-amd64 config /tmp/zot_test/zxp_config2.json &
curl -u observability:password http://127.0.0.1:5001/metrics
```

[zxp_config2.json](https://github.com/user-attachments/files/21253207/zxp_config2.json)
[zot_config2.json](https://github.com/user-attachments/files/21253862/zot_config2.json)

### Expected behavior

I expect to receive the full set of metrics, as would happen if the `accessControl` and `auth` sections of the config were removed

### Screenshots

_No response_

### Additional context

_No response_

AkashKumar7902:
**What type of PR is this?**

bug

Fixes #3263

When zot is configured with authentication (e.g. htpasswd), the `zxp` node exporter could not scrape zot's `/metrics` endpoint because it always issued an unauthenticated request. The JSON decoder silently consumed the 401 error body, so the resulting Prometheus output was missing every `zot_*` metric (only `go_*`/`process_*` showed up) and `zot_up` never flipped to 0 to surface the failure.

This PR makes the exporter forward the `Authorization` header from the inbound `/metrics` request to the upstream zot scrape:

- `pkg/exporter/api/exporter.go` swaps the global `prometheus.Register` + `promhttp.Handler()` wiring for a per-request handler. Each request now builds a fresh `prometheus.Registry` and registers a collector clone scoped to that request via `Collector.WithHeaders`, which carries the forwarded `Authorization` header(s) into the scrape.
- `pkg/extensions/monitoring/minimal_client.go` adds `MetricsClient.WithHeaders` and copies any configured headers onto every outgoing request in `makeGETRequest`. Non-2xx responses are now treated as scrape failures instead of being decoded as JSON, so `zot_up` correctly reports 0 and the underlying error is logged.

Behavior with no auth on zot is unchanged because no `Authorization` header is forwarded when none is sent.

N/A — see #3263 for repro steps and logs.

- Added `TestMetricsHandlerForwardsAuthorizationHeader` and `TestMetricsHandlerReportsZotDownWithoutAuthorizationHeader` in `pkg/exporter/api/exporter_internal_test.go` which spin up an httptest server that requires a Basic Authorization header. The first test asserts the exporter forwards the header (200 + `zot_up 1`); the second asserts that without the header the scrape is treated as a failure (`zot_up 0`).
- Added `TestMetricsClientForwardsHeadersAndRejectsNonSuccessStatus` in `pkg/extensions/monitoring/minimal_client_test.go` which verifies that an unauthenticated `GetMetrics()` call now returns `unexpected status code 401` (instead of decoding the error body) and that calling `WithHeaders` produces a client whose request carries the `Authorization` header.
- `go build ./pkg/exporter/... ./pkg/extensions/monitoring/...`
- `go test ./pkg/exporter/api/... ./pkg/extensions/monitoring/...` — both packages pass locally.
- `gofmt -l` and `go vet` clean on the touched files.

**Automation added to e2e**:

No e2e change. The fix is fully covered by the new unit/internal tests against an httptest server that mirrors a 401-protected zot. An e2e test that boots a real htpasswd-protected zot under zxp can be tracked separately if desired.

No. The wire format of the `/metrics` endpoint is unchanged. Existing zxp deployments that do not pass an `Authorization` header continue to work exactly as before; the only behavior change is that scrape failures (any non-2xx from zot) now correctly surface as `zot_up 0` rather than an empty success.

Yes — operators running `zxp` against a zot with authentication enabled can now scrape it by forwarding their Prometheus credentials (e.g. `basic_auth` in the Prometheus scrape config pointed at `zxp`). zxp passes the inbound `Authorization` header straight through to zot.

```release-note
fix(exporter): zxp now forwards the inbound `Authorization` header to zot's /metrics endpoint, so it can scrape zot when authentication (e.g. htpasswd) is enabled. Non-2xx responses from zot are also treated as scrape failures and reported via `zot_up 0`.
```

### DIFF
--- a/pkg/exporter/api/exporter.go
+++ b/pkg/exporter/api/exporter.go
@@ -29,6 +29,13 @@ type Collector struct {
 	invalidChars *regexp.Regexp
 }
 
+func (zc *Collector) WithHeaders(headers http.Header) *Collector {
+	clone := *zc
+	clone.Client = zc.Client.WithHeaders(headers)
+
+	return &clone
+}
+
 // Describe implements prometheus.Collector interface.
 func (zc Collector) Describe(ch chan<- *prometheus.Desc) {
 	for _, metricDescription := range zc.MetricsDesc {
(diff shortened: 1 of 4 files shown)


rchincha:
@AkashKumar7902 do you plan to work on these PRs? Some of them have been moved out of draft state (ones we like).

AkashKumar7902:
@rchincha yes

rchincha:
@AkashKumar7902 pls fix the merge conflicts

PR #4028, opened 2026-05-01
fix(auth): refine OIDC identity handling and claim-mapping logs

andaaron:
Rename getOpenIDUsername to getOpenIDIdentity and thread "identity" through bearer OIDC, Basic-auth parsing, OAuth2Callback, and log fields.

Only fall back (and warn) to the default email claim when the configured username claim is non-default but missing or empty.

Stop emitting Info logs when groups are absent on only UserInfo or only ID token claims; log once at Debug when no groups remain after merging both.

Update ClaimMapping docs to mention username and groups claims; fix mTLS extractIdentity comment typo; clarify GetAuthUserFromRequestSession doc.

### DIFF
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -602,7 +602,7 @@ func bearerAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 			}
 
 			// Try OIDC authentication first if configured
-			var username string
+			var identity string
 
 			var groups []string
 
@@ -611,21 +611,21 @@ func bearerAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 
 				var authenticated bool
 
-				username, groups, authenticated, err = oidcAuthorizer.AuthenticateRequest(request.Context(), header)
+				identity, groups, authenticated, err = oidcAuthorizer.AuthenticateRequest(request.Context(), header)
 				if err == nil && authenticated {
 					// OIDC authentication succeeded
-					ctlr.Log.Debug().Str("username", username).Msg("the OIDC bearer authentication was successful")
+					ctlr.Log.Debug().Str("identity", identity).Msg("the OIDC bearer authentication was successful")
 
 					// Set user context for authorization
 					userAc := reqCtx.NewUserAccessControl()
-					userAc.SetUsername(username)
+					userAc.SetUsername(identity)
 					userAc.AddGroups(groups)
 					userAc.SaveOnRequest(request)
 
 					// Update user groups in MetaDB if available
 					if ctlr.MetaDB != nil {
 						if err := ctlr.MetaDB.SetUserGroups(request.Context(), groups); err != nil {
-							ctlr.Log.Error().Err(err).Str("username", username).Msg("failed to update user profile")
+							ctlr.Log.Error().Err(err).Str("identity", identity).Msg("failed to update user profile")
 							response.WriteHeader(http.StatusInternalServerError)
 
 							return
@@ -1027,10 +1027,10 @@ func getUsernamePasswordBasicAuth(request *http.Request) (string, string, error)
 		return "", "", zerr.ErrParsingAuthHeader
 	}
 
-	username := pair[0]
+	identity := pair[0]
 	passphrase := pair[1]
 
-	return username, passphrase, nil
+	return identity, passphrase, nil
 }
 
 func GetGithubUserInfo(ctx context.Context, client *github.Client, log log.Logger) (string, []string, error) {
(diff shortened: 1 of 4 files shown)


pkg/api/authn.go
```diff
@@ -1283,24 +1288,24 @@ func OAuth2Callback(ctlr *Controller, w http.ResponseWriter, r *http.Request, st
...
+	if err := saveUserLoggedSession(ctlr.CookieStore, w, r, identity, provider, secure, ctlr.Log); err != nil {
 		return "", err
 	}
 
 	if err := ctlr.MetaDB.SetUserGroups(r.Context(), groups); err != nil {
-		ctlr.Log.Error().Err(err).Str("identity", email).Msg("failed to update the user profile")
+		ctlr.Log.Error().Err(err).Str("identity", identity).Msg("failed to update the user profile")
 
 		return "", err
 	}
```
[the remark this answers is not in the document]
andaaron: This is not the case.

pkg/api/authn.go
```diff
@@ -1196,73 +1196,78 @@ func appendOpenIDGroups(groups []string, claims map[string]any, claimName string
...
 			Str("provider", providerName).
-			Str("username", username).
-			Msg("using email as username (fallback)")
+			Str("identity", identity)
+		if fellBackToDefaultClaim {
+			event.Msg("using email as identity (fallback)")
+		} else {
+			event.Msg("using email as identity (default)")
+		}
 	}
```
vrajashkr: nit: maybe we can combine these to an unconditional logger? Something like

```
logger.Debug().
			Str("provider", providerName).
			Str("claim", usernameClaim).
			Str("identity", identity).
            Bool("fellBackToDefault", fellBackToDefaultClaim),
			Msg("extracted identity")
```

pkg/api/authn.go
```diff
@@ -1196,73 +1196,78 @@ func appendOpenIDGroups(groups []string, claims map[string]any, claimName string
...
-		groupsFound bool
-	)
+	var groups []string
 
 	if info != nil {
-		groups, groupsFound = appendOpenIDGroups(groups, info.Claims, groupsClaim)
-		if !groupsFound {
-			logger.Info().Msgf("failed to find any %q claim for user %s in UserInfo", groupsClaim, username)
-		}
+		groups, _ = appendOpenIDGroups(groups, info.Claims, groupsClaim)
```
vrajashkr, line 1238: nit: if we don't need the `groupsFound` return anymore, should we consider only returning groups from `appendOpenIDGroups`?

pkg/api/authn.go
```diff
@@ -1196,73 +1196,78 @@ func appendOpenIDGroups(groups []string, claims map[string]any, claimName string
...
+// extractOpenIDIdentity resolves identity and groups for an OIDC callback
 // based on the provider's configured claim mapping. It returns the resolved
-// username, the deduplicated/sorted groups, and a boolean reporting whether
-// the username could be resolved at all (false means callers should reject).
+// identity string, the deduplicated/sorted groups, and a boolean reporting whether
+// identity could be resolved at all (false means callers should reject).
 func extractOpenIDIdentity(logger log.Logger, authConfig *config.AuthConfig, providerName string,
 	info *oidc.UserInfo, idTokenClaims map[string]any,
 ) (string, []string, bool) {
 	usernameClaim, groupsClaim, usernameConfigured := getOpenIDClaimMapping(authConfig, providerName)
```
vrajashkr: Should we rename this to `identityClaim` and `identityConfigured` since other places have been updated?

pkg/api/authn_internal_test.go
```diff
@@ -161,14 +163,107 @@ func TestGetOpenIDUsername(t *testing.T) {
...
-				t.Fatalf("expected username %q, got %q", test.expected, username)
+			identity := getOpenIDIdentity(test.info, test.claim)
+			if identity != test.expected {
+				t.Fatalf("expected identity %q, got %q", test.expected, identity)
 			}
 		})
 	}
 }
 
+func TestExtractOpenIDIdentity(t *testing.T) {
```
vrajashkr: Maybe these test cases could be added to the `TestGetOpenIDClaimMapping` test that was added here - https://github.com/project-zot/zot/pull/3999/changes#diff-bcc0c291483488732eb23559295f001f24be9a59cbaca57cbcbc54ee04ee6c09R11

pkg/api/authn.go
andaaron: It’s useful if we ever want to log/debug that again.

vrajashkr (approved):
LGTM

rchincha (approved):
lgtm

PR #4030, opened 2026-05-02
refactor(zli): add typed ~/.zot config layer and strict validation

andaaron:
Introduce pkg/cli/client/config.go with ZliConfigFile/ZliConfig and ReadZliConfigFile, replacing the loose map[string]any load/save path in config_cmd.go.

Parsing now rejects malformed JSON with ErrCliBadConfig and requires a non-null configs array (ErrCliMissingConfigsField when wrapped). Each profile must have non-empty _name and url.

Config commands delegate to typed helpers (Find, AddEntry, RemoveEntry, GetVar/SetVar/ResetVar, FormatNames, WriteFile). Fresh or minimal files still behave as empty via isFreshCliRead (ErrEmptyJSON or missing configs).

Tests: prefer t.Setenv("HOME", t.TempDir()) where CLI resolution uses --url only; align CVE/client/search tests with mandatory profile URL and HOME isolation.

### DIFF
--- a/pkg/cli/client/config.go
+++ b/pkg/cli/client/config.go
@@ -1,0 +1,329 @@
+//go:build search
+
+package client
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"os"
+	"slices"
+	"strconv"
+	"strings"
+	"text/tabwriter"
+
+	jsoniter "github.com/json-iterator/go"
+
+	zerr "zotregistry.dev/zot/v2/errors"
+)
+
+const (
+	defaultConfigPerms = 0o644
+	defaultFilePerms   = 0o600
+
+	nameKey           = "_name"
+	showspinnerConfig = "showspinner"
+	verifyTLSConfig   = "verify-tls"
+)
+
+// ZliConfigFile is the on-disk JSON shape for ~/.zot (zli CLI registry profiles).
+type ZliConfigFile struct {
+	Configs []ZliConfig `json:"configs"`
+}
+
+// ZliConfig is one named registry profile inside ZliConfigFile.Configs.
+type ZliConfig struct {
+	Name        string `json:"_name"` //nolint:tagliatelle // persisted ~/.zot schema uses `_name`
+	URL         string `json:"url"`
+	ShowSpinner any    `json:"showspinner,omitempty"`
+	VerifyTLS   any    `json:"verify-tls,omitempty"` //nolint:tagliatelle // hyphenated ~/.zot key
+}
+
+// isConfigUnavailable reports errors that mean there is no usable ~/.zot config yet (empty file
+// or JSON without a "configs" field). Those cases are treated like an empty config for some commands.
+func isConfigUnavailable(err error) bool {
+	return errors.Is(err, zerr.ErrEmptyJSON) || errors.Is(err, zerr.ErrCliMissingConfigsField)
+}
+
+// ReadZliConfigFile reads and validates ~/.zot JSON from path.
+//
+// If the path does not exist yet, OpenFile(..., O_CREATE) creates an empty file first so later
+// ReadFile sees zero bytes and returns [ErrEmptyJSON] ("fresh" CLI state). Adding entries via
+// `zli config add` still works without relying on this side effect.
+func ReadZliConfigFile(filePath string) (*ZliConfigFile, error) {
+	file, err := os.OpenFile(filePath, os.O_RDONLY|os.O_CREATE, defaultConfigPerms)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := file.Close(); err != nil {
+		return nil, err
+	}
+
+	data, err := os.ReadFile(filePath)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(bytes.TrimSpace(data)) == 0 {
+		return nil, zerr.ErrEmptyJSON
+	}
+
+	json := jsoniter.ConfigCompatibleWithStandardLibrary
+
+	var jsonMap map[string]any
+	if unmarshalErr := json.Unmarshal(data, &jsonMap); unmarshalErr != nil {
+		return nil, fmt.Errorf("%w: %w", zerr.ErrCliBadConfig, unmarshalErr)
+	}
+
+	if _, ok := jsonMap["configs"]; !ok || jsonMap["configs"] == nil {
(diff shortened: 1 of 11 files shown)


pkg/cli/client/image_cmd_internal_test.go
```diff
@@ -35,7 +33,7 @@ func TestSearchImageCmd(t *testing.T) {
 	Convey("Test image help", t, func() {
 		args := []string{"--help"}
 
-		_ = makeConfigFile(t, "")
+		t.Setenv("HOME", t.TempDir())
```
vrajashkr, line 36: The previous behavior here was that an empty file would be created. The new code doesn't appear to be doing the same. Is that alright?

The individual tests themselves seem to be creating their own config files as needed, but just wanted to double check.

pkg/cli/client/config.go
```diff
@@ -0,0 +1,329 @@
...
+type ZliConfig struct {
+	Name        string `json:"_name"` //nolint:tagliatelle // persisted ~/.zot schema uses `_name`
+	URL         string `json:"url"`
+	ShowSpinner any    `json:"showspinner,omitempty"`
+	VerifyTLS   any    `json:"verify-tls,omitempty"` //nolint:tagliatelle // hyphenated ~/.zot key
+}
+
+// isFreshCliRead reports errors that mean there is no usable ~/.zot config yet (empty file or
+// JSON without a "configs" field). Those cases are treated like an empty config for some commands.
+func isFreshCliRead(err error) bool {
```
vrajashkr: nit: maybe can be renamed to `isConfigNotAvailable` or something similar might better indicate the work that this function does.

pkg/cli/client/config.go
```diff
@@ -0,0 +1,329 @@
...
+				zerr.ErrCliBadConfig, i, v)
+		}
+
+		c, err := zliConfigFromMap(configMap, i)
+		if err != nil {
+			return nil, err
+		}
+
+		out.Configs = append(out.Configs, c)
+	}
```
vrajashkr, line 106: Just curious - is there any reason we can't directly Unmarshal the the config into a `ZliConfigFile` object?

pkg/cli/client/config_cmd_internal_test.go
```diff
@@ -0,0 +1,363 @@
...
+}
+
+func TestGetConfigValue(t *testing.T) {
+	t.Parallel()
+
+	validProfile := `{"configs":[{"_name":"a","url":"https://example.com"}]}`
+
+	tests := []struct {
+		name           string
+		raw            string // empty => do not write file before call (missing ~/.zot)
```
vrajashkr: nit: could be renamed perhaps to something like "cfgContents" or similar?

Similar comment for the other test tables below.

pkg/cli/client/config_test.go
```diff
@@ -0,0 +1,221 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	zerr "zotregistry.dev/zot/v2/errors"
+	"zotregistry.dev/zot/v2/pkg/cli/client"
+)
+
+func writeZotFile(t *testing.T, dir, json string) string {
```
vrajashkr, line 17: nit: I think this function is duplicated in a few test files. Maybe we could have them in just one file and re-use them across. I think they are in the same package.

pkg/cli/client/config_test.go
```diff
@@ -0,0 +1,221 @@
...
+			assert.ErrorIs(t, cfg.ResetVar(errCase.key), errCase.wantErr)
+		})
+	}
+
+	t.Run("clears_verify_tls", func(t *testing.T) {
+		t.Parallel()
+
+		cfg := client.ZliConfig{Name: "main", URL: "https://example.com", VerifyTLS: false}
+		require.NoError(t, cfg.ResetVar("verify-tls"))
+		assert.Nil(t, cfg.VerifyTLS)
```
vrajashkr, line 219: Just curious - since VerifyTLS has a default value, shouldn't reset set it back to the default and not unset the value entirely?

vrajashkr (commented):
Looks good to me overall. Left a few tiny comments.

pkg/cli/client/image_cmd_internal_test.go
andaaron: We don't need a zli config to run the `--help` command.
Actually, I could have skipped this entire t.Setenv("HOME", t.TempDir()) here since it would be useful only if zli made calls to a server, but kept it just in case.

In the other tests, which are supposed to make requests, I removed the file creation in some cases where the `--url` command line parameter is used. The config doesn't need to exist there and it could potentially interfere with the `--url` parameter testing.

pkg/cli/client/config.go
andaaron: Because I want to make sure the values are valid before returning, and this way we can provide custom ErrCliBadConfig messages and strict rules (configs array shape, _name/url rules) that a bare unmarshal won’t express as clearly to the user.

pkg/cli/client/config_test.go
andaaron: packages client_test vs client cannot share unexported helpers, I don't think it's worth it to export it

pkg/cli/client/config_test.go
andaaron: Today reset clears optional fields and ApplyDefaults fills defaults on read via Find, so behavior is “unset → default when used.” If we change it to persist default after reset, that’s a behavior change

pkg/cli/client/config.go
```diff
@@ -0,0 +1,329 @@
...
+
+// ReadZliConfigFile reads and validates ~/.zot JSON from path.
+//
+// If the path does not exist yet, OpenFile(..., O_CREATE) creates an empty file first so later
+// ReadFile sees zero bytes and returns [ErrEmptyJSON] ("fresh" CLI state). Adding entries via
+// `zli config add` still works without relying on this side effect.
+func ReadZliConfigFile(filePath string) (*ZliConfigFile, error) {
+	file, err := os.OpenFile(filePath, os.O_RDONLY|os.O_CREATE, defaultConfigPerms)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
andaaron: I don't want to update already existing file permissions users may be relying on.

vrajashkr (approved):
LGTM

PR #3711, opened 2026-01-17
feat(oidc): introduce support for OIDC workload identity federation

matheuscscp:
Closes: #3704

Builds on/Supersedes: #3705

* Support multiple OIDC issuers. This is crucial for a single Zot instance to support multiple Kubernetes clusters, since each cluster has its own issuer URL.
* Support CEL expressions for a more powerful ability of validation and extraction of username and groups from the OIDC claims. This is critical for security integrations like this as users normally have to implement additional assertions to comply with security requirements and properly isolate identities. Two strong examples of this mechanism are the GCP Workload Identity Federation [feature](https://docs.cloud.google.com/iam/docs/workload-identity-federation#mapping) where you can use CEL expressions to do exactly what I'm proposing here, and also what we implemented in Flux Operator for OIDC SSO, [same thing](https://fluxoperator.dev/docs/web-ui/web-config-api/#claims-processing-with-cel).
* Load the OIDC issuer lazily. This is crucial for the Zot server not to crash on startup due to an OIDC issuer server that is temporarily unavailable. This feature should not impact the Zot server health, as there are other authentication mechanisms available. We implement the same strategy in Flux Operator.

PR #4036, opened 2026-05-05
Introduce expiration field for `accessConfig` policies

matheuscscp:
### Use case

Similar to #3755 (implemented via #3761), but now for the `accessConfig` policies. It turns out that issuing customer licenses as simple "ID cards" i.e. without the permissions embedded directly on them is much better (i.e. use OIDC federation), because we don't have to re-issue a license for a customer every time their permissions change. This feature allows us to configure the customer access on the server side via static Zot config (e.g. giving a customer limited access to a container image as a free trial, for example), and they can always use the same license, which is now just an attestation of their identity.

### Why this feature belongs in Zot from a separation of concerns perspective

This is an important feature for OIDC federation users: OIDC's responsibility is to simply identify a user/identity (authn), and the relying party (Zot) is fully responsible for authz. When OIDC federation was introduced (via #3711), we agreed that it would entirely rely on `accessConfig`. So to implement a rule like this for OIDC, the only system that can do it is Zot, the relying party. There's nowhere else  where this rule can be enforced from OIDC perspective. So the separation of concerns is:

* Zot delegates authn to the configured OIDC provider
* Zot takes full responsibility for the authz part

Furthermore, a feature like this for access control is very common and it would simply enrich Zot's capabilities. All the three major cloud providers have this feature, see below.

AWS:

https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_examples_aws-dates.html

```json
{
    "Version":"2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": "service-prefix:action-name",
            "Resource": "*",
            "Condition": {
                "DateGreaterThan": {"aws:CurrentTime": "2020-04-01T00:00:00Z"},
                "DateLessThan": {"aws:CurrentTime": "2020-06-30T23:59:59Z"}
            }
        }
    ]
}
```

GCP:

https://docs.cloud.google.com/iam/docs/conditions-overview#example-date-time

```cel
request.time < timestamp('2021-01-01T00:00:00Z')
```

Azure:

https://learn.microsoft.com/en-us/azure/role-based-access-control/pim-integration

```json
{
    "properties": {
        "principalId": "aaaaaaaa-bbbb-cccc-1111-222222222222",
        "roleDefinitionId": "/subscriptions/.../providers/Microsoft.Authorization/roleDefinitions/c8d4ff99-41c3-41a8-9f60-21dfdad59608",
        "requestType": "AdminAssign",
        "scheduleInfo": {
            "startDateTime": "2020-04-01T00:00:00Z",
            "expiration": {
                "type": "AfterDateTime",
                "endDateTime": "2020-06-30T23:59:59Z"
            }
        }
    }
}
```

### DIFF
--- a/pkg/api/authz.go
+++ b/pkg/api/authz.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net/http"
 	"slices"
+	"time"
 
 	glob "github.com/bmatcuk/doublestar/v4"
 	"github.com/gorilla/mux"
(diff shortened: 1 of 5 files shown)


matheuscscp:
@rchincha @andaaron Ramkumar told me in Slack that you were debating internally if this feature makes sense to be in Zot at all from a separation of concerns perspective. I think it does, as it is a simple and common access control feature, all the major cloud providers have it, please see the updated PR description.

matheuscscp:
Superseded by: #4040

PR #4037, opened 2026-05-05
feat(zli): add config list/show/get/set/reset and isolate deprecated syntax

andaaron:
Introduce first-class subcommands for listing profiles, showing a profile, getting and setting keys, and resetting optional keys (alongside existing add/remove). The parent command now resolves ~/.zot via zliUserConfigPath(), documents that profile names must not clash with subcommand names, and states that positional/--list/--reset usage is deprecated and will be removed soon.

Legacy behavior is delegated to config_cmd_deprecated.go with stderr warnings for old flags and positional get/set. Examples and inline help point users at the new commands. FormatNames/FormatListedVars comments reference config list/show.

Tests are split, so config_cmd_test.go exercises the supported subcommands while config_cmd_deprecated_test.go retains coverage for the deprecated paths under renamed TestConfigCmdDeprecated* entries.

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -87,6 +87,7 @@ var (
 	ErrScanNotSupported                 = errors.New("scanning is not supported for given media type")
 	ErrCLITimeout                       = errors.New("query timed out while waiting for results")
 	ErrDuplicateConfigName              = errors.New("cli config name already added")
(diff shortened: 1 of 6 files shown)


rchincha (approved):
lgtm

PR #4048, opened 2026-05-09
chore: fix dependabot alerts

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


pkg/api/authn.go
```diff
@@ -76,12 +76,13 @@ func (amw *AuthnMiddleware) sessionAuthn(ctlr *Controller, userAc *reqCtx.UserAc
...
-			Name:    "session",
-			Value:   "",
-			Path:    "/",
-			Expires: time.Unix(0, 0),
-
+			Name:     "session",
+			Value:    "",
+			Path:     "/",
+			Expires:  time.Unix(0, 0),
+			Secure:   true,
```
andaaron: ## Findings

### 1) `Secure: true` unconditionally is risky (likely breaks HTTP/local-dev)

- **Positive intent**: session cookies *should* be `Secure` on HTTPS deployments.
- **Issue**: Setting `Secure` unconditionally means browsers will typically **ignore** `Set-Cookie: ...; Secure` when the response is served over **plain HTTP**.
  - Consequence: the invalid/expired session cookie may **not be cleared/overwritten** at all.
  - Resulting behavior: clients keep sending the old `session` cookie, repeatedly hitting the “invalid/expired” flow.

This is especially problematic because:

- The `session` cookie is **`HttpOnly`**, so clients cannot reliably remove it via JS.
- If the server can’t overwrite/delete it (because the cookie write is ignored due to `Secure`), the client can be stuck in a “permanent 401 + logout loop” pattern.

### 2) `SameSite=Lax` on the *deletion cookie* is inconsistent with the *session cookie*

Elsewhere, zot sets session cookie options via `session.Options` (e.g. `saveUserLoggedSession`) and uses:

- `SameSite: http.SameSiteDefaultMode`
- `Secure: secure` (derived from config/request)

Adding `SameSite=Lax` only on the invalidation cookie doesn’t establish a consistent cookie policy; it just makes the “delete cookie” different from the “set cookie” path.

## zui client impact (zui is the cookie consumer)

### How zui uses cookies

- zui’s authenticated API calls rely on the backend-managed **`session`** cookie.
  - zui enables `axios` `withCredentials` **only when UI origin != API origin** (cross-origin) and auth is enabled.
- zui’s “logged in” UI state is based on a separate **non-HttpOnly** `user` cookie (readable from `document.cookie`).
- On 401s, zui calls `/zot/auth/logout`, clears the `user` cookie via JS, and redirects to `/login` (or an IdP end-session URL).

### Why unconditional `Secure: true` is a zui-specific footgun

- zui **cannot** delete `session` (it’s `HttpOnly`), so it depends on the backend successfully clearing it.
- In common dev setups (HTTP, localhost ports, or non-TLS reverse proxies), browsers may ignore Secure cookie updates → **backend cannot clear the session** → zui can get stuck repeatedly receiving 401s and attempting logout/redirect while the `session` cookie remains.

## Recommendations

### A) Don’t hardcode `Secure: true` in `sessionAuthn` invalidation

Set `Secure` based on the same decision used when creating/saving the real session cookie:

- config-driven secure decision (e.g. `UseSecureSession()`), and/or
- request TLS / forwarded proto logic (if that’s how zot is intended to behave behind proxies)

### B) Align `SameSite` policy consistently

If the goal is to tighten cookie policy, apply it at the point where the **session cookie is set** (e.g. `saveUserLoggedSession` and other session-save paths), not only when clearing it.

andaaron:
This depends on the change in https://github.com/project-zot/zot/pull/4049 for the dockerfile/stacker changes.

andaaron (approved):

PR #4056, opened 2026-05-11
fix: downgrade expected missing-blob HEAD logging from error to debug

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/local/local_test.go
+++ b/pkg/storage/local/local_test.go
@@ -893,6 +893,92 @@ func FuzzCheckBlob(f *testing.F) {
 	})
 }
 
+func TestMissingBlobChecksDoNotLogErrors(t *testing.T) {
+	t.Parallel()
+
+	newImageStore := func(t *testing.T) (storageTypes.ImageStore, *bytes.Buffer) {
+		t.Helper()
+
+		var buf bytes.Buffer
+
+		logger := zlog.NewLoggerWithWriter("debug", &buf)
+		metrics := monitoring.NewMetricsServer(false, logger)
+		t.Cleanup(metrics.Stop)
+
+		dir := t.TempDir()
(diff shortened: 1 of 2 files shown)


pkg/storage/local/local_test.go
```diff
@@ -893,6 +893,92 @@ func FuzzCheckBlob(f *testing.F) {
...
+		t.Cleanup(metrics.Stop)
+
+		dir := t.TempDir()
+		cacheDriver, err := storage.Create("boltdb", cache.BoltDBDriverParameters{
+			RootDir:     dir,
+			Name:        "cache",
+			UseRelPaths: true,
+		}, logger)
+		if err != nil {
+			t.Fatalf("create cache driver: %v", err)
```
andaaron, line 915: Can we use `github.com/stretchr/testify/assert` for assertions instead of this pattern?

andaaron (approved):

pkg/storage/local/local_test.go
rchincha: But we are using t.Fatalf() everywhere. This should be a project-wide refactor, to be fair.

vrajashkr: IMO, we could choose one approach and use it consistently in new code. We could offload the refactor to Copilot perhaps.

I like the testify API since it allows for fewer lines of code, but it does add the additional dependency. I'm good either way though.

andaaron: > But we are using t.Fatalf() everywhere. This should be a project-wide refactor, to be fair.

Not everywhere. There's just a few files which were mostly added by AI PRs in the past year or so. Let's not persist in this.
So at least for NEW tests, let's use `github.com/stretchr/testify/assert`.

PR #4063, opened 2026-05-12
Pr/stream blobs on demand

mrudnoru:
**What type of PR is this?**
 
feature
 
N/A
 
Related discussions:
 
- [On-Demand Proxy Download Progress](https://github.com/project-zot/zot/discussions?discussions_q=On-Demand+Proxy+Download+Progress)
- [Partial mirror aka download blobs on demand](https://github.com/project-zot/zot/discussions?discussions_q=Partial+mirror+aka+download+blobs+on+demand)

Adds optional blob streaming for sync on-demand registries.
 
When `"stream": true` is set on a sync registry entry, zot can stream a missing blob from the upstream registry to the client while also writing it into local storage. Without this, a client requesting an uncached blob may wait until zot fully downloads and stores the blob before receiving data.
 
If `"stream"` is unset or `false`, existing behavior is unchanged.
 
Large blob pulls through on-demand sync can leave clients or intermediaries idle while zot fetches and stores content from the upstream registry. For multi-GB layers, this can interact badly with client, proxy, or runtime timeouts because no response body bytes are sent while zot is still fetching the blob.
 
```console
$ go test -tags 'sync lint' ./pkg/extensions/sync -run TestOnDemandBlobStreaming -count=1
ok  	zotregistry.dev/zot/v2/pkg/extensions/sync	2.776s
```
 
Manual validation was done with a streaming-enabled build in a multi-replica pull-through cache environment. We tested large image pulls through zot's on-demand mirror path while tracking pull success, retries, HTTP/2 stream errors, context cancellations, panics, and HTTP 5xx responses. Test pulls completed successfully, and zot logs showed no panics or stream errors during the rollout/test window.
 
Broader validation of the custom build showed repeated successful large-image and fan-out pulls; blob streaming specifically covers the blob GET miss path and is verified by the unit test above.
 
**Automation added to e2e**:
 
No e2e automation added. Unit coverage was added for streamed on-demand blob coordination.
 
No. The new `stream` config field is optional and defaults to disabled.
 
Yes. A new optional sync registry config field is added: `stream`.
 
Added optional `stream` support for sync on-demand registries. When enabled, zot can stream missing blobs from upstream to the client while caching them locally.
 
---
 

mrudnoru:
We've been running this patch in production-shaped testing on our zot deployment and wanted to share results in case they're useful to reviewers.

## Environment

- 2 Kubernetes clusters, 3 zot replicas each (`Deployment`)
- S3 (Ceph) blob storage + Redis cache
- On-demand sync from Docker Hub
- Kubelet + containerd as the pull client

## Test images

- `nvidia/cuda:12.8.1-devel-ubuntu24.04` - 5.2 GB (3 GB single layer)
- `pytorch:2.5.1-cuda12.4-cudnn9-devel` - 7.4 GB (3.3 GB single layer)
- `golang:1.23` - 324 MB (fan-out)

## Cold-pull procedure per run

1. Delete cached manifests + S3 blobs for the target image
2. Clear containerd cache on the target node
3. Rolling-restart zot pods
4. `kubectl run pull-test` on a chosen node
5. Record:
   - wall clock to `Running`
   - kubelet attempts
   - HTTP/2 errors
   - `context canceled`
   - regclient retries
   - EOF/RST counts

## Results - cold pulls with streaming enabled

| Image | Size | Runs | Success | Time range |
|---|---:|---:|:---:|---:|
| CUDA | 5.2 GB | 13 | 100% | 105–764s |
| PyTorch | 7.4 GB | 3 | 100% | 443–497s |
| Golang (3-node fan-out) | 324 MB | 3 | 100% | 130–142s |
| CUDA (2-node fan-out) | 5.2 GB | 2 | 100% | 792s |

**21 runs total** - zero crashes and zero stream-related errors.

## A/B isolating the `stream` flag`

(Same binary, only the flag toggled)

| stream | Tag | Time | Result |
|---|---|---:|:---:|
| `false` | `cuda:13.2.1` | 865s | SUCCESS |
| `true` | `cuda:13.2.0` | 701s | SUCCESS |


Happy to share more detail on any specific run.

rchincha:
| We've been running this patch in production-shaped testing on our zot deployment and wanted to share results in case they're useful to reviewers.

@mrudnoru first, thanks for this PR.

I would expect improvements to be most pronounced when container images are very large (AI model use cases etc)

From your results, it appears there is a 20% improvement in overall time.

One other datapoint to track is time-to-first-byte since clients could still timeout despite the overall improvement.

curl -o /dev/null -s -w "TTFB: %{time_starttransfer}\n" https://example.com

If it helps, we also build a `zb` tool available via `make bench` which also runs various tests and measures ttfb.
https://github.com/project-zot/zot/actions/runs/25718628429/job/75521582453?pr=4063

Pls also take a look at: https://github.com/project-zot/zot/pull/3778 (draft, use as reference)

rchincha:
` git commit --amend --no-edit -S -s`

to sign and sign-off your commits (but would wait till the end)

rchincha:
https://github.com/project-zot/zot/pull/4065

here is an addendum PR to help address copilot reviews

examples/README.md
```diff
@@ -1266,8 +1266,9 @@ Configure each registry sync:
 				"urls": ["https://index.docker.io"],
 				"onDemand": true,                     # doesn't have content, don't periodically pull, pull just on demand.
 				"tlsVerify": true,
-				"maxRetries": 3,                      
-				"retryDelay": "15m"
+				"maxRetries": 3,
+				"retryDelay": "15m",
+				"stream": true                        # stream blobs to clients while fetching from upstream on-demand (requires onDemand: true)
```
rchincha: | requires onDemand



Can be enforced in pkg/cli/server/root.go#L1105
LoadConfiguration().

Either error out or automatically turn onDemand on.

pkg/extensions/sync/on_demand.go
```diff
@@ -30,16 +41,119 @@ process just the first one, also keep track of all background retrying routines.
...
+}
+
+func (onDemand *BaseOnDemand) IsStreamEnabled() bool {
+	return onDemand.streamEnabled
+}
+
+func (onDemand *BaseOnDemand) SyncBlobOnDemand(ctx context.Context, repo string,
+	digest godigest.Digest, imgStore storageTypes.ImageStore,
+) (io.ReadCloser, int64, bool, <-chan struct{}, error) {
+	key := repo + "@" + digest.String()
```
rchincha, line 91: 👍

pkg/extensions/sync/on_demand.go
```diff
@@ -30,16 +41,119 @@ process just the first one, also keep track of all background retrying routines.
...
+
+	onDemand.blobInflightMu.Lock()
+
+	if ok, _, _ := imgStore.CheckBlob(repo, digest); ok {
+		onDemand.blobInflightMu.Unlock()
```
rchincha: any particular reason not using the defer()?

rchincha:
@mrudnoru are you able to finish this PR? We have an upcoming release and would like to include it.

pkg/extensions/sync/on_demand.go
mrudnoru: Its a good point - Ive changed to defer as its generally safer for locking

mrudnoru:
> @mrudnoru are you able to finish this PR? We have an upcoming release and would like to include it.

Yes Ive applied changes requested from copilot and some minor fixes from your review @rchincha

examples/README.md
mrudnoru: Yes 100% agree Ive enforced it so either both are true or we throw ErrBadConfig

rchincha:
|  Error: pkg/api/routes_test.go:2484:14: do not define dynamic errors, use wrapped static errors instead: "errors.New(\"cache commit failed\")" (err113)
  	cacheErr := errors.New("cache commit failed")
  	
  If specific to this test file, would define this error in routes_test.go itself.

rchincha:
Missing sign-off(s):

	dfe54a8377d228e61f7eac28863c172391f4291a
		no sign-off found
	caa7a50d51c0d03d196258c477d41ef95bb32f46
		no sign-off found
	7a8fca3a5505d2d97a3b573abb094be486d4fd55
		no sign-off found
	1d261cff8160aa4252cab7386691195c369ab32f
		no sign-off found
	7b774a8c1617f534d876a794c525ca7dfae46837
		no sign-off found
	5d48ae9ef1477a2087cd8f2102927e6461c74437
		no sign-off found
		
		
`git commit --amend --no-edit -S -s`

to sign and sign-off your commits (but would wait till the end)

rchincha:
Pls also consider adding a BATS test under `test/blackbox`

You will need to update `test/ports.json` and `test/blackbox/ci.sh#L18`

pkg/extensions/sync/on_demand.go
```diff
@@ -30,16 +58,134 @@ process just the first one, also keep track of all background retrying routines.
...
+			Msg("blob already being downloaded, waiting on channel")
+
+		select {
+		case <-inf.ready:
+			if inf.err != nil {
+				return nil, 0, false, nil, inf.err
+			}
+		case <-ctx.Done():
+			return nil, 0, false, nil, ctx.Err()
+		}
```
vrajashkr, line 126: If my understanding is correct, I believe that in a use case where there are multiple clients downloading the same blob from zot at once, clients that are not the first client would have to wait for the download to complete before they can start downloading the image. Is there a way we can have the other clients stream as well by "latching on" to the file currently being downloaded so that the experience for a client is uniform?

For example, if a new image is being rolled out to multiple Kubernetes clusters, each node will start an image download within a short duration of each other. The first one will stream the blobs while the others wait.

test/blackbox/sync_stream_on_demand.bats
```diff
@@ -0,0 +1,195 @@
...
+}
+
+function teardown_file() {
+    zot_stop_all
+}
+
+@test "stream on-demand pulls image manifest from upstream" {
+    zot_port1=`cat ${BATS_FILE_TMPDIR}/zot.port1`
+    zot_port2=`cat ${BATS_FILE_TMPDIR}/zot.port2`
+
```
vrajashkr, line 118: nit: to make things a bit more readable in the test code that follows, it might be helpful to name these as "zot_upstream_port" and "zot_port".

Similar comment for the other instances where the ports are being read in this file.

vrajashkr (commented):
Thanks for the PR! I've left a couple of comments.

rchincha:
@mrudnoru 

do you have data on this?
https://github.com/project-zot/zot/pull/4063#issuecomment-4433064463

mrudnoru:
> @mrudnoru
> 
> do you have data on this? [#4063 (comment)](https://github.com/project-zot/zot/pull/4063#issuecomment-4433064463)

Yes indeed this improvement is for large images only.

As requested Ive ran an A/B with `zb` to pin down TTFB. 

### Setup

Two local zot processes (no S3, single host — pure feature measurement):
- Upstream zot on `:8081`
- Downstream zot on `:8080` with `sync.onDemand=true`, toggling `stream`
- Cold data dir for each downstream run
- `zb -d /tmp/zb --test-regex "^On-demand Sync.*" --upstream-server-url http://127.0.0.1:8081 http://127.0.0.1:8080`

### `On-demand Sync 100MB` — 3 runs each

| Run | Mode | Manifest HEAD TTFB | Manifest GET TTFB | Layer TTFB |
|-----|------|--------------------|-------------------|------------|
| 1 | `stream=true`  | 306.94 ms | 1.14 ms | 1.83 ms |
| 2 | `stream=true`  | 311.87 ms | 1.26 ms | 1.90 ms |
| 3 | `stream=true`  | 266.69 ms | 1.74 ms | 1.74 ms |
| 1 | `stream=false` | 262.40 ms | 1.26 ms | 2.16 ms |
| 2 | `stream=false` | 265.88 ms | 1.34 ms | 1.76 ms |
| 3 | `stream=false` | 263.61 ms | 1.14 ms | 1.85 ms |

### `On-demand Sync 1GB` — 1 run each

| Mode | Manifest HEAD TTFB | Manifest GET TTFB | Layer TTFB |
|------|--------------------|-------------------|------------|
| `stream=true`  | 3.945 s | 2.01 ms | 1.90 ms |
| `stream=false` | 4.055 s | 2.14 ms | 5.95 ms |

### Conclusion

Im not sure what those zb tests are suppose to show in context of this feature but for image that small and without S3 attached the results wont be great. The streaming only runs when a client asks for a specific blob by its digest without first fetching the image manifest. At least we have confirm that feature is non-regressing.

### Why this is a big win on our K8s cluster

Without streaming, zot has to pull the whole image, save it to S3, and only then send it to the client. For big images on S3 sometimes that takes longer than our timeouts allow, so the pull just fails. With `stream=true`, zot starts sending bytes to the client right away and saves to S3 in the background so the pull finishes instead of timing out.

pkg/extensions/sync/on_demand.go
mrudnoru: Yes u are correct. If we have 100 nodes in the cluster only the one that arrives **first** to zot claims the slot and starts streaming from upstream. The rest waits. I think its a good suggestion to "latch onto" the stream that Node 1 is already pulling.
I think the solution would be to extend the current solution so that when extra clients show up mid-download they first read whatever bytes are already in the local cache then hook into the live stream from upstream for the rest instead of waiting for the first client to finish. If a client is too slow to keep up we drop it the same way the code already handles a broken leader connection. That avoids adding a memory buffer or changing the storage interface.
BUT I'm not sure if this should be part of this PR?

A simpler alternative but would be to leave the leaders code alone and just change what late clients do: instead of waiting for the full download to finish they read the partial cache file every 100ms and send any new bytes to the client until the blob is complete. It avoids the multi-writer logic entirely but it works only when storage is local filesystem for S3 backed deployments we either keep todays wait-then-read or add a small "read partial blob" hook to the storage interface later.

What do you think?

rchincha:
| With stream=true, zot starts sending bytes to the client right aw ay and saves to S3 in the background so the pull finishes instead of timing out.

The caveat is that validity of images is unknown until the image is downloaded. So clients could be getting invalid images. Without streaming, zot always checks everything.

rchincha:
Another concern is the there is too much going in pkg/api/routes.go wrt storage etc.
We should move all that code out to pkg/api/blobstream.go

We are still debating pros and cons wrt https://github.com/project-zot/zot/pull/3778 which is a competing implementation.

WMP:
Hi, I worked with @mrudnoru on this PR. The point you mentioned: "The caveat is that the validity of images is unknown until the image is downloaded. So clients could be getting invalid images. Without streaming, zot always checks everything," is indeed true. However, in our specific use case, Zot serves as a cache for Docker when dealing with large images.

Without streaming, downloading large images without a local cache caused timeouts in kubelet, leading to errors. kubelet would then restart the download from scratch (I don’t recall exactly whether this consumes Docker’s rate limits). Meanwhile, Zot was still working on fetching the first image - downloading layers, validating, and uploading to S3. This cycle caused further kubelet errors. In extreme cases, it took over ten minutes just to successfully pull the entire image.

Enabling the streaming option simply makes it work. As @mrudnoru demonstrated in her first comment on this PR, there are no errors. kubelet is aware that the image is being pulled and waits patiently, avoiding timeouts because at least some data is being sent.

I believe that in future PRs, we should consider handling edge cases, such as errors occurring while downloading specific layers.

vrajashkr:
@mrudnoru @WMP in your implementation, I see that you have the stream setting per upstream registry. Is there a particular reason you would like that control per upstream registry or would it work for you if zot did streaming for all upstream registries?

The streaming behavior is technically a zot functionality and not related to the upstream as such so I was curious about your use case.

mrudnoru:
> @mrudnoru @WMP in your implementation, I see that you have the stream setting per upstream registry. Is there a particular reason you would like that control per upstream registry or would it work for you if zot did streaming for all upstream registries?
> 
> The streaming behavior is technically a zot functionality and not related to the upstream as such so I was curious about your use case.

We're migrating from Harbor to zot and during the transition we run zot in front of two upstreams -> **Harbor** (our existing internal registry) and **Docker Hub**. Harbor has known issues with HTTP/2 and we've seen similar instability around streaming responses through it, so for that upstream we'd rather keep the proven download-then-serve path. Docker Hub is where we want streaming (large layers, transcontinental hops, where TTFB matters most). Per-registry gives us that split during the migration.
That said if the preference is a global setting, we can live with that and just toggle it globally once Harbor is fully retired. 
Let me know which direction you'd like and I'll push the change.

vrajashkr:
> > @mrudnoru @WMP in your implementation, I see that you have the stream setting per upstream registry. Is there a particular reason you would like that control per upstream registry or would it work for you if zot did streaming for all upstream registries?
> > The streaming behavior is technically a zot functionality and not related to the upstream as such so I was curious about your use case.
> 
> We're migrating from Harbor to zot and during the transition we run zot in front of two upstreams -> **Harbor** (our existing internal registry) and **Docker Hub**. Harbor has known issues with HTTP/2 and we've seen similar instability around streaming responses through it, so for that upstream we'd rather keep the proven download-then-serve path. Docker Hub is where we want streaming (large layers, transcontinental hops, where TTFB matters most). Per-registry gives us that split during the migration. That said if the preference is a global setting, we can live with that and just toggle it globally once Harbor is fully retired. Let me know which direction you'd like and I'll push the change.

Thanks for the details! We've been working on #3778 which implements blob streaming and has multi-client streaming support out of box. We have some tests with kubelet (kind) and multiple streaming clients (skopeo) passing. There is still some hardening needed and some additional work to round it up, but, if you happen to have a test setup, might you be able to try out the changes in #3778 and let us know how it goes? Any feedback on it would be super helpful. I'd like to know if it works well for your usecase.

test/blackbox/sync_stream_on_demand.bats
mrudnoru: Done in b1b40085 btw

mrudnoru:
Hi @rchincha @vrajashkr I did some cleanup meaning rebased + signed off + lint fix. Can u approve ci run?

PR #4068, opened 2026-05-14
fix: skip DynamoDB table creation when tables exist

### LINKED ISSUE #4066: [Bug]: DynamoDB no creation
### zot version

v2.1.16

### Describe the bug

When starting Zot, it tries to create the DynamoDB before starting. If the instance don't have access to creation action, it will exit with the following error message: 

> {"time":"2026-05-13T13:44:37.139713741Z","level":"error","message":"failed to init controller","error":"operation error DynamoDB: CreateTable, https response error StatusCode: 400,

### To reproduce

1. Create a dynamoDB table in AWS
2. In an AWS instance with DynamodB data read only, try to run Zot with a config similar to this
3. Get the error message previously mentioned

``` json
{
    "distSpecVersion": "1.1.1",
    "storage": {
        "rootDirectory": "/tmp/zot",
        "dedupe": true,
        "storageDriver": {
            "name": "s3",
            "rootdirectory": "/",
            "region": "ca-central-1",
            "bucket": "mybucket",
            "forcepathstyle": true,
            "secure": true,
            "skipverify": false
        },
        "cacheDriver": {
            "name": "dynamodb",
            "endpoint": "https://dynamodb.us-east-1.amazonaws.com",
            "region": "us-east-1",
            "cacheTablename": "zot-blob",
            "repoMetaTablename": "zot-repometadata",
            "imageMetaTablename": "zot-imagemeta",
            "repoBlobsInfoTablename": "zot-repoblobinfo",
            "userDataTablename": "zot-userdata",
            "apiKeyTablename": "zot-apikey",
            "versionTablename": "zot-version"
        }
    },
    "http": {
        "address": "0.0.0.0",
        "port": "8080"
    },
    "log": {
        "level": "debug"
    }
}

```

### Expected behavior

Zot should try a describe table and a data RO call instead of trying a creation. This avoid granting too much privileged when DynamoDB tables are created before deploying Zot and let user to keep instances with least privileged when needed. 

This call should be fine 
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_DescribeTable.html

### Screenshots

_No response_

### Additional context

The code here try the creation and manage the exception if the table already exists instead of describe the table first and try to create is not exists

https://github.com/project-zot/zot/blob/main/pkg/meta/dynamodb/dynamodb.go#L2269
https://github.com/project-zot/zot/blob/main/pkg/meta/dynamodb/dynamodb.go#L2328

mturac:
## Summary
- check whether DynamoDB tables already exist before attempting creation
- skip CreateTable when DescribeTable confirms the table is present
- keep create behavior for missing tables

Fixes #4066.

## Tests
- go test ./pkg/meta/dynamodb -run TestWrapperErrors/Create -count=1
- go test ./pkg/meta/dynamodb -count=1

### DIFF
--- a/pkg/meta/dynamodb/dynamodb.go
+++ b/pkg/meta/dynamodb/dynamodb.go
@@ -63,33 +63,59 @@ func New(client *dynamodb.Client, params DBDriverParameters, log log.Logger,
 		return nil, err
 	}
 
-	err = dynamoWrapper.createTable(dynamoWrapper.RepoMetaTablename)
+	err = dynamoWrapper.createTableIfNotExists(dynamoWrapper.RepoMetaTablename, func() error {
+		return dynamoWrapper.createTable(dynamoWrapper.RepoMetaTablename)
+	})
 	if err != nil {
 		return nil, err
 	}
 
-	err = dynamoWrapper.createTable(dynamoWrapper.RepoBlobsTablename)
+	err = dynamoWrapper.createTableIfNotExists(dynamoWrapper.RepoBlobsTablename, func() error {
+		return dynamoWrapper.createTable(dynamoWrapper.RepoBlobsTablename)
+	})
 	if err != nil {
 		return nil, err
 	}
 
-	err = dynamoWrapper.createTable(dynamoWrapper.ImageMetaTablename)
+	err = dynamoWrapper.createTableIfNotExists(dynamoWrapper.ImageMetaTablename, func() error {
+		return dynamoWrapper.createTable(dynamoWrapper.ImageMetaTablename)
+	})
 	if err != nil {
 		return nil, err
 	}
 
-	err = dynamoWrapper.createTable(dynamoWrapper.UserDataTablename)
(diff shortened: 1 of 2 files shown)


rchincha:
@mturac thanks for the PR, pls take a look at the comments.

mturac:
Thanks, addressed in the latest push.

The version table now goes through `createVersionTable()` directly, so pre-created empty version tables still get the initial `DBVersion` row while already-populated version rows remain untouched. I also added a regression test for `New()` with a pre-existing version table missing the version row.

I amended the commit with a `Signed-off-by` trailer for DCO as well.

Tests:
- `go test ./pkg/meta/dynamodb -run 'TestWrapperErrors/Create version table behavior' -count=1`
- `go test ./pkg/meta/dynamodb -count=1`

rchincha:
@mturac pls take a look at copilot review comments.

andaaron:
Let's continue development on https://github.com/project-zot/zot/pull/4120

PR #4075, opened 2026-05-18
 feat(sync): use HTTP/2 with automatic HTTP/1.1 fallback on framing errors

mrudnoru:
## What type of PR is this?

feature

## Which issue does this PR fix

N/A

## What does this PR do / Why do we need it

Adds automatic HTTP/1.1 fallback for sync's outbound HTTP client when the upstream registry produces HTTP/2 framing errors (`PROTOCOL_ERROR`, `INTERNAL_ERROR`, `REFUSED_STREAM`, GOAWAY, raw SETTINGS frames on an HTTP/1.1 connection).

- Sync dials upstream preferring HTTP/2. The primary transport is `http.DefaultTransport.Clone()` configured with `golang.org/x/net/http2.ConfigureTransport()`, so framing failures surface as typed `http2.StreamError` / `*http2.GoAwayError` values and are detected with `errors.As`.
- One case stays a substring match: `malformed HTTP response`, produced when an HTTP/1.1 connection receives raw HTTP/2 SETTINGS frames. `net/http` returns an unexported `badStringError` for that path, so there is no typed error to match.
- On a framing error the request is retried transparently on an HTTP/1.1-only transport. Requests whose body cannot be rewound (`GetBody == nil`) are not retried.
- The fallback decision is remembered per host for 15 minutes (`http2FallbackStickyTTL`); after the TTL the host gets another HTTP/2 attempt, so a temporary upstream issue does not pin a host to HTTP/1.1 forever.
- Because the wrapper is an opaque `http.RoundTripper`, regclient cannot inject its per-host TLS settings into it (regclient only does that for plain `*http.Transport` values). The wrapper therefore builds and carries the TLS material itself, mirroring regclient's behavior: root CAs from `<certDir>/<host>/*.crt` and `/etc/docker/certs.d/<host>/*.crt`, the registry CA and client cert pair from the flat `certDir`, and `InsecureSkipVerify` for `tlsVerify: false`. Without this, sync against upstreams that need a custom CA, mutual TLS, or `tlsVerify: false` would break.

If the upstream registry handles HTTP/2 cleanly, behavior is unchanged.

## If an issue # is not available please add repro steps and logs showing the issue

Some upstream registries break HTTP/2 framing, particularly behind certain LB/CDN paths. The failure we hit in production-shaped testing (details and numbers in [this comment](https://github.com/project-zot/zot/pull/4075#issuecomment-4481827436)): Docker Hub's LB intermittently sends raw HTTP/2 SETTINGS frames on connections Go's `net/http` opened as HTTP/1.1:

```text
net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x00\x00\x12\x04..."
```

The fallback also covers the typed HTTP/2 framing errors the x/net transport can surface (representative forms, exact Go formats):

```text
stream error: stream ID N; PROTOCOL_ERROR
http2: server sent GOAWAY and closed the connection; LastStreamID=N, ErrCode=INTERNAL_ERROR, debug=""
```

These errors surface as failed sync operations: the client request (manifest fetch, blob pull) fails, regclient enters its backoff cycle, and the user must retry the entire pull. The same upstream URLs succeed over HTTP/1.1.

## Testing done on this change

```console
$ go test -tags '...,sync' -run 'TestTLS$|TestIsHTTP2FramingError|TestHTTP2Fallback|TestFallbackTLSConfig|TestNewHTTP2FallbackTransportTLS' ./pkg/extensions/sync/ -v
=== RUN   TestIsHTTP2FramingError
--- PASS: TestIsHTTP2FramingError (0.00s)
=== RUN   TestHTTP2FallbackTransportRoundTrip
--- PASS: TestHTTP2FallbackTransportRoundTrip (0.00s)
=== RUN   TestHTTP2FallbackRealTransport
--- PASS: TestHTTP2FallbackRealTransport (0.02s)
=== RUN   TestHTTP2FallbackStickyPerHost
--- PASS: TestHTTP2FallbackStickyPerHost (0.00s)
=== RUN   TestFallbackTLSConfig
--- PASS: TestFallbackTLSConfig (0.58s)
=== RUN   TestNewHTTP2FallbackTransportTLS
--- PASS: TestNewHTTP2FallbackTransportTLS (0.00s)
=== RUN   TestTLS
--- PASS: TestTLS (2.77s)
PASS
ok  	zotregistry.dev/zot/v2/pkg/extensions/sync	7.224s
```

- `TestHTTP2FallbackRealTransport` exercises the real transport boundary: an `httptest` TLS server aborts HTTP/2 streams; the test asserts the `ConfigureTransport`'d primary returns the typed x/net errors and that the wrapper retries over HTTP/1.1 end-to-end.
- `TestFallbackTLSConfig` / `TestNewHTTP2FallbackTransportTLS` cover the TLS material carried by the wrapped transports.
- `TestTLS` covers the `certDir`/mutual-TLS sync path through the new transport. It previously hung in an unbounded wait when the wrapper dropped the TLS settings (this is what made the "Run zot with extensions tests" CI job time out); the wait now has a 30s deadline and dumps the downstream log on failure.

Production-shaped A/B results (2 Kubernetes clusters, 3 zot replicas each, S3/Ceph + Redis, on-demand sync from Docker Hub through kubelet/containerd) are documented in [this comment](https://github.com/project-zot/zot/pull/4075#issuecomment-4481827436).

## Automation added to e2e

`test/blackbox/sync_http2_fallback.bats` is a happy-path smoke test (a downstream zot on-demand-syncs from a zot upstream through the new transport). It is wired into `test/blackbox/ci.sh` (run by `make run-blackbox-ci` in the ecosystem-tools job), with port ranges reserved in `test/ports.json`. The fallback branch itself (framing-error detection, HTTP/1.1 retry) is covered by the unit tests above, since a clean zot upstream does not emit the errors that trigger it.

## Will this break upgrades or downgrades?

No.

There is no new configuration field; the fallback is an internal behavior of sync's HTTP client. Deployments that never encounter HTTP/2 framing errors will see no behavioral change.

## Does this PR introduce any user-facing change?

Yes.

Previously failing sync operations caused by upstream HTTP/2 framing errors now transparently recover by retrying over HTTP/1.1. No configuration changes are required.

---


mrudnoru:
We've been running this patch in production-shaped testing on our zot deployment and wanted to share results in case they're useful to reviewers.

## Environment

- 2 Kubernetes clusters, 3 zot replicas each (Deployment)
- S3 (Ceph) blob storage + Redis cache
- On-demand sync from Docker Hub
- Kubelet + containerd as the pull client

## Failure mode against stock zot

Docker Hub's load balancer intermittently sends raw HTTP/2 SETTINGS frames on connections Go's `net/http` opened as HTTP/1.1:

```text
net/http: HTTP/1.x transport connection broken:
  malformed HTTP response "\x00\x00\x12\x04..."
```

`regclient` interprets this as a server error and enters its configured `retryDelay` backoff.

Combined with kubelet's independent 5-minute CRI deadline, a single framing error effectively guarantees a pull timeout.

Kubelet timeline on `nvidia/cuda:12.8.1-devel-ubuntu24.04` (5.2 GB) against stock zot:

```text
[0s]   Pulling image...
[302s] ErrImagePull — context canceled on manifest HEAD
[315s] ImagePullBackOff
[628s] ErrImagePull — retry also hit CRI timeout
[783s] Running — finally pulled from zot's S3 cache after backoff
```

13 minutes wall-clock for a pull that should have taken ~3-5 minutes.

For `pytorch:2.5.1-cuda12.4-cudnn9-devel` (7.4 GB, single 3.3 GB layer) the failure was permanent — every test node exceeded the 600s pull timeout and never converged.

## Test images

- `nvidia/cuda:12.8.1-devel-ubuntu24.04` — 5.2 GB (3 GB single layer)
- `pytorch:2.5.1-cuda12.4-cudnn9-devel` — 7.4 GB (3.3 GB single layer)
- `golang:1.23` — 324 MB (fan-out)

## Cold-pull procedure per run

1. Delete cached manifests + S3 blobs for the target image
2. Clear containerd cache on the target node
3. Rolling-restart zot pods (resets regclient's in-memory backoff state)
4. `kubectl run pull-test` on a chosen node
5. Record:
   - wall clock to `Running`
   - kubelet attempts
   - HTTP/2 framing errors
   - `context canceled`
   - regclient retries
   - EOF / RST counts

## Results — cold pulls

`nvidia/cuda:12.8.1-devel-ubuntu24.04` (5.2 GB, single 3 GB layer):

| Transport              | Runs | Success | Time range | Mean |
|-------------------------|-----:|--------:|------------:|-----:|
| forced HTTP/1.1        |    4 |    100% |   150–764s | 361s |
| HTTP/2 + auto-fallback |    9 |    100% |   105–129s | 115s |

`pytorch:2.5.1-cuda12.4-cudnn9-devel` (7.4 GB, single 3.3 GB layer), HTTP/2 + auto-fallback:

| Run | Node | Time | Attempts | Errors | Result  |
|----:|------|-----:|---------:|:------:|----------|
|   1 | n04  | 485s |        4 |   0    | SUCCESS |
|   2 | n05  | 443s |        1 |   0    | SUCCESS |
|   3 | n06  | 497s |        2 |   0    | SUCCESS |

The same image against stock zot failed permanently on all three nodes.

## Fan-out validation

Multiple nodes pulling the same uncached image concurrently (`HTTP/2 + auto-fallback`):

| Image                   | Nodes | Stagger | Time per node | Result       |
|-------------------------|------:|--------:|---------------:|---------------|
| `golang:1.23` 324 MB   |     3 |    10s  |     130–142s  | 100% SUCCESS |
| `cuda` 5.2 GB          |     2 |    10s  |          792s | 100% SUCCESS |

## Aggregate

**21 cold-pull runs total** across CUDA, PyTorch, and fan-out scenarios on images that previously failed consistently against stock zot.

Observed results:

- zero crashes
- zero HTTP/2 framing errors surfaced to clients
- 100% successful convergence across all validation runs

rchincha:
@sudo-bmitch

Does dist-spec have an opinion wrt http/1 and http/2?

sudo-bmitch:
> @sudo-bmitch
> 
> Does dist-spec have an opinion wrt http/1 and http/2?

None that I'm aware of.

rchincha:
sync_http2_fallback.bats is not adding a lot a value?

mrudnoru:
> sync_http2_fallback.bats is not adding a lot a value?

yes it was added added because in previous pr You have requested it. But indeed in this pr it does not make any sense. Ive removed it.

pkg/extensions/sync/service.go
```diff
@@ -893,27 +893,8 @@ func newClient(opts syncconf.RegistryConfig, credentials syncconf.CredentialsFil
 		regOpts = append(regOpts, reg.WithDelay(*opts.RetryDelay, *opts.RetryDelay))
 	}
 
-	// Configure transport with timeouts to prevent indefinite hangs.
```
rchincha, line 1002: This comment no longer makes sense?

mrudnoru: moved to http2_fallback.go and modified

rchincha:
@mrudnoru pls take a look at copilot reviews.

mrudnoru:
> @mrudnoru pls take a look at copilot reviews.

yes done 👍

rchincha:
https://github.com/project-zot/zot/pull/4075#discussion_r3262592568

https://github.com/project-zot/zot/pull/4075#discussion_r3262592607

rchincha:
@mrudnoru pls try to finish this PR. This one looks good.

rchincha:
pkg/extensions/sync/sync_test.go:2694

```
deadline := time.Now().Add(30 * time.Second)

for time.Now().Before(deadline) {
	var currentIndex ispec.Index

	destBuf, err := os.ReadFile(path.Join(destDir, testImage, "index.json"))
	if err == nil {
		if err := json.Unmarshal(destBuf, &currentIndex); err == nil && len(currentIndex.Manifests) > 0 {
			destIndex = currentIndex
			break
		}
	}

	time.Sleep(500 * time.Millisecond)
}

if len(destIndex.Manifests) == 0 {
	logData, _ := os.ReadFile(dctlr.Config.Log.Output)
	t.Fatalf("timed out waiting for TLS sync to populate %s/index.json; downstream log:\n%s",
		testImage, string(logData))
}
```

(OR) poll with timeout upstream registry

```
deadline := time.Now().Add(30 * time.Second)

for time.Now().Before(deadline) {
	resp, err := destClient.R().Get(destBaseURL + "/v2/" + testImage + "/manifests/" + testImageTag)
	if err == nil && resp.StatusCode() == http.StatusOK {
		break
	}

	time.Sleep(500 * time.Millisecond)
}
```

pkg/extensions/sync/http2_fallback.go
```diff
@@ -0,0 +1,141 @@
...
+func isHTTP2FramingError(err error) bool {
+	var streamErr http2.StreamError
+	if errors.As(err, &streamErr) {
+		return true
+	}
+
+	var goAwayErr *http2.GoAwayError
+	if errors.As(err, &goAwayErr) {
+		return true
+	}
```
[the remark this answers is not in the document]
mrudnoru: Fixed in 964b670b ("fix(sync): detect typed HTTP/2 framing errors via ConfigureTransport"). `newHTTP2FallbackTransport` now calls `golang.org/x/net/http2.ConfigureTransport()` on the cloned primary transport, which registers the x/net HTTP/2 implementation in `TLSNextProto`. HTTP/2 requests on the primary are handled by x/net, so framing failures surface as typed `http2.StreamError` / `*http2.GoAwayError` values that the `errors.As` checks match (the stdlib's bundled HTTP/2 only auto-configures when `TLSNextProto` is nil at first use; `ConfigureTransport` populates it, so that no longer happens here).

`TestHTTP2FallbackRealTransport` covers this at the real transport boundary: an `httptest` TLS server aborts HTTP/2 streams, and the test asserts the error returned by the actual `ConfigureTransport`'d transport satisfies the typed `errors.As` checks and that the wrapper then retries over HTTP/1.1. If `ConfigureTransport` ever fails, a warning is logged and classification degrades to the substring match.

pkg/extensions/sync/http2_fallback_internal_test.go
```diff
@@ -0,0 +1,264 @@
...
+		Convey("wrapped http2.StreamError returns true", func() {
+			err := fmt.Errorf("transport failure: %w", http2.StreamError{
+				StreamID: 1, Code: http2.ErrCodeProtocol,
+			})
+			So(isHTTP2FramingError(err), ShouldBeTrue)
+		})
+
+		Convey("http2.GoAwayError returns true", func() {
+			err := &http2.GoAwayError{ErrCode: http2.ErrCodeInternal, DebugData: "shutdown"}
+			So(isHTTP2FramingError(err), ShouldBeTrue)
```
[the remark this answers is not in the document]
mrudnoru: Addressed in 964b670b: `TestHTTP2FallbackRealTransport` now exercises the compatibility boundary this comment is about. It builds the primary transport mirroring the production construction (`http.DefaultTransport.Clone()` + `golang.org/x/net/http2.ConfigureTransport`), points it at a real TLS server that aborts HTTP/2 streams, and asserts (1) the error coming out of the real transport is the typed x/net error matched by `isHTTP2FramingError`, and (2) the full fallback wrapper retries the request over HTTP/1.1 end-to-end. The synthesized/stub cases remain for classifier edge cases (wrapped and unrelated errors) and for RoundTrip-level behavior that doesn't need a live server (body rewind, sticky-host state).

test/blackbox/sync_http2_fallback.bats
andaaron, line 1: This test doesn't trigger fallback and largely duplicates sync.bats.
Let's drop it, atm it's just waste of CI resoruces.

Header says this exercises the HTTP/2 primary path, but upstream is `http://`

pkg/extensions/sync/sync_test.go
```diff
@@ -2692,15 +2692,27 @@ func TestTLS(t *testing.T) {
 		defer dcm.StopServer()
 
 		// wait till ready
-		for {
-			destBuf, _ := os.ReadFile(path.Join(destDir, testImage, "index.json"))
-			_ = json.Unmarshal(destBuf, &destIndex)
+		deadline := time.Now().Add(30 * time.Second)
```
andaaron: 30s deadline may still be tight on slow CI when TLS + sync + index write run together. #4114 used 2 minutes for index wait and added manifest polling — consider matching if this flakes.

pkg/extensions/sync/sync_test.go
```diff
@@ -2692,15 +2692,27 @@ func TestTLS(t *testing.T) {
...
+
+					break
+				}
 			}
+
+			time.Sleep(500 * time.Millisecond)
+		}
+
+		if len(destIndex.Manifests) == 0 {
+			logData, _ := os.ReadFile(dctlr.Config.Log.Output)
```
andaaron, line 2727: On timeout, log dump is helpful. Consider also failing fast on cert trust errors in the log (as #4114 does) to distinguish TLS regressions from slow sync.

andaaron (commented):
@mrudnoru @rchincha 

Can we get a resolution on https://github.com/project-zot/zot/pull/4075 vs https://github.com/project-zot/zot/pull/4114? Which one are we proceeding with?

test/blackbox/sync_http2_fallback.bats
mrudnoru: Done in ae66c8e5

pkg/extensions/sync/sync_test.go
mrudnoru: Done. Deadline is now 2 * time.Minute

pkg/extensions/sync/sync_test.go
mrudnoru: Done. Wait loop scans for "certificate signed by unknown authority" and fatals with a tls specific message

mrudnoru:
@andaaron @rchincha regarding #4075 vs #4114 I would like to finish this one. The branch is rebased on latest main, all review feedback is addressed. Could you approve the workflow runs so CI can run. Also is it planned for v2.1.19? I could fold in anything you prefer from #4114

seniorquico:
We ran into the same `malformed HTTP response "\x00\x00\x12\x04..."` signature while working on sync throughput against an h2-capable upstream, and traced it to a client-side ALPN mismatch that reproduces against a perfectly well-behaved server. Flagging it here because I think it affects the fallback transport in this PR (and in #4114, which carries the same code).

`http.DefaultTransport.Clone()` itself runs `onceSetNextProtoDefaults` on the global, which populates the process-wide `TLSClientConfig` with `NextProtos: ["h2", "http/1.1"]`. No prior HTTPS request is needed, the first clone in the process is enough, and every clone inherits that list.

Disabling HTTP/2 on the clone doesn't undo it. `onceSetNextProtoDefaults` early-returns at `if !protocols.HTTP2() && !protocols.UnencryptedHTTP2() { return }`, which is before the `adjustNextProtos` call that would strip h2 from ALPN. So the transport unregisters the h2 RoundTripper but keeps advertising h2 in the handshake. The server selects h2, the client speaks HTTP/1.1, and the first SETTINGS frame gets read as a status line. This is true for both ways of turning h2 off, the documented `TLSNextProto = map[...]{}` idiom and Go 1.24's newer `Transport.Protocols`:

```go
tr := http.DefaultTransport.(*http.Transport).Clone()
tr.TLSClientConfig.InsecureSkipVerify = true   // test server
mutate(tr)                                     // one of the four rows below
resp, err := (&http.Client{Transport: tr}).Get(srv.URL)
```

Against an `httptest` server started with `EnableHTTP2 = true` and `NextProtos: []string{"h2", "http/1.1"}` (both advertised):

```text
default (h2 allowed)              ALPN=[h2 http/1.1]  => server saw HTTP/2.0
Protocols=HTTP1 only              ALPN=[h2 http/1.1]  => ERROR malformed HTTP response "\x00\x00\x1e\x04..."
TLSNextProto={} only              ALPN=[h2 http/1.1]  => ERROR malformed HTTP response "\x00\x00\x1e\x04..."
Protocols=HTTP1 + ALPN pin        ALPN=[http/1.1]     => server saw HTTP/1.1
```

Only pinning `TLSClientConfig.NextProtos = []string{"http/1.1"}` actually gets you an HTTP/1.1 connection.

In `newHTTP2FallbackTransport`, `TLSClientConfig` is only replaced when `tlsConf != nil`:

```go
fallbackTransport := clonedTransport(opts)
if tlsConf != nil {
    fallbackTransport.TLSClientConfig = tlsConf.Clone()
}

fallbackTransport.TLSNextProto = make(map[string]func(string, *cryptotls.Conn) http.RoundTripper)
fallbackTransport.ForceAttemptHTTP2 = false
```

For a registry with no `certDir` and `tlsVerify` left on, `tlsConf` is nil, so the fallback keeps the `DefaultTransport` clone's `NextProtos: ["h2","http/1.1"]` and advertises h2. If the upstream selects it, the fallback request fails with the same `malformed HTTP response` string the fallback is matching on, and `http2FallbackStickyTTL` then keeps that host on the broken path for 15 minutes.

I read this off the diff, I haven't run your branch, so I may be missing something that makes `tlsConf` non-nil in practice? The unit test looks like it would miss this, though:

```go
So(fallback.TLSClientConfig.NextProtos, ShouldNotContain, "h2")
```

This passes because the test supplies a non-nil `tlsConf` whose `NextProtos` is empty? The same assertion with `tlsConf == nil` should fail.

Maybe pin ALPN on the fallback transport after ensuring the config is non-nil?

```go
if fallbackTransport.TLSClientConfig == nil {
    fallbackTransport.TLSClientConfig = &cryptotls.Config{MinVersion: cryptotls.VersionTLS12}
}

fallbackTransport.TLSClientConfig.NextProtos = []string{"http/1.1"}
```

`Transport.Clone` deep-copies `TLSClientConfig`, so this shouldn't leak back into the process-global one. Adding a `tlsConf == nil` case to the existing test covers the regression.

For what it's worth, I hit this from the other end. Our first read of the error matched the one in the PR description, thinking the upstream must be putting SETTINGS frames on a connection we had opened as HTTP/1.1. But it turned out to be the client, the connection genuinely having been negotiated as h2, because we were still advertising h2 in ALPN and had only swapped out the RoundTripper.

PR #4092, opened 2026-05-23
feat(storage): redirect blob pulls to backend URLs

### LINKED ISSUE #2752: [Feat]: Ability to generate s3 presign link for blob

rchincha:
**What type of PR is this?**

Closes #2752

**Automation added to e2e**:

```release-note

```


examples/README.md
```diff
@@ -1032,6 +1032,7 @@ The following AWS policy is required by zot for push and pull. Make sure to repl
     "storage": {
         "rootDirectory": "/tmp/zot",  # local path used to store dedupe cache database
         "dedupe": true,
+        "redirect": true,
```
vrajashkr: IMO, I think just "redirect" may not capture the full usage of the feature. Perhaps "redirectBlobURLs" or something along those lines might help make it easier to understand.

pkg/api/config/config.go
```diff
@@ -28,6 +28,7 @@ type StorageConfig struct {
 	MaxRepos      int
 	Dedupe        bool
 	RemoteCache   bool
+	Redirect      bool
```
vrajashkr: similar comment about naming - may need a bit more in the name to help indicate that it redirects blob URLs. Maybe `RedirectBlobURL`?

pkg/api/routes.go
```diff
@@ -1418,6 +1418,41 @@ func writeMultipartRanges(
...
+	storePath := rh.c.StoreController.GetStorePath(name)
+
+	if storePath != storage.DefaultStorePath {
+		if subPathConfig, ok := storageConfig.SubPaths[storePath]; ok {
+			return subPathConfig.Redirect
+		}
+	}
+
+	return storageConfig.Redirect
+}
```
vrajashkr: Can we move this to the config package instead of being in routes.go?

pkg/api/routes.go
```diff
@@ -1418,6 +1418,41 @@ func writeMultipartRanges(
 	}
 }
 
+func (rh *RouteHandler) isBlobRedirectEnabled(name string) bool {
```
vrajashkr: Might be useful to rename it to `isBlobRedirectEnabledForRepo`

vrajashkr (changes requested):
Looks functionally sound. Left a few comments. Needs BATS tests for S3 redirect URL.

examples/README.md
rchincha: fixed

pkg/api/config/config.go

pkg/api/routes.go
rchincha: there is storage.go coupling which makes this tricky, and looks like there are other funcs have similar coupling in this file - listStorageRepositories() for example

pkg/api/routes.go
rchincha: "name" is always a repo right? Does the "ForRepo" add more clarity?

examples/README.md
```diff
@@ -1032,6 +1032,7 @@ The following AWS policy is required by zot for push and pull. Make sure to repl
     "storage": {
         "rootDirectory": "/tmp/zot",  # local path used to store dedupe cache database
         "dedupe": true,
+        "redirectBlobURL": true,
```
andaaron, line 1102: Shouldn't this be a setting under the storageDriver key itself? Given it only makes sense for S3 (and maybe GCS)? Or is it here because we do some magic with the keys under storageDriver to pass them along to the 3rd party library?

pkg/api/routes.go
andaaron: Or
```suggestion
func (rh *RouteHandler) isBlobRedirectEnabled(repo string) bool {
```
?

Or maybe this method should be attached to the store controller instead of having it in this file.

pkg/api/routes.go
```diff
@@ -1473,6 +1508,25 @@ func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Requ
 		}
 	}
 
+	if rh.isBlobRedirectEnabled(name) {
+		redirectURL, err := imgStore.GetBlobRedirectURL(request, name, digest)
+		if err != nil {
+			writeBlobError(err)
```
andaaron, line 1515: Has there been a discussion about pros and cons of returning an error vs falling back silently to calling the older storage APIs

dene14:
Hello @rchincha !
Any plans to finalize this work? it will be sad if we'll get another closed PR with that feature (4th if I remember correct).

examples/README.md
rchincha: no, since the map is passed as-is to upstream driver.

pkg/cli/server/root.go
```diff
@@ -606,6 +612,12 @@ func validateStorageConfigSection(
 	// enforce s3/gcs driver on subpaths in case of using storage driver
 	if len(storageConfig.SubPaths) > 0 {
 		for route, subStorageConfig := range storageConfig.SubPaths {
+			if subStorageConfig.RedirectBlobURL && len(subStorageConfig.StorageDriver) == 0 {
+				logger.Warn().Str("subpath", route).Bool("redirectBlobURL", subStorageConfig.RedirectBlobURL).
+					Str("storageDriver", storageConstants.LocalStorageDriverName).
+					Msg("redirectBlobURL is enabled for non-s3/gcs storage; blob pulls will be proxied")
+			}
```
vrajashkr, line 628: Should zot return an error for this? redirect with non-s3 or gcs isn't supported. I don't see a point to allowing it to be a valid config if zot ignores the setting for local storage.

Similar question about the global config on line 586 above.

vrajashkr:
There don't seem to be any tests added for the changes in the drivers.
Additionally `GetBlobRedirectURL` in imagestore.go doesn't appear to be covered by unit tests.
Please consider adding tests for these.

https://app.codecov.io/gh/project-zot/zot/pull/4092

pkg/cli/server/root.go
andaaron: Yes, I don't understand what it means by "will be proxied"? What is this proxy it mentions for local?
Why not error?

rchincha: fixed

pkg/api/routes.go
```diff
@@ -1443,9 +1443,13 @@ func normalizeBlobRedirectURL(rawURL string) (string, bool) {
 		return "", false
 	}
 
-	parsedURL.Scheme = scheme
+	i := strings.Index(rawURL, ":")
+	if i <= 0 {
+		return "", false
+	}
```
vrajashkr: If the URL doesn't have a colon, wouldn't it have already failed the URL parsing on line 1432 above?

pkg/api/routes.go
```diff
@@ -1443,9 +1443,13 @@ func normalizeBlobRedirectURL(rawURL string) (string, bool) {
...
 
-	return parsedURL.String(), true
+	// Preserve original URL bytes for signed URLs; only normalize scheme casing.
+	return scheme + rawURL[i:], true
```
vrajashkr: Just curious - is this an actual issue? I feel that the signed URL should be returned to the client as is. Fixing the casing of the scheme seems like an oddly specific problem.

IMO, that's something zot should probably only handle if it is a reported issue with a particular storage provider and not right away.

pkg/api/routes_internal_test.go
```diff
@@ -104,13 +104,23 @@ func TestParseRangeHeader(t *testing.T) {
 }
 
 func TestNormalizeBlobRedirectURL(t *testing.T) {
-	normalized, ok := normalizeBlobRedirectURL("HTTPS://storage.example.com/blob")
+	rawURL := "HTTPS://storage.example.com/blob?X-Amz-Signature=a%2Fb%2Bc"
```
vrajashkr: As in the other comment, IMO, this is an oddly specific problem to be solving for. I'm curious whether this has been observed with providers in the past.

Even then, it should be something the upstream fixes and not zot unless it happens frequently enough that it makes sense to have zot correct for it.

pkg/api/routes_test.go
```diff
@@ -976,6 +976,113 @@ func TestRoutes(t *testing.T) {
...
+
+					rthdlr.GetBlob(response, request)
+
+					resp := response.Result()
+					defer resp.Body.Close()
+
+					So(resp.StatusCode, ShouldEqual, http.StatusBadRequest)
+				})
+
+				Convey("with redirect disabled", func() {
```
vrajashkr: I feel like this is a test case that should already exist somewhere in the codebase.
Given the zero value of boolean is false, I'm wondering whether we really need this test case (assuming it is already covered elsewhere).

pkg/cli/server/root_test.go
```diff
@@ -1570,6 +1570,35 @@ storage:
...
+		tmpfile := MakeTempFileWithContent(t, "zot-test.json", content)
+
+		cfg := config.New()
+		err := cli.LoadConfiguration(cfg, tmpfile)
+		So(err, ShouldNotBeNil)
+		So(err.Error(), ShouldContainSubstring,
+			"invalid storage config, redirectBlobURL is supported only for s3/gcs storage")
+	})
+
+	Convey("Test redirectBlobURL error for empty storageDriver map", t, func(c C) {
```
vrajashkr, line 1588: Could we add a few tests for subpaths as well since we have support for this feature at the subpath level too?

pkg/api/routes_internal_test.go
```diff
@@ -102,3 +102,29 @@ func TestParseRangeHeader(t *testing.T) {
 		})
 	}
 }
+
+func TestNormalizeBlobRedirectURL(t *testing.T) {
```
vrajashkr, line 110: It would be great to have a few tests covering the other scenarios in normalize:
1. allowed schemes (http and https only)
2. failed parsing
3. CRLF in URL (I'm on the fence about this case IMO, but it feels like a good check to have so we can keep it).

vrajashkr (commented):
Left a few additional comments.

pkg/cli/server/root_test.go
rchincha: Added dedicated subpath-only validation coverage for both scenarios:

subpath redirectBlobURL=true with local storage (no storageDriver)
subpath redirectBlobURL=true with empty storageDriver map
This ensures we exercise the subpath validation branch directly rather than failing on global storage first.

pkg/api/routes_internal_test.go
rchincha: These scenarios are covered in the normalizeBlobRedirectURL table tests:
allows http
rejects disallowed scheme
rejects parse failure
rejects CRLF injection
along with signed-URL byte preservation and missing-host rejection.

vrajashkr (approved):
LGTM

vrajashkr:
Closes #2752

andaaron (approved):

PR #4094, opened 2026-05-23
feat(cli): support default config name

### LINKED ISSUE #129: cli: add a default config if not specified

AkashKumar7902:
## Summary
- add top-level `defaultConfigName` support to the zli `~/.zot` config file
- add `zli config set-default <name>` and `zli config clear-default`
- use the default profile only when neither `--url` nor `--config` is provided
- show the default in `zli config list` / `zli config show`, and clear it automatically when removing that profile

Supersedes #3996.
Fixes #129.

## Validation
- `git diff --check`
- `GOEXPERIMENT=jsonv2 go test -tags search ./pkg/cli/client -run 'TestConfig'`\n- `GOEXPERIMENT=jsonv2 go test -tags search ./pkg/cli/client -run 'Test(ReadZliConfigFile_errors|ZliConfigFile_DefaultConfigName|ZliConfigFile_DefaultNameMissingProfile|ConfigCmdMain|ConfigCmdBasics|GetConfigValue|ResetConfigValue|SetConfigValue|DefaultConfigValue|ConfigCmd_listFreshAndFindErrors|Utils)$'`

### DIFF
--- a/pkg/cli/client/config.go
+++ b/pkg/cli/client/config.go
@@ -21,14 +21,16 @@ const (
 	defaultConfigPerms = 0o644
 	defaultFilePerms   = 0o600
 
-	nameKey           = "_name"
-	showspinnerConfig = "showspinner"
-	verifyTLSConfig   = "verify-tls"
+	defaultConfigNameKey = "defaultConfigName"
+	nameKey              = "_name"
+	showspinnerConfig    = "showspinner"
+	verifyTLSConfig      = "verify-tls"
 )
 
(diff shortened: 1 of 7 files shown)


andaaron:
I don't have additional comments. Please fix the CI failures, and the copilot comments.

andaaron:
Superseeded by https://github.com/project-zot/zot/pull/4143.
@AkashKumar7902 thank you for the initial commit.

Issue #4096, opened 2026-05-25
[Feat]: Need a way to disable CVE only but keep UI & Search


davidkim83:
### Is your feature request related to a problem? Please describe.

I would like a way to be able to configure disabling CVE entirely (not download trivy DB at all) at 1st run.
The use case I have is the following:
- airgapped environment
- need a UI for operators
- need to use S3-compatible storage

Currently, enabling remote storage S3 requires CVE be disabled.
CVE scanning is usually handled elsewhere by someone else, so it's fine if Zot doesn't do it itself.
**But**, currently the only way to disable CVE is to disable the search extension itself.
(correct me if I'm wrong here)
And If I disable search extension, UI extension must also be disabled, and these two are something I don't want to disable.

### Describe the solution you'd like

A way to enable the UI and Remote Storage together.
Just disabling CVE while keeping UI/Search & remote storage enabled will be fine.

If the application doesn't support this right now, can you put this in the roadmap somewhere?

### Describe alternatives you've considered

One of the reasons I am considering switching to Zot is because I'm trying to reduce the number of containers I run to host a private registry.
(it's either this, or CNCF Distribution + a separate UI, or Nexus+Postgres, which I'm trying to avoid going back to)

rchincha:
@davidkim83 

airgap is something we are beginning to look at: https://github.com/project-zot/zot/issues/4051

rchincha:
@davidkim83 are you able to try and test this PR?
https://github.com/project-zot/zot/pull/4100

davidkim83:
@rchincha If you have an image tag I can pull, I'll test it now.

andaaron:
@davidkim83 having UI and search enabled with CVE scanning disabled should already be supported by existing zot.

Have you tried simply omitting the CVE from the zot config?
```
  "extensions": {
    "search": { "enable": true },
    "ui": { "enable": true }
  }
```

davidkim83:
@andaaron this was what I needed.
Now the UI, remote storage and search all work.
I would not have raised this issue if I knew it worked like this.

Can you perhaps improve the documentations, starting with these two?
- https://zotregistry.dev/v2.1.17/articles/cve-scanning/
- https://zotregistry.dev/v2.1.17/admin-guide/admin-configuration/#scheduling-background-tasks

Especially this part, since users may expect CVE to be configurable via `extensions.search.cve.enabled="false"`.

[...]

PR #4102, opened 2026-05-26
feat: allow disabling CVE independently from search

rchincha:
fixes issue #4096

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/config/config_test.go
+++ b/pkg/extensions/config/config_test.go
@@ -44,6 +44,22 @@ func buildSearchConfigWithCVE(enabled bool) *config.ExtensionConfig {
 	return ext
 }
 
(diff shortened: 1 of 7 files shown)


pkg/extensions/config/config_test.go
```diff
@@ -44,6 +44,22 @@ func buildSearchConfigWithCVE(enabled bool) *config.ExtensionConfig {
 	return ext
 }
 
+func buildSearchConfigWithCVEDisabled(enabled bool) *config.ExtensionConfig {
```
andaaron, line 47:
```suggestion
func buildSearchConfig(enableSearch, enableCVE bool) *config.ExtensionConfig {
```

PR #4110, opened 2026-06-03
feat(metrics): anonymous access when enabled in accessControl config

uaggarwa:
**What type of PR is this?**

This PR is an additional feature to enable anonymous access to the /metrics endpoint.

This PR adds support for anonymous access to the `/metrics` endpoint, which allows Prometheus scrapers to scrape metrics when metrics.AnonymousAccess is enabled in the accessControl config. 

**Changes:**

Manual testing was performed against a locally built binary (`-tags metrics`) using two configurations:

**1. htpasswd auth + `metrics.anonymousAccess: true` (`config-metrics-authn-anonymous-access.json`)**

```
# Anonymous scrape — should succeed
$ curl -s -o /dev/null -w "%{http_code}" http://127.0.0.1:8080/metrics
200 ✅

# Wrong credentials — should be rejected at authn
$ curl -s -o /dev/null -w "%{http_code}" -u wronguser:wrongpass http://127.0.0.1:8080/metrics
401 ✅

# Valid credentials — should succeed
$ curl -s -o /dev/null -w "%{http_code}" -u testuser:testpass http://127.0.0.1:8080/metrics
200 ✅

# Anonymous access to registry — still protected (no anonymousPolicy on repos)
$ curl -s -o /dev/null -w "%{http_code}" http://127.0.0.1:8080/v2/
401 ✅
```

**2. No auth + `metrics.anonymousAccess: true` (`config-metrics-anonymous-access.json`)**

```
# Config validation passes without anonymousPolicy in repos
$ ./bin/zot-local verify examples/config-metrics-anonymous-access.json
{"time":"2026-06-03T11:56:34.07278-04:00","level":"info","message":"config file is valid","config":"examples/config-metrics-anonymous-access.json"} ✅

# Anonymous scrape — should succeed
$ curl -s -o /dev/null -w "%{http_code}" http://127.0.0.1:8080/metrics
200 ✅
```

**Unit tests added**:

New test cases added to `pkg/extensions/monitoring/monitoring_test.go`:

- `basic auth: metrics.anonymousAccess=true` — verifies anonymous `GET /metrics` returns 200, wrong credentials returns 401, valid credentials return 200, and anonymous `GET /v2/` remains 401.
- `with basic auth: metrics.anonymousAccess=false` — verifies the default behavior is unchanged when `anonymousAccess` is not set.
- `TestMetricsAnonymousAccessNoAuth` — verifies anonymous `GET /metrics` returns 200 with no auth configured and `metrics.anonymousAccess: true`

**Automation added to e2e**:

There were no e2e tests are added.

No

No

```release-note

```

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -867,6 +867,45 @@ Behaviour-based action list
 }
 ```
 
+##### Metrics access control
+
+The `metrics` key inside `accessControl` controls access to the Prometheus scrape endpoint independently of repository policies. It supports two fields:
+
+- `users` - list of named authenticated users allowed to scrape. Requires authentication (e.g. htpasswd) to be configured.
+- `anonymousPolicy` - set to `["read"]` to allow unauthenticated access to the metrics endpoint when authentication is configured for other routes.
+
+To restrict scraping to specific named users:
+
+```
+"accessControl": {
+    "metrics": {
+        "users": ["prometheus"]
+    }
+}
+```
+
+See [config-metrics-authz.json](config-metrics-authz.json) for a complete example combining htpasswd authentication with repository policies.
+
+When authentication is configured and repositories have non-anonymous policies, `anonymousPolicy` on `metrics` allows unauthenticated scrapers to reach the metrics endpoint while keeping repository routes protected:
+
+```
+"http": {
+    "auth": {
+        "htpasswd": { "path": "test/data/htpasswd" }
+    },
+    "accessControl": {
+        "metrics": {
+            "anonymousPolicy": ["read"]
+        },
+        "repositories": {
+            "**": { "defaultPolicy": ["read", "create"] }
+        }
+    }
+}
+```
+
+See [config-metrics-authn-anonymous-access.json](config-metrics-authn-anonymous-access.json) for a complete example.
+
 ##### Conditional access on policies
 
 Policy entries can carry an optional list of `conditions`: CEL boolean
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -40,6 +40,7 @@ import (
 	"zotregistry.dev/zot/v2/pkg/api/constants"
 	apiErr "zotregistry.dev/zot/v2/pkg/api/errors"
 	zcommon "zotregistry.dev/zot/v2/pkg/common"
+	extconf "zotregistry.dev/zot/v2/pkg/extensions/config"
 	"zotregistry.dev/zot/v2/pkg/log"
 	reqCtx "zotregistry.dev/zot/v2/pkg/requestcontext"
 )
@@ -432,6 +433,10 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
 			accessControlConfig := ctlr.Config.CopyAccessControlConfig()
 			allowAnonymous := accessControlConfig != nil && accessControlConfig.AnonymousPolicyExists()
 
+			// Allow anonymous access to the metrics endpoint only if configured.
+			extensionsConfig := ctlr.Config.CopyExtensionsConfig()
+			isMetricsRequestedWithAnonymousAccess := isAnonymousMetricsRequest(request, accessControlConfig, extensionsConfig)
+
 			// build user access control info
 			userAc := reqCtx.NewUserAccessControl()
 			// if it will not be populated by authn handlers, this represents an anonymous user
@@ -461,7 +466,7 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
 				// If session authentication fails, but anonymous or management access is allowed,
 				// treat the request as authenticated. This fallback is necessary because the session
 				// header may be present for anonymous or management requests.
-				authenticated = authenticated || allowAnonymous || isMgmtRequested
+				authenticated = authenticated || allowAnonymous || isMgmtRequested || isMetricsRequestedWithAnonymousAccess
 
 			// Try mTLS authentication if client certificates are present
 			case ctlr.Config.IsMTLSAuthEnabled() && request.TLS != nil && len(request.TLS.PeerCertificates) > 0:
@@ -471,8 +476,8 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
 			case !authConfig.IsBasicAuthnEnabled() && !ctlr.Config.IsMTLSAuthEnabled():
 				authenticated = true
 
-			// If no credentials provided - check for anonymous / mgmt requests
-			case allowAnonymous || isMgmtRequested:
+			// If no credentials provided - check for anonymous / mgmt / metrics requests
+			case allowAnonymous || isMgmtRequested || isMetricsRequestedWithAnonymousAccess:
 				// Docker workaround: force 401 on /v2/ when anonymous policies coexist with
 				// authenticated-only policies. Otherwise Docker treats 200 on /v2/ as "no auth"
 				// and will not send stored credentials for protected repositories.
@@ -577,6 +582,16 @@ func bearerAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
 			header := request.Header.Get("Authorization")
 
 			if isAuthorizationHeaderEmpty(request) && isMgmtRequested {
+				next.ServeHTTP(response, request)
+
+				return
+			}
+
+			// Allow anonymous access to the metrics endpoint only if configured.
+			accessControlConfig := ctlr.Config.CopyAccessControlConfig()
+			extensionsConfig := ctlr.Config.CopyExtensionsConfig()
+
+			if isAnonymousMetricsRequest(request, accessControlConfig, extensionsConfig) {
 				next.ServeHTTP(response, request)
 
 				return
@@ -975,6 +990,18 @@ func authFail(w http.ResponseWriter, r *http.Request, realm string, delay int) {
 
 	w.Header().Set("Content-Type", "application/json")
 	zcommon.WriteJSON(w, http.StatusUnauthorized, apiErr.NewError(apiErr.UNAUTHORIZED))
+}
+
+func isAnonymousMetricsRequest(request *http.Request, accessControlConfig *config.AccessControlConfig,
+	extensionsConfig *extconf.ExtensionConfig,
+) bool {
+	prometheusConfig := extensionsConfig.GetMetricsPrometheusConfig()
+
+	return isAuthorizationHeaderEmpty(request) &&
+		slices.Contains(accessControlConfig.GetMetrics().AnonymousPolicy, constants.ReadPermission) &&
+		extensionsConfig.IsMetricsEnabled() &&
+		prometheusConfig != nil &&
+		strings.HasPrefix(request.URL.Path, prometheusConfig.Path)
 }
 
 func isAuthorizationHeaderEmpty(request *http.Request) bool {
--- a/pkg/api/authz.go
+++ b/pkg/api/authz.go
@@ -742,6 +742,12 @@ func MetricsAuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
 			}
 
 			metricsConfig := accessControlConfig.GetMetrics()
+			if slices.Contains(metricsConfig.AnonymousPolicy, constants.ReadPermission) {
+				next.ServeHTTP(response, request)
+
+				return
+			}
+
 			if len(metricsConfig.Users) == 0 {
 				log := ctlr.Log
 				log.Warn().Msg("auth is enabled but no metrics users in accessControl: /metrics is unaccesible")
--- a/pkg/api/config/config.go
+++ b/pkg/api/config/config.go
@@ -639,6 +639,22 @@ func (config *AccessControlConfig) GetMetrics() Metrics {
 	return config.Metrics
 }
 
+// ContainsOnlyMetricsAnonymousPolicy identifies an access control configuration
+// that contains only an anonymous policy for metrics.
+func (config *AccessControlConfig) ContainsOnlyMetricsAnonymousPolicy() bool {
+	if config == nil {
+		return false
+	}
+	admin := config.GetAdminPolicy()
+	if len(admin.Actions)+len(admin.Users)+len(admin.Groups) > 0 {
+		return false
+	}
+
+	return len(config.GetRepositories()) == 0 &&
+		len(config.GetGroups()) == 0 &&
+		slices.Contains(config.GetMetrics().AnonymousPolicy, "read")
+}
+
 // GetGroups safely gets a copy of the groups configuration.
 func (config *AccessControlConfig) GetGroups() Groups {
 	if config == nil {
@@ -728,7 +744,8 @@ type Condition struct {
 }
 
 type Metrics struct {
-	Users []string
+	Users           []string
+	AnonymousPolicy []string
 }
 
 type Config struct {
--- a/pkg/cli/server/root.go
+++ b/pkg/cli/server/root.go
@@ -783,8 +783,10 @@ func validateAuthzPolicies(config *config.Config, logger zlog.Logger) error {
 
 	logger.Info().Msg("checking if anonymous authorization is the only type of authorization policy configured")
 
+	// if no authentication is configured, policies must be anonymous-only;
 	if !authConfig.IsBasicAuthnEnabled() && !config.IsMTLSAuthEnabled() && !authConfig.IsBearerAuthEnabled() &&
-		!accessControlConfig.ContainsOnlyAnonymousPolicy() {
+		!accessControlConfig.ContainsOnlyAnonymousPolicy() &&
+		!accessControlConfig.ContainsOnlyMetricsAnonymousPolicy() {
 		msg := "access control config requires one of htpasswd, ldap, openid or mTLS authentication " +
 			"or using only 'anonymousPolicy' policies"
 		logger.Error().Err(zerr.ErrBadConfig).Msg(msg)
--- a/examples/config-metrics-authn-anonymous-access.json
+++ b/examples/config-metrics-authn-anonymous-access.json
@@ -1,0 +1,31 @@
+{
+    "distSpecVersion": "1.1.1",
+    "storage": {
+        "rootDirectory": "/tmp/zot"
+    },
+    "http": {
+        "address": "127.0.0.1",
+        "port": "8080",
+        "auth": {
+            "htpasswd": {
+                "path": "/tmp/zot-test-htpasswd"
+            }
+        },
+        "accessControl": {
+            "metrics": {
+                "anonymousPolicy": ["read"]
+            }
+        }
+    },
+    "log": {
+        "level": "debug"
+    },
+    "extensions": {
+        "metrics": {
+            "enable": true,
+            "prometheus": {
+                "path": "/metrics"
+            }
+        }
+    }
+}
--- a/pkg/extensions/monitoring/monitoring_test.go
+++ b/pkg/extensions/monitoring/monitoring_test.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"math/rand"
 	"net/http"
+	"net/url"
 	"os"
 	"path"
 	"testing"
@@ -26,10 +27,12 @@ import (
 	"zotregistry.dev/zot/v2/pkg/scheduler"
 	common "zotregistry.dev/zot/v2/pkg/storage/common"
 	"zotregistry.dev/zot/v2/pkg/storage/gc"
+	authutils "zotregistry.dev/zot/v2/pkg/test/auth"
 	test "zotregistry.dev/zot/v2/pkg/test/common"
 	. "zotregistry.dev/zot/v2/pkg/test/image-utils"
 	"zotregistry.dev/zot/v2/pkg/test/mocks"
 	ociutils "zotregistry.dev/zot/v2/pkg/test/oci-utils"
+	tlsutils "zotregistry.dev/zot/v2/pkg/test/tls"
 )
 
 func TestExtensionMetrics(t *testing.T) {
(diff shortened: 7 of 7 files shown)


examples/config-metrics-authz.json
```diff
@@ -13,7 +13,8 @@
         },
         "accessControl": {
             "metrics":{
-                "users": ["metrics"]
+                "users": ["metrics"],
+                "anonymousAccess": true
```
rchincha: "anonymousPolicy": [
                        "read"
                    ], 

why not align with this? although you really only need "read" but keeps things consistent.

pkg/api/authn.go
```diff
@@ -432,6 +432,13 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
 			accessControlConfig := ctlr.Config.CopyAccessControlConfig()
 			allowAnonymous := accessControlConfig != nil && accessControlConfig.AnonymousPolicyExists()
 
+			// Allow anonymous access to the metrics endpoint only if configured
+			prometheusConfig := ctlr.Config.CopyExtensionsConfig().GetMetricsPrometheusConfig()
+			isMetricsAnonymousAccess := isAuthorizationHeaderEmpty(request) &&
+				accessControlConfig.GetMetrics().AnonymousAccess &&
+				prometheusConfig != nil &&
+				strings.HasPrefix(request.RequestURI, prometheusConfig.Path)
```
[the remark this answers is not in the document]
uaggarwa: I have added the extensionsConfig.IsMetricsEnabled() and the request.URL.Path as mentioned in the suggestion.

pkg/api/authn.go
```diff
@@ -582,6 +589,19 @@ func bearerAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
...
 			}
 
+			// Allow anonymous access to the metrics endpoint only if configured
+			accessControlConfig := ctlr.Config.CopyAccessControlConfig()
+			prometheusConfig := ctlr.Config.CopyExtensionsConfig().GetMetricsPrometheusConfig()
+
+			if isAuthorizationHeaderEmpty(request) &&
+				accessControlConfig.GetMetrics().AnonymousAccess &&
+				prometheusConfig != nil &&
+				strings.HasPrefix(request.RequestURI, prometheusConfig.Path) {
```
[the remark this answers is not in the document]
uaggarwa: I have added the extensionsConfig.IsMetricsEnabled() and the request.URL.Path as mentioned in the suggestion.

pkg/cli/server/root.go
```diff
@@ -784,9 +784,10 @@ func validateAuthzPolicies(config *config.Config, logger zlog.Logger) error {
 	logger.Info().Msg("checking if anonymous authorization is the only type of authorization policy configured")
 
 	if !authConfig.IsBasicAuthnEnabled() && !config.IsMTLSAuthEnabled() && !authConfig.IsBearerAuthEnabled() &&
-		!accessControlConfig.ContainsOnlyAnonymousPolicy() {
+		!accessControlConfig.ContainsOnlyAnonymousPolicy() &&
+		!accessControlConfig.GetMetrics().AnonymousAccess {
 		msg := "access control config requires one of htpasswd, ldap, openid or mTLS authentication " +
-			"or using only 'anonymousPolicy' policies"
+			"or using only 'anonymousPolicy' policies, or metrics.anonymousAccess"
```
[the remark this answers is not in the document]
uaggarwa: I have updated the validation condition to now cover the edge case. 
```
1. Valid without Auth details.
This had to added to consider ```
"http": {
        "address": "127.0.0.1",
        "port": "8080",
        "accessControl": {
            "metrics": {
                "anonymousAccess": false
            }
        }
    },
```


2. Invalid without Auth details.
```
And     "http": {
        "address": "127.0.0.1",
        "port": "8080",
        "accessControl": {
            "repositories": {
                "**": {
                    "policies": [
                        {
                            "users": ["alice"],
                            "actions": ["read"]
                        }
                    ]
                }
            },
            "metrics": {
                "anonymousAccess": true
            }
        }
    },
```

pkg/api/authn.go
```diff
@@ -471,8 +478,8 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
 			case !authConfig.IsBasicAuthnEnabled() && !ctlr.Config.IsMTLSAuthEnabled():
 				authenticated = true
 
-			// If no credentials provided - check for anonymous / mgmt requests
-			case allowAnonymous || isMgmtRequested:
+			// If no credentials provided - check for anonymous / mgmt / metrics requests
+			case allowAnonymous || isMgmtRequested || isMetricsAnonymousAccess:
```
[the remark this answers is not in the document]
uaggarwa: Added. 
```
authenticated = authenticated || allowAnonymous || isMgmtRequested || isMetricsRequestedWithAnonymousAccess
```
isMetricsRequestedWithAnonymousAccess is set to true only if all of the gates above are true.

examples/config-metrics-authz.json
uaggarwa: Sounds better. Let me update the field to "anonymousPolicy".

uaggarwa: I have moved it to anonymousPolicy.

examples/config-metrics-anonymous-access.json
vrajashkr, line 1: There is a README file in the examples directory - https://github.com/project-zot/zot/blob/main/examples/README.md

This details the various configuration options. It would be great if you could update that file to describe the accessControl config for metrics.

pkg/api/authn.go
```diff
@@ -582,6 +591,21 @@ func bearerAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
...
+			// Allow anonymous access to the metrics endpoint only if configured.
+			accessControlConfig := ctlr.Config.CopyAccessControlConfig()
+			extensionsConfig := ctlr.Config.CopyExtensionsConfig()
+			prometheusConfig := extensionsConfig.GetMetricsPrometheusConfig()
+
+			if isAuthorizationHeaderEmpty(request) &&
+				len(accessControlConfig.GetMetrics().AnonymousPolicy) > 0 &&
+				extensionsConfig.IsMetricsEnabled() &&
+				prometheusConfig != nil &&
+				strings.HasPrefix(request.URL.Path, prometheusConfig.Path) {
```
vrajashkr: This condition for allowing anonymous metrics access seems to be duplicated from line 438 above.
I'm wondering if it might reduce duplication if the logic could be moved to a reusable function instead.

vrajashkr (commented):
Thank you for the PR! Left a couple of comments.

examples/config-metrics-anonymous-access.json
uaggarwa: Sure, let me add that.

pkg/api/authn.go
uaggarwa: Sure. I believe the lint findings should be be fixed after I make that change.

pkg/api/authn.go
```diff
@@ -582,6 +587,16 @@ func bearerAuthHandler(ctlr *Controller) mux.MiddlewareFunc {
...
 
+			// Allow anonymous access to the metrics endpoint only if configured.
+			accessControlConfig := ctlr.Config.CopyAccessControlConfig()
+			extensionsConfig := ctlr.Config.CopyExtensionsConfig()
+
+			if isAnonymousMetricsRequest(request, accessControlConfig, extensionsConfig) {
+				next.ServeHTTP(response, request)
+
+				return
+			}
```
vrajashkr, line 598: This code appears to need unit test coverage - https://app.codecov.io/gh/project-zot/zot/pull/4110

rchincha:
Error: pkg/api/authn.go:995:1: The line is 152 characters long, which exceeds the maximum of 120 characters. (lll)
  func isAnonymousMetricsRequest(request *http.Request, accessControlConfig *config.AccessControlConfig, extensionsConfig *extconf.ExtensionConfig) bool {
  ^

pkg/api/authn.go
uaggarwa: Thanks! Added additional unit tests and also fixed the (lll) issue.

pkg/api/authn.go
```diff
@@ -977,6 +992,18 @@ func authFail(w http.ResponseWriter, r *http.Request, realm string, delay int) {
...
+func isAnonymousMetricsRequest(request *http.Request, accessControlConfig *config.AccessControlConfig,
+	extensionsConfig *extconf.ExtensionConfig,
+) bool {
+	prometheusConfig := extensionsConfig.GetMetricsPrometheusConfig()
+
+	return isAuthorizationHeaderEmpty(request) &&
+		len(accessControlConfig.GetMetrics().AnonymousPolicy) > 0 &&
+		extensionsConfig.IsMetricsEnabled() &&
+		prometheusConfig != nil &&
+		strings.HasPrefix(request.URL.Path, prometheusConfig.Path)
```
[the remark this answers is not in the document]
uaggarwa: Updated to exclusively check for Read permission.

pkg/api/config/config.go
```diff
@@ -639,6 +639,12 @@ func (config *AccessControlConfig) GetMetrics() Metrics {
 	return config.Metrics
 }
 
+// ContainsOnlyMetricsAnonymousPolicy identifies a pure metrics-only deployment
+// that is valid without any authentication configured.
+func (config *AccessControlConfig) ContainsOnlyMetricsAnonymousPolicy() bool {
+	return len(config.GetRepositories()) == 0 && len(config.GetMetrics().AnonymousPolicy) > 0
+}
```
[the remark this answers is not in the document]
uaggarwa: added check for adminPolicy and groups.

pkg/api/authz.go
```diff
@@ -742,6 +742,12 @@ func MetricsAuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
 			}
 
 			metricsConfig := accessControlConfig.GetMetrics()
+			if len(metricsConfig.AnonymousPolicy) > 0 {
+				next.ServeHTTP(response, request)
+
+				return
+			}
```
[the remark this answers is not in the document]
uaggarwa: Updated to exclusively check for Read permission.

examples/config-metrics-authz.json
```diff
@@ -13,7 +13,8 @@
         },
         "accessControl": {
             "metrics":{
-                "users": ["metrics"]
+                "users": ["metrics"],
+                "anonymousPolicy": ["read"]
             },
```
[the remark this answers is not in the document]
uaggarwa: removed.

pkg/cli/server/root.go
```diff
@@ -783,8 +783,11 @@ func validateAuthzPolicies(config *config.Config, logger zlog.Logger) error {
...
 	logger.Info().Msg("checking if anonymous authorization is the only type of authorization policy configured")
 
+	// if no authentication is configured, repo policies must be anonymous-only;
+	// the only exception is a pure metrics-only config (no repo policies + metrics.anonymousPolicy)
 	if !authConfig.IsBasicAuthnEnabled() && !config.IsMTLSAuthEnabled() && !authConfig.IsBearerAuthEnabled() &&
-		!accessControlConfig.ContainsOnlyAnonymousPolicy() {
+		!accessControlConfig.ContainsOnlyAnonymousPolicy() &&
+		!accessControlConfig.ContainsOnlyMetricsAnonymousPolicy() {
 		msg := "access control config requires one of htpasswd, ldap, openid or mTLS authentication " +
 			"or using only 'anonymousPolicy' policies"
```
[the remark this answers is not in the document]
uaggarwa: `'or using only 'anonymousPolicy' policies'` is a blanket statement. And now that we have moved metrics to also follow 'anonymousPolicy' structure. I believe the statement covers that. Thoughts?

pkg/api/authn.go
```diff
@@ -977,6 +992,18 @@ func authFail(w http.ResponseWriter, r *http.Request, realm string, delay int) {
...
+	extensionsConfig *extconf.ExtensionConfig,
+) bool {
+	prometheusConfig := extensionsConfig.GetMetricsPrometheusConfig()
+
+	return isAuthorizationHeaderEmpty(request) &&
+		slices.Contains(accessControlConfig.GetMetrics().AnonymousPolicy, constants.ReadPermission) &&
+		extensionsConfig.IsMetricsEnabled() &&
+		prometheusConfig != nil &&
+		strings.HasPrefix(request.URL.Path, prometheusConfig.Path)
+}
```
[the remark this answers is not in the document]
uaggarwa: The default path for prometheusConfig.Path is set to /metrics in the code.
```
type PrometheusConfig struct {
	Path string // default is "/metrics"
}
```.
Your thoughts on this suggestion?

File: pkg/api/config/config.go [hunk 14]
```diff
@@ -639,6 +639,18 @@ func (config *AccessControlConfig) GetMetrics() Metrics {
...
+// that is valid without any authentication configured.
+func (config *AccessControlConfig) ContainsOnlyMetricsAnonymousPolicy() bool {
+	admin := config.GetAdminPolicy()
+	if len(admin.Actions)+len(admin.Users)+len(admin.Groups) > 0 {
+		return false
+	}
+	return len(config.GetRepositories()) == 0 &&
+		len(config.GetGroups()) == 0 &&
+		slices.Contains(config.GetMetrics().AnonymousPolicy, "read")
+}
```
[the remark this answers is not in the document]
uaggarwa: let me add this.

pkg/cli/server/root.go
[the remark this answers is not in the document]
vrajashkr: IMO, I think this suggestion can be skipped since the existing message will suffice.

pkg/api/authn.go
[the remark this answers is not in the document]
vrajashkr: IMO, I think this suggestion can be skipped too.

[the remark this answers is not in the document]
uaggarwa: Thanks!

pkg/api/config/config.go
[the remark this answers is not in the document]
uaggarwa: Added.

vrajashkr:
Linter detection:

```
 Error: pkg/api/config/config.go:652:2: return with no blank line before (nlreturn)
  	return len(config.GetRepositories()) == 0 &&
  	^
  1 issues:
  * nlreturn: 1
```

uaggarwa:
> Linter detection:
> 
> ```
>  Error: pkg/api/config/config.go:652:2: return with no blank line before (nlreturn)
>   	return len(config.GetRepositories()) == 0 &&
>   	^
>   1 issues:
>   * nlreturn: 1
> ```
Added a new line. Local Run now:
```
./hack/tools/bin/golangci-lint run --build-tags metrics ./pkg/api/...         
0 issues.
```

vrajashkr:
Thank you for your persistence with this PR!

I think codecov might be a bit happier if you could add a small test the cover the new code that checks if config is nil.

uaggarwa:
> Thank you for your persistence with this PR!
> 
> I think codecov might be a bit happier if you could add a small test the cover the new code that checks if config is nil.

Added. No issues at all. I appreciate the guidance along the way.

vrajashkr (dismissed):
LGTM

uaggarwa:
@vrajashkr The merging is blocked from my end, can you merge it for me? or is there anything I need to change?

vrajashkr:
> @vrajashkr The merging is blocked from my end, can you merge it for me? or is there anything I need to change?

One of the maintainers would take a look at the changes next and share their feedback if any.

uaggarwa:
> > @vrajashkr The merging is blocked from my end, can you merge it for me? or is there anything I need to change?
> 
> One of the maintainers would take a look at the changes next and share their feedback if any.

Got itt.

pkg/api/config/config.go
```diff
@@ -639,6 +639,22 @@ func (config *AccessControlConfig) GetMetrics() Metrics {
 	return config.Metrics
 }
 
+// ContainsOnlyMetricsAnonymousPolicy identifies a pure metrics-only deployment
+// that is valid without any authentication configured.
```
andaaron: What is the use case for this? If no authentication or authorization is configured on the server we shouldn't expect authorization for the metrics endpoint?

examples/config-metrics-anonymous-access.json
```diff
@@ -0,0 +1,26 @@
...
+    "distSpecVersion": "1.1.1",
+    "storage": {
+        "rootDirectory": "/tmp/zot"
+    },
+    "http": {
+        "address": "127.0.0.1",
+        "port": "8080",
+        "accessControl": {
+            "metrics": {
+                "anonymousPolicy": ["read"]
```
andaaron: What is the need of supporting this? If there is no authentication or authorization configured for zot in general we shouldn't need this configuration.

pkg/cli/server/root.go
```diff
@@ -783,8 +783,11 @@ func validateAuthzPolicies(config *config.Config, logger zlog.Logger) error {
 
 	logger.Info().Msg("checking if anonymous authorization is the only type of authorization policy configured")
 
+	// if no authentication is configured, repo policies must be anonymous-only;
+	// the only exception is a pure metrics-only config (no repo policies + metrics.anonymousPolicy)
```
andaaron: How is this a valid case? Can you elaborate?

amold1 (approved):

examples/config-metrics-anonymous-access.json
uaggarwa: the main usecase: I am trying to support with the PR is `/metrics` be scraped anonymously when authentication of authorization configured. examples/config-metrics-authn-anonymous-access.json

`examples/config-metrics-anonymous-access.json` is added only as a **blanket usecase**, to make sure zot does not reject such a configuration as well. 

As you mentioned correctly, that is the behaviour by default.

Without this change, if I remember correctly the "validation of the config" was failing.

pkg/api/config/config.go
uaggarwa: My intention was to mirror/extend the existing usecase https://github.com/project-zot/zot/blob/main/examples/config-anonymous-authz.json (which I used as a reference).

If no auth is defined: only anonymous policies can be defined in the access control
```
"http": {
        "address": "127.0.0.1",
        "port": "8080",
        "accessControl": {
            "metrics": {
                "anonymousPolicy": ["read"]
            }
        }
    },
```
Therefore this a valid config from that standpoint, but it is rejected by zot validator based on how the `ContainsOnlyAnonymousPolicy() func ` is defined in the code.

pkg/cli/server/root.go
uaggarwa: It comes down to the fact, whether the intended default behaviour, but in a config form is straight up rejected in the validator or not.
```
"http": {
        "address": "127.0.0.1",
        "port": "8080",
        "accessControl": {
            "metrics": {
                "anonymousPolicy": ["read"]
            }
        }
    },

```
Would love to know your thoughts on this? If we want to reject this config in zot, happy to make the required changes.

examples/config-metrics-anonymous-access.json
andaaron: This is also the implicit configuration I am fine with it.

I don't want to be forced into adding accessControl config with metrics on a setup which worked fine without this patch in a previous zot version.

pkg/api/config/config.go
andaaron: For context https://github.com/project-zot/zot/blob/main/examples/config-anonymous-authz.json is an example of how to configure zot to reject re-tagging different manifests with an existing tag regardless of who attempts it. The idea is to have immutable tags.


I guess we can keep this if this is also the implicit behavior. We don't want everyone to have to update their existing configs to allow reads on metrics.
But I would drop the example json and the Readme example for no auth, we don't need to advertise this.

pkg/cli/server/root.go
andaaron: I think this is fine if it is the same as implicit behavior.

examples/config-metrics-anonymous-access.json
andaaron: I think we can drop this example. https://github.com/project-zot/zot/blob/main/examples/config-anonymous-authz.json is given for a very specific use case (immutable tags) I explained in another comment.

examples/README.md
```diff
@@ -867,6 +867,57 @@ Behaviour-based action list
...
 ```
 
+##### Metrics access control
+
+The `metrics` key inside `accessControl` controls access to the Prometheus scrape endpoint independently of repository policies. It supports two fields:
+
+- `users` - list of named authenticated users allowed to scrape. Requires authentication (e.g. htpasswd) to be configured.
+- `anonymousPolicy` - set to `["read"]` to allow unauthenticated access to the metrics endpoint.
+
+To allow anonymous scraping with no authentication required:
```
andaaron: Let's drop this example. By default we allow this implicitly.

uaggarwa: Understood. We do not want to advertise this configuration in any way and silently support the edge case. Removed the config file example and also removed its reference in the documentation.

andaaron (approved):

rchincha:
metrics path prefix can become an authn bypass if misconfigured
Affected files:

/home/runner/work/zot/zot/project-zot/zot/pkg/api/authn.go
/home/runner/work/zot/zot/project-zot/zot/pkg/extensions/extension_metrics.go
The new helper uses:

Go
strings.HasPrefix(request.URL.Path, prometheusConfig.Path)
and metrics routing is registered with:

Go
router.PathPrefix(prometheusConfig.Path)
If prometheusConfig.Path is accidentally configured as /, /v2, or another broad/reserved prefix, then anonymous metrics authn bypass logic can apply to non-metrics routes that share that prefix.

Downstream authz may still block many registry operations, but authn should not be bypassed based on a broad prefix that can overlap protected routes.

Recommended mitigation:

Reject empty, /, and reserved/overlapping metrics paths during config validation.
Prefer exact path matching for metrics authn bypass, or normalize and compare against a dedicated metrics endpoint.
At minimum, require prometheusConfig.Path to be a non-root path and only match the exact path or path boundary.

rchincha:
if we have anonymous enabled and a user in a users list, a different authenticated user not in the list should be rejected. Even if unauthenticated users should be allowed.

uaggarwa:
> metrics path prefix can become an authn bypass if misconfigured Affected files:
> 
> /home/runner/work/zot/zot/project-zot/zot/pkg/api/authn.go /home/runner/work/zot/zot/project-zot/zot/pkg/extensions/extension_metrics.go The new helper uses:
> 
> Go strings.HasPrefix(request.URL.Path, prometheusConfig.Path) and metrics routing is registered with:
> 
> Go router.PathPrefix(prometheusConfig.Path) If prometheusConfig.Path is accidentally configured as /, /v2, or another broad/reserved prefix, then anonymous metrics authn bypass logic can apply to non-metrics routes that share that prefix.
> 
> Downstream authz may still block many registry operations, but authn should not be bypassed based on a broad prefix that can overlap protected routes.
> 
> Recommended mitigation:
> 
> Reject empty, /, and reserved/overlapping metrics paths during config validation. Prefer exact path matching for metrics authn bypass, or normalize and compare against a dedicated metrics endpoint. At minimum, require prometheusConfig.Path to be a non-root path and only match the exact path or path boundary.

I see that there is a default value for the prometheusConfig.Path in the code
```
type PrometheusConfig struct {
	Path string // default is "/metrics"
}

```
Do you suggest to add a `validateMetricsConfig func` that rejects `"", /, /v2 ` for prometheusConfig.Path first hand in configuration details?
or valueMatch path value to `/metrics` in `isAnonymousMetricsRequest func` exclusively ?

uaggarwa:
> if we have anonymous enabled and a user in a users list, a different authenticated user not in the list should be rejected. Even if unauthenticated users should be allowed.

I am not sure if I understood this one, can you elaborate on this one please.

andaaron:
> > if we have anonymous enabled and a user in a users list, a different authenticated user not in the list should be rejected. Even if unauthenticated users should be allowed.
> 
> I am not sure if I understood this one, can you elaborate on this one please.

This config
```
"metrics": {
  "users": ["prometheus"],
  "anonymousPolicy": ["read"]
}
```

Should:
- allow read by unauthenticated users
- allow read by authenticted user prometheus
- block reads for other authenticated users.

Current implementation seems to allow all authenticated users as well for `"anonymousPolicy": ["read"]`

rchincha:
@uaggarwa we can tackle improvements in a separate PR. Let's merge this first.

rchincha:
| Merging is blocked

^ it might help to squash your commits (very carefully) and then sign and push again.
`git commit --amend --no-edit -S -s`

uaggarwa:
> > > if we have anonymous enabled and a user in a users list, a different authenticated user not in the list should be rejected. Even if unauthenticated users should be allowed.
> > 
> > 
> > I am not sure if I understood this one, can you elaborate on this one please.
> 
> This config
> 
> ```
> "metrics": {
>   "users": ["prometheus"],
>   "anonymousPolicy": ["read"]
> }
> ```
> 
> Should:
> 
> * allow read by unauthenticated users
> * allow read by authenticted user prometheus
> * block reads for other authenticated users.
> 
> Current implementation seems to allow all authenticated users as well for `"anonymousPolicy": ["read"]`

Now that I think about it, "users" and "anonymousPolicy" should exist in either or relationship in the access control.
Otherwise `"anonymousPolicy": ["read"]` seems to intuitively supersede the "users" configuration. 
But if we want to support the following workflow:
* allow read by unauthenticated users
* allow read by authenticted user prometheus
* block reads for other authenticated users.

Let me see, if I can differentiate bw unauthenticated request vs authenticated request at a authorization layer.

andaaron:
We already do this for the repo-level authorization. anonymous users are only the unauthenticated users there.

andaaron:
> @uaggarwa we can tackle improvements in a separate PR. Let's merge this first.

merged this one

uaggarwa:
@andaaron Thank you!

PR #4114, opened 2026-06-05
feat(sync): [DC-216839]: use HTTP/2 with automatic HTTP/1.1 fallback on framing errors

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


test/blackbox/sync_http2_fallback.bats
```diff
@@ -0,0 +1,132 @@
+# Smoke test for the HTTP/2 → HTTP/1.1 fallback transport used by sync.
+# The fallback only kicks in when the upstream emits malformed HTTP/2 framing
```
andaaron, line 2: sync_http2_fallback.bats is largely redundant with existing CI tests for sync. It doesn’t test fallback, and the smoke path it does run is already exercised by sync.bats (and friends).

Can we drop this?

pkg/extensions/sync/http2_fallback.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	return true
+}
+
+func (t *http2FallbackTransport) markHostStuck(host string) {
+	t.stickyHost.Store(host, t.now().Add(t.stickyTTL))
+}
+
+func isHTTP2FramingError(err error) bool {
```
andaaron, line 121: This matches all StreamError / GoAwayError codes, not just protocol/framing mismatches (e.g. REFUSED_STREAM, ENHANCE_YOUR_CALM). For Docker Hub that’s probably fine, but it could mask genuine HTTP/2 issues on other registries by silently downgrading to HTTP/1.1.

I suggest this is refactored in 3 separate functions

```
func shouldRetryOnHTTP11Fallback(err error) bool {
    return isHTTP2SessionError(err) || isHTTP1HTTP2Mismatch(err)
}

func isHTTP2SessionError(err error) bool { /* StreamError, GoAwayError */ }

func isHTTP1HTTP2Mismatch(err error) bool {
    return strings.Contains(err.Error(), "malformed HTTP response")
}
```

andaaron:
@mrudnoru Please review this. I see you left some comments on https://github.com/project-zot/zot/pull/4075

PR #4119, opened 2026-06-06
feat: enhance config sanitization to mask sensitive keys in storage a…

### LINKED ISSUE #4117: Startup "configuration settings" log leaks S3 storageDriver accesskey/secretkey in plaintext
## Summary

On startup, zot logs its **full rendered configuration** at `log.level: info` in the `configuration settings` line, and for an S3 `storageDriver` this includes the `accesskey` and `secretkey` **in plaintext**. Any log-aggregation pipeline (Loki, Elastic, CloudWatch, etc.) then stores the S3 credentials at rest.

## Version

`zot v2.1.17` (`ghcr.io/project-zot/zot-linux-amd64:v2.1.17`). The dump comes from `Controller.Init` (around `pkg/api/controller.go:331`).

## Reproduce

Run zot with an S3 storage driver, e.g.:

```json
{
  "storage": {
    "rootDirectory": "/var/lib/zot",
    "storageDriver": {
      "name": "s3",
      "region": "us-east-1",
      "bucket": "my-bucket",
      "regionendpoint": "https://s3.example.com",
      "accesskey": "REDACTED_ACCESS_KEY",
      "secretkey": "REDACTED_SECRET_KEY",
      "forcepathstyle": true
    }
  },
  "http": { "address": "0.0.0.0", "port": "5000" },
  "log": { "level": "info", "output": "/dev/stdout" }
}
```

The first `info` line on boot is:

```json
{"level":"info","message":"configuration settings","params":{ ... "StorageDriver":{"accesskey":"REDACTED_ACCESS_KEY", ... ,"secretkey":"REDACTED_SECRET_KEY", ...} ... },"caller":"...pkg/api/controller.go:331","func":"...(*Controller).Init"}
```

## Impact

- The S3 credentials are written to stdout and therefore into any centralized log store at rest.
- Operators can't simply avoid this by lowering the log level: `info` carries other operationally-important lines (e.g. GC sweep activity) that we want to keep, so dropping to `warn` to suppress the credential dump is not an acceptable trade-off.
- This isn't limited to S3 keys — the same config dump would expose any other secret material present in the config (e.g. an htpasswd path's contents aren't dumped, but bearer/OpenID client secrets, session keys, sync credentials, etc. are candidates worth auditing).

## Suggested fix

Redact known-sensitive fields before emitting the `configuration settings` line, replacing their values with a placeholder such as `***` / `[REDACTED]`. At minimum:

- `storage.storageDriver.secretkey`, `storage.storageDriver.accesskey`
- subpath storage drivers (`storage.subPaths.*.storageDriver.*`)
- any cache driver / sync / auth credentials carried in the config

A small allow/deny list of field names (or a `redact:"true"` struct tag honored by a custom marshaler used only for the startup dump) would keep `info`-level config visibility while keeping secrets out of logs.

## Workaround (for reference)

We currently redact these fields at the log-collector layer (a Promtail `replace` stage masking `"(access|secret)key":"..."`), but a native redaction in zot would be the durable fix and would protect every zot operator by default.

rchincha:
…nd session drivers

Fixes issue #4117

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/config/config.go
+++ b/pkg/api/config/config.go
@@ -859,7 +859,7 @@ func sanitizeMapValues(values map[string]any) {
 }
 
 func sanitizeSliceValues(values []any) {
-	for idx, value := range values {
+	for _, value := range values {
 		switch typedVal := value.(type) {
 		case map[string]any:
 			sanitizeMapValues(typedVal)
@@ -867,8 +867,6 @@ func sanitizeSliceValues(values []any) {
 			sanitizeStringMapValues(typedVal)
 		case []any:
 			sanitizeSliceValues(typedVal)
-		default:
-			values[idx] = value
 		}
 	}
 }
--- a/pkg/api/config/config.go
+++ b/pkg/api/config/config.go
@@ -5,6 +5,7 @@ import (
 	"maps"
 	"os"
 	"slices"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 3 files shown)


pkg/api/config/config.go
```diff
@@ -819,6 +820,59 @@ func (c *Config) isTagsRetentionEnabled(tagRetentionPolicy KeepTagsPolicy) bool
...
+		case map[string]any:
+			sanitizeMapValues(typedVal)
+		case map[string]string:
+			sanitizeStringMapValues(typedVal)
+		case []any:
+			sanitizeSliceValues(typedVal)
+		default:
+			values[idx] = value
+		}
+	}
```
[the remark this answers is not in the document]
rchincha: fixed

pkg/api/config/config.go
```diff
@@ -819,6 +820,57 @@ func (c *Config) isTagsRetentionEnabled(tagRetentionPolicy KeepTagsPolicy) bool
 	return false
 }
 
+func isSensitiveConfigMapKey(key string) bool {
+	normalized := strings.ToLower(strings.ReplaceAll(key, "_", ""))
+
+	switch normalized {
+	case "accesskey", "secretkey", "password", "secret", "token", "clientsecret", "sessionhashkey", "sessionencryptkey":
```
andaaron: Can you add `sentinelpassword` to the case? This is for Redis config.

pkg/api/config/config.go
```diff
@@ -881,6 +957,7 @@ func (c *Config) Sanitize() *Config {
 			}
 
 			sanitizedConfig.Extensions.Events.Sinks[i].Credentials.Password = "******"
+			sanitizedConfig.Extensions.Events.Sinks[i].Credentials.Token = "******"
```
andaaron: Password and token are theoretically mutually exclusive, we should not log a sanitized config showing both Password = "****** and Token = "******", that would be misleading.

```suggestion			
			if sanitizedConfig.Extensions.Events.Sinks[i].Credentials.Password != "" {
                sanitizedConfig.Extensions.Events.Sinks[i].Credentials.Password = "******"
            }
            if sanitizedConfig.Extensions.Events.Sinks[i].Credentials.Token != "" {
                sanitizedConfig.Extensions.Events.Sinks[i].Credentials.Token = "******"
            }
```

pkg/api/config/config.go

pkg/api/config/config.go

andaaron (approved):

vrajashkr (commented):
LGTM

PR #4132, opened 2026-06-14
refactor(test/blackbox): extract shared push/pull helpers

### LINKED ISSUE #3727: [Refactor]: BATS tests under tests/blackbox have lot of duplicate code and test cases
Analyze test code under tests/blackbox and try to refactor into common re-usable code.
There is way too much duplicate code there.

andaaron:
See commit messages. Closes #3727.


test/blackbox/helpers_pushpull.bash
```diff
@@ -0,0 +1,570 @@
...
+        return 1
+    fi
+
+    if ! command -v jq >/dev/null; then
+        echo "you need to install jq as a prerequisite to running the tests" >&3
+        return 1
+    fi
+
+    return 0
+}
```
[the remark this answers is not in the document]
andaaron: Overkill, not needed

test/blackbox/helpers_pushpull_authn.bash
```diff
@@ -0,0 +1,325 @@
...
+
+    if [ "${PUSHPULL_AUTHN_FIPS_MODE:-0}" = 1 ]; then
+        if ! command -v mkpasswd >/dev/null; then
+            echo "you need to install mkpasswd as a prerequisite to running the tests" >&3
+            return 1
+        fi
+    fi
+
+    return 0
+}
```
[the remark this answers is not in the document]
andaaron: We're not going down this rabbit hole again :)

test/blackbox/helpers_pushpull.bash
```diff
@@ -0,0 +1,578 @@
...
+
+    if ! command -v git >/dev/null; then
+        echo "you need to install git as a prerequisite to running the tests" >&3
+        return 1
+    fi
+
+    if ! command -v docker >/dev/null; then
+        echo "you need to install docker as a prerequisite to running the tests" >&3
+        return 1
+    fi
```
vrajashkr: nit: wondering if this could become a loop.

test/blackbox/helpers_pushpull.bash
```diff
@@ -0,0 +1,578 @@
...
+        log_output | jq 'contains("fips140 is currently enabled")?' | grep true
+    fi
+}
+
+function pushpull_teardown() {
+    if [ "${PUSHPULL_FIPS_MODE:-0}" = 1 ]; then
+        cat "${BATS_FILE_TMPDIR}/zot/zot-log.json"
+    else
+        cat "${BATS_FILE_TMPDIR}/zot.log"
+    fi
```
vrajashkr: Same comment about the log file name difference carries down here too.

test/blackbox/helpers_pushpull.bash
```diff
@@ -0,0 +1,578 @@
...
+
+    local zot_root_dir=${BATS_FILE_TMPDIR}/zot
+    local zot_config_file=${BATS_FILE_TMPDIR}/zot_config.json
+    local oci_data_dir=${BATS_FILE_TMPDIR}/oci
+    local log_file
+
+    mkdir -p "${zot_root_dir}" "${oci_data_dir}"
+
+    if [ "${PUSHPULL_FIPS_MODE:-0}" = 1 ]; then
+        log_file=${zot_root_dir}/zot-log.json
```
vrajashkr: I didn't quite understand the reasoning for the different log file names between fips and non fips mode. Maybe we can have them write to the same file?

test/blackbox/helpers_pushpull.bash
```diff
@@ -0,0 +1,578 @@
...
+        cat "${BATS_FILE_TMPDIR}/zot.log"
+    fi
+}
+
+function pushpull_teardown_file() {
+    zot_stop_all
+
+    if [ "${PUSHPULL_FIPS_MODE:-0}" = 1 ]; then
+        unset GODEBUG
+    fi
```
vrajashkr, line 90: I assume the tests run in a subshell. I'm wondering if we really need to unset the environment variable here.

test/blackbox/helpers_pushpull.bash
```diff
@@ -0,0 +1,578 @@
...
+
+    run crictl pull "localhost:${zot_port}/test-regclient"
+    [ "${status}" -eq 0 ]
+}
+
+function helper_push_oci_artifact_with_regclient() {
+    local zot_port
+    zot_port=$(get_zot_port)
+
+    run regctl artifact put "localhost:${zot_port}/artifact:demo" <<TXT
```
vrajashkr: Just curious - the pull helper below pulls the manifest followed by the artifact. Does the push not need to push the manifest as well?

test/blackbox/helpers_pushpull.bash
```diff
@@ -0,0 +1,578 @@
...
+    [ "$(echo "${lines[-1]}" | jq '.manifests | length')" -eq "${expected_count}" ]
+    run regctl artifact list --filter-artifact-type "application/vnd.example.icecream.v1" "localhost:${zot_port}/artifact-ref:demo" --format raw-body
+    [ "${status}" -eq 0 ]
+    [ "$(echo "${lines[-1]}" | jq '.manifests | length')" -eq "${expected_count}" ]
+    run regctl artifact list --filter-artifact-type "application/invalid" "localhost:${zot_port}/artifact-ref:demo" --format raw-body
+    [ "${status}" -eq 0 ]
+    [ "$(echo "${lines[-1]}" | jq '.manifests | length')" -eq 0 ]
+}
+
+function helper_push_docker_image() {
```
vrajashkr: Should this be push and pull? The test appears to do both.

test/blackbox/helpers_pushpull.bash
```diff
@@ -0,0 +1,578 @@
...
+    dockerfile=${BATS_TEST_TMPDIR}/Dockerfile
+
+    cat > "${dockerfile}" <<DOCKERFILE
+FROM ghcr.io/project-zot/test-images/busybox-docker:1.37
+RUN echo "hello world" > /testfile
+DOCKERFILE
+    run sh -c "unset GODEBUG; docker build -f '${dockerfile}' -t localhost:${zot_port}/test '${BATS_TEST_TMPDIR}'"
+    [ "${status}" -eq 0 ]
+    run docker push "localhost:${zot_port}/test"
+    [ "${status}" -eq 1 ]
```
vrajashkr, line 590: Is this push expected to fail? Didn't quite catch why the status code is expected to be 1.

test/blackbox/helpers_pushpull.bash
```diff
@@ -0,0 +1,578 @@
...
+    [ "${status}" -eq 0 ]
+}
+
+function helper_push_image_with_regclient() {
+    local zot_port
+    zot_port=$(get_zot_port)
+
+    run regctl registry set "localhost:${zot_port}" --tls disabled
+    [ "${status}" -eq 0 ]
+    run regctl image copy "ocidir://${TEST_DATA_DIR}/golang:1.20" "localhost:${zot_port}/test-regclient"
```
vrajashkr: Maybe the image names could be arguments to the function (with defaults)?

test/blackbox/helpers_pushpull.bash
```diff
@@ -0,0 +1,578 @@
...
+    [ "${status}" -eq 0 ]
+}
+
+function helper_pull_image_with_regclient() {
+    local zot_port
+    zot_port=$(get_zot_port)
+
+    run regctl image copy "localhost:${zot_port}/test-regclient" "ocidir://${TEST_DATA_DIR}/golang:1.20"
```
vrajashkr: Here too - perhaps it would be helpful to have the image names as args to the function with defaults.

test/blackbox/helpers_pushpull.bash
```diff
@@ -0,0 +1,578 @@
...
+    [ "${status}" -eq 0 ]
+    helper_assert_output_contains_line latest
+}
+
+function helper_push_manifest_with_regclient() {
+    local zot_port manifest
+    zot_port=$(get_zot_port)
+    manifest=$(regctl manifest get "localhost:${zot_port}/test-regclient" --format=raw-body)
+
+    run regctl manifest put "localhost:${zot_port}/test-regclient:1.0.0" \
```
vrajashkr: Might be useful to have these image names as args.

vrajashkr: Similar comment for the other functions that have hardcoded names.

test/blackbox/helpers_wait.bash
```diff
@@ -26,3 +26,22 @@ function wait_file() {
...
+    shift 2
+
+    while [ "${attempts}" -gt 0 ]; do
+        run "$@"
+        if [ "${status}" -eq 0 ]; then
+            return 0
+        fi
+        attempts=$((attempts - 1))
+        if [ "${attempts}" -eq 0 ]; then
+            return 1
```
vrajashkr: Should this return the last status of the command being run instead of 1?

test/blackbox/helpers_pushpull_authn.bash
```diff
@@ -0,0 +1,325 @@
...
+    wait_zot_reachable "${zot_port}"
+
+    if [ "${PUSHPULL_AUTHN_FIPS_MODE:-0}" = 1 ]; then
+        log_output | jq 'contains("fips140 is currently enabled")?' | grep true
+    fi
+}
+
+function authn_teardown() {
+    if [ "${PUSHPULL_AUTHN_FIPS_MODE:-0}" = 1 ]; then
+        cat "${BATS_FILE_TMPDIR}/zot/zot-log.json"
```
vrajashkr: Similar question about the variation in the log file name between fips and non fips.

test/blackbox/helpers_pushpull_authn.bash
```diff
@@ -0,0 +1,325 @@
...
+    local pass=${2}
+
+    helper_authn_regctl_tls_disabled
+    run regctl registry login "localhost:$(get_zot_port)" -u "${user}" -p "${pass}"
+    [ "${status}" -eq 0 ]
+}
+
+function helper_authn_push_image_with_regclient() {
+    helper_authn_regctl_login "${AUTH_USER}" "${AUTH_PASS}"
+    run regctl image copy "ocidir://${TEST_DATA_DIR}/busybox:1.36" \
```
vrajashkr: Might be useful to have args with defaults perhaps.

test/blackbox/helpers_pushpull_authn.bash
```diff
@@ -0,0 +1,325 @@
...
+}
+
+# Args: $1=username, $2=password, $3=source_repo, $4=dest_ref
+function helper_authn_pull_image_with_auth() {
+    local user=${1}
+    local pass=${2}
+    local source_repo=${3}
+    local dest_ref=${4}
+
+    helper_authn_regctl_login "${user}" "${pass}"
```
vrajashkr, line 269: Since this is logging in here as part of the function, should it also logout after the image copy?

test/blackbox/helpers_upgrade.bash
```diff
@@ -632,10 +167,10 @@ function test_new_attach_oras_artifacts() {
     helper_attach_oras_artifacts golang 1.20
 }
 
-# Note: expected_count parameter allows different counts for full vs minimal zot
+# Args: $1 = expected artifact count. Full and minimal zot can have different counts.
 function test_new_discover_oras_artifacts() {
     local expected_count=${1:-4}
-    helper_discover_oras_artifacts golang 1.20 ${expected_count}
+    helper_discover_oras_artifacts golang 1.20 "${expected_count}"
 }
```
vrajashkr: IMO, these short functions might be adding an extra layer of indirection. Is it possible to directly use `helper_discover_oras_artifacts` inside the upgrade test? What do you think?

vrajashkr (commented):
Looks good overall. Thanks for taking a look at this. Left some questions and comments.

vrajashkr:
We could consider adding "Closes" for issue https://github.com/project-zot/zot/issues/3727

test/blackbox/helpers_pushpull.bash
andaaron: Probably not. I'd rather be on the safe side.

test/blackbox/helpers_pushpull.bash
andaaron: Will fix this in next push.

test/blackbox/helpers_pushpull.bash
andaaron: Will fix this in next push.

test/blackbox/helpers_pushpull.bash
andaaron: Will fix this in next push.

test/blackbox/helpers_pushpull_authn.bash
andaaron: Will fix in next push.

test/blackbox/helpers_wait.bash
andaaron: Will make the change in next push.

test/blackbox/helpers_pushpull.bash
andaaron: It is not, see https://github.com/regclient/regclient/blob/main/cmd/regctl/artifact.go#L698

test/blackbox/helpers_upgrade.bash
andaaron: I agree. Will make the change.

test/blackbox/helpers_pushpull.bash
andaaron: Because the zot compat for docker is supposed to be off.
Will rename the function.

test/blackbox/helpers_pushpull.bash
andaaron: Will rename in next push.

test/blackbox/helpers_pushpull_authn.bash
andaaron: I'll refactor this to have the login per suite (the data is held on a file in the user home by default).
That way it will "logout" when the suite ends.

test/blackbox/helpers_pushpull_authn.bash
andaaron: went with required args at call sites instead of defaults in helpers, I think it's easier for the person writing the suite, since the tests depend on one another. For example, in case of separate push and pull helpers, using defaults requires you to check the values used by both helpers in a different file and make sure they match.

test/blackbox/helpers_pushpull.bash
andaaron: went with required args at call sites instead of defaults in helpers, I think it's easier for the person writing the suite, since the tests depend on one another. For example, in case of separate push and pull helpers, using defaults requires you to check the values used by both helpers in a different file and make sure they match.

test/blackbox/helpers_pushpull.bash
andaaron: went with required args at call sites instead of defaults in helpers, I think it's easier for the person writing the suite, since the tests depend on one another. For example, in case of separate push and pull helpers, using defaults requires you to check the values used by both helpers in a different file and make sure they match.

test/blackbox/helpers_pushpull.bash
andaaron: went with required args at call sites instead of defaults in helpers, I think it's easier for the person writing the suite, since the tests depend on one another. For example, in case of separate push and pull helpers, using defaults requires you to check the values used by both helpers in a different file and make sure they match.

rchincha (approved):

vrajashkr (commented):
LGTM

PR #4134, opened 2026-06-15
feat(cli): add zli config set-default / clear-default commands

BRGOVIND:
﻿## Problem

Users who work primarily against a single zot registry must pass `--config <name>` on every `zli` command. There is no way to designate one profile as the default so that commands "just work" without a flag.

Issue: #129

## Solution

Add a `defaultConfigName` field to the `~/.zot` JSON file and two new subcommands:

```
zli config set-default <name>   # mark a profile as default
zli config clear-default        # remove the default designation
```

When `--url` and `--config` are both absent, `GetServerURLFromFlags` and `GetCliConfigOptions` now fall back to the default profile automatically.

`zli config list` appends `(default)` to the designated profile so users can see which one is active at a glance.

## Changes

| File | What changed |
|------|-------------|
| `pkg/cli/client/config.go` | `ZliConfigFile` gets `DefaultConfig string` (`defaultConfigName` JSON key, omitempty). Added `SetDefault(name)` / `ClearDefault()` methods. `FormatNames()` marks the default entry. |
| `pkg/cli/client/config_cmd.go` | `NewConfigSetDefaultCommand()`, `NewConfigClearDefaultCommand()`, `setDefaultConfig()`, `clearDefaultConfig()` helpers. Commands registered in `NewConfigCommand()`. Examples updated. |
| `pkg/cli/client/utils.go` | `GetServerURLFromFlags` and `GetCliConfigOptions` call `readDefaultConfigName()` when `--config` is not given. New `readDefaultConfigName()` reads the `defaultConfigName` field from `~/.zot`. |

## Example

```
$ zli config add main https://zot.example.com:8080
$ zli config set-default main
$ zli config list
main    https://zot.example.com:8080    (default)

$ zli image list          # no --config needed
...
```

Signed-off-by: B R GOVIND <brgovind2005@gmail.com>

PR #4143, opened 2026-06-17
feat(cli): support default config name

andaaron:
See previous work on https://github.com/project-zot/zot/issues/129
- https://github.com/project-zot/zot/pull/4094
- https://github.com/project-zot/zot/pull/4134

### DIFF
--- a/pkg/cli/client/config.go
+++ b/pkg/cli/client/config.go
@@ -21,14 +21,16 @@ const (
 	defaultConfigPerms = 0o644
 	defaultFilePerms   = 0o600
 
-	nameKey           = "_name"
-	showspinnerConfig = "showspinner"
-	verifyTLSConfig   = "verify-tls"
+	defaultConfigNameKey = "defaultConfigName"
+	nameKey              = "_name"
+	showspinnerConfig    = "showspinner"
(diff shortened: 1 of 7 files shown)


pkg/cli/client/config.go
```diff
@@ -21,14 +21,16 @@ const (
 	defaultConfigPerms = 0o644
 	defaultFilePerms   = 0o600
 
-	nameKey           = "_name"
-	showspinnerConfig = "showspinner"
-	verifyTLSConfig   = "verify-tls"
+	defaultConfigNameKey = "defaultConfigName"
```
rchincha, line 24: why not just call this "default"

andaaron: I think it's easier to expand the functionality in the future. 
Like adding other keys containing common settings for all configs. For example. IMHO, `ShowSpinner` should have been a common setting/key in this struct as opposed as a per-config setting.
Maybe one of those future settings would be a list or some other type of objects for the user to choose from, and simply "default" could be ambiguous. 

The configs are actually under a "configs" key. We could have named it "defaultConfig", but that would have created an expectation for it to be a config type struct.

rchincha (approved):

PR #4137, opened 2026-06-17
refactor(storage)!: add a common blobstore to store all blobs (#4136)

rchincha:
* feat(storage): add a common blobstore to store all blobs (#3906)

Currently, zot uses one of the existing repos as the master copy for a blob to achieve dedupe, which complicates dedupe tracking logic. Furthermore, we have a global storage lock which is becoming a bottleneck. In order to move to a per-repo lock, we first need to simplify this logic.

Now use a single hidden global repo (_blobstore/) as a blob store instead.

* fix(storage): address blobstore cache correctness issues

---------

**What type of PR is this?**

**Which issue does this PR fix**:

**What does this PR do / Why do we need it**:

**If an issue # is not available please add repro steps and logs showing the issue**:

**Testing done on this change**:

**Automation added to e2e**:

**Will this break upgrades or downgrades?**

**Does this PR introduce any user-facing change?**:

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #2964, opened 2025-02-14
[Bug]:  Significant performance hit due to excessive image store lock


vanhtuan0409:
### zot version

v2.1.2

### Describe the bug

We've identified a significant performance bottleneck in the image store layer due to a global lock.  This lock appears to be causing excessive contention and dramatically impacting overall system performance.  Especially during GC period

### To reproduce

1. Enable GC and image retention
2. Create a lot of ephemeral image
3. Observe any request took 10s latency

### Expected behavior

Single lock per repo

### Additional context

Code reference: https://github.com/project-zot/zot/blob/main/pkg/storage/imagestore/imagestore.go#L45

Image store use a single mutex object and it is called a lot

rchincha:
@vanhtuan0409 could you comment on what your scale is?

Yes, we are aware of some of these limitations, and there is work to address/mitigate this.

If you insist on keeping a single instance zot,
https://github.com/project-zot/zot/pull/2600

else,
https://zotregistry.dev/v2.1.2/articles/scaleout/

andaaron:
> If you insist on keeping a single instance zot, [#2600](https://github.com/project-zot/zot/pull/2600)

That is an experiment I did not get to finish. It was getting stuck, probably a deadlock.

andaaron:
@vanhtuan0409 is this something you started experiencing in v2.1.2? That lock was part of the implementation for a very log time.

What storage type and cache driver are you using?

vanhtuan0409:
@rchincha @andaaron  I do have a single node setup with 80% of repositories are ephemeral. Storage driver is s3 and cache driver is dynamodb

To my understanding, if we scale up the Zot service. Each instance will still perform GC scan through all repositories. Repositories clean up will render Zot service unusable due to the write lock on Image Storage

vanhtuan0409:
@andaaron I believe this issue has manifested in the code base for a very long time. Due to the nature, our application uses mostly ephemeral repositories and cant disable deduplication

vanhtuan0409:
Approx 400 repositories. Each repositories have 1-5 tags. 90% of those are ephemeral
Around 30 concurrent clients with 1-2 rps
Latency currently is up to 10s

rchincha:
https://zotregistry.dev/v2.1.2/articles/pprofiling

If possible can you generate a profile/flamegraph for us.

Also, pls do share the zot config (anonymized)

rchincha:
[...]

PR #4166, opened 2026-06-26
fix(api): use repo-local blob stat on HEAD and ranged GET reads

andaaron:
When dedupe is enabled, zot can store a blob in one repository and "mount" it into another via POST .../blobs/uploads/?mount=<digest>. The destination repo gets a placeholder; the real bytes live in the global dedupe cache keyed by digest.

`CheckBlob` is the write-oriented path: if the repo-local file is missing or zero-length, it looks up the digest in the dedupe cache and **re-materializes** the blob into that repository via `copyBlob`. That is correct for mount uploads, but wrong for read APIs — a client that DELETEs the blob from the destination repo must see 404 on subsequent reads there, even though the source repo (and cache) still have the content.

OCI distribution-spec v2 conformance exercises exactly this scenario ("atomic delete"): mount blob A→B, DELETE from B, HEAD B must be 404. HEAD (and ranged GET) previously called `CheckBlob`, which silently re-created the blob in B and returned 200.

Switch blob existence checks on read paths to `StatBlob`, which only consults repo-local storage (`originalBlobInfo`). Full GET was already repo-local via `GetBlob` → `originalBlobInfo`; no change needed there.

| Method        | Handler        | Storage call                          | Cross-repo dedupe on read? |
|---------------|----------------|---------------------------------------|----------------------------|
| HEAD          | CheckBlob      | StatBlob → originalBlobInfo           | No                         |
| GET (no Range)| GetBlob        | GetBlob → originalBlobInfo            | No (unchanged)             |
| GET (Range)   | GetBlob        | StatBlob, then GetBlobPartial → …     | No                         |
| POST mount    | CreateBlobUpload | CheckBlob (unchanged)               | Yes (intended)             |

`originalBlobInfo` may follow an in-repo zero-byte dedupe link via `checkCacheBlob`, but only when the repo-local path still exists. If the blob was deleted from that repo, stat fails and the API returns 404 — it does not pull bytes back from another repository.

Add integration test TestBlobReadRepoLocalAfterMountDelete covering HEAD, full GET, and ranged GET across mount + delete.

Fixes OCI distribution-spec v2 conformance failures for Blob/AtomicDelete (HEAD after cross-repo mount and delete).

### DIFF
--- a/examples/config-conformance.json
+++ b/examples/config-conformance.json
@@ -3,7 +3,7 @@
   "storage": {
     "rootDirectory": "/tmp/zot",
     "gc": true,
-    "dedupe": false
+    "dedupe": true
   },
   "http": {
     "address": "0.0.0.0",
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -5811,11 +5811,16 @@ func TestAuthorizationMountBlob(t *testing.T) {
 		So(err, ShouldBeNil)
 		So(resp.StatusCode(), ShouldEqual, http.StatusAccepted)
 
-		/* a HEAD request by user1 on blob digest (found in user1Repo) should return 200
-		because user1 has permission to read user1Repo */
-		resp, err = userClient1.R().Head(baseURL + fmt.Sprintf("/v2/%s/blobs/%s", username1+"/"+"mysecondrepo", blobDigest))
+		/* a HEAD request by user1 on the blob in repoName1 should return 200;
+		blob reads are repo-local (StatBlob), not satisfied via cross-repo cache */
+		resp, err = userClient1.R().Head(baseURL + fmt.Sprintf("/v2/%s/blobs/%s", repoName1, blobDigest))
 		So(err, ShouldBeNil)
 		So(resp.StatusCode(), ShouldEqual, http.StatusOK)
+
+		/* the same digest is not known to a different repository until mounted there */
+		resp, err = userClient1.R().Head(baseURL + fmt.Sprintf("/v2/%s/blobs/%s", username1+"/"+"mysecondrepo", blobDigest))
+		So(err, ShouldBeNil)
+		So(resp.StatusCode(), ShouldEqual, http.StatusNotFound)
 
 		// user2 can upload without dedupe
 		err = UploadImageWithBasicAuth(img, baseURL, repoName2, tag, username2, password2)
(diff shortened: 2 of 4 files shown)


renedewaele:
We have production evidence suggesting that, in addition to fixing OCI conformance, this change may significantly improve push performance for S3-backed Zot registries with `dedupe: true`. This appears to be a concrete `HEAD`/`CheckBlob` manifestation of the broader global image-store lock problem tracked in #2964.

We operate Zot v2.1.17 with an S3-compatible backend and `dedupe: true`. During a single BuildKit multi-tag push, successful authenticated `HEAD 200` requests for existing blobs took 5–40 seconds in Zot, while new blob/config/attestation writes and manifest `PUT 201` requests were reported by Zot with 0s latency.

For the same runner and image, BuildKit spent 32.0s and 51.5s pushing layers to Zot for two tags, versus 1.9s and 0.6s against GHCR.

Looking at the code, for mount-authorized requests the current `HEAD` path calls the write-oriented `CheckBlob`, takes the exclusive storage lock when dedupe is enabled, performs S3 metadata operations, and may walk repository manifests to resolve the descriptor. Concurrent BuildKit blob checks therefore appear to serialize behind remote storage work. The same path also appears to remain present in v2.1.20.

This PR’s switch to repo-local `StatBlob` seems likely to avoid the exclusive lock, descriptor walk, and possible re-materialization during a read request. The broader storage refactor in #4137 may address the global locking design, but this PR appears capable of removing one particularly hot read path from the exclusive lock sooner.

Is reducing this lock/S3 amplification an expected side effect of this change? If useful, we can provide sanitized request timings and compare `zot_storage_lock_latency_seconds` on an isolated build with and without this patch.

examples/config-conformance.json
```diff
@@ -3,7 +3,7 @@
   "storage": {
     "rootDirectory": "/tmp/zot",
     "gc": true,
-    "dedupe": false
+    "dedupe": true
```
vrajashkr, line 6: Should we have conformance tests run for both dedupe enabled and disabled?

vrajashkr (commented):
Code looks good to me. Nice PR!
One question about running conformance with both variations of dedupe.

andaaron:
Superseded by https://github.com/project-zot/zot/pull/4363.
It behaves the same as in this PR by default. The user can enable the old behavior using storage configuration options.

examples/config-conformance.json
andaaron: I guess we could, but I would expect most users would dedupe on.
It is the more complex of the 2 flows.

PR #4175, opened 2026-07-01
fix(sync): enable exponential HTTP retry backoff via maxRetryDelay

### LINKED ISSUE #4158: [Feat]: Retries with exponential backoff

For edge locations networks can become unresponsive and disconnected. These poor or disconnected network states could exist for long periods, e.g. weeks or months or for shorter periods, e.g. minute or hours.

Zot's mirror sync could be extended with an exponential backoff of retries. The technique retries an operation, with an exponentially increasing wait time up to a maximum of 24 hours and will then remain at that retry period until the operation completes successfully again. This technique embraces the fact that cloud and edge resources might intermittently be unavailable for more than a few seconds for any reason.

Here’s how you might implement a basic exponential backoff in Python. Making it maintain a maximum retry period (e.g. ~24 hours) could be done.
```python
import time
import random

def fetch_with_backoff(url, max_retries=5):
    retry_delay = 1  # Initial delay in seconds
    for attempt in range(max_retries):
        try:
            response = requests.get(url)
            response.raise_for_status()
            return response.json()
        except requests.RequestException:
            time.sleep(retry_delay)
            retry_delay *= 2  # Double the delay for the next attempt
            retry_delay += random.uniform(0, 1)  # Add jitter

    raise Exception("Maximum retry attempts reached")
``` 
In this example, the delay before the next retry attempt doubles each time a request fails. The addition of a random "jitter" avoids the scenario where many clients, following the same exponential backoff algorithm, retry simultaneously.

Best Practices and Considerations
While implementing retries and exponential backoff, there are several best practices to consider:

- Limit the Number of Retries: Always set a maximum limit/time period on retries to prevent infinite loops. If you do need to retry indefinitely, ensure that there is a maximum retry_delay to avoid the exponential backoff from growing too large (weeks, months, etc.).
- Consider the Type of Error: Not all errors are worth retrying. For example, retrying after a 4xx 404 (Not Found) error usually doesn’t make sense. Its a permanent problem.
- Use Jitter: Jitter helps prevent synchronized retries from many clients, which can create additional load at regular intervals.
- Be Mindful of Timeout Settings: Ensure that your overall timeout settings for the operation are adjusted to account for the time spent in retries.
- Logging and Monitoring: Log retry attempts and monitor the rate of retries to understand the health of the external services and the network.

-

### Additional context

_No response_

andaaron:
Add optional maxRetryDelay to sync registry config and pass separate init/max values to regclient WithDelay instead of using retryDelay for both. When maxRetryDelay is unset, it defaults to retryDelay so existing configs keep a fixed HTTP retry interval. Validate that maxRetryDelay requires retryDelay and is not less than it.

Update examples and docs for the new option.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1358,7 +1358,8 @@ Configure each registry sync:
 				"tlsVerify": true,                  # whether or not to verify tls (default is true)
 				"certDir": "/home/user/certs",      # use certificates at certDir path similar to Docker's /etc/docker/certs.d., if not specified then use the default certs dir,
 				"maxRetries": 5,                    # maxRetries in case of temporary errors (default: no retries)
-				"retryDelay": "10m",                # delay between retries, retry options are applied for both on demand and periodically sync and retryDelay is mandatory when using maxRetries.
+				"retryDelay": "1s",                 # initial HTTP retry delay; mandatory when using maxRetries
+				"maxRetryDelay": "30s",             # max HTTP retry backoff; optional, defaults to retryDelay (fixed interval). Set higher than retryDelay for exponential backoff.
 				"onlySigned": true,                 # sync only signed images (either notary or cosign)
 				"content":[                         # which content to periodically pull, also it's used for filtering ondemand images, if not set then periodically polling will not run
 					{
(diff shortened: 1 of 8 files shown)


rchincha:
| enable exponential HTTP retry

I see the max bounds but where is this exponential backoff logic being implemented?

andaaron:
> | enable exponential HTTP retry
> 
> I see the max bounds but where is this exponential backoff logic being implemented?

In regclient https://github.com/regclient/regclient/blob/v0.11.5/internal/reghttp/http.go#L344

rchincha (approved):

PR #4190, opened 2026-07-06
feat(auth): allow all authentication features to work together

### LINKED ISSUE #4033: [Bug]: Failed to use openid.oidc with bearer.oidc authorization being configured
### zot version

2.1.16

### Describe the bug

I configured openid authorization via Vault OIDC and it was working until I have added Workload OIDC authorization via bearer.oidc.
Sign in with Vault stopped to work with 400 error.

### To reproduce

1. Configuration
```yaml
http:
  auth:
    apikey: true
    htpasswd:
      path: /run/registry/htpasswd
    bearer:
      realm: zot
      service: zot
      oidc:
        - issuer: https://kube.mydomain:6443
          audiences:
            - https://kubernetes.default.svc.cluster.local
          claimMapping:
            username: claims.sub
    openid:
      providers:
        oidc:
          name: Vault
          issuer: https://vault.mydomain/v1/identity/oidc/provider/zot
          credentialsfile: /run/registry/oidc_creds.json
          claimMapping: { username: username }
          scopes: [openid, user]
```

2. Logs
```json
{"time":"2026-05-04T21:55:30.999245456+07:00","level":"error","message":"failed to authenticate due to unrecognized openid provider","caller":"zotregistry.dev/zot/v2/pkg/api/authn.go:811","func":"zotregistry.dev/zot/v2/pkg/api.(*RouteHandler).SetupRoutes.(*RouteHandler).AuthURLHandler.func2","goroutine":64}
{"time":"2026-05-04T21:55:30.999602002+07:00","level":"info","message":"HTTP API","module":"http","component":"session","clientIP":"127.0.0.1:42416","method":"GET","path":"/zot/auth/login?callback_ui=https%3A%2F%2Fzot.mydomain/home&provider=oidc","statusCode":400,"latency":"0s","bodySize":0,"headers":{}}
```

### Expected behavior

Working authentication via both methods

### Screenshots

_No response_

### Additional context

Just noted that API token login also stopped to work, tried to do docker login using zak_* token and failed with 401 Unauthorized, until I removed bearer token auth configuration block.

### LINKED ISSUE #4173: [Bug]: After configuration, anonymous accounts cannot pull images
### zot version

v2.1.18

### Describe the bug

After configuration, anonymous accounts cannot pull images

### To reproduce

1. Configuration
```
 cat config.json
{
    "distSpecVersion": "1.1.1",
    "storage": {
        "rootDirectory": "/var/lib/registry",
        "gc": true,
        "commit": true,
        "dedupe": true,
        "gcDelay": "12h",
        "gcInterval": "24h"
    },
    "http": {
        "address": "0.0.0.0",
        "port": "5000",
        "externalUrl": "https://xxxx.com",
        "compat": ["docker2s2"],
        "realm": "zot",
        "accessControl": {
            "repositories": {
                "**": {
                    "anonymousPolicy": [ "read" ],
                    "defaultPolicy": [ "read" ]
                }
            },
            "metrics": {
                 "users": ["xxx"]
            },
            "adminPolicy": {
                "users": ["xxx"],
                "actions": ["read", "create", "update", "delete"]
            }
        },
        "auth": {
            "apikey": true,
            "openid": {
                "providers": {
                    "oidc": {
                        "credentialsFile": "/etc/zot/oidc-credentials.json",
                        "issuer": "https://xxx/application/o/zot/",
                        "keypath": "",
                        "scopes": ["openid", "profile", "email"]
                    }
                }
            }
        }


    },
    "extensions": {
        "ui": {
            "enable": true
        },
        "scrub": {
            "enable": true,
            "interval": "24h"
        },
        "mgmt": {
            "enable": true
        },
        "search": {
            "enable": true,
            "cve": {
            "updateInterval": "24h"
            }
        },
        "sync": {
            "enable": true,
            "registries": [
              {
                "urls": [ "https://docker.io" ],
                "onDemand": true,
                "preserveDigest": false,
                "content": [{
                    "prefix": "**",
                    "destination": "/docker"
                }]
              },
              {
                "urls": [ "https://ghcr.io" ],
                "onDemand": true,
                "preserveDigest": false,
                "content": [{
                        "prefix": "**",
                        "destination": "/ghcr.io"
                }]
              }
            ]
        }
    },
    "log": {
        "level": "info"
    }
}
```
2. Client tool used
3. Seen error
```
docker pull xxxx.com/docker/portainer/portainer-ce:2.40.0-alpine
Error response from daemon: Head "https://xxxx.com/v2/docker/portainer/portainer-ce/manifests/2.40.0-alpine": no basic auth credentials
```

### Expected behavior

Help: Is this a configuration issue or a code issue?

### Screenshots

_No response_

### Additional context

_No response_

matheuscscp:
Fixes: #4033
Fixes: #4173

Implements the plan described here: https://github.com/project-zot/zot/issues/4033#issuecomment-4883700669

> Generated with codex/gpt-5.5:
>
> - Remove the bearer-auth short-circuit (the root cause of this issue).
> - Initialize OpenID/session/API-key/basic/mTLS auth even when bearer auth is configured.
> - Route broad Zot OCI/API requests by credential shape:
>   - `Authorization: Bearer ...` -> `bearer.oidc`, then traditional bearer.
>   - `Authorization: Basic ...` -> htpasswd/LDAP/API key.
>   - UI session header + cookie -> OpenID/session.
>   - client certificate -> mTLS.
>   - no credentials -> anonymous when policy allows it.
> - Preserve anonymous as Zot's existing empty-username model, not as a configured username/group.
> - Refine Docker `/v2/` challenge behavior so anonymous pulls still work when `anonymousPolicy` grants read.

### DIFF
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -52,6 +52,7 @@ const (
 type AuthnMiddleware struct {
 	htpasswd   *HTPasswd
 	ldapClient *LDAPClient
+	bearerAuth *BearerAuth
 	log        log.Logger
 }
 
@@ -67,16 +68,19 @@ func AuthHandler(ctlr *Controller) AuthSetup {
 	}
 
 	authConfig := ctlr.Config.CopyAuthConfig()
+
+	var tokenHandler http.HandlerFunc
+
 	if authConfig.IsBearerAuthEnabled() {
 		bearerAuth := NewBearerAuth(authConfig, ctlr.Log)
-
-		return AuthSetup{
-			Middleware:   bearerAuth.Middleware(ctlr),
-			TokenHandler: bearerAuth.TokenExchangeHandler(),
-		}
+		authnMiddleware.bearerAuth = bearerAuth
+		tokenHandler = bearerAuth.TokenExchangeHandler()
 	}
 
-	return AuthSetup{Middleware: authnMiddleware.tryAuthnHandlers(ctlr)}
+	return AuthSetup{
+		Middleware:   authnMiddleware.tryAuthnHandlers(ctlr),
+		TokenHandler: tokenHandler,
+	}
 }
 
 func (amw *AuthnMiddleware) sessionAuthn(ctlr *Controller, userAc *reqCtx.UserAccessControl,
@@ -425,13 +429,6 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
 			}
 
 			isMgmtRequested := request.RequestURI == constants.FullMgmt
-			isV2Requested := strings.TrimSuffix(request.URL.Path, "/") == constants.RoutePrefix
-			// Match Docker daemon-proxied requests regardless of the upstream client tool.
-			// The Docker daemon always prefixes its UA with "docker/<version>" when proxying,
-			// while the upstream tool (docker CLI, compose, buildx, etc.) appears inside
-			// "UpstreamClient(...)". Direct Docker CLI requests use "Docker-Client/...".
-			ua := request.Header.Get("User-Agent")
-			isDockerClient := strings.Contains(ua, "Docker-Client") || strings.HasPrefix(ua, "docker/")
 
 			// Get auth config safely
 			authConfig := ctlr.Config.CopyAuthConfig()
@@ -461,8 +458,14 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
 			case hasMultipleAuthorizationHeaders(request):
 				authenticated = false
 
-			// The authorization header presence is an explicit attempt to use basic authentication
-			case !isAuthorizationHeaderEmpty(request) && authConfig.IsBasicAuthnEnabled():
+			// Bearer credentials are owned by the bearer backends.
+			case hasBearerAuthorizationHeader(request) && amw.bearerAuth != nil:
+				amw.bearerAuth.Middleware(ctlr)(next).ServeHTTP(response, request)
+
+				return
+
+			// Basic credentials are owned by htpasswd/LDAP/API key authentication.
+			case hasBasicAuthorizationHeader(request) && authConfig.CanAuthenticateWithBasicCredentials():
 				authenticated, err = amw.basicAuthn(ctlr, userAc, response, request)
 
 			// The session header is an explicit attempt to use session authentication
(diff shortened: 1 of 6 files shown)


rchincha:
Blocker: /v2/ no longer challenges in mixed anonymous/authenticated configs

pkg/api/routes.go:292-295
The new anonymousRead := ... HasAnonymousReadPolicy() suppresses WWW-Authenticate whenever any repo grants anonymous read.
This breaks Docker/skopeo credential negotiation for authenticated push/private repos: clients see /v2/ as unauthenticated and do not send stored credentials later.
CI already shows this: Check client tools fails in anonymous_policy.bats because authenticated push gets authentication required after /v2/ returned 200.
Suggested fix: preserve the mixed-policy Docker challenge behavior; only suppress /v2/ challenge in truly anonymous-only/read-only cases where no authenticated operations need negotiation.
Mixed Basic + ****** still advertise only ****** unauthenticated clients

pkg/api/authn.go:507-508, pkg/api/bearer_auth.go:236-238
With both Basic-capable auth and ****** no-credential requests are delegated to BearerAuth.Middleware, so clients receive only a ******
The new test uses preemptive SetBasicAuth, so it does not cover normal Docker/client negotiation.
Suggested fix: advertise both challenges, or otherwise ensure Basic/API-key clients receive a Basic challenge when those backends are enabled.
Lower priority: OpenID-only configs block anonymous fallback for unrelated Authorization schemes

pkg/api/authn.go:886-889
IsBasicAuthnEnabled() includes OpenID, but OpenID/session auth does not own the Authorization header. This contradicts the comment and can reject anonymous access for OpenID-only configs with stray/unsupported Authorization.
Suggested fix: use CanAuthenticateWithBasicCredentials() there instead of IsBasicAuthnEnabled().

rchincha:
I don't believe there is an easy solution that works which is interoperable with current ecosystem tooling.
cc: @andaaron

andaaron:
> I don't believe there is an easy solution that works which is interoperable with current ecosystem tooling. cc: @andaaron

No, not easy. But if we can capture the existing client workflows in a document such as https://github.com/project-zot/zot/issues/4033#issuecomment-4900400211 we should be able to find a solution which would work for most of them.
Using bearer for most clients seems to work fine for Harbor. We could go a similar path and generate our own tokens, even tokens containing anonymous permissions.

That leaves out some clients such as pyoci, for which we'd need to find a solution.

matheuscscp:
One way to solve #4173 is not considering it a bug but a feature request, we make it opt-in. Since changing this behavior will change the set of compatible clients, the users who want this could still try and see if it works for them, and by making it opt-in we ship this feature without breaking default functionality.

matheuscscp:
But honestly I'm not too concerned with #4173... Even ECR has two separate services for public and private audiences. As far as I know, ECR only works with authentication, and ECR Public is a separate service, with a separate domain / host, that accepts anonymous pulls. And I'm not sure you can restrict ECR Public to not be public.

Honestly I think I'd defer this use case to have more people interested in it and have time to think it carefully.

#4033 On the other hand sounds super important if we can do something about it... In that case authentication is always required, and it's super desirable that both humans and machines can access the registry.

matheuscscp:
> But honestly I'm not too concerned with #4173... Even ECR has two separate services for public and private audiences. As far as I know, ECR only works with authentication, and ECR Public is a separate service, with a separate domain / host, that accepts anonymous pulls. And I'm not sure you can restrict ECR Public to not be public.
> 
> Honestly I think I'd defer this use case to have more people interested in it and have time to think it carefully.
> 
> #4033 On the other hand sounds super important if we can do something about it... In that case authentication is always required, and it's super desirable that both humans and machines can access the registry.

Hey @rchincha @andaaron, I'm working on a separate PR to address #4033 and a few other things, but not #4173. I really don't see a way out for #4173 right now, except for the suggestion above about splitting Zot in two separate deployments (e.g. making the anonymous-facing one read-only, for example).

Going to close this one.

PR #4191, opened 2026-07-07
feat(retention): add untagged retention configuration

### LINKED ISSUE #4148: [Feat]: retention rules ignore untagged manifests, evicting digest-only pull-through cache items

When zot runs as an on-demand pull-through cache and clients pull by digest, the cached manifests are stored untagged. Garbage collection only protects untagged manifests by a blob-age timer (`retention.delay`) and never considers pull activity, so images that are actively being pulled get evicted and immediately re-synced on the next request. There is no configuration that prevents this, because retention rules (`pulledWithin`, `mostRecentlyPulledCount`, etc.) apply to tags only.

```mermaid
sequenceDiagram
    autonumber
    participant C as OCI client
    participant Z as zot (on-demand cache, gc=on)
    participant U as Upstream OCI registry

    C->>Z: GET manifest by digest (sha256:…)
    Z->>U: cache miss → on-demand sync (fetch by digest)
    U-->>Z: manifest bytes (a by-digest fetch carries no tag)
    Z->>Z: store manifest UNTAGGED in index.json
    Z-->>C: 200 — served
    Note over C,Z: many subsequent clients pull the same digest — fast cache hits

    Z->>Z: GC pass — manifest is untagged & blob older than delay → DELETE
    C->>Z: GET same digest (still in active use)
    Z->>U: cache miss again → on-demand re-sync
    Note over C,U: cache thrash — an in-use image is repeatedly evicted and re-fetched
```

The cached `index.json` descriptor has no `org.opencontainers.image.ref.name` annotation, confirming the manifest is untagged on the pull-through cache even when it is tagged upstream. A by-digest fetch never transfers the upstream tag (the tag lives in the upstream index, not in the manifest).

We have little control over how clients build references; pulling by digest is legitimate and recommended for immutability, but it produces untagged cache entries. The blunt levers (`gc: false` or `deleteUntagged: false`) stop the thrash but also stop all reclamation, defeating some of our goals with the cache server.

Allow retention to keep untagged manifests based on pull activity, e.g. by letting `pulledWithin` / `mostRecentlyPulledCount` apply to untagged manifests (opt-in per repo?), and by making untagged-manifest GC consult the last-pull timestamp instead of only blob age.

zot already records per-digest pull stats, including for digest pulls — `UpdateStatsOnDownload` updates `Statistics[<digest>].LastPullTimestamp`, and `SetRepoReference` creates a `Statistics` entry for every manifest (tagged or not). The retention/GC path simply never reads it for untagged manifests.

A few pointers in the v2.1.17 code base:

- Retention candidates are built only from tags — untagged manifests are never evaluated by any rule. `pkg/retention/candidate.go` → `GetCandidates` iterates `repoMeta.Tags`.
- Untagged manifests go down a separate path whose only control is a blob-age timer, not pull activity. `pkg/storage/gc/gc.go` → `removeUntaggedManifests` → `gcManifest` → `isBlobOlderThan(retention.delay)` (uses blob mod-time via `StatBlob`; a `GET` never refreshes mod-time).
- By default `deleteUntagged` is `true` and, when GC is on, `retention.delay` defaults to `gcDelay` — so untagged manifests are reaped aggressively with no opt-out short of disabling GC entirely or `deleteUntagged: false` (which then keeps everything and disables reclamation).

I'm not really sure what other alternatives are possible, but I'm definitely open to discussing other ideas!

### Additional context

_No response_

rchincha:
Fixes https://github.com/project-zot/zot/issues/4148

Digest-only pull-through cache entries are stored as untagged manifests, so existing tag-based retention rules do not protect actively pulled content. This adds explicit untagged retention rules that use the same pull/push activity signals as tag retention.

Configuration
Added keepUntagged under retention policies.
Reuses mostRecentlyPushedCount, mostRecentlyPulledCount, pulledWithin, and pushedWithin. Keeps deleteUntagged as the master switch.
{
  "repositories": ["infra/*", "tmp/**"],
  "deleteUntagged": true,
  "keepTags": [{
    "patterns": ["v2.*", ".*-prod"],
    "mostRecentlyPulledCount": 10,
    "pulledWithin": "720h"
  }],
  "keepUntagged": {
    "mostRecentlyPulledCount": 10,
    "pulledWithin": "720h"
  }
}
Retention evaluation

Builds untagged candidates from index descriptors without tag annotations. Joins candidates with MetaDB digest statistics.
Evaluates tag and untagged pools independently.
Garbage collection

Skips untagged manifests retained by keepUntagged. Falls back to existing delay-based cleanup when untagged retention cannot use MetaDB statistics. Docs and schema

Updated retention examples and README guidance.
Extended generated schema coverage for keepUntagged.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -198,6 +198,7 @@ You can also force a full reparse with the `--force-reparse` flag to `zot serve`
 ## Retention
 
 You can define tag retention rules that govern how many tags of a given repository to retain, or for how long to retain certain tags.
+You can also define `keepUntagged` rules for untagged manifests, including manifests cached by digest-only pull-through requests.
 
 There are 4 possible rules for tags:
 
@@ -206,7 +207,12 @@ mostRecentlyPulledCount: x - top x most recently pulled tags
 pulledWithin: x hours - tags pulled in the last x hours
 pushedWithin: x hours - tags pushed in the last x hours
 
-If ANY of these rules are met by a tag, then it will be retained, in other words there is an OR logic between them
+The same 4 rules can be used under `keepUntagged`; for digest-only cached content, pushed means cached locally at the descriptor's push timestamp.
+Tagged and untagged manifests are evaluated separately, so counts in `keepTags` do not compete with counts in `keepUntagged`.
+An empty `keepUntagged: {}` does not retain all untagged manifests; it is equivalent to omitting `keepUntagged`.
+`keepUntagged` rules require the metadata database. If it is unavailable, zot ignores `keepUntagged` and uses the existing delay-based untagged cleanup.
+
+If ANY of these rules are met by a tag or untagged manifest, then it will be retained, in other words there is an OR logic between them
 
 repositories uses glob patterns
 tag patterns uses regex
@@ -236,7 +242,11 @@ tag patterns uses regex
                         "patterns": ["v1.*"],           // all the other tags will be removed
                         "pulledWithin": "168h",      
                         "pushedWithin": "168h"
-                    }]
+                    }],
+                    "keepUntagged": {                   // untagged manifests pulled or cached locally within the last 168h are retained
+                        "pulledWithin": "168h",
+                        "pushedWithin": "168h"
+                    }
                 },
                 {
                     "repositories": ["**"],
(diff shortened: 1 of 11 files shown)


rchincha:
https://github.com/project-zot/zot/issues/4148

@seniorquico

This PR is our current thinking around ^

andaaron (approved):

PR #4201, opened 2026-07-08
fix(storage): only promote dedupe origin when the deleted blob is the origin (boltdb)

archy-rock3t-cloud:
**What type of PR is this?**
bug

The BoltDB cache driver promoted a duplicate blob into the origin bucket even when the deleted path was not the current origin. On S3 dedupe, where duplicates are zero-size stubs, this points the content holder at a stub and the blob can be lost on a later delete. The DynamoDB and Redis drivers already guard this with an origin == path check; this makes BoltDB match them.

Added TestBoltDBDeleteNonOriginDuplicate. It fails without the fix and passes with it.

No.

No.


andaaron (approved):

andaaron:
@archy-rock3t-cloud good catch, thank you for the pr

PR #4202, opened 2026-07-08
fix(imagetrust): bind cosign signature verification to the manifest digest

archy-rock3t-cloud:
**What type of PR is this?**
bug

VerifyCosignSignature checked the signature over the payload but never checked that the payload's docker-manifest-digest matches the manifest being verified, which is the claim check cosign performs in CheckClaims. A signature valid for one image could be attached to another image signed by the same key and reported as trusted. This verifies the claimed digest before returning trusted.

Added TestCosignSignatureDigestBinding. It signs a real image, then verifies that signature against a different manifest digest: trusted without the fix, rejected with it.

No.

An image whose recorded cosign signature does not reference its own manifest digest is no longer reported as trusted.

### DIFF
--- a/pkg/extensions/imagetrust/cosign.go
+++ b/pkg/extensions/imagetrust/cosign.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"crypto"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -20,6 +21,7 @@ import (
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	sigstoreSigs "github.com/sigstore/sigstore/pkg/signature"
 	"github.com/sigstore/sigstore/pkg/signature/options"
+	sigPayload "github.com/sigstore/sigstore/pkg/signature/payload"
 
 	zerr "zotregistry.dev/zot/v2/errors"
 )
@@ -117,6 +119,16 @@ func VerifyCosignSignature(
 		err = verifier.VerifySignature(bytes.NewReader(signature), bytes.NewReader(payload),
 			options.WithContext(context.Background()))
 		if err == nil {
+			var signedPayload sigPayload.SimpleContainerImage
+
+			if err := json.Unmarshal(payload, &signedPayload); err != nil {
+				continue
+			}
+
+			if signedPayload.Critical.Image.DockerManifestDigest != digest.String() {
+				continue
+			}
+
 			return string(pubKeyContent), true, nil
 		}
 	}
--- a/pkg/extensions/imagetrust/image_trust_test.go
+++ b/pkg/extensions/imagetrust/image_trust_test.go
@@ -548,6 +548,114 @@ func TestVerifySignatures(t *testing.T) {
 	})
 }
 
+func TestCosignSignatureDigestBinding(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/extensions/imagetrust/cosign.go
```diff
@@ -117,6 +118,22 @@ func VerifyCosignSignature(
 		err = verifier.VerifySignature(bytes.NewReader(signature), bytes.NewReader(payload),
 			options.WithContext(context.Background()))
 		if err == nil {
+			var signedPayload struct {
```
andaaron: Is there any reason not to us the official struct?


```suggestion
import sigPayload "github.com/sigstore/sigstore/pkg/signature/payload"

var signedPayload sigPayload.SimpleContainerImage
if err := json.Unmarshal(payload, &signedPayload); err != nil {
    continue
}
if signedPayload.Critical.Image.DockerManifestDigest != digest.String() {
    continue
}
```

pkg/extensions/imagetrust/image_trust_test.go
```diff
@@ -548,6 +548,112 @@ func TestVerifySignatures(t *testing.T) {
...
+			if manifest.Digest != image.Digest() {
+				blobContent, err := ctlr.StoreController.DefaultStore.GetBlobContent(repo, manifest.Digest)
+				So(err, ShouldBeNil)
+
+				var cosignSig ispec.Manifest
+
+				err = json.Unmarshal(blobContent, &cosignSig)
+				So(err, ShouldBeNil)
+
+				sigKey = cosignSig.Layers[0].Annotations[zcommon.CosignSigKey]
```
andaaron, line 629: Maybe we should add 
`So(sigKey, ShouldNotBeEmpty)`

pkg/extensions/imagetrust/image_trust_test.go
```diff
@@ -548,6 +548,112 @@ func TestVerifySignatures(t *testing.T) {
...
+
+				rawSignature, err = ctlr.StoreController.DefaultStore.GetBlobContent(repo, cosignSig.Layers[0].Digest)
```
andaaron, line 632: Maybe we should add `So(rawSignature, ShouldNotBeEmpty)`

andaaron:
@archy-rock3t-cloud, this is a very good catch, please look at the comments.
And also fix the lint issues.

archy-rock3t-cloud:
@andaaron sure, will do

pkg/extensions/imagetrust/image_trust_test.go
```diff
@@ -548,6 +548,115 @@ func TestVerifySignatures(t *testing.T) {
...
+		cwd, err := os.Getwd()
+		So(err, ShouldBeNil)
+
+		So(os.Chdir(cosignDir), ShouldBeNil)
+
+		t.Setenv("COSIGN_PASSWORD", "")
+		err = generate.GenerateKeyPairCmd(context.TODO(), "", "cosign", nil)
+		So(err, ShouldBeNil)
+
+		So(os.Chdir(cwd), ShouldBeNil)
```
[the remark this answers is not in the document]
andaaron: THe suggested code is wrong.

andaaron:
@archy-rock3t-cloud I think you forgot to push the update for my last 2 comments (the assertions). Please also rebase, I merged your other PRs.

andaaron (approved):

PR #4209, opened 2026-07-09
fix(sync): guard registry credentials map against concurrent refresh

archy-rock3t-cloud:
**What type of PR is this?**
bug

refreshRegistryTemporaryCredentials wrote service.credentials and reinitialized the client without holding clientLock, racing with concurrent on-demand syncs and risking a fatal concurrent map write. This holds clientLock during refresh.

Added TestRefreshRegistryTemporaryCredentialsConcurrent. Run with -race it reports a data race without the fix.

No.

No.


andaaron:
Superseded by https://github.com/project-zot/zot/pull/4272

PR #4210, opened 2026-07-09
fix(events): apply default HTTP sink timeout when unset

archy-rock3t-cloud:
**What type of PR is this?**
bug

NewHTTPSink built the request context deadline from config.Timeout without a default, so an unset timeout produced an already-expired deadline and every event failed to deliver. This applies the default timeout when none is configured.

Added a TestHTTPSink case that emits an event through a sink with no timeout set and expects it to be delivered. It fails without the fix.

No.

No.

### DIFF
--- a/pkg/extensions/events/http_sink.go
+++ b/pkg/extensions/events/http_sink.go
@@ -30,6 +30,10 @@ func NewHTTPSink(config eventsconf.SinkConfig) (*HTTPSink, error) {
 		return nil, zerr.ErrEventSinkAddressEmpty
 	}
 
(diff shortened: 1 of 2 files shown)


andaaron (approved):

andaaron:
@archy-rock3t-cloud please rebase and fix lint issues. The test issues are unrelated, should be fixed on main.

rchincha:
@archy-rock3t-cloud pls fix the ci failures

andaaron:
Superseded by https://github.com/project-zot/zot/pull/4277

PR #4211, opened 2026-07-09
fix(api): reject backslash in OIDC callback redirect to prevent open redirect

archy-rock3t-cloud:
**What type of PR is this?**
bug

ValidateCallbackUI accepted callbacks such as /\example.com. This rejects callbacks containing a backslash.

Added backslash cases to TestValidateCallbackUI. They redirect to the safe default only with the fix.

No.

No.

### DIFF
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -593,6 +593,10 @@ func ValidateCallbackUI(callbackUI string, allowOrigins []string) string {
 		return "/"
 	}
 
+	if strings.Contains(callbackUI, "\\") {
+		return "/"
+	}
+
 	parsed, err := url.Parse(callbackUI)
 	if err != nil {
 		return "/"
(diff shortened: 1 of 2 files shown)


pkg/api/authn.go
```diff
@@ -593,6 +593,10 @@ func ValidateCallbackUI(callbackUI string, allowOrigins []string) string {
 		return "/"
 	}
 
+	if strings.Contains(callbackUI, "\\") {
```
andaaron, line 596: Thanks for the fix — this matches the backslash open-redirect issue in ValidateCallbackUI (`/\evil.com` passing validation and ending up in the post-OIDC http.Redirect).

There is also the case of callback_ui `%2F%5Cevil.attacker.com`.
That input has no literal \, so it still passes validation and is returned as /%5Cevil.com. Browsers decode %5C to \ and can normalize the Location header the same way.

Please extend the fix to also reject percent-encoded backslashes, e.g.:
```
strings.Contains(strings.ToUpper(callbackUI), "%5C") {
    return "/"
}
```

(or check parsed.Path after url.Parse for `\`, which also catches single-encoded %5C).

rchincha:
@archy-rock3t-cloud will you be able to work on your PR?

andaaron:
Superseded by https://github.com/project-zot/zot/pull/4238

PR #4220, opened 2026-07-11
test: migrate extensions and pkg/test from GetFreePort to port "0"

andaaron:
Replace pre-allocated ports with kernel-assigned binding across pkg/extensions (except sync) and pkg/test integration helpers to avoid TOCTOU bind races under parallel CI.

Extensions: imagetrust, lint, monitoring, scrub, search, and parent extension tests use conf.HTTP.Port = "0" with baseURL := cm.StartAndWait(). Cosign and notation tests resolve localhost:port refs from cm.Port() after bind. Search CVE subtests that inject a mock scanner after Init() poll ctlr.GetPort() before WaitTillServerReady().

pkg/test: migrate upload, oci-utils, notation, and utils_test helpers the same way. One no-server upload subtest keeps GetFreePort() to hit a port with nothing listening (no bind, so no TOCTOU risk).

This is a pattern added in https://github.com/project-zot/zot/pull/4216

And one of the extension failures we fis is highlighted at: https://github.com/project-zot/zot/actions/runs/28934135670/job/85843668079

### DIFF
--- a/pkg/extensions/extension_events_disabled_test.go
+++ b/pkg/extensions/extension_events_disabled_test.go
@@ -18,14 +18,13 @@ import (
 func TestEventsExtension(t *testing.T) {
 	Convey("event generation is skipped when extension is disabled", t, func() {
 		conf := config.New()
-		port := test.GetFreePort()
 
 		globalDir := t.TempDir()
 		defaultValue := true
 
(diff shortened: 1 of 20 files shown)


vrajashkr (approved):
LGTM. Thanks for making this change.

PR #4221, opened 2026-07-12
fix(search): use index descriptor platform when config platform is empty

andaaron:
Multi-arch image indexes (including ORAS artifact indexes) could carry OS/arch on index manifest descriptors rather than in the config blob. GraphQL search omitted platform for those manifests, leaving the UI OS/Arch dropdown empty.

Fall back to the index descriptor when config platform is unset, while still preferring config when both are present. Apply the same resolution when resolving index tags by platform or digest in resolveImageData.

See symptoms below.

Before:
image

After:
image

### DIFF
--- a/pkg/extensions/search/convert/convert_internal_test.go
+++ b/pkg/extensions/search/convert/convert_internal_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 
 	"github.com/99designs/gqlgen/graphql"
+	godigest "github.com/opencontainers/go-digest"
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
 	. "github.com/smartystreets/goconvey/convey"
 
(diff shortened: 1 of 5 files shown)


rchincha (approved):
reasonable compromise. What happens if index platform is empty also?

vrajashkr (approved):

andaaron:
> reasonable compromise. What happens if index platform is empty also?

When both are empty, behavior is unchanged: GraphQL leaves OS/Arch empty (convert_test.go “leaves platform empty without index descriptor platform”), and resolveImageData returns ErrImageNotFound when filtering by platform. Digest-only lookup still works but returns empty platform.

See https://github.com/project-zot/zot/pull/4221/changes#diff-e49883642a0d16dddf14f67bf1bbc5e1826387e087b593ad6aded8274d9f2aaeR569

PR #4223, opened 2026-07-13
feat: restrict periodic storage GC to a configurable gcTimeWindow

### LINKED ISSUE #3001: [Feat]: Implement gcTimeWindow in order to reduce storage locking during active usage time

shcherbak:
Adds a gcTimeWindow storage config option (e.g. "01:00-08:00") that restricts periodic garbage collection to a daily off-peak window, avoiding lock contention on storage during high-load periods. Resolves [#3001](https://github.com/project-zot/zot/issues/3001).

- New gcTimeWindow field on StorageConfig (top-level and per-subpath), start inclusive / end exclusive, wraps past midnight (e.g. "22:00-06:00"). Applied at server startup, same as the existing
  gcDelay/gcInterval settings - a config file reload updates the stored value but doesn't affect already-running periodic GC tasks; a restart is required to pick up changes.
- gc.ValidateGCTimeWindow validates the "HH:MM-HH:MM" format at config load time; windows that wrap past midnight (e.g. "22:00-06:00") are supported.
- GCTaskGenerator.IsReady now also checks the configured window before allowing a GC run to proceed.
- Unit tests for parsing, validation, wrap-around containment, and generator behavior; config-load tests for valid/invalid windows and the case where a window is set without GC enabled.
- Docs and example config updated (examples/README.md, examples/config-gc.json).

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -136,6 +136,16 @@ Orphan blobs are removed if they are older than gcDelay.
         "gcDelay": "2h"
 ```
 
+On high-load registries, garbage collection can add lock contention on storage
+while it runs. To restrict periodic GC runs to a daily time-of-day window
+(e.g. only during off-peak hours), set `gcTimeWindow` to a "HH:MM-HH:MM" range
+in local time. A window may wrap past midnight (e.g. `"22:00-06:00"`). Leaving
+it unset (the default) allows GC to run at any time:
+
+```
+        "gcTimeWindow": "01:00-08:00"
+```
+
 To limit the maximum number of repositories that can be created, set:
 
 ```
--- a/examples/README.md
+++ b/examples/README.md
@@ -139,12 +139,16 @@ Orphan blobs are removed if they are older than gcDelay.
 On high-load registries, garbage collection can add lock contention on storage
 while it runs. To restrict periodic GC runs to a daily time-of-day window
 (e.g. only during off-peak hours), set `gcTimeWindow` to a "HH:MM-HH:MM" range
-in local time. A window may wrap past midnight (e.g. `"22:00-06:00"`). Leaving
-it unset (the default) allows GC to run at any time:
+in local time (start inclusive, end exclusive). A window may wrap past midnight
+(e.g. `"22:00-06:00"`). Leaving it unset (the default) allows GC to run at any time:
 
 ```
         "gcTimeWindow": "01:00-08:00"
 ```
+
+`gcTimeWindow` is applied when periodic GC tasks are scheduled at server startup;
+changing it via a config file reload updates the stored value but does not affect
+already-running periodic GC tasks, so a restart is required for the change to take effect.
 
 To limit the maximum number of repositories that can be created, set:
 
--- a/examples/README.md
+++ b/examples/README.md
@@ -139,12 +139,13 @@ Orphan blobs are removed if they are older than gcDelay.
 On high-load registries, garbage collection can add lock contention on storage
 while it runs. To restrict when a periodic GC sweep over all repositories may
 start to a daily time-of-day window (e.g. only during off-peak hours), set
-`gcTimeWindow` to a "HH:MM-HH:MM" range in local time (start inclusive, end
-exclusive). A window may wrap past midnight (e.g. `"22:00-06:00"`). A sweep
-that starts inside the window is allowed to run to completion even past the
-window's end, so GC work stays amortized instead of stalling mid-sweep on a
-large registry until the window reopens the next day. Leaving it unset (the
-default) allows GC to run at any time:
+`gcTimeWindow` to a "HH:MM-HH:MM" range in UTC (start inclusive, end
+exclusive) - not the server's local time zone, to keep behavior unambiguous
+regardless of how the host or container is configured. A window may wrap
+past midnight (e.g. `"22:00-06:00"`). A sweep that starts inside the window 
+is allowed to run to completion even past the window's end, so GC work stays 
+amortized instead of stalling mid-sweep on a large registry until the window
+reopens the next day. Leaving it unset (the default) allows GC to run at any time:
 
 ```
         "gcTimeWindow": "01:00-08:00"
--- a/examples/README.md
+++ b/examples/README.md
@@ -137,10 +137,14 @@ Orphan blobs are removed if they are older than gcDelay.
 ```
 
 On high-load registries, garbage collection can add lock contention on storage
-while it runs. To restrict periodic GC runs to a daily time-of-day window
-(e.g. only during off-peak hours), set `gcTimeWindow` to a "HH:MM-HH:MM" range
-in local time (start inclusive, end exclusive). A window may wrap past midnight
-(e.g. `"22:00-06:00"`). Leaving it unset (the default) allows GC to run at any time:
+while it runs. To restrict when a periodic GC sweep over all repositories may
+start to a daily time-of-day window (e.g. only during off-peak hours), set
+`gcTimeWindow` to a "HH:MM-HH:MM" range in local time (start inclusive, end
+exclusive). A window may wrap past midnight (e.g. `"22:00-06:00"`). A sweep
+that starts inside the window is allowed to run to completion even past the
+window's end, so GC work stays amortized instead of stalling mid-sweep on a
+large registry until the window reopens the next day. Leaving it unset (the
+default) allows GC to run at any time:
 
 ```
         "gcTimeWindow": "01:00-08:00"
--- a/examples/config-gc.json
+++ b/examples/config-gc.json
@@ -4,7 +4,8 @@
         "rootDirectory": "/tmp/zot",
         "gc": true,
         "gcDelay": "2h",
-        "gcInterval": "1h"
+        "gcInterval": "1h",
+        "gcTimeWindow": "01:00-08:00"
     },
     "http": {
         "address": "127.0.0.1",
--- a/pkg/api/config/config.go
+++ b/pkg/api/config/config.go
@@ -36,9 +36,12 @@ type StorageConfig struct {
 	Commit          bool
 	GCDelay         time.Duration // applied for blobs
 	GCInterval      time.Duration
-	Retention       ImageRetention
-	StorageDriver   map[string]any `mapstructure:",omitempty"`
-	CacheDriver     map[string]any `mapstructure:",omitempty"`
+	// GCTimeWindow restricts periodic garbage-collection runs to a daily time-of-day
+	// window, e.g. "01:00-08:00". Empty means GC can run at any time.
+	GCTimeWindow  string
+	Retention     ImageRetention
+	StorageDriver map[string]any `mapstructure:",omitempty"`
+	CacheDriver   map[string]any `mapstructure:",omitempty"`
 
 	// GCMaxSchedulerDelay is the maximum random delay for GC task scheduling
 	// This field is not configurable by the end user
@@ -1070,6 +1073,7 @@ func (c *Config) UpdateReloadableConfig(newConfig *Config) {
 	c.Storage.RedirectBlobURL = newConfig.Storage.RedirectBlobURL
 	c.Storage.GCDelay = newConfig.Storage.GCDelay
 	c.Storage.GCInterval = newConfig.Storage.GCInterval
+	c.Storage.GCTimeWindow = newConfig.Storage.GCTimeWindow
 
 	// Only update retention if we have a metaDB already in place
 	if c.isRetentionEnabledInternal() {
@@ -1088,6 +1092,7 @@ func (c *Config) UpdateReloadableConfig(newConfig *Config) {
 		subPathConfig.RedirectBlobURL = storageConfig.RedirectBlobURL
 		subPathConfig.GCDelay = storageConfig.GCDelay
 		subPathConfig.GCInterval = storageConfig.GCInterval
+		subPathConfig.GCTimeWindow = storageConfig.GCTimeWindow
 
 		// Only update retention if we have a metaDB already in place
 		if c.isRetentionEnabledInternal() {
--- a/pkg/api/config/config.go
+++ b/pkg/api/config/config.go
@@ -37,8 +37,8 @@ type StorageConfig struct {
 	GCDelay         time.Duration // applied for blobs
 	GCInterval      time.Duration
 	// GCTimeWindow restricts periodic garbage-collection runs to a daily time-of-day
-	// window, e.g. "01:00-08:00". Empty means GC can run at any time.
-	GCTimeWindow  string
+	// window, e.g. "01:00-08:00". The zero value means GC can run at any time.
+	GCTimeWindow  GCTimeWindow
 	Retention     ImageRetention
 	StorageDriver map[string]any `mapstructure:",omitempty"`
 	CacheDriver   map[string]any `mapstructure:",omitempty"`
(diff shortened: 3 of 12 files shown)


examples/README.md
```diff
@@ -136,6 +136,16 @@ Orphan blobs are removed if they are older than gcDelay.
         "gcDelay": "2h"
 ```
 
+On high-load registries, garbage collection can add lock contention on storage
+while it runs. To restrict periodic GC runs to a daily time-of-day window
+(e.g. only during off-peak hours), set `gcTimeWindow` to a "HH:MM-HH:MM" range
+in local time. A window may wrap past midnight (e.g. `"22:00-06:00"`). Leaving
+it unset (the default) allows GC to run at any time:
```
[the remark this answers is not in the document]
shcherbak: agree

pkg/api/controller.go
```diff
@@ -628,6 +628,7 @@ func RunGCTasks(conf *config.Config, storeController storage.StoreController, me
 			Delay:             storageConfig.GCDelay,
 			ImageRetention:    storageConfig.Retention,
 			MaxSchedulerDelay: storageConfig.GCMaxSchedulerDelay,
+			TimeWindow:        storageConfig.GCTimeWindow,
 		}, audit, logger, metrics)
```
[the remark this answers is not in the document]
shcherbak: agree

pkg/api/controller.go
```diff
@@ -628,6 +628,7 @@ func RunGCTasks(conf *config.Config, storeController storage.StoreController, me
...
```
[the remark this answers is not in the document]
shcherbak: readme and pr description are upadted

pkg/storage/gc/gc.go
```diff
@@ -1151,7 +1233,15 @@ func (gen *GCTaskGenerator) IsDone() bool {
...
-	return time.Now().After(gen.nextRun)
+	if !time.Now().After(gen.nextRun) {
+		return false
+	}
+
+	if gen.timeWindow != nil && !gen.timeWindow.contains(time.Now()) {
+		return false
+	}
+
+	return true
```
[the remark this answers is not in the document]
shcherbak: edge case, bu is valid, agree

rchincha:
The downside of this approach is that the gc work is no longer amortized but has to finish within that time window - which it may not.

shcherbak:
> The downside of this approach is that the gc work is no longer amortized but has to finish within that time window - which it may not.

Good catch - updated the design so gcTimeWindow only gates when a new sweep is allowed to start, not every repo-task within it. Once a sweep begins inside the window it runs to completion even past the window's end, so GC work stays fully amortized and won't stall mid-sweep on a large registry until the next day's window. Docs and tests updated accordingly.

pkg/cli/server/root.go
```diff
@@ -1597,6 +1598,13 @@ func validateGC(config *config.Config, logger zlog.Logger) error {
 			zerr.ErrBadConfig, storageConfig.GCInterval)
 	}
 
+	if err := gc.ValidateGCTimeWindow(storageConfig.GCTimeWindow); err != nil {
+		logger.Error().Err(err).Str("gcTimeWindow", storageConfig.GCTimeWindow).
+			Msg("invalid garbage-collect time window specified")
+
+		return err
```
andaaron:
```suggestion
		return fmt.Errorf("%w: invalid garbage-collect time window specified %s",
			zerr.ErrBadConfig, storageConfig.GCTimeWindow)
```

Same for substores.

pkg/cli/server/root.go
```diff
@@ -1625,6 +1638,15 @@ func validateGC(config *config.Config, logger zlog.Logger) error {
...
 		}
 
+		if err := gc.ValidateGCTimeWindow(subPath.GCTimeWindow); err != nil {
+			logger.Error().Err(err).
+				Str("subPath", name).
+				Str("gcTimeWindow", subPath.GCTimeWindow).
+				Msg("invalid garbage-collect time window specified")
+
+			return err
+		}
```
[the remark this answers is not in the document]
andaaron: Seems to be correct.

pkg/api/config/config.go
```diff
@@ -36,9 +36,12 @@ type StorageConfig struct {
 	Commit          bool
 	GCDelay         time.Duration // applied for blobs
 	GCInterval      time.Duration
-	Retention       ImageRetention
-	StorageDriver   map[string]any `mapstructure:",omitempty"`
-	CacheDriver     map[string]any `mapstructure:",omitempty"`
+	// GCTimeWindow restricts periodic garbage-collection runs to a daily time-of-day
+	// window, e.g. "01:00-08:00". Empty means GC can run at any time.
+	GCTimeWindow  string
```
andaaron: `GCTimeWindow` is a raw string parsed again in validateGC and again in newGCTimeWindow.
How about making this a typed field (e.g. gc.TimeWindow) decoded at LoadConfiguration time via a mapstructure decode hook alongside StringToTimeDurationHookFunc (root.go:1261-1263). That matches how gcDelay/gcInterval work and avoids double-parsing.

pkg/cli/server/root.go
```diff
@@ -1625,6 +1638,15 @@ func validateGC(config *config.Config, logger zlog.Logger) error {
 				zerr.ErrBadConfig, subPath.GCDelay)
 		}
 
+		if err := gc.ValidateGCTimeWindow(subPath.GCTimeWindow); err != nil {
+			logger.Error().Err(err).
+				Str("subPath", name).
+				Str("gcTimeWindow", subPath.GCTimeWindow).
+				Msg("invalid garbage-collect time window specified")
+
+			return err
```
andaaron: Return fmt.Errorf("%w: invalid garbage-collect time window specified %s", zerr.ErrBadConfig, subPath.GCTimeWindow) instead of returning the wrapped parse error directly. Consider including the subpath name in the message for easier debugging.

pkg/storage/gc/gc.go
```diff
@@ -43,9 +44,99 @@ type Options struct {
 	// Defaults to 30 seconds if not specified
 	MaxSchedulerDelay time.Duration
 
+	// TimeWindow restricts when a new periodic GC sweep over all repositories may
+	// start to a daily time-of-day window, e.g. "01:00-08:00". A sweep that starts
+	// inside the window is allowed to run to completion even past the window's end,
+	// so GC work stays amortized and orphaned blobs don't outpace a narrow window.
+	// Empty means no restriction.
+	TimeWindow string
```
andaaron: Once config decodes to a typed window, pass it through Options directly instead of re-parsing the string in newGCTimeWindow (:124-137).

pkg/storage/gc/gc.go
```diff
@@ -43,9 +44,99 @@ type Options struct {
...
+	if window == "" {
+		return nil
+	}
+
+	parsed, err := parseGCTimeWindow(window)
+	if err != nil {
+		log.Error().Err(err).Str("gcTimeWindow", window).
+			Msg("invalid gcTimeWindow, ignoring and running GC without a time restriction")
+
+		return nil
```
andaaron: After config-load validation, newGCTimeWindow's "log and ignore invalid window" path is likely unreachable in normal operation.

pkg/storage/gc/gc.go
```diff
@@ -1151,7 +1244,23 @@ func (gen *GCTaskGenerator) IsDone() bool {
...
+	}
+
+	// Only gate the start of a new sweep on the configured time window. Once a sweep
+	// has begun (processedRepos is non-empty), let it run to completion regardless of
+	// the window, so GC work stays amortized instead of stalling mid-sweep until the
+	// window reopens the next day.
+	startingNewSweep := len(gen.processedRepos) == 0 && gen.nextRun.IsZero()
+
+	if startingNewSweep && gen.timeWindow != nil && !gen.timeWindow.contains(now) {
+		return false
```
andaaron, line 1184: When returning false because we're outside the window, consider a Debug log ("GC sweep deferred, outside gcTimeWindow") so operators can tell why periodic GC hasn't started.

pkg/cli/server/root.go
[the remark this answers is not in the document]
shcherbak: agree

andaaron:
@shcherbak I forgot to ask: what timezone is this supposed to be in? I think it would be very ambiguous for the user as well

pkg/api/config/config.go
shcherbak: Good catch - agreed the double-parse isn't ideal. Only wrinkle: gc already imports config (for ImageRetention), so a gc.TimeWindow type would create an import cycle. Fix would be defining the typed window in config instead (like ImageRetention already is), decoded via a custom mapstructure hook next to StringToTimeDurationHookFunc, with gc.Options.TimeWindow taking that type directly and parseGCTimeWindow/newGCTimeWindow/ValidateGCTimeWindow going away. That's a decent-sized change for this PR's scope. Do we want it here, or as a follow-up?

pkg/storage/gc/gc.go
shcherbak: Agree - with the current call graph it's unreachable, since validateGC already rejects the config before the server ever calls this. Kept it as a safety net rather than removing it: gc is a library package with an exported Options.TimeWindow string, and if a direct caller skips validation, silently trusting the parse would leave a zero-value window (startMin=0, endMin=0), which makes contains() always false - GC would stop running permanently and silently, which is worse than the current logged fallback. Added a comment explaining why it's there.

pkg/storage/gc/gc.go
shcherbak: Same suggestion as the one on config.go:39-41

shcherbak:
> @shcherbak I forgot to ask: what timezone is this supposed to be in? I think it would be very ambiguous for the user as well

Fixed: the window is now explicitly evaluated in UTC. Docs and comments updated to say UTC explicitly.

thees:
> The downside of this approach is that the gc work is no longer amortized but has to finish within that time window - which it may not.

I created this fix to speed up gc massively: https://github.com/project-zot/zot/pull/4236

pkg/cli/server/root.go
andaaron: This comment is still valid.

pkg/api/config/config.go
andaaron: What I meant is defining a GCTimeWindow struct in this package (e.g. in config.go), with a custom mapstructure decode hook in the same package — similar in spirit to SinkConfigDecoderHook in pkg/extensions/config/events/decoder.go.

Register that hook in root.go next to StringToTimeDurationHookFunc(). Move parsing/validation (currently in parseGCTimeWindow, Contains, empty string = no restriction) into pkg/api/config; invalid values would then fail at unmarshal time.

gc.Options.TimeWindow would take config.GCTimeWindow directly (like it already takes config.ImageRetention), so pkg/storage/gc would not re-parse or validate — it would only consume the already-decoded value.

This should be cleaner and require no extra packages to import in root.go or gc.go

pkg/storage/gc/gc.go
andaaron: See my other comment in config.go.

rchincha:
@shcherbak pls take care of the review comments and ci failures.

We can include this PR in the next release.

andaaron:
> What I meant is defining a GCTimeWindow struct in this package (e.g. in config.go), with a custom mapstructure decode hook in the same package — similar in spirit to SinkConfigDecoderHook in pkg/extensions/config/events/decoder.go.
> 
> Register that hook in root.go next to StringToTimeDurationHookFunc(). Move parsing/validation (currently in parseGCTimeWindow, Contains, empty string = no restriction) into pkg/api/config; invalid values would then fail at unmarshal time.
> 
> gc.Options.TimeWindow would take config.GCTimeWindow directly (like it already takes config.ImageRetention), so pkg/storage/gc would not re-parse or validate — it would only consume the already-decoded value.
> 
> This should be cleaner and require no extra packages to import in root.go or gc.go

Please look into this

pkg/storage/gc/gc.go
```diff
@@ -43,9 +44,103 @@ type Options struct {
...
+	parsed, err := time.Parse("15:04", value)
+	if err != nil {
+		return 0, fmt.Errorf("%w: invalid time %q, expected 24h format \"HH:MM\"", zerr.ErrBadConfig, value)
+	}
+
+	return parsed.Hour()*minutesInHour + parsed.Minute(), nil
+}
+
+// ValidateGCTimeWindow returns an error if window is non-empty and not a valid
+// "HH:MM-HH:MM" daily time-of-day range. An empty window is valid and means no restriction.
```
rchincha: 00:00-00:00?

examples/config-gc.json
```diff
@@ -4,7 +4,8 @@
         "rootDirectory": "/tmp/zot",
         "gc": true,
         "gcDelay": "2h",
-        "gcInterval": "1h"
+        "gcInterval": "1h",
+        "gcTimeWindow": "01:00-08:00"
```
rchincha, line 8: Not including this value means it defaults to a 24h window?

pkg/storage/gc/gc.go
shcherbak: "00:00-00:00" is rejected, not treated as "whole day" - start==end is ambiguous, and "no restriction" already has an unambiguous form (empty string). Unchanged from the pre-refactor behavior, just relocated into config.GCTimeWindow.

examples/config-gc.json
shcherbak: Omitting it means no restriction, not a literal 24h window - and a literal 24h window isn't even expressible in HH:MM-HH:MM format anyway. gcTimeWindow stays needed for restricting to less than a full day.

shcherbak:
> > What I meant is defining a GCTimeWindow struct in this package (e.g. in config.go), with a custom mapstructure decode hook in the same package — similar in spirit to SinkConfigDecoderHook in pkg/extensions/config/events/decoder.go.
> > Register that hook in root.go next to StringToTimeDurationHookFunc(). Move parsing/validation (currently in parseGCTimeWindow, Contains, empty string = no restriction) into pkg/api/config; invalid values would then fail at unmarshal time.
> > gc.Options.TimeWindow would take config.GCTimeWindow directly (like it already takes config.ImageRetention), so pkg/storage/gc would not re-parse or validate — it would only consume the already-decoded value.
> > This should be cleaner and require no extra packages to import in root.go or gc.go
> 
> Please look into this

this needs deeper refactoring, ok

pkg/api/config/config.go
andaaron, line 839: Shouldn't this also look at the GC time window?

andaaron (commented):
Looks good. I left one question

pkg/api/config/config.go
andaaron: Only used in tests.

andaaron (dismissed):

rchincha:
| Merging is blocked

`git commit --amend -S -s` for all your commits

pkg/api/config/config.go
shcherbak: Agree

rchincha:
> | Merging is blocked Commits must have verified signatures.
> 
> `git commit --amend -S -s` for all your commits

Your PR has multiple commits, so `git rebase -i HEAD~5 --exec "git commit --amend -S -s --no-edit"` for "last 5" commits for example

rchincha:
| This user has not yet uploaded their public signing key.

@shcherbak pls also upload your public signing key to your GH profile - a one-time thing that will be useful for the future as well.

andaaron (approved):

PR #4227, opened 2026-07-13
fix(storage): skip reserved dirs in GetNextRepository walk

### LINKED ISSUE #4226: [Bug]: GC silently stops enumerating repositories when a ghost/empty .uploads prefix aborts the storage walk (GCS)
### zot version

2.1.18

### Describe the bug

1. `GarbageCollect`'s task generator drives a GC run by repeatedly calling
   `ImageStore.GetNextRepository(processedRepos)` until it returns `""`,
   which is interpreted as "no more repositories" → `done`
   (`pkg/storage/gc/gc.go:1124-1147`).
2. `GetNextRepository` walks the storage root
   (`pkg/storage/imagestore/imagestore.go:417-481`). Unlike its sibling
   walkers `GetRepositories` (`imagestore.go:308-313`) and the paginated
   lister (`imagestore.go:384-389`), its walk callback is **missing the
   `driver.ErrSkipDir` guard** for the reserved `.sync`/`blobs`/`.uploads`
   sub-directories — so the walk descends into every already-processed
   repo's full subtree, including `<repo>/.uploads/`.
3. On the GCS backend, listing a prefix with no visible objects returns
   `PathNotFoundError` ("Treat empty response as missing directory",
   distribution `registry/storage/driver/gcs/gcs.go:700-702`). An abandoned
   upload session under `<repo>/.uploads/` (or a prefix concurrently emptied
   by GC) is exactly such a ghost prefix.
4. `storagedriver.WalkFallback` aborts the entire walk when a nested `List`
   returns an error (distribution `registry/storage/driver/walk.go:83-86`);
   it only tolerates a nested `Stat` PathNotFound (`walk.go:101-110`).
5. `GetNextRepository` maps **any** `PathNotFoundError` bubbling out of the
   walk to "empty rootDir" and returns `""`
   (`pkg/storage/imagestore/imagestore.go:469-473`) — GC marks itself done
   and all remaining repositories are silently skipped.

### To reproduce

1. Configuration: zot with the GCS storage driver and periodic GC enabled
   (`storage.gc: true`); a bucket with several repositories, where one
   repository that sorts alphabetically early contains a ghost/empty
   `.uploads/` prefix (e.g. left behind by an abandoned/interrupted push
   whose upload session objects were cleaned, or a prefix emptied
   concurrently by GC).
2. Client tool used: any (bug is server-side; trigger is the leftover
   `.uploads` state, e.g. from an interrupted `oras cp` / `docker push`).
3. Seen error: none - that is the problem. The GC run logs completion after
   the first few repositories; repositories sorting after the poisoned one
   never appear in the GC logs and their untagged/expired content is never
   deleted, while `GetNextRepository`'s walk received a
   `PathNotFoundError` that was swallowed as "no more repositories".

### Expected behavior

A ghost/empty `.uploads` (or `blobs`/`.sync`) prefix inside an
already-processed repository must not terminate repository enumeration.

GC must visit every repository under the root, on every storage backend.

### Screenshots

_No response_

### Additional context

- Production impact observed on GKE with a GCS bucket: GC processed only the
  first 4 repositories alphabetically for weeks; bucket grew to ~1.5 TB.

- The bug is invisible in logs because the `PathNotFoundError` is translated
  into the normal "no more repositories" termination path - no error or
  warning is emitted.

thees:
GetNextRepository descended into every already-processed repo's subtree, including <repo>/.uploads/. On backends where listing an empty prefix returns PathNotFoundError (e.g. GCS: a ghost prefix left by an abandoned upload session), the nested List error aborted the whole walk and was mapped to "no more repositories", so periodic GC silently stopped and never collected any repository sorting after the poisoned one.

Add the same ErrSkipDir guard for .sync/blobs/.uploads that the sibling walkers (GetRepositories, GetRepositoriesWithPagination) already use, so enumeration never descends into or lists reserved directories. Covered by a regression test that exercises the real storagedriver.WalkFallback path used by the GCS driver.

Fixes: https://github.com/project-zot/zot/issues/4226

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/gcs/nextrepo_walk_abort_test.go
+++ b/pkg/storage/gcs/nextrepo_walk_abort_test.go
@@ -1,0 +1,234 @@
+package gcs_test
+
+// Regression test for a GC bug that was observed in production against a GCS
+// bucket: the GC task generator stopped after the first few repositories and
+// silently skipped all remaining ones.
+//
+// Chain that was affected (all REAL code):
+//   ImageStore.GetNextRepository (pkg/storage/imagestore/imagestore.go)
+//     -> zot gcs.Driver.Walk (pkg/storage/gcs/driver.go)
+//       -> distribution storagedriver.WalkFallback (v3.1.1 walk.go)
+//         -> StorageDriver.List / Stat
+//
+// The distribution GCS driver returned PathNotFoundError whenever a prefix
+// listed empty ("Treat empty response as missing directory",
+// registry/storage/driver/gcs/gcs.go). That happened for prefixes whose only
+// objects were filtered out client-side (in-flight upload sessions in
+// .uploads/) or whose objects were deleted concurrently by GC between the
+// parent List and the child List. doWalkFallback propagated that nested List
+// error up, and GetNextRepository mapped ANY PathNotFoundError from the walk to
+// "empty rootDir" -> returned "" -> GCTaskGenerator set done. Result: every
+// repository sorting after the poisoned one was never garbage collected.
+//
+// The bug is now fixed by the .uploads/blobs/.sync ErrSkipDir guard in
+// GetNextRepository (pkg/storage/imagestore/imagestore.go): those reserved
+// sub-directories are never descended into, so a ghost/empty prefix can no
+// longer abort repository enumeration. This test guards against a regression by
+// asserting that ALL repositories are still enumerated despite a ghost .uploads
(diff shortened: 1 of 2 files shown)


andaaron (approved):

rchincha (approved):
lgtm

PR #4232, opened 2026-07-13
ci: relax dco check for dependabot alerts

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/.github/workflows/dco.yaml
+++ b/.github/workflows/dco.yaml
@@ -8,6 +8,7 @@ permissions: read-all
 
 jobs:
   check:
+    if: github.event.pull_request.user.login != 'dependabot[bot]'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0


rchincha:
conflicting copilot reviews

.github/workflows/dco.yaml
vrajashkr, line 26: I think we can use the dco-check tool's CLI args/ environment variables instead.
https://github.com/christophebedard/dco-check/blob/master/dco_check/dco_check.py#L172

```
 help=(
            'exclude regular expresssion matched author emails from checks '
            '(commits with an author email matching regular expression pattern will be ignored)'
        ),
```

vrajashkr: Exclude emails might be better than pattern matching perhaps as we know the exact email that the bot uses:
https://github.com/christophebedard/dco-check/blob/master/dco_check/dco_check.py#L163

```
parser.add_argument(
        '-e', '--exclude-emails', metavar='EMAIL[,EMAIL]',
        action=EnvDefaultOption, env_var=ENV_VAR_EXCLUDE_EMAILS,
        default=None,
        help=(
            'exclude a comma-separated list of author emails from checks '
            '(commits with an author email matching one of these emails will be ignored)'
        ),
    )
```

rchincha: can a author sneak this through with bogus values?

andaaron (approved):
I think this is enough.
For authenticity we already have the GPG signature.

.github/workflows/dco.yaml
vrajashkr: Yeah, that's a good point. IMO, we can stick with the current approach of the PR author which can't be messed with.

vrajashkr (approved):

PR #4234, opened 2026-07-14
fix(meta): derive push timestamp from blob mod time on storage parse

thees:
The push timestamp only lives in the metaDB. When the metaDB is lost (e.g. a non-persistent cache backend restarts) the storage parse recreated every statistics entry with time.Now(), stamping all images with the startup time. Retention policies based on push time (pushedWithin, mostRecentlyPushedCount) then saw one identical fake timestamp and expired every tag at the same moment.

Pass the manifest blob's storage mod time as a hint through ImageMeta and prefer it over time.Now() when a new statistics entry is created. On object storage blobs are immutable, so the mod time is the original upload time and a metaDB rebuild reconstructs the real push dates.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/meta/parse.go
+++ b/pkg/meta/parse.go
@@ -535,6 +535,13 @@ func SetImageMetaFromInput(ctx context.Context, repo, reference, mediaType strin
 		return nil
 	}
 
+	// The push time only lives in the metaDB; when it is lost, the manifest blob's storage
+	// mod time is the closest remaining record of when the image was pushed. Pass it as a
+	// hint so a metaDB rebuild doesn't stamp every image with the current time.
+	if found, _, modTime, err := imageStore.StatBlob(repo, digest); err == nil && found {
+		imageMeta.PushTimestamp = modTime
+	}
+
 	err := metaDB.SetRepoReference(ctx, repo, reference, imageMeta)
 	if err != nil {
 		log.Error().Err(err).Str("component", "metadb").Msg("failed to set repo meta")
--- a/pkg/meta/parse.go
+++ b/pkg/meta/parse.go
@@ -538,11 +538,12 @@ func SetImageMetaFromInput(ctx context.Context, repo, reference, mediaType strin
 	// The push time only lives in the metaDB; when it is lost, the manifest blob's storage
 	// mod time is the closest remaining record of when the image was pushed. Pass it as a
 	// hint so a metaDB rebuild doesn't stamp every image with the current time.
+	setOpts := []mTypes.SetRepoReferenceOption{}
 	if found, _, modTime, err := imageStore.StatBlob(repo, digest); err == nil && found {
-		imageMeta.PushTimestamp = modTime
+		setOpts = append(setOpts, mTypes.WithPushTimestamp(modTime))
 	}
 
-	err := metaDB.SetRepoReference(ctx, repo, reference, imageMeta)
+	err := metaDB.SetRepoReference(ctx, repo, reference, imageMeta, setOpts...)
 	if err != nil {
 		log.Error().Err(err).Str("component", "metadb").Msg("failed to set repo meta")
 
--- a/pkg/meta/parse.go
+++ b/pkg/meta/parse.go
@@ -294,8 +294,17 @@ func parseRepo(repo string, metaDB mTypes.MetaDB, storeController stypes.StoreCo
 			reference = manifest.Digest.String()
 		}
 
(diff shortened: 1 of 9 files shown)


pkg/meta/types/types.go
```diff
@@ -241,6 +241,10 @@ type ImageMeta struct {
 	Size      int64           // Size refers to the image descriptor, a manifest or a index (if multiarch)
 	Index     *ispec.Index    // If the image is multiarch the Index will be non-nil
 	Manifests []ManifestMeta  // All manifests under the image, 1 for simple images and many for multiarch
+	// PushTimestamp is an optional hint for when the image was pushed, used only when a new statistics
+	// entry is created (e.g. the manifest blob storage mod time on a storage re-parse). Zero means
+	// unknown and the metaDB falls back to the current time.
+	PushTimestamp time.Time
```
andaaron, line 252: Note ImageMeta is common for all instances of this manifest, not a specific instance in a specific repo with a specific tag.

That is why we have a DescriptorStatistics with PushTimestamp which is per repo and digest. Why not use that one?

thees: ImageMeta.PushTimestamp is only a transient transport hint and is never persisted as manifest-level data. 

The value ends up in the per-repo DescriptorStatistics.PushTimestamp as before.
It rides on ImageMeta because SetRepoReference creates the statistics entry internally and offers no other channel for a per-call hint without changing the MetaDB interface.

Alternatively, the hint could be made explicit by extending the signature to SetRepoReference(ctx, repo, reference, imageMeta, pushTimestamp) (zero value = unknown → time.Now()).
That touches the interface, all backends (boltdb/dynamodb/redis), the mocks, and one production call site.

I decided this way, because the alternative would have a much bigger impact on code and I wanted it to be surgical.

thees: @andaaron have a look and the second commit. claude was so nice to do the mess for me ;-) however, I think this change is more that it should be if you ask me.

pkg/meta/parse.go
```diff
@@ -535,6 +535,13 @@ func SetImageMetaFromInput(ctx context.Context, repo, reference, mediaType strin
 		return nil
 	}
 
+	// The push time only lives in the metaDB; when it is lost, the manifest blob's storage
+	// mod time is the closest remaining record of when the image was pushed. Pass it as a
+	// hint so a metaDB rebuild doesn't stamp every image with the current time.
+	if found, _, modTime, err := imageStore.StatBlob(repo, digest); err == nil && found {
+		imageMeta.PushTimestamp = modTime
+	}
```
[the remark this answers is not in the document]
thees: ok this is correct.

However, wrapping the call in RLock/RUnlock inside SetImageMetaFromInput would not be safe either: the storage-parse path (parseRepo) already holds the store RLock, and recursively read-locking a sync.RWMutex can deadlock once a writer is queued between the two acquisitions.

Fixed instead by moving the StatBlob call up into parseRepo, which already holds the RLock and passing the mod time through SetImageMetaFromInput as a SetRepoReference option. As a side effect the push and sync paths no longer stat the blob at all.
the hint is only gathered on a storage re-parse, which is the only place it matters; regular pushes fall back to time.Now() as before.

pkg/meta/parse_test.go
```diff
@@ -436,6 +436,37 @@ func TestParseStorageDynamoWrapper(t *testing.T) {
...
+		image := CreateRandomImage() //nolint:staticcheck
+
+		err := WriteImageToFileSystem(image, repo, "tag1", storeController)
+		So(err, ShouldBeNil)
+
+		// Backdate the manifest blob: after a metaDB loss the real push time is gone,
+		// so the storage mod time is the only remaining record of when it was pushed.
+		manifestDigest := image.ManifestDescriptor.Digest
+		blobPath := path.Join(rootDir, repo, "blobs", manifestDigest.Algorithm().String(), manifestDigest.Encoded())
+		pushedAt := time.Now().Add(-72 * time.Hour)
```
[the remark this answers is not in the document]
thees: test now resolves the path via imageStore.BlobPath(repo, manifestDigest) instead of hard-coding the on-disk layout, consistent with how other tests in the repo do it.

pkg/meta/types/types.go
andaaron: Thanks

andaaron (approved):

Issue #4235, opened 2026-07-15
[Bug]: GC CleanupRepo re-reads all manifests per deleted blob (quadratic on remote storage)


thees:
### zot version

2.1.18

### Describe the bug

`CleanupRepo` checks every candidate blob against the repo index by re-reading **all** manifests from storage once per blob (`IsBlobReferenced` performs a full manifest traversal per digest). This makes a single GC run cost **O(deleted blobs × manifests)** storage reads.

On remote storage (GCS/S3) with many tags this is catastrophic:

- In our example, a repo with **519 manifests** took **~20s per deleted blob** (~500 sequential manifest GETs each).
- The repo lock is held for the entire run, so a single cleanup can stretch into a **multi-day** operation and **blocks all pushes** to that repo in the meantime.

### To reproduce

1. **Configuration**: zot with a remote storage backend (GCS or S3) and GC enabled.
2. **Client tool used**: push many tags/manifests into a single repo (e.g. ~500 manifests), then delete enough tags to leave a large batch of orphan blobs to be collected.
3. **Seen error**: GC (`CleanupRepo`) crawls at tens of seconds per deleted blob; the repo lock stays held for hours, and concurrent pushes to that repo block until GC finishes.

### Expected behavior

The referenced-blob set should be computed **once per repo** per cleanup run (refreshed only when a manifest deletion actually rewrites `index.json`), so GC cost is roughly **O(manifests + deleted blobs)** instead of quadratic, and the repo lock is released in a timely manner.

### Additional context

Proposed fix:

- Add `common.GetReferencedBlobs`, which walks `index.json` and reads each (possibly nested) manifest **exactly once** to build the full set of referenced digests (manifest, config, and layer digests). It mirrors `IsBlobReferencedInImageIndex`'s traversal and skips unreadable/missing manifests identically.
- Rework `CleanupRepo` to build that set once and reuse it for every blob, refreshing it only after a manifest deletion rewrites `index.json` (blobs referenced solely by the deleted manifest then become collectible).
- Introduce `deleteBlobChecked(repo, digest, isReferenced func() bool)` so GC can pass the precomputed set, while the `DeleteBlob` API path keeps the per-blob `IsBlobReferenced` check unchanged.

Behavior preserved: same referencing semantics and same handling of unreadable manifests; only the read pattern changes.

Tests added:

[...]

PR #4236, opened 2026-07-15
fix(gc): collect referenced blobs once per repo during cleanup

thees:
CleanupRepo checked every blob against the repo index by re-reading all manifests from storage (IsBlobReferenced per blob), making one GC run cost O(deleted blobs x manifests) storage reads. On remote storage (GCS/S3) with many tags this throttles GC to tens of seconds per deleted blob and keeps the repo lock held for hours, blocking pushes: a repo with 519 manifests took ~20s per blob (~500 sequential manifest GETs each), turning a single cleanup into a multi-day run.

Build the referenced-blob set once per CleanupRepo call and refresh it only after a manifest deletion rewrites index.json. The per-blob check is preserved for the DeleteBlob API path.

**What type of PR is this?**

https://github.com/project-zot/zot/issues/4235

**Automation added to e2e**:

```release-note

```


pkg/storage/imagestore/imagestore.go
```diff
@@ -1961,6 +1973,16 @@ func (is *ImageStore) CleanupRepo(repo string, blobs []godigest.Digest, removeRe
 			}
 
 			count++
+
+			// the manifest deletion above rewrote index.json, so blobs referenced
+			// only by that manifest are now deletable: refresh the set
+			referenced, err = common.GetReferencedBlobs(is, repo, is.log)
```
rchincha: How large can this grow? We also have a requirement that we run on resource constrained devices so cpu and memory tradeoffs.

pkg/storage/common/common.go
```diff
@@ -619,6 +619,69 @@ func IsBlobReferenced(imgStore storageTypes.ImageStore, repo string,
...
+
+	return referenced, nil
+}
+
+func collectReferencedBlobs(imgStore storageTypes.ImageStore, repo string,
+	index ispec.Index, referenced map[godigest.Digest]struct{}, log zlog.Logger,
+) {
+	for _, desc := range index.Manifests {
+		referenced[desc.Digest] = struct{}{}
+
```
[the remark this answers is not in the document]
andaaron: Valid point.

pkg/storage/imagestore/imagestore.go
andaaron: Repo shape | Unique digests | Approx. memory
-- | -- | --
519 tags, ~10 layers each, heavy sharing | ~5k–15k | ~300 KB–1 MB
10k tags, same manifest | ~12 (1 manifest + config + layers) | < 1 KB

pkg/storage/imagestore/imagestore.go
```diff
@@ -1937,11 +1937,23 @@ the caller function MUST lock from outside.
 func (is *ImageStore) CleanupRepo(repo string, blobs []godigest.Digest, removeRepo bool) (int, error) {
```
andaaron, line 1942: I think the fix for #4235 belongs in the GC package, not in `CleanupRepo`.

Today:
- `cleanRepo` updates `index.json` first (`gc.go:175-187`)
- `removeUnreferencedBlobs` builds `refBlobs`, filters `gcBlobs`, then calls `CleanupRepo` (`gc.go:892-948`)
- `CleanupRepo` re-checks every blob (`IsBlobReferenced` today; `GetReferencedBlobs` in this PR)

That duplicates work and keeps blob-GC policy split across `gc` and `imagestore`.

**Proposal (GC-owned cleanup):**
1. In `removeUnreferencedBlobs`, replace the shallow `addIndexBlobsToReferences` pass with one **complete** reference collection (this PR's `GetReferencedBlobs`, or fix `addImageIndexBlobsToReferences` to recurse into platform manifests).
2. Filter `gcBlobs` from that set (unchanged).
3. Delete blobs in GC (loop currently inside `CleanupRepo`) — stat, dedupe/cache checks, delete — using the precomputed map for reference checks.
4. Leave `CleanupRepo` as a thin helper for non-GC callers/tests, or remove the reference-check + `deleteImageManifest` fallback from the GC path entirely.

This fixes B×T without a second manifest traversal inside `imagestore`, and keeps "what is referenced?" and "delete orphans" in one place.

Note: `addImageIndexBlobsToReferences` (`gc.go:1010-1012`) only records platform manifest digests today, not config/layers — any unified collector must be recursive like `GetReferencedBlobs`, not a reuse of today's GC helpers as-is.

pkg/storage/common/common.go
```diff
@@ -619,6 +619,69 @@ func IsBlobReferenced(imgStore storageTypes.ImageStore, repo string,
...
+		return nil, err
+	}
+
+	referenced := map[godigest.Digest]struct{}{}
+	collectReferencedBlobs(imgStore, repo, index, referenced, log)
+
+	return referenced, nil
+}
+
+func collectReferencedBlobs(imgStore storageTypes.ImageStore, repo string,
```
andaaron, line 671: This looks a lot like gc,addIndexBlobsToReferences(). Is the intention to move some of that over to storage package?

pkg/storage/imagestore/imagestore.go
andaaron: Under a normal worflow the blobs never includes manifests which were not previously removed from the index.json.

Someone added this defensive check `errors.Is(err, zerr.ErrBlobReferenced):` in case the function was improperly called before the blob (in this case a manifest) was removed to the index. So `is.deleteImageManifest` should be called above.

I may be a relic from a time when GC was running immediately after a manifest was deleted, which is no longer the case.

So now we need this call here to accomodate for a case which is not part of normal zot call flow.

pkg/storage/imagestore/imagestore.go
```diff
@@ -1937,11 +1937,23 @@ the caller function MUST lock from outside.
...
 	count := 0
 
+	// Collect the referenced blobs once per repo instead of re-reading every
+	// manifest for each deleted blob (quadratic on remote storage).
+	referenced, err := common.GetReferencedBlobs(is, repo, is.log)
```
andaaron: If GC owns cleanup, this call should not live in `CleanupRepo`. Build the referenced set once in `removeUnreferencedBlobs` and delete blobs there. `CleanupRepo` should not re-walk `index.json` on every GC run.

pkg/storage/imagestore/imagestore.go
andaaron: This branch tries `deleteImageManifest` when blob delete is blocked because the digest is still referenced. On designed GC flow, indexed manifests should already be removed in `cleanRepo`; this is defensive/legacy behavior in a low-level API.

If we move orphan blob deletion into `gc.removeUnreferencedBlobs`, GC should not rely on this path. Consider whether `CleanupRepo` still needs manifest-index mutation at all, or only direct blob deletes for non-GC callers.

andaaron (commented):
I think the change belongs in `pkg/storage/gc`, not `CleanupRepo`.

Please:
- collect referenced digests once in gc package, in `removeUnreferencedBlobs` (complete traversal, fix multi-arch gap)
- delete orphan blobs there using that map
- avoid a second `GetReferencedBlobs` in `imagestore`
- drop refresh / `deleteImageManifest` fallback from the GC path

This issue is real, this PR is an improvement, and we can make it even better.

pkg/storage/common/common.go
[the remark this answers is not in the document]
thees: has been addressed with 1f08e76c0d2b1275aadbed6e2bfb6300f8e5d662

pkg/storage/common/common.go
```diff
@@ -619,6 +619,69 @@ func IsBlobReferenced(imgStore storageTypes.ImageStore, repo string,
...
+			}
+
+			collectReferencedBlobs(imgStore, repo, indexImage, referenced, log)
+		case ispec.MediaTypeImageManifest:
+			manifestContent, err := GetImageManifest(imgStore, repo, desc.Digest, log)
+			if err != nil {
+				log.Warn().Err(err).Str("repository", repo).Str("digest", desc.Digest.String()).
+					Msg("skipping unreadable manifest blob while collecting referenced blobs")
+
+				continue
```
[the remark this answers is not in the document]
thees: has been addressed with 1f08e76c0d2b1275aadbed6e2bfb6300f8e5d662

pkg/storage/imagestore/imagestore_test.go
```diff
@@ -254,3 +255,155 @@ func TestCleanupRepoFailsOnDeleteImageManifest(t *testing.T) {
...
+			manifestBodies[store.BlobPath(repo, manifestDigest)] = body
+			index.Manifests = append(index.Manifests, ispec.Descriptor{
+				MediaType: ispec.MediaTypeImageManifest,
+				Digest:    manifestDigest,
+				Size:      int64(len(body)),
+			})
+		}
+
+		indexBody, err := json.Marshal(index)
+		So(err, ShouldBeNil)
```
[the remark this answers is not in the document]
thees: has been addressed with 1f08e76c0d2b1275aadbed6e2bfb6300f8e5d662

pkg/storage/imagestore/imagestore.go
thees: I'll see if I can change the implementation towards your proposal

pkg/storage/imagestore/imagestore.go
thees: Agreed with @andaaron's estimate. The redesign in 443ce893 also improves this: the referenced set is now built once in the GC package (removeUnreferencedBlobs) instead of twice (previously gc's refBlobs plus a second set inside CleanupRepo). It's a single map[godigest.Digest]struct{} of the repo's unique referenced digests, held only during one repo's GC run (GC processes one repo at a time under the repo lock) and freed afterward. Worst case matches the table (~hundreds of KB–1 MB for a large, heavily-shared repo; <1 KB when many tags share one manifest, thanks to the dedup). Happy to cap/stream it if you consider that transient footprint too high for constrained devices.

pkg/storage/imagestore/imagestore.go
```diff
@@ -1937,11 +1937,23 @@ the caller function MUST lock from outside.
...
+	if err != nil {
+		is.log.Warn().Err(err).Str("repository", repo).
+			Msg("failed to collect referenced blobs, treating all blobs as unreferenced")
+	}
```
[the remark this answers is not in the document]
thees: Initially addressed in 1f08e76c with a per-blob IsBlobReferenced fallback. With the GC-owned redesign in 443ce893 the concern no longer applies on this path: the referenced set is built in gc.removeUnreferencedBlobs, and a build failure returns the error and aborts the run before any deletion. CleanupRepo no longer builds a set or deletes against a nil set, and GC dry-run is now fully non-destructive.

pkg/storage/imagestore/imagestore.go
```diff
@@ -1961,6 +1973,16 @@ func (is *ImageStore) CleanupRepo(repo string, blobs []godigest.Digest, removeRe
 			}
 
 			count++
+
+			// the manifest deletion above rewrote index.json, so blobs referenced
+			// only by that manifest are now deletable: refresh the set
+			referenced, err = common.GetReferencedBlobs(is, repo, is.log)
+			if err != nil {
+				is.log.Warn().Err(err).Str("repository", repo).
```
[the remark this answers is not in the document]
thees: The post-manifest-delete refresh (and its fallback) was removed entirely in 443ce893. CleanupRepo no longer re-walks index.json or refreshes a set mid-run; GC computes the orphan set once under the repo lock, so this path no longer exists.

pkg/storage/imagestore/imagestore.go
thees: Done in 443ce893: reference collection + orphan filtering moved into gc.removeUnreferencedBlobs (single complete common.GetReferencedBlobs pass, multi-arch gap fixed), the second traversal and the deleteImageManifest/refresh fallback removed, and GC dry-run made non-destructive. CleanupRepo is now a thin unchecked-delete helper.

pkg/storage/common/common.go
thees: Yes, exactly. 443ce893 makes common.GetReferencedBlobs the single complete collector used by gc.removeUnreferencedBlobs, and removes the GC helper trio (addIndexBlobsToReferences / addImageIndexBlobsToReferences / addImageManifestBlobsToReferences).

This also closed a multi-arch gap: the GC helper only recorded nested  platform-manifest digests, not their config/layers. the common collector recurses and now also covers the docker-compat media types.

pkg/storage/imagestore/imagestore.go
thees: OK. removed from CleanupRepo in 443ce893. One caveat surfaced in testing (TestGarbageCollectImageUnknownManifest): index entries with an unknown/unsupported media type were previously pruned by exactly that fallback.
To preserve that without the fallback, cleanRepo now prunes unknown-media-type descriptors from index.json before blob GC (removeUnknownMediaTypeManifestEntries), so the manifest and its config/layers are collected together instead of orphaning the blobs behind a dangling entry.

pkg/storage/imagestore/imagestore.go
thees: Done in 443ce893: removeUnreferencedBlobs builds the referenced set once via common.GetReferencedBlobs; CleanupRepo no longer walks index.json. it just deletes the already-filtered orphan blobs GC passes it.

pkg/storage/imagestore/imagestore.go
thees: The reference-check + deleteImageManifest fallback is dropped from the GC path in 443ce893. 

The physical blob delete stays in imagestore (it needs unexported state,.  Cache/dedupe handling, storeDriver and the repo lock is non-reentrant, so GC can't call the public DeleteBlob without deadlocking), but GC now owns the policy: it computes the complete referenced set and the orphan list, and CleanupRepo deletes exactly those, unchecked.

pkg/storage/gc/gc.go
```diff
@@ -927,7 +975,7 @@ func (gc GarbageCollect) removeUnreferencedBlobs(repo string, delay time.Duratio
 			return 0, err
 		}
 
-		if _, ok := refBlobs[digest.String()]; !ok {
+		if _, ok := refBlobs[digest]; !ok {
```
andaaron, line 986: Why was this needed?

pkg/storage/gc/gc.go
```diff
@@ -183,16 +193,23 @@ func (gc GarbageCollect) cleanRepo(ctx context.Context, repo string) error {
...
-	}
+	var blobsDeleted, uploadsDeleted int
 
-	// gc old blob uploads
-	uploadsDeleted, err := gc.removeBlobUploads(repo, gc.opts.Delay)
-	if err != nil {
-		return err
+	// DryRun must be a non-destructive simulation: index.json was left untouched above, and blob GC
+	// must not delete anything either - it re-reads the on-disk index.json, so running it here would
+	// delete blobs for real (e.g. orphaned by manifest-pass edits) while their index entries survive.
```
andaaron, line 206: Nome notes:
- dryRun is a retention feature (to observe what manifests would be deleted based on a specific retention configuration), not a GC feature. It is configured under retention, which has separate config settings from GC.
- from GC perspective garbage is garbage, it's not something which needs a config review or can be debated with someone else looking at the zot config, so there is no need for a dry-run.

There is one case where I disagree with the above, which is the zot command line to validate retention policies, it also has the side effect of deleting the garbage, which I don't like (but I think we have this documented).
Will probably need to update some tests there if we are to keep this change in the PR.

So I don't have a strong position for or against. @rchincha, any thoughts on this?

andaaron (commented):
Overall PR is in good shape.
Probably needs some test fixes for the dry-run behavior change besides the lint fixes.

pkg/storage/imagestore/imagestore.go
andaaron: Thanks

pkg/storage/imagestore/imagestore.go
andaaron: Makes sense.

pkg/storage/imagestore/imagestore.go
andaaron: Thanks, much cleaner now.

pkg/storage/gc/gc.go
thees: It's just no longer needed.

refBlobs is now the map[godigest.Digest]struct{} returned by common.GetReferencedBlobs (it replaced the locally-built  map[string]bool). 

he key is already a godigest.Digest, so the lookup no longer needs .String(). Pure consequence of reusing the shared collector.

pkg/storage/gc/gc.go
thees: I only gated blob-GC under dry-run to avoid a corruption I hit otherwise: with the unknown-media-type prune skipped under dry-run but blob-GC still running, the unknown manifest's config/layers were deleted for real while its index entry survived (dangling entry). If we prefer "GC always collects garbage, dry-run only concerns retention," I can drop the GC-side gating — but then the unknown-media-type prune must run (and persist) even under dry-run to keep the index consistent.

Happy to go either way and update the retention-CLI tests accordingly

pkg/storage/common/common.go
```diff
@@ -619,6 +619,99 @@ func IsBlobReferenced(imgStore storageTypes.ImageStore, repo string,
 	return IsBlobReferencedInImageIndex(imgStore, repo, digest, index, log)
 }
 
+/*
+GetReferencedBlobs returns the set of all blob digests referenced by the repo's index.json:
+manifest digests themselves plus the config and layer digests of every (possibly nested) manifest.
+
+It mirrors the traversal of IsBlobReferencedInImageIndex, but reads each manifest only once,
+so callers checking many digests (e.g. GC) avoid re-reading every manifest per digest.
+Unreadable or missing manifests are skipped, matching IsBlobReferencedInImageIndex.
```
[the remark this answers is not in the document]
thees: fixed

pkg/storage/common/common.go
```diff
@@ -619,6 +619,99 @@ func IsBlobReferenced(imgStore storageTypes.ImageStore, repo string,
...
+
+		if _, ok := seen[desc.Digest]; ok {
+			continue
+		}
+
+		seen[desc.Digest] = struct{}{}
+
+		switch {
+		case desc.MediaType == ispec.MediaTypeImageIndex || compat.IsCompatibleManifestListMediaType(desc.MediaType):
+			indexImage, err := GetImageIndex(imgStore, repo, desc.Digest, log)
```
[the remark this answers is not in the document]
thees: Fixed in 70930c1f:
 IsBlobReferencedInImageIndex now routes Docker manifest-list/manifest media types through the same recurse/manifest handling as GetReferencedBlobs (reusing the shared IsImageIndexMediaType/IsImageManifestMediaType helpers), so DeleteBlob and GC agree on Docker-referenced blobs. Added TestIsBlobReferencedInImageIndexDockerCompat covering the nested config/layer case.

pkg/storage/gc/gc.go
thees: I updated the four retention-verify tests in 4eb91c4a to match the non-destructive dry-run behavior.

thees: also your mentioned linting fixes are in 944183f67e8100b7b75d4c0b987ba753c03fa7d3

pkg/storage/common/common.go
```diff
@@ -584,7 +584,7 @@ func IsBlobReferencedInImageIndex(imgStore storageTypes.ImageStore, repo string,
 			}
 
 			found, _ = IsBlobReferencedInImageIndex(imgStore, repo, digest, indexImage, log)
-		case ispec.MediaTypeImageManifest:
+		case IsImageManifestMediaType(desc.MediaType):
 			found, _ = isBlobReferencedInImageManifest(imgStore, repo, digest, desc.Digest, log)
```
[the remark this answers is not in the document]
thees: Fixed in 54d783ae.

IsBlobReferencedInImageIndex now propagates the errors it previously discarded - both from the recursive index call and from isBlobReferencedInImageManifest (the found, _ = sites). On any non-not-found error it returns (false, err), matching what collectReferencedBlobs/GetReferencedBlobs already do. The skip-on-missing semantics are unchanged: the callees still collapse zerr.ErrBlobNotFound / driver.PathNotFoundError to (false, nil), so a genuinely missing blob is still skipped rather than turned into a hard error.

While here I applied the same fix to the one other caller that discarded this error, the DeleteImageManifest index guard, so all three call sites now fail closed consistently.

Added TestIsBlobReferencedInImageIndexNestedRecursionError and TestIsBlobReferencedInImageIndexManifestReadError covering the nested (:586/:588) error paths - the existing TestIsBlobReferencedInImageIndexNonMissingError only exercised the direct read - plus a regression asserting the not-found path still returns (false, nil).

pkg/storage/imagestore/imagestore.go
```diff
@@ -1995,6 +1985,17 @@ func (is *ImageStore) CleanupRepo(repo string, blobs []godigest.Digest, removeRe
 }
 
 func (is *ImageStore) deleteBlob(repo string, digest godigest.Digest) error {
+	return is.deleteBlobChecked(repo, digest, func() bool {
+		ok, _ := common.IsBlobReferenced(is, repo, digest, is.log)
+
+		return ok
+	})
+}
```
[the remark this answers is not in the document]
thees: Fixed in 54d783ae.

deleteBlob no longer discards the common.IsBlobReferenced error. The shared deleteBlobChecked predicate was widened func() bool → func() (bool, error), and the reference check is now evaluated before the delete: on error it aborts and returns the error (fail closed) instead of defaulting ok to false and deleting with unknown reference status. The error surfaces through DeleteBlob and maps to the existing HTTP 500 handling in routes.go, so no route change was needed.

CleanupRepo's batch path is unaffected - it still passes an always-false predicate (func() (bool, error) { return false, nil }) since it deletes an already-filtered orphan set; GC behavior is unchanged.

Covered by a new TestDeleteBlobsInUse case asserting the blob is not deleted and the error is returned when the reference check fails.

pkg/storage/common/common.go
```diff
@@ -619,6 +627,110 @@ func IsBlobReferenced(imgStore storageTypes.ImageStore, repo string,
...
+			continue
+		}
+
+		seen[desc.Digest] = struct{}{}
+
+		switch {
+		case IsImageIndexMediaType(desc.MediaType):
+			indexImage, err := GetImageIndex(imgStore, repo, desc.Digest, log)
+			if err != nil {
+				var pathNotFoundErr driver.PathNotFoundError
```
andaaron, line 687: If you can, cover the error path with a test case.

andaaron (dismissed):
Looks good.

andaaron:
@rchincha 

# PR #4236 workflow comparison: current PR vs `main`

This document compares garbage-collection storage I/O and behavior for [PR #4236](https://github.com/project-zot/zot/pull/4236) (**head `1ee6bbd6`**, 2026-07-16) against current `main`.

**Short answer:** the PR moved orphan-blob policy into the GC package (Approach B). One full reference walk per repo replaces the old **shallow GC pass + quadratic `CleanupRepo` re-check**. That fixes [#4235](https://github.com/project-zot/zot/issues/4235) and also closes several correctness gaps (multi-arch nested blobs, Docker on DeleteBlob, DryRun, fail-closed deletes).

> Earlier drafts of this doc described an intermediate design where `CleanupRepo` itself called `GetReferencedBlobs` and refreshed after `deleteImageManifest`. That is **no longer** what the PR does.

---

## Snapshot

| | `main` | PR #4236 |
|--|--------|----------|
| Who decides “referenced”? | Shallow GC helpers + per-blob `IsBlobReferenced` in `CleanupRepo` | `common.GetReferencedBlobs` once in `gc.removeUnreferencedBlobs` |
| Dominant cost for B orphans, T manifests | **O(B × T)** manifest GETs in `CleanupRepo` | **O(U)** GETs once (U = unique manifest digests; visited set) |
| Multi-arch config/layers | GC shallow pass records platform digests only; nested config/layers rely on `CleanupRepo` / fail-open DeleteBlob | Collected by recursive `collectReferencedBlobs` |
| Docker schema2 on DeleteBlob | OCI-only switch in `IsBlobReferencedInImageIndex` | Shared `IsImage*MediaType` helpers include docker-compat |
| `CleanupRepo` | Re-checks every blob; `ErrBlobReferenced` → `deleteImageManifest` | Unconditional delete of pre-filtered orphans |
| DryRun | Skips `PutIndexContent` but still runs blob/upload GC | Skips blob and upload GC too |
| Ref-check errors | Often discarded (`ok, _ := …`) → fail open | Propagated → fail closed (this **repo** only) |

---

## Actors and terms

| Symbol | Meaning |
|--------|---------|
| **T** | Number of manifest descriptors in `index.json` (tags + untagged entries) |
| **U** | Number of **unique** manifest/index digests reached by the walk (U ≤ T) |
| **B** | Number of orphan digests passed to `CleanupRepo` (`len(gcBlobs)`) |
| **P** | Platform manifests under multi-arch indexes (extra GETs only on the PR’s full walk) |

**Storage GET** = a remote read of `index.json`, a manifest blob, or an image index blob (the expensive part on S3/GCS).

---

## Full GC pipeline

Both sides still do retention / untagged / stale pruning under the repo lock, then blob GC. The **blob-reference** phase is what changed.

### `main`

```
cleanRepo (repo locked)
  │
  ├─ GetIndex / retention / stale prune
  ├─ PutIndexContent                 (skipped if DryRun)
  │
  └─ removeUnreferencedBlobs         ← ALWAYS runs (even on DryRun)
        ├─ GetIndex                  → 1 × index.json GET
        ├─ addIndexBlobsToReferences → ~T GETs, SHALLOW for multi-arch
        ├─ GetAllBlobs               → list
        └─ CleanupRepo(gcBlobs)
              └─ × B: deleteBlob → IsBlobReferenced → up to T GETs each
                    └─ on ErrBlobReferenced: deleteImageManifest (rare on happy path)
```

### PR #4236

```
cleanRepo (repo locked)
  │
  ├─ GetIndex / retention / unknown-media prune / stale prune
  ├─ PutIndexContent                 (skipped if DryRun)
  │
  └─ if !DryRun:
        removeUnreferencedBlobs
          ├─ GetReferencedBlobs      → 1 × index GET + ~U (+ nested) GETs, FULL walk
          ├─ GetAllBlobs             → list
          └─ CleanupRepo(gcBlobs)    → Stat + delete only (no reference re-check)
        removeBlobUploads
```

---

## Reference collection: shallow vs full

### `main` — `addIndexBlobsToReferences`

| Entry type | What gets marked referenced |
|------------|-----------------------------|
| Image manifest (OCI or docker-compat) | Manifest + config + layers + **subject** |
| Image index / manifest list | Index digest + **subject** + **platform digests only** |
| Nested platform manifests | **Not** opened for config/layers |

That shallow multi-arch gap is why `CleanupRepo` still walked the full tree with `IsBlobReferenced` — and paid **B × T** for it.

### PR — `GetReferencedBlobs` / `collectReferencedBlobs`

| Behavior | Detail |
|----------|--------|
| Full recurse | Indexes and manifests (OCI + docker-compat via shared helpers) |
| Visited set | Shared digest read at most once → **~U** GETs, not **~T** when tags share a digest |
| Contents recorded | Descriptor digests + config + layers of every reached manifest |
| Subjects | **Not** added as referenced (unlike main’s GC helpers) |
| Missing blobs | `ErrBlobNotFound` / `PathNotFound` skipped with warn |
| Other errors | Returned → abort this repo’s blob GC |

---

## Call counts (orphan blob GC)

Assume a typical remote-store GC: many orphans, many tags, mostly single-arch or already-expanded refs.

### Inside the expensive path

| Operation | `main` | PR |
|-----------|--------|-----|
| GC reference walk | ~T shallow GETs | **~U** full GETs (configs/layers included) |
| `CleanupRepo` reference checks | **B ×** full `IsBlobReferenced` ≈ **B × T** GETs | **0** |
| Per-blob `Stat` / delete | B | B |

**Dominant term**

| | Cost |
|--|------|
| `main` | **O(B × T)** |
| PR | **O(U)** (one walk), independent of B |

### Example: issue #4235 (T ≈ 519, R≈U)

| B | `main` manifest GETs (CleanupRepo alone) | PR manifest GETs (one collect) |
|---|------------------------------------------|--------------------------------|
| 100 | ~100 × 519 ≈ **51,900** | **~519** |
| 1,000 | ~**519,000** | **~519** |
| 10,000 | ~**5.19M** | **~519** |

Full-run rough totals (GC walk + CleanupRepo):

| | `main` | PR |
|--|--------|-----|
| B=1000, T=519 | ~T + B×T ≈ **519,519** | **~U ≈ 519** (one walk; no second scan) |

The earlier “follow-up: merge the two scans” opportunity is **already done** on the PR: there is only one reference walk.

---

## Side-by-side: `CleanupRepo`

| | `main` | PR |
|--|--------|-----|
| Input | `gcBlobs` from shallow filter | `gcBlobs` from full referenced set |
| Per blob | `deleteBlob` → `IsBlobReferenced` | `deleteBlobChecked(…, always false)` |
| `ErrBlobReferenced` | Try `deleteImageManifest` | N/A (should not occur for true orphans) |
| Trust model | Re-verify under lock | Trust GC filter under same lock |

Physical delete / dedupe / cache handling stays in imagestore on both sides.

---

## Behavior changes vs `main` (not just perf)

| Change | Effect |
|--------|--------|
| Multi-arch nested config/layers | **Protected** by the GC collect itself (improvement; main relied on CleanupRepo or could mis-delete Docker nested blobs) |
| Docker-compat on DeleteBlob | **Aligned** with GC collect via `IsImageIndexMediaType` / `IsImageManifestMediaType` |
| Unknown media-type index entries | Explicitly **pruned** before blob GC so digests are not stuck “referenced” forever while configs/layers disappear |
| Subject digests | Main GC marked subjects referenced; PR collect does **not** — subjects may become GC-eligible sooner (likely over-protection on main) |
| DryRun | **Non-destructive** for blobs/uploads (main could delete blobs while leaving `index.json` unchanged) |
| Fail-closed ref checks | Corrupt/partial reads abort delete / collect instead of treating blobs as free |
| Error scope | Collect / CleanRepo failure fails **that repo’s** task; the scheduler logs and continues other repos |

---

## Error / abort scope

Periodic GC submits **one scheduler task per repo** (`GCTaskGenerator` → `CleanRepo(ctx, repo)`).

If `GetReferencedBlobs` (or anything else in `CleanRepo`) returns an error:

1. That repo’s GC run stops for this interval.
2. The worker logs the task error and keeps consuming other tasks.
3. The failed repo is already in `processedRepos`, so the generator still advances.

So fail-closed behavior does **not** abort GC for the whole image store — only the current repo for that run.

---

## Duplicate tags

| | `main` `CleanupRepo` | PR collect |
|--|----------------------|------------|
| Shared digest across many tags | Re-read up to **T** times **per blob** | Read **once** (`seen` map) |

If 519 tags point at one manifest (U = 1): main still pays **B × 519**; PR pays **1** manifest GET for that digest.

---

## What the PR does **not** change

| Item | Notes |
|------|-------|
| Repo write lock for the whole `cleanRepo` | Still held for the collect + delete loop |
| Retention / untagged / referrer policy passes | Same high-level structure (plus unknown-media prune) |
| Need for a delay before orphan deletes | Age gate unchanged |
| API `DeleteBlob` still checks references | Now fail-closed and docker-aware |

---

## Verdict

| Scenario | Benefit vs `main` |
|----------|-------------------|
| Many orphans (B ≫ 1) + many manifests + remote storage | **Large** — removes **B × T** (#4235) |
| Multi-arch images | **Correctness** — nested config/layers collected once |
| DryRun / retention verify | **Safer** — no real blob GC under DryRun |
| DeleteBlob + Docker images | **Safer** — configs/layers recognized as referenced |
| Tiny repos, B ≤ 1 | Perf ~wash; still one full walk vs shallow + one full check |

**Bottom line:** vs `main`, the PR is no longer “cache references inside `CleanupRepo`.” It makes GC own a single full referenced-blob set, turns `CleanupRepo` into a thin delete helper, drops the quadratic re-walk, and tightens several fail-open / DryRun / media-type edges.

---

## Call graph cheat sheet

### `main` — B blobs in `CleanupRepo`

```
removeUnreferencedBlobs
  ├─ addIndexBlobsToReferences          [~T GETs, shallow multi-arch]
  └─ CleanupRepo
       └─ × B deleteBlob
            └─ IsBlobReferenced         [1 index GET + up to T manifest GETs]
                 └─ (optional) deleteImageManifest
```

### PR — one repo blob GC

```
removeUnreferencedBlobs
  ├─ GetReferencedBlobs                 [1 index GET + ~U GETs, full + visited]
  └─ CleanupRepo
       └─ × B deleteBlobChecked(always unreferenced)
            └─ Stat + physical delete   [0 reference GETs]
```

---

## References

- Issue: [#4235](https://github.com/project-zot/zot/issues/4235)
- PR: [#4236](https://github.com/project-zot/zot/pull/4236) (head `1ee6bbd6`)
- Code (PR): `pkg/storage/gc/gc.go` (`removeUnreferencedBlobs`, unknown-media prune, DryRun gate), `pkg/storage/common/common.go` (`GetReferencedBlobs`, media-type helpers, fail-closed `IsBlobReferencedInImageIndex`), `pkg/storage/imagestore/imagestore.go` (`CleanupRepo`, `deleteBlobChecked`)
- Code (`main`): same paths with `addIndexBlobsToReferences*` and per-blob `IsBlobReferenced` inside `CleanupRepo`

rchincha:
Let's not merge this into 2.1.x release train.

I like what the PR is doing, however if we get this wrong for whatever reason, we will cause data loss and late into the mature 2.1.x train.

Apologies if AI reviews were thrown in the mix but was just to make sure we have all our bases covered.

Let's target the upcoming 2.2.x release train for this.

pkg/storage/common/common.go
thees: Added three mock-based unit tests in pkg/storage/common/common_test.go that drive the image-index error branches of collectReferencedBlobs through the exported common.GetReferencedBlobs:

- TestGetReferencedBlobsPropagatesCorruptIndexError: a corrupt nested index (valid read, invalid JSON) yields a non-not-found error, so it is propagated (err != nil, referenced == nil) instead of being silently skipped.
- TestGetReferencedBlobsSkipsMissingIndex: a nested index missing with zerr.ErrBlobNotFound is skipped (the errors.Is arm) and recursion continues, so a healthy sibling's config+layer digests are still collected.
- TestGetReferencedBlobsSkipsMissingIndexPathNotFound: same for driver.PathNotFoundError (the errors.As arm at common.go:688, which no prior test exercised).

pkg/storage/gc/gc.go
```diff
@@ -204,6 +220,45 @@ func (gc GarbageCollect) cleanRepo(ctx context.Context, repo string) error {
...
+		if isKnownManifestMediaType(desc.MediaType) {
+			kept = append(kept, desc)
+
+			continue
+		}
+
+		if err := gc.syncStaleManifestRemoval(repo, desc); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
thees: syncStaleManifestRemoval was parametrized (now syncManifestRemoval(repo, desc, reason, message)). removeUnknownMediaTypeManifestEntries now emits reason="unknownMediaTypePrune" / "pruned unknown media type manifest entry from index", so log/audit consumers can  distinguish it from stale-retention prunes. The two stale-prune callsites keep reason="staleManifestPrune" (output unchanged). Added TestGCUnknownMediaTypePruneReason + TestGCStaleManifestPruneReason, which capture the audit log and assert the emitted reason.

pkg/storage/common/common.go
```diff
@@ -619,6 +627,110 @@ func IsBlobReferenced(imgStore storageTypes.ImageStore, repo string,
...
+
+				log.Error().Err(err).Str("repository", repo).Str("digest", desc.Digest.String()).
+					Msg("failed to read multiarch(index) image while collecting referenced blobs")
+
+				return err
+			}
+
+			if err := collectReferencedBlobs(imgStore, repo, indexImage, referenced, seen, log); err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
thees: fixed

pkg/storage/common/common.go
```diff
@@ -619,6 +627,110 @@ func IsBlobReferenced(imgStore storageTypes.ImageStore, repo string,
...
+					Msg("failed to read manifest image while collecting referenced blobs")
+
+				return err
+			}
+
+			referenced[manifestContent.Config.Digest] = struct{}{}
+
+			for _, layer := range manifestContent.Layers {
+				referenced[layer.Digest] = struct{}{}
+			}
```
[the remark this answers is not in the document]
thees: fixed

andaaron:
Please revert 345c4888 — subjects should not be in the referenced set.

OCI defines `subject` as a weak association to a separate DAG; missing subjects are allowed. Marking `Subject.Digest` as referenced was a bug in the old GC helpers, not behavior we should restore.

The lifetime dependency is image → signature, not signature → image. If a manifest is removed from `index.json`, we should GC that manifest and its blobs even if a referrer/signature still points at it via `subject`. Dangling referrers are handled by the existing DeleteReferrer policy when enabled; they are not a reason to pin the subject forever.

Also:
- This only records the subject digest and does not walk its config/layers, so the “protection” is incomplete anyway.
- It reintroduces an inconsistency with `IsBlobReferenced` / DeleteBlob, which still do not treat `subject` as protecting the subject digest (manifest/config/layers only). The new GetReferencedBlobs doc even calls this out.

Please drop subject digests from `collectReferencedBlobs` again (and the tests that assert the opposite).

rchincha:
@thees 

https://github.com/project-zot/zot/pull/4137 @82c9cbaf34fb3c0d1393796e9c49c9beff890b5b
is beginning to converge.

Pls. feel free to test.
Makes sense to pair your PR with that.

thees:
> @thees
> 
> #4137 @82c9cbaf34fb3c0d1393796e9c49c9beff890b5b is beginning to converge.
> 
> Pls. feel free to test. Makes sense to pair your PR with that.

@rchincha 
 
I ran a trial rebase of this PR onto refactor/storage (4c2c6a35) on a throwaway branch: the net diff applies cleanly except for one adjacency conflict in imagestore_test.go (this PR drops TestCleanupRepoFailsOnDeleteImageManifest, yours adds TestNewImageStoreFailsWhenMigrationFails right next to it). common, gc and imagestore tests all pass on top.


Looks good imho: we both widened deleteBlobChecked to func() (bool, error), and your cross-repo reclaim (ShouldDeleteGlobalBlob /  isDigestReferencedAcrossRepos) lives inside deleteBlobChecked independent of the caller's predicate. So this PR's GC-owned CleanupRepo passing an always-false predicate just skips the redundant per-repo re-check while your global-blobstore reclaim still guards the shared copy. WithBlobstoreAndRepoLock and the _blobstore special-casing survive the rebase untouched.

So, how to proceed here?

thees:
> > @thees
> > #4137 @82c9cbaf34fb3c0d1393796e9c49c9beff890b5b is beginning to converge.
> > Pls. feel free to test. Makes sense to pair your PR with that.
> 
> @rchincha
> 
> I ran a trial rebase of this PR onto refactor/storage ([4c2c6a3](https://github.com/project-zot/zot/commit/4c2c6a35d50f70db9d87fe0f24d4331e36e7caec)) on a throwaway branch: the net diff applies cleanly except for one adjacency conflict in imagestore_test.go (this PR drops TestCleanupRepoFailsOnDeleteImageManifest, yours adds TestNewImageStoreFailsWhenMigrationFails right next to it). common, gc and imagestore tests all pass on top.
> 
> Looks good imho: we both widened deleteBlobChecked to func() (bool, error), and your cross-repo reclaim (ShouldDeleteGlobalBlob / isDigestReferencedAcrossRepos) lives inside deleteBlobChecked independent of the caller's predicate. So this PR's GC-owned CleanupRepo passing an always-false predicate just skips the redundant per-repo re-check while your global-blobstore reclaim still guards the shared copy. WithBlobstoreAndRepoLock and the _blobstore special-casing survive the rebase untouched.
> 
> So, how to proceed here?

rebased after merge of [#4223](https://github.com/project-zot/zot/pull/4223)

andaaron:
@thees, please rebase this PR, we're considering merging it now.

thees:
@andaaron done

andaaron (approved):

andaaron:
@thees, thanks again for the fix.

Issue #4240, opened 2026-07-18
[Bug]: storage(gc): CleanupRepo repo-removal guard ignores in-progress .sync on-demand staging


seniorquico:
### zot version

fea81a5 (most recent trunk commit)

### Describe the bug

`ImageStore.CleanupRepo` will delete an entire repository directory when GC finds every
committed blob unreferenced. The safety guard that is supposed to prevent removing a repo
with in-flight writes only checks the `.uploads/` directory (regular blob-upload
sessions). It does **not** account for the sync staging directory
(`<repo>/.sync/<uuid>/`), so a repo can be removed out from under an on-demand pull that is
actively staging into it.

This applies to the default sync configuration (no `downloadDir`): `NewService` then reuses
the serving store controller as the temp store (`pkg/extensions/sync/service.go`), placing
staging sessions inside the repo directory. With a configured `downloadDir`, staging lives
outside the repo and is not destroyed by this path.

```go
// pkg/storage/imagestore/imagestore.go
func (is *ImageStore) CleanupRepo(repo string, blobs []godigest.Digest, removeRepo bool) (int, error) {
    ...
    blobUploads, _ := is.ListBlobUploads(repo)   // only inspects <repo>/.uploads

    // if removeRepo flag is true and we cleanup all blobs and there are no blobs currently being uploaded.
    if removeRepo && count == len(blobs) && count > 0 && len(blobUploads) == 0 {
        is.log.Info().Str("repository", repo).Msg("removed all blobs, removing repo")
        if err := is.storeDriver.Delete(path.Join(is.rootDir, repo)); err != nil {  // rm -rf <root>/<repo>, incl. .sync/<uuid>
            ...
        }
    }
}
```

The on-demand sync destination stages downloads into a session directory **inside** the
repo folder, then commits and removes only that session:

```go
// pkg/extensions/sync/oci_layout.go — GetImageReference
tempSyncPath   := path.Join(localImageStore.RootDir(), repo, constants.SyncBlobUploadDir) // <root>/<repo>/.sync
sessionRepoPath := path.Join(tempSyncPath, uuid.String())                                 // <root>/<repo>/.sync/<uuid>
sessionRepo     := path.Join(sessionRepoPath, repo)                                        // <root>/<repo>/.sync/<uuid>/<repo>
```

`GetAllBlobs` only lists `<repo>/blobs/`, so GC never counts the staged blobs, and
[...]

PR #4241, opened 2026-07-19
fix(gc): respect in-progress .sync sessions when removing repos

seniorquico:
**What type of PR is this?**

bug

#4240

`ImageStore.CleanupRepo` deletes an entire repository directory when garbage collection finds every committed blob unreferenced. Its "no in-flight writes" guard only checked the `.uploads/` directory (`ListBlobUploads`) and ignored the on-demand sync staging directory `.sync/<uuid>/`. As a result, GC could remove a repository out from under a live on-demand pull that is staging blobs into it (the default sync config with no `downloadDir`), corrupting the in-flight sync.

This PR closes that race with two mirrored changes:

1. Guard (`pkg/storage/imagestore/imagestore.go`): `CleanupRepo` now also skips repo removal while `<repo>/.sync/` contains any session directory, exactly as it already does for a non-empty `.uploads/`.
2. Reaper (`pkg/storage/gc/gc.go`): a new `removeStaleSyncSessions` deletes `.sync/<uuid>` staging sessions older than the GC delay, mirroring the existing `removeBlobUploads` reaper. Without it, a session left behind by a crashed/interrupted sync would block repo removal forever (and leak disk). It runs at the end of `cleanRepo`, after `removeUnreferencedBlobs` (and thus after the `CleanupRepo` guard), so a freshly created session's blobs are never reaped out from under it within the same pass.

Known limitation (disclosed): a `.sync/<uuid>` directory's mtime only advances when its _direct_ children change, so a single download running longer than the GC delay could still be reaped mid-flight. Severity is low — the GC delay defaults to 1 hour (`DefaultGCDelay`), well above realistic staging times — and this is strictly better than the prior behavior, which deleted live sessions unconditionally via repo removal. A related follow-up is intentionally out of scope: `CommitAll` (`pkg/extensions/sync/destination.go`) silently returns `nil` when its session directory has vanished; it should fail loudly instead.

New unit tests were added covering both halves of the fix, exercised against the real on-disk staging layout (`<repo>/.sync/<uuid>/`):

- `TestCleanupRepoToleratesSyncSessions` (`pkg/storage/imagestore`) — an in-progress `.sync` session blocks repo removal; removal proceeds when none exists; an empty `.sync/` dir does not block removal.
- `TestRemoveStaleSyncSessions` (`pkg/storage/gc`) — a stale session (old mtime) is reaped while a fresh one is kept.
- `TestRemoveStaleSyncSessionsNoSessions` — no `.sync` directory ⇒ zero deletions, no error.
- `TestGCReaperSyncSessionsWithRepoRemoval` — a repo with an unreferenced blob plus a stale session: the guard blocks repo removal while the session is present, the blob is GC'd, and the reaper removes the stale session (empty repos are intentionally left in place).

```bash
$ go test ./pkg/storage/gc/... ./pkg/storage/imagestore/...
ok      zotregistry.dev/zot/v2/pkg/storage/gc          15.4s
ok      zotregistry.dev/zot/v2/pkg/storage/imagestore  0.1s

$ go test ./pkg/storage/gc/... -run 'TestRemoveStaleSyncSessions|TestGCReaperSyncSessions' -v
--- PASS: TestRemoveStaleSyncSessions
--- PASS: TestRemoveStaleSyncSessionsNoSessions
--- PASS: TestGCReaperSyncSessionsWithRepoRemoval

$ go test ./pkg/storage/imagestore/... -run TestCleanupRepoToleratesSyncSessions -v
--- PASS: TestCleanupRepoToleratesSyncSessions
```

No. This changes only runtime garbage-collection behavior. There is no on-disk format change, no config schema change, and no API change — the .sync/<uuid> staging layout already exists. A binary on either side of this change reads and writes the same storage. Safe to upgrade or downgrade.

No API or configuration change. Behaviorally, garbage collection no longer deletes a repository while an on-demand sync is staging into it, and it now reaps stale `.sync` staging sessions. GC deletion metrics gain one new label value, `syncSession`, alongside the existing `manifest/blob/upload`.

```release-note
Garbage collection no longer removes a repository directory while an on-demand sync is
staging blobs into it, and now reaps stale sync staging sessions (`.sync/<uuid>`) left
behind by interrupted syncs. GC deletion metrics gain a new `syncSession` label value.
```

### DIFF
--- a/pkg/storage/gc/gc.go
+++ b/pkg/storage/gc/gc.go
@@ -20,6 +20,7 @@ import (
 	zcommon "zotregistry.dev/zot/v2/pkg/common"
 	"zotregistry.dev/zot/v2/pkg/compat"
 	"zotregistry.dev/zot/v2/pkg/extensions/monitoring"
+	"zotregistry.dev/zot/v2/pkg/extensions/sync/constants"
 	zlog "zotregistry.dev/zot/v2/pkg/log"
 	mTypes "zotregistry.dev/zot/v2/pkg/meta/types"
 	"zotregistry.dev/zot/v2/pkg/retention"
@@ -195,15 +196,23 @@ func (gc GarbageCollect) cleanRepo(ctx context.Context, repo string) error {
 		return err
 	}
 
+	// gc stale sync staging sessions
+	sessionsDeleted, err := gc.removeStaleSyncSessions(repo, gc.opts.Delay)
+	if err != nil {
+		return err
+	}
+
 	if !gc.opts.ImageRetention.DryRun {
 		monitoring.IncGCDeleted(gc.metrics, "manifest", manifestsDeleted)
 		monitoring.IncGCDeleted(gc.metrics, "blob", blobsDeleted)
 		monitoring.IncGCDeleted(gc.metrics, "upload", uploadsDeleted)
+		monitoring.IncGCDeleted(gc.metrics, "syncSession", sessionsDeleted)
 	}
 
 	return nil
 }
 
+// removeStaleManifestEntries removes manifest/index descriptors whose blobs no longer exist in storage.
 func (gc GarbageCollect) removeStaleManifestEntries(repo string, index *ispec.Index) error {
 	if gc.opts.ImageRetention.DryRun {
 		return nil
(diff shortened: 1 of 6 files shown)


andaaron:
Thanks for coming forward with a PR to fix this.

**Blocking**: sync sessions are always local; the store driver may not be. They are not part of serving storage — only a temporary staging area before the image is validated and pushed into actual storage.

Sync staging always goes through ocidir:// + a local image store (pkg/extensions/sync/oci_layout.go, getImageStore). For remote storage (S3/GCS/Azure), config already requires extensions.sync.downloadDir, so sessions live under that local directory — not under the remote store-driver prefix.

This PR’s ListSyncSessions / StatSyncSession / DeleteSyncSession go through the serving ImageStore’s storeDriver, like .uploads. That only matches the default local layout (no downloadDir), where .sync sits under <root>/<repo>/.

For remote + downloadDir (and local + downloadDir):
- the repo-wipe race does not apply the same way (remote Delete does not remove the local staging tree);
- the new guard/reaper on the serving store look at the wrong backend;
- orphan cleanup under downloadDir remains unaddressed.
- 
Unlike .uploads (real store-driver objects, including on remote), .sync sessions are not remote objects.
Extending the ImageStore / driver API for them conflates serving storage with always-local ocidir staging and sets the wrong expectation.

I don’t think we should land this as-is. Sync-session lifecycle needs to stay on the local staging root, not on the remote-capable serving store driver. Happy to discuss an approach that keeps the local in-repo guard without putting these APIs on ImageStore.

seniorquico:
Thanks, @andaaron! Yes, I'm happy to keep working on this with some guidance. What is the best way to workshop an alternative design that guards only the local storage?

andaaron:
Here's what I propose:

No new ImageStore APIs, drop ListSyncSessions / StatSyncSession / DeleteSyncSession from types.ImageStore. Those belong to local staging, not serving storage.

1. In CleanupRepo, before Delete(repo):
- if the store driver is not local → keep current behavior (remote Delete cannot wipe local downloadDir staging);
- if local → check the filesystem under `filepath.Join(RootDir(), repo, syncConstants.SyncBlobUploadDir)` with os/filepath (not storeDriver.List);
- if any session dir exists → skip repo removal (same idea as .uploads presence).

2. Add a reaper task generator, but not in GC.

Sync already removes session dirs on the normal path (defer CleanupImage in syncImage, and os.RemoveAll in CommitAll).
Orphans only matter after crash/kill/CleanupImage never running.

If we want cleanup for orphans, a better fit is a separate scheduled task generator in the sync extension that:
- walks known local staging roots (downloadDir if set, else each local store RootDir);
- deletes `<repo>/.sync/<uuid>/` sessions older than a delay - I suggest using the same value as for GC delay.
- uses local FS ops only — never the remote store driver.

andaaron:
Hi @seniorquico, are you still working on this?

seniorquico:
Hi @andaaron! The short answer, not really.

I did try with Claude and your notes to work on refactoring it, but it quickly ran past my level of understanding. I set aside more time to gain a better, overall understanding of the storage providers. That led to a couple other recent issues/PRs. I haven't yet had time to get get back into this specific issue, though.

andaaron:
Superseded by https://github.com/project-zot/zot/pull/4383.

PR #4171, opened 2026-06-28
feat(oidc): support login flow for OIDC tokens

matheuscscp:
Closes: #4169

The main use case for this feature is a Kubernetes image pull Secret with the following format (`username` is ignored and can be any string):

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: zot-oidc-pull-secret
  namespace: zot-oidc-test
type: kubernetes.io/dockerconfigjson
stringData:
  .dockerconfigjson: |
    {
      "auths": {
        "zot-oidc-wid:5000": {
          "username": "zot",
          "password": "<kubernetes-service-account-token>"
        }
      }
    }
```

This will allow kubelet to pull images from Zot authenticating with any OIDC token trusted by Zot in the `password` field, including ServiceAccount tokens (which are OIDC tokens).

PR #4243, opened 2026-07-20
feat(auth): allow bearer with API keys and OIDC for humans

### LINKED ISSUE #4033: [Bug]: Failed to use openid.oidc with bearer.oidc authorization being configured

matheuscscp:
Fixes: #4033

This PR introduces support for Bearer tokens (including OIDC) to work together with OIDC for humans and API keys.

The solution implemented here is enhancing the token endpoint introduced in #4171 to work also for API keys. API keys are traditionally sent via `Authorization: Basic base64(user:zak_...)` and this continues to work, but now the token endpoint can also detect API keys. When it does, it wraps the `user:zak_...` in a (non-signed!) Bearer token and sends in the token endpoint response. This causes clients to send the `user:zak_...` on every request like before, but via `Authorization: Bearer <envelopment>` instead.

This way users can enable OIDC for machines and humans at the same time, and also API keys e.g. created by a human that logged in via OIDC in the browser:

* Machines can use OIDC
* Humans can use OIDC in the browser
* Humans can create API keys in the browser after logging in via OIDC
* Humans can use CLI clients with the API keys they created on the browser


pkg/api/bearer_auth.go
```diff
@@ -240,8 +243,36 @@ func (b *BearerAuth) Middleware(ctlr *Controller) mux.MiddlewareFunc {
...
-	if b == nil || !b.authConfig.IsOIDCBearerAuthEnabled() || b.oidc == nil {
+func requestedBearerResourceAction(request *http.Request) *ResourceAction {
+	if request.RequestURI == "/v2/" {
+		return nil
+	}
+
+	vars := mux.Vars(request)
+	name := vars["name"]
+
+	var action string
```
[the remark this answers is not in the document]
matheuscscp: Fixed, thanks!

matheuscscp:
Can you pls trigger the CI again? :pray: Hopefully copilot will get satisfied now :pray:

matheuscscp:
omg how pesky, copilot, pls shut up a lil bit

pkg/api/authn.go
```diff
@@ -533,6 +571,28 @@ func setBearerAuthChallenge(
...
+	requestedAccess *ResourceAction,
+) {
+	if hasSessionHeader(request) {
+		return
+	}
+
+	setBearerAuthChallenge(response, authConfig, requestedAccess)
+}
+
+func hasBearerChallengeConfig(authConfig *config.AuthConfig) bool {
```
andaaron: Shouldn't this be a method of authConfig?

pkg/api/authn.go
```diff
@@ -533,6 +571,28 @@ func setBearerAuthChallenge(
...
+	return authConfig != nil && authConfig.Bearer != nil &&
+		authConfig.Bearer.Realm != "" && authConfig.Bearer.Service != ""
+}
+
+func shouldAdvertiseBearerChallenge(authConfig *config.AuthConfig) bool {
```
andaaron: Shouldn't this be a method of authConfig?

pkg/api/authn.go
```diff
@@ -67,16 +66,25 @@ func AuthHandler(ctlr *Controller) AuthSetup {
...
 	}
 
-	return AuthSetup{Middleware: authnMiddleware.tryAuthnHandlers(ctlr)}
+	return AuthSetup{
+		Middleware:   authnMiddleware.tryAuthnHandlers(ctlr),
+		TokenHandler: tokenHandler,
+	}
+}
+
+func shouldInitializeBearerAuth(authConfig *config.AuthConfig) bool {
```
andaaron: Shouldn't this be a method of AuthConfig?

Also I think it could use a comment about the rationale behind these specific conditions.

pkg/api/authn.go
```diff
@@ -271,72 +279,77 @@ func (amw *AuthnMiddleware) basicAuthn(ctlr *Controller, userAc *reqCtx.UserAcce
...
-				ctlr.Log.Info().Err(err).Msgf("failed to find any user info for hashed key %s in DB", hashedKey)
+	trimmedAPIKey := strings.TrimPrefix(apiKey, constants.APIKeysPrefix)
 
-				return false, nil
-			}
+	hashedKey := hashUUID(trimmedAPIKey)
 
-			ctlr.Log.Error().Err(err).Msgf("failed to get user info for hashed key %s in DB", hashedKey)
+	storedIdentity, err := ctlr.MetaDB.GetUserAPIKeyInfo(hashedKey)
+	if err != nil {
```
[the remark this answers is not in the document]
andaaron: Let's add this just in case? Theoretically it is not possible to get to this point.

pkg/api/authn.go
```diff
@@ -445,6 +458,17 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
 			// Allow anonymous access to the metrics endpoint only if configured.
 			extensionsConfig := ctlr.Config.CopyExtensionsConfig()
 			isMetricsRequestedWithAnonymousAccess := isAnonymousMetricsRequest(request, accessControlConfig, extensionsConfig)
+			anonymousAllowed := false
+			anonymousAllowedComputed := false
+			anonymousAccessAllowed := func() bool {
```
andaaron: I think this is hard to follow, there are too many variables with similar names,

pkg/api/authn.go
```diff
@@ -848,6 +937,81 @@ func isAnonymousMetricsRequest(request *http.Request, accessControlConfig *confi
...
+}
+
+func distSpecAuthzActionForRequest(ctlr *Controller, request *http.Request, repository, reference string,
+	hasReference bool,
+) string {
+	switch request.Method {
+	case http.MethodGet, http.MethodHead:
+		return constants.ReadPermission
+	case http.MethodPut, http.MethodPatch, http.MethodPost:
+		if hasReference {
```
andaaron: What is this about? Seems like a behavior change.

pkg/api/authn.go
```diff
@@ -848,6 +937,81 @@ func isAnonymousMetricsRequest(request *http.Request, accessControlConfig *confi
...
+	default:
+		return ""
+	}
+}
+
+func hasBasicAuthorizationHeader(request *http.Request) bool {
+	return hasAuthorizationScheme(request, "basic")
+}
+
+func hasEmptyBasicAuthorizationHeader(request *http.Request) bool {
```
andaaron, line 944: Is this a valid case?

pkg/api/authn.go
```diff
@@ -1242,10 +1406,8 @@ func OAuth2Callback(ctlr *Controller, w http.ResponseWriter, r *http.Request, st
 }
 
 func hashUUID(uuid string) string {
-	digester := sha256.New()
-	digester.Write([]byte(uuid))
-
-	return godigest.NewDigestFromEncoded(godigest.SHA256, hex.EncodeToString(digester.Sum(nil))).Encoded()
+	// codeql[go/weak-sensitive-data-hashing] API keys are random token material hashed for deterministic lookup.
+	return godigest.FromBytes([]byte(uuid)).Encoded()
```
andaaron: What does this change?

pkg/api/authn_internal_test.go
```diff
@@ -334,3 +342,293 @@ func TestAppendOpenIDGroups(t *testing.T) {
 		})
 	}
 }
+
+func TestAuthorizationSchemeHelpers(t *testing.T) {
+	t.Parallel()
+
+	request := httptest.NewRequest("GET", "/v2/", nil)
+	request.Header.Set("Authorization", "Basic")
+	if hasBasicAuthorizationHeader(request) {
```
andaaron: can we use github.com/stretchr/testify/assert for new tests?

pkg/api/authn.go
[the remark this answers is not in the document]
matheuscscp: Done in c525a096. authenticateAPIKeyCredential now returns an error when ctlr or MetaDB is nil before any DB access, so callers respond with 500 instead of panicking. Added TestAuthenticateAPIKeyCredentialRequiresMetaDB for this path.

wizardist:
`podman login` doesn't work with this patch when OIDC (eg GitHub) is used for interactive auth and API key is generated via the GUI.

```
Error: authenticating creds for "zot-test.example.org": Get "zot-test?account=test%40example.org&service=zot-test-service": unsupported protocol scheme ""
```

However, the patch fixes the inability (400 Bad Request) to sign in in the browser with OIDC when both `auth.bearer` and `auth.openid` are configured.

**Upd.** Apologies, docs got out of sync with `README-OIDC-WORKLOAD-IDENTITY.md`. `podman login` works fine. :+1:

matheuscscp:
@andaaron @rchincha PTAL :pray:

pkg/api/authn.go
```diff
@@ -848,6 +921,34 @@ func isAnonymousMetricsRequest(request *http.Request, accessControlConfig *confi
...
+		return false
+	}
+
+	parts := strings.SplitN(header, " ", 2)
+	if len(parts) != 2 {
+		return false
+	}
+
+	return strings.EqualFold(parts[0], scheme)
+}
```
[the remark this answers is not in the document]
matheuscscp: Fixed, thanks!

pkg/api/token_exchange.go
```diff
@@ -95,6 +142,69 @@ func readTokenRequestFormBody(body io.Reader) ([]byte, error) {
...
+}
+
+func bearerTokenFromAuthHeader(header string) (string, bool) {
+	parts := strings.SplitN(strings.TrimSpace(header), " ", 2)
+	if len(parts) != 2 || !strings.EqualFold(parts[0], "bearer") {
+		return "", false
+	}
+
+	return strings.TrimSpace(parts[1]), true
+}
```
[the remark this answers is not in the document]
matheuscscp: Fixed, thanks!

pkg/api/bearer_auth.go
```diff
@@ -280,14 +330,43 @@ func (b *BearerAuth) TokenExchangeHandler() http.HandlerFunc {
...
+
+					return
+				}
+
+				b.log.Debug().Msg("api key token exchange failed")
+
+				continue
+			}
+
+			switch localOIDCTokenOwnerForCredential(credential.Secret, b.authConfig) {
```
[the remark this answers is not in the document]
matheuscscp: Fixed, thanks!

matheuscscp:
> ## Pull request overview
> Copilot reviewed 11 out of 11 changed files in this pull request and generated no new comments.

Unbelievable :face_holding_back_tears:

matheuscscp:
@andaaron @rchincha Please trigger the CI :pray:

matheuscscp:
@andaaron @rchincha One more time pls :pray: I think fixed the CodeQL issue. The stress test seems unrelated.

matheuscscp:
Hmmmm..... https://www.githubstatus.com/

matheuscscp:
@andaaron @rchincha I'm back from vacations and would like to wrap this up :pray:

pkg/api/token_exchange.go
```diff
@@ -38,12 +52,19 @@ type unverifiedJWTClaims struct {
 	audiences []string
 }
 
+type wrappedCredentialBearer struct {
+	Version  int    `json:"v"`
+	Type     string `json:"typ"`
```
andaaron: Is `"typ"` a typo?

pkg/api/authn.go
```diff
@@ -49,9 +47,12 @@ const (
 	relyingPartyCookieMaxAge = 120
 )
 
+var errAPIKeyMetadataDBRequired = errors.New("metadata database is required for api key authentication")
```
andaaron: Can we move the error definitions to the errors package under the root of the repository?
This applies to multiple such instances.

pkg/api/authn.go
```diff
@@ -455,17 +472,25 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
...
+			// Switch authentication methods based on the credential transport the client chose.
 			switch {
-			// Reject requests with multiple Authorization headers as a security measure
+			// Reject requests with multiple Authorization headers as a security measure.
 			case hasMultipleAuthorizationHeaders(request):
 				authenticated = false
 
-			// The authorization header presence is an explicit attempt to use basic authentication
-			case !isAuthorizationHeaderEmpty(request) && authConfig.IsBasicAuthnEnabled():
+			case hasBearerAuthorizationHeader(request) && amw.bearerAuth != nil:
```
andaaron, line 479: The changes in this file are very hard to follow.

Can you add some documentation to this PR in the codebase? It should cover at least:
- the supported combinations of authentication - I suggest to also include a flow chart.
- the real world scenarios with use cases and what 3rd party integrations we added them for.

pkg/api/bearer_auth.go
```diff
@@ -233,15 +242,56 @@ func (b *BearerAuth) Middleware(ctlr *Controller) mux.MiddlewareFunc {
...
+	if ctlr != nil && ctlr.MetaDB != nil {
+		return true
+	}
+
+	b.log.Error().Msg("failed to authenticate api key because metadata database is not initialized")
+
+	return false
+}
+
+func (b *BearerAuth) TokenExchangeHandler(ctlr *Controller) http.HandlerFunc {
```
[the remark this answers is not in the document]
andaaron: - **Bug:** Client sends a wrapped API-key Bearer (`zot_cred_v1.…`). Token handler does not treat it as local, so with an upstream token URL configured it **forwards the request (including `Authorization`)** and the upstream sees the API key inside the unsigned wrapper.
- **Why:**
  1. Credential collection only reads Basic + form — not `Authorization: Bearer` — `token_exchange.go:62-100`
  2. “Local” detection only matches `zak_…`, not `zot_cred_v1.…` — `token_exchange.go:130-133`
  3. Proxy copies `Authorization` (not on the skip list) — `bearer_auth.go:388-450`, `541-547`
- **Also:** Same leak if the wrapper is posted as form `password` / `token` (still not `zak_`).
- **Not this bug:** Bare `zak_…` without username is already blocked from proxy — `bearer_auth.go:333-340`
- **Fix:** Treat `zot_cred_v1.` as locally owned (never proxy); add a regression test

pkg/api/authn.go
```diff
@@ -455,17 +472,25 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
 
 			var err error
 
-			// Switch authentication methods based on provided request context
+			// Switch authentication methods based on the credential transport the client chose.
 			switch {
-			// Reject requests with multiple Authorization headers as a security measure
+			// Reject requests with multiple Authorization headers as a security measure.
```
andaaron, line 475: - **Scenario:** Registry has anonymous (or OpenID/htpasswd) but **no** BearerAuth (`bearerAuth == nil`). Client still sends `Authorization: Bearer …`.
- **Gap in the switch**:
  - Bearer is handled only when `hasBearer… && bearerAuth != nil` (`481-484`)
  - If `bearerAuth` is nil, that case is skipped
  - Request then hits `allowsUnauthenticatedRequest` (`520-530`) → treated as anonymous → **200**
- **main behavior:** Any non-empty `Authorization` with basic authn enabled went through `basicAuthn` and failed → **401**
- **Where to change:** In `tryAuthnHandlers`, **before** the anonymous case (`520`), add an explicit fail for “client sent Authorization we did not handle”, e.g.:
  - `case hasBearerAuthorizationHeader(request): authenticated = false` (covers Bearer when BearerAuth is off), **or**
  - broader: any non-empty `Authorization` that fell through the scheme cases → `authenticated = false`
- **Do not** put this inside the anonymous branch — keep anonymous only for requests with **no** usable credentials

pkg/api/authn.go
```diff
@@ -455,17 +472,25 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
...
 			case hasMultipleAuthorizationHeaders(request):
 				authenticated = false
 
-			// The authorization header presence is an explicit attempt to use basic authentication
-			case !isAuthorizationHeaderEmpty(request) && authConfig.IsBasicAuthnEnabled():
+			case hasBearerAuthorizationHeader(request) && amw.bearerAuth != nil:
+				amw.bearerAuth.Middleware(ctlr)(next).ServeHTTP(response, request)
```
andaaron: - **Problem:** Rebuilds a full middleware per call; re-runs OPTIONS / multi-header / mgmt / metrics; challenge rules can drift from the outer switch
- **Fix:** Extract `BearerAuth.Authenticate(…)` and call it from the switch (keep a thin `Middleware` only if needed for tests)

pkg/api/authn.go
```diff
@@ -475,18 +500,24 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
...
-			// Try mTLS authentication if client certificates are present
+			// Try mTLS authentication if client certificates are present.
 			case ctlr.Config.IsMTLSAuthEnabled() && request.TLS != nil && len(request.TLS.PeerCertificates) > 0:
 				authenticated, err = amw.mTLSAuthn(ctlr, userAc, request)
 
-			// If no auth methods enabled at all - then just authenticate anything
-			case !authConfig.IsBasicAuthnEnabled() && !ctlr.Config.IsMTLSAuthEnabled():
+			case isAuthorizationHeaderEmpty(request) && amw.shouldChallengeWithBearer(authConfig) &&
+				(isV2Requested || !allowsUnauthenticatedRequest):
+				amw.bearerAuth.Middleware(ctlr)(next).ServeHTTP(response, request)
```
andaaron: - **Problem:** Rebuilds a full middleware per call; re-runs OPTIONS / multi-header / mgmt / metrics; challenge rules can drift from the outer switch
- **Fix:** Extract `BearerAuth.Authenticate(…)` and call it from the switch (keep a thin `Middleware` only if needed for tests)

matheuscscp:
@andaaron Just fixed all your comments and rebased with main, pls allow the CI to run :pray:

matheuscscp:
@andaaron I think the only job that failed is a flake, not related. Can you pls retrigger that job? :pray:

examples/README-COMBINED-AUTHENTICATION.md
```diff
@@ -0,0 +1,115 @@
...
+  BearerLocal -->|yes| BearerAuth[Authenticate as wrapped API key, workload OIDC, or traditional Bearer]
+  BearerLocal -->|no| Reject
+  BearerHeader -->|no| BasicHeader{Username/password credential?}
+  BasicHeader -->|yes| BasicAuth[Authenticate with htpasswd, LDAP, or API key]
+  BasicHeader -->|no| SessionHeader{Browser session client?}
+  SessionHeader -->|yes| SessionAuth[Authenticate session cookie]
+  SessionHeader -->|no| ClientCert{Client TLS certificate?}
+  ClientCert -->|yes| MTLSAuth[Authenticate with mTLS identity]
+  ClientCert -->|no| NoCreds{No credentials?}
+  NoCreds -->|yes| Challenge{Bearer challenge needed?}
```
andaaron, line 40: `ReturnChallenge` / `Reject` nodes do not say which challenge header is emitted. In code the rule is config-global (prefer Bearer when advertiseable), not per-credential-style:

| When | Challenge |
|---|---|
| `ShouldAdvertiseBearerChallenge()` (realm+service set **and** bearer.oidc/traditional **or** API keys) | **`WWW-Authenticate: Bearer …`** |
| Else, if htpasswd/LDAP/API key Basic can auth | **`WWW-Authenticate: Basic …`** |
| UI session client (`X-ZOT-API-CLIENT: zot-ui`) | **neither** (authn/authz suppress challenge) |
| Token-exchange local 401 (`oidcTokenExchangeUnauthorized`) | **Basic** realm (token URL itself), not Bearer |

**Doc fix:** Replace bare `ReturnChallenge` with a decision node, or add a short “Challenge advertisement” subsection with the table above.

examples/README-COMBINED-AUTHENTICATION.md
```diff
@@ -0,0 +1,115 @@
...
+  }
+}
+```
+
+Zot handles these token exchange requests locally:
+
+- API keys that start with Zot's API-key prefix.
+- Wrapped API-key Bearer credentials issued by Zot.
+- OIDC workload tokens trusted by `bearer.oidc`.
+- Browser OpenID/OAuth2 tokens from configured `openid.providers`; these are rejected locally because they are for the browser login flow, not the OCI token service flow.
```
andaaron: A `zot_cred_v1.…` on the **token** endpoint is refused with **401** (never re-minted, never proxied):

```350:355:pkg/api/bearer_auth.go
		if token, ok := bearerTokenFromAuthHeader(...); ok && isWrappedCredentialBearerToken(token) {
			oidcTokenExchangeUnauthorized(...)
			return
		}
```

Form/Basic carrying the wrapper is also locally owned → 401 (`380-384`, `442-445`). Successful use of the wrapper is on **`/v2/` via `Authenticate`**, not on `/zot/auth/token`.

**Doc fix:** Split “successfully exchange” (`zak_`, workload OIDC) vs “recognize as Zot-owned and never proxy (401 if sent to token endpoint again); use on registry as Bearer.”

examples/README-COMBINED-AUTHENTICATION.md
```diff
@@ -0,0 +1,115 @@
...
+| `bearer.oidc` and `openid` | Yes | Workloads use OIDC tokens as Bearer credentials; humans use browser SSO sessions. |
+| `bearer.oidc` and `apikey` | Yes | Workloads can use secret-less OIDC tokens, while users or automation that cannot get workload tokens can use API keys. |
+| `bearer.oidc` and traditional Bearer JWTs | Yes | Zot tries locally trusted workload OIDC tokens first. Traditional Bearer JWTs still work when configured with a certificate or AWS Secrets Manager verification keys. |
+| `bearer.oidc` and an external token service | Yes | Advertise Zot's `/zot/auth/token` endpoint and set `upstreamTokenEndpoint` so Zot can forward only token requests that do not belong to a local Zot credential source. |
+| `mtls` with other authentication methods | Yes | Client certificates remain available as a separate credential style for clients that present TLS certificates. |
+| Anonymous policies with authentication | Yes | Requests without credentials can be allowed by anonymous, management, or metrics policy. Requests with invalid credentials are rejected instead of being treated as anonymous. |
```
andaaron: This sentence overstates: failed/expired UI sessions can still fall through to
anonymous (or mgmt/metrics) when those policies allow it (authn session case).
Empty Docker-style Basic can too. Non-empty bad Authorization is rejected.

Please rephrase to: invalid Authorization credentials are rejected; browser UI
clients with a missing/expired session may still use anonymous policy when
configured. No need to document the internal session header name for OCI users.

examples/README-COMBINED-AUTHENTICATION.md
```diff
@@ -0,0 +1,115 @@
...
+  Start[Request reaches Zot] --> Multi{Multiple Authorization headers?}
+  Multi -->|yes| Reject[Reject request]
+  Multi -->|no| BearerHeader{Bearer credential?}
+  BearerHeader -->|yes| BearerLocal{Zot Bearer support enabled?}
+  BearerLocal -->|yes| BearerAuth[Authenticate as wrapped API key, workload OIDC, or traditional Bearer]
+  BearerLocal -->|no| Reject
+  BearerHeader -->|no| BasicHeader{Username/password credential?}
+  BasicHeader -->|yes| BasicAuth[Authenticate with htpasswd, LDAP, or API key]
+  BasicHeader -->|no| SessionHeader{Browser session client?}
+  SessionHeader -->|yes| SessionAuth[Authenticate session cookie]
```
andaaron, line 35: Mermaid shows `SessionAuth → Authz` only. Add a fail edge to Anonymous when policy allows (matches `515`).

andaaron:
@matheuscscp implementation looks good, please look at the documentation comments.

matheuscscp:
@andaaron All comments fixed :pray:

andaaron (dismissed):

pkg/api/authn.go
```diff
@@ -475,18 +512,28 @@ func (amw *AuthnMiddleware) tryAuthnHandlers(ctlr *Controller) mux.MiddlewareFun
...
-			// Try mTLS authentication if client certificates are present
+			// Try mTLS authentication if client certificates are present.
 			case ctlr.Config.IsMTLSAuthEnabled() && request.TLS != nil && len(request.TLS.PeerCertificates) > 0:
 				authenticated, err = amw.mTLSAuthn(ctlr, userAc, request)
 
-			// If no auth methods enabled at all - then just authenticate anything
-			case !authConfig.IsBasicAuthnEnabled() && !ctlr.Config.IsMTLSAuthEnabled():
+			case isAuthorizationHeaderEmpty(request) && amw.shouldChallengeWithBearer(authConfig) &&
+				(isV2Requested || !allowsUnauthenticatedRequest):
+				bearerAuthFail(response, request, authConfig, delay)
```
andaaron, line 523:
```suggestion
            case isAuthorizationHeaderEmpty(request) && amw.shouldChallengeWithBearer(authConfig):
            	// Challenge when anonymous is not allowed at all, or on /v2/ when
            	// anonymous coexists with authenticated-only policies (so clients
            	// obtain Bearer credentials before hitting protected repos).
            	// Pure anonymous-only configs must not force a Bearer handshake on /v2/:
            	// /zot/auth/token cannot mint an anonymous token.
            	needsBearerChallenge := !allowsUnauthenticatedRequest ||
            		(isV2Requested && accessControlConfig.HasMixedAnonymousAndAuthenticatedPolicies())
            	if needsBearerChallenge {
            		bearerAuthFail(response, request, authConfig, delay)
            
            		return
            	}
```

I think the question is if we want to force bearer for anonymous. @matheuscscp  Is this intended?

Other related changes:
In `examples/README-COMBINED-AUTHENTICATION.md` Challenge Advertisement / Anonymous row, note:

- Bearer is advertised on `/v2/` when mixed anonymous + authenticated policies exist.
- Anonymous-only + Bearer advertise does **not** force `/v2/` challenge (anonymous pull works without token exchange).

Suggested tests:
1. Config: API keys + `bearer.realm`/`service` + **only** anonymous `read` on `**` → `GET /v2/` without creds → **200**, no (or unused) forced failure before anonymous.
2. Config: anonymous `read` on `public/**` + authenticated-only on `private/**` + Bearer advertise → `GET /v2/` without creds → **401** with Bearer `WWW-Authenticate`.
3. Keep existing tags-list / API-key exchange coverage.

pkg/api/bearer_auth.go
```diff
@@ -140,113 +145,193 @@ func (b *BearerAuth) Middleware(ctlr *Controller) mux.MiddlewareFunc {
...
 
-					// Use BEARER_OIDC to enable authorization via accessControl config.
-					// Unlike traditional bearer tokens (which contain 'access' claims with permissions),
-					// OIDC tokens contain identity only, so authorization must come from the config.
-					amCtx := acCtrlr.getAuthnMiddlewareContext(BEARER_OIDC, request)
-					next.ServeHTTP(response, request.WithContext(amCtx)) //nolint:contextcheck
+	if wrapped, ok, err := wrappedCredentialFromBearerAuthHeader(header); ok {
+		if err != nil {
+			ctlr.Log.Debug().Err(err).Msg("failed to parse wrapped bearer credential")
+		} else if wrapped.Type == wrappedCredentialTypeAPIKey && b.authConfig.IsAPIKeyEnabled() {
```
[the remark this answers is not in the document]
andaaron: Agreed this is a real hot-reload gap for the new wrapped-credential path, but route/`BearerAuth` re-init is not required for the important case.

Please use a per-request live `ctlr.Config.CopyAuthConfig()` (helper on BearerAuth is fine) for:
- `IsAPIKeyEnabled()` in `Authenticate` (wrapped path) — currently bearer_auth.go:184
- API-key redeem + `localOIDCTokenOwnerForCredential` in `TokenExchangeHandler` — :387, :422

Keep construction-time `b.oidc` / traditional authorizer as-is.

Also: if live API keys are off but the secret still has the `zak_` prefix, treat it as locally owned and 401 (never proxy).

**Out of scope here: mounting `/zot/auth/token` when API keys were disabled at process start.** That still needs restart (or a larger reload re-init follow-up).

Please add a regression test: enable → mint wrapper → reload APIKey false → wrapper Bearer rejected.

rchincha:
@matheuscscp pls take a look at comments above. Being a little conservative here since in critical authn/authz path.

Pls also rebase.

matheuscscp:
@andaaron @rchincha Fixed all comments and rebased.

pkg/api/controller_test.go
andaaron: `TestBearerAuthWithAllowReadAccess` still expects `401` on anonymous `GET /v2/` (`:3744`), but this fixture is anonymous-only (`:3726-3730`). After the Comment B gate (`authn.go:473-474`, `:529-531`), that request correctly returns `200` without a forced Bearer handshake.

Please update the test to expect `200` on `/v2/` (no challenge) and keep Bearer challenge coverage on the write paths (`:3773+`). Don’t revert the anonymous-only behavior — classic Docker’s `/v2/` ping limitation is why we still force challenges for **mixed** policies, not anonymous-only (see moby#50896 / docs rows `:59-60`).

This should fix the failing CI jobs

matheuscscp:
@andaaron @rchincha The reds are probably all flakes, can you please retrigger them?

pkg/api/bearer_auth.go
```diff
@@ -262,10 +363,17 @@ func (b *BearerAuth) TokenExchangeHandler() http.HandlerFunc {
 			return
 		}
 
+		if token, ok := bearerTokenFromAuthHeader(request.Header.Get("Authorization")); ok &&
+			isWrappedCredentialBearerToken(token) {
+			oidcTokenExchangeUnauthorized(response, authConfig)
+
+			return
+		}
```
andaaron: Local secrets can still be forwarded to the upstream token service

We already block proxying when a client sends a wrapped API-key token (`zot_cred_v1…`) as `Authorization: Bearer` on the token endpoint (`bearer_auth.go:366-371`). That was the earlier leak fix.

There is a leftover gap for the same endpoint:

1. Local ownership is decided only from credentials collected as **Basic auth** or **form fields** (`token_exchange.go:56-66`).
2. A raw API key (`zak_…`) or a Zot-owned OIDC JWT sent **only** as `Authorization: Bearer …` is never classified as local.
3. If an upstream token URL is configured, the request is proxied (`bearer_auth.go:470-476`) and the `Authorization` header is copied upstream (`bearer_auth.go:589-612`).

So a client (or buggy tool) that puts a Zot API key / local OIDC token in the Bearer header on `/zot/auth/token` can accidentally send that secret to the upstream IdP/token service.

**Ask:** Treat Bearer Authorization the same as Basic/form for “is this ours?” — if it looks like `zak_…`, a local OIDC token, or `zot_cred_v1…`, redeem or return `401`, never proxy. Please add a regression test that sends these **only** as Bearer and asserts zero upstream calls.

andaaron:
@matheuscscp, they are not, please see my comment.

matheuscscp:
@andaaron Fixed both, thanks

andaaron (dismissed):

matheuscscp:
@andaaron Sorry, just fixed one more issue

andaaron (dismissed):

matheuscscp:
@andaaron Any pending actions from my side?

andaaron:
> @andaaron Any pending actions from my side?

Waiting on @rchincha's approval.
Please also rebase.

matheuscscp:
Rebase done.

pkg/api/bearer_auth.go
```diff
@@ -140,119 +154,206 @@ func (b *BearerAuth) Middleware(ctlr *Controller) mux.MiddlewareFunc {
...
+func requestedBearerResourceAction(request *http.Request) *ResourceAction {
+	if strings.TrimSuffix(request.URL.Path, "/") == constants.RoutePrefix {
+		return nil
+	}
+
+	vars := mux.Vars(request)
+	name := vars["name"]
+	if name == "" {
+		return nil
+	}
```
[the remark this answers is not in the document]
andaaron: @matheuscscp what is your take on this?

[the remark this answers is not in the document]
matheuscscp: Great catch! Fixed!

matheuscscp:
I can't take this non-deterministic loop with copilot anymore! Please merge this PR and let's move on! If someone files a bug we fix it!!! :weary: :weary: :weary:

andaaron:
> I can't take this non-deterministic loop with copilot anymore! Please merge this PR and let's move on! If someone files a bug we fix it!!! 😩 😩 😩

Sorry for re-requesting copilot review, but auth part is critical and sometimes it finds real bugs.

matheuscscp:
Sure but this PR is already huge, so copilot will never stop finding things! Let's do a post-merge review and file small fixes before releasing!

pkg/api/bearer_auth.go
```diff
@@ -140,119 +154,207 @@ func (b *BearerAuth) Middleware(ctlr *Controller) mux.MiddlewareFunc {
...
-						if err := ctlr.MetaDB.SetUserGroups(request.Context(), groups); err != nil {
-							ctlr.Log.Error().Err(err).Str("identity", identity).Msg("failed to update user profile")
-							response.WriteHeader(http.StatusInternalServerError)
+// Authenticate verifies Bearer credentials and saves request-scoped auth details needed downstream.
+func (b *BearerAuth) Authenticate(
+	ctlr *Controller, userAc *reqCtx.UserAccessControl, request *http.Request,
+) (bearerAuthenticationResult, error) {
+	authConfig := b.effectiveAuthConfig(ctlr)
+	result := bearerAuthenticationResult{authConfig: authConfig}
+	header := request.Header.Get("Authorization")
```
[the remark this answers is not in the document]
andaaron: This is a nit.

pkg/api/token_exchange.go
```diff
@@ -62,24 +81,62 @@ func normalizeTokenExchangeRequest(request *http.Request) (*tokenExchangeRequest
...
 	}
 
 	return tokenRequest, nil
 }
 
-func (request *tokenExchangeRequest) addCredential(credential string) {
-	credential = strings.TrimSpace(credential)
-	if credential == "" || slices.Contains(request.credentials, credential) {
+func (request *tokenExchangeRequest) addCredential(username, secret string) {
+	credential := tokenExchangeCredential{
```
rchincha, line 103: # Whitespace Normalization Bypasses Username Validation

Location: token_exchange.go:104-106, authn.go:1011-1012
Issue: API key validation uses the trimmed username (TrimSpace(username)), but identity matching later uses the stored username without trimming:
```golang
credential := tokenExchangeCredential{
    Username: strings.TrimSpace(username),  // Normalized for API key
    Secret:   strings.TrimSpace(secret),
}
```

```
// Later: storedIdentity, err := ctlr.MetaDB.GetUserAPIKeyInfo(hashedKey)
// if storedIdentity != identity {  // This compares ORIGINAL identity to stored
```
Attack: Attacker can use " user " (with spaces) in Basic auth, which gets trimmed during token exchange but might not match identity validation in authenticateAPIKeyCredential() (line 317-318 in authn.go)
Bypass: Could allow unauthorized API key use if identity comparison is inconsistent

rchincha:
### **Most Critical Vulnerabilities:**

| Severity | Issue | Impact | Attack |
|----------|-------|--------|--------|
| 🔴 CRITICAL | **Plaintext API Key in Wrapped Token** | Token easily decrypted to extract credentials | Base64 decode `zot_cred_v1.<base64>` to leak API key |
| 🔴 CRITICAL | **Unsigned Tokens Allow Forgery** | Attackers can craft valid wrapped tokens without auth | Create own `wrappedCredentialBearer{...}` and base64 it with `zot_cred_v1.` prefix |
| 🔴 CRITICAL | **Whitespace Bypass to Upstream** | API keys leaked to third-party token servers | Use `Authorization: Basic   <base64(user:zak_xxx)>` (extra space) to bypass local detection |
| 🔴 CRITICAL | **Username Trim Breaks Identity Match** | API key valid for different users if whitespace differs | Create key for `" user "`, exchange with `"user"` (or vice versa) |

### **Recommended Tests**
1. Can an attacker forge a wrapped token without calling the token endpoint?
2. Does using `Authorization: Basic   <base64>` (extra whitespace) bypass API key detection and leak to upstream?
3. Are wrapped tokens logged anywhere in plaintext?
4. Can multiple wrapped tokens for the same API key be generated?
5. Is there token expiration or revocation?

rchincha:
@matheuscscp 

Let me take over this PR.

matheuscscp:
Please!

matheuscscp:
Thanks!

andaaron:
@matheuscscp, @rchincha 

## `rchincha` CRITICAL table (assessed)

### 1–2. Plaintext / unsigned `zot_cred_v1` — **not vulnerabilities relative to API keys**

```155:163:pkg/api/token_exchange.go
func newWrappedCredentialBearerToken(username, secret string) string {
	payload, _ := json.Marshal(wrappedCredentialBearer{ //nolint:errchkjson,gosec // Fixed scalar struct.
		Version:  wrappedCredentialVersion,
		Type:     wrappedCredentialTypeAPIKey,
		Username: username,
		Secret:   secret,
	})

	return wrappedCredentialBearerPrefix + base64.RawURLEncoding.EncodeToString(payload)
}
```

- PR description already says **non-signed** wrapper.
- Possession of wrapper ≡ possession of `user:zak_…` (same as Basic). Forging a wrapper without the secret does nothing useful; forging *with* the secret is equivalent to using Basic.
- Auth still goes through `authenticateAPIKeyCredential` (hash lookup, identity match, expiry) at `bearer_auth.go:191-209` / `authn.go:317-333`.
- **Do not treat as merge-blocking CVEs** unless product wants HMAC/JWT wrapping as a follow-up (orthogonal hardening).

### 3. Whitespace Basic → upstream leak — **real footgun, overstated severity**

Token exchange collects Basic via Go `request.BasicAuth()`:

```59:62:pkg/api/token_exchange.go
username, password, ok := request.BasicAuth()
if ok && password != "" {
	tokenRequest.addCredential(username, password)
}
```

Verified: `Basic  <b64>` and `Basic\t<b64>` → `BasicAuth()` **ok=false**. Those headers are **not** added as credentials, so `hasAPIKeySecret()` never runs; if upstream is configured, `proxyOIDCBearerTokenExchange` copies `Authorization` (`bearer_auth.go:526`, skip list does **not** drop it at `616-622`).

Meanwhile registry Basic parsing uses `strings.Fields` (`authn.go:1017-1023`) and **would** accept the same header.

- **Preconditions:** upstream token endpoint + non-canonical Basic whitespace + `zak_` in that header.
- Normal Docker/Podman clients use single-space `Basic `.
- Attacker who already has the key can self-exfiltrate; this is not credential theft from others.
- **Reasonable ask (non-blocking or soft block per rchincha):** also parse Basic with `splitAuthorizationHeader` / treat undecodable Basic that Fields-parses as `zak_` as locally owned before proxy — same class as prior “never proxy local secrets” work.

### 4. Username TrimSpace identity bypass — **incorrect / fail-closed**

```102:106:pkg/api/token_exchange.go
func (request *tokenExchangeRequest) addCredential(username, secret string) {
	credential := tokenExchangeCredential{
		Username: strings.TrimSpace(username),
		Secret:   strings.TrimSpace(secret),
	}
```

Exchange passes **already-trimmed** `credential.Username` into `authenticateAPIKeyCredential` (`bearer_auth.go:419`). Mismatch with stored identity → `false` (`authn.go:317-318`), not privilege escalation. Unresolved thread at `token_exchange.go:103` should be closed with that explanation (optional: document that exchange normalizes; direct Basic path does not trim — deny on `" user "`).

andaaron:
IMHO we can merge as is, and fix point 3 in a follow up PR. @rchincha

andaaron (approved):

rchincha:
@matheuscscp thanks for being patient and persevering with this PR. A lot more scrutiny given that it is in the authn path.

matheuscscp:
@rchincha No problem! I will do a post-review and file small hardening follow-ups if needed. Will let you know otherwise. Let's hold on cutting a release until I get back with my conclusions.

PR #4257, opened 2026-07-23
feat: send an http event when an image has been scanned

### LINKED ISSUE #4251: [Feat]: Send an http event when an image has been scanned

There is currently no way to be notified via http when Zot finished an image scan.

Harbor can send a webhook event when an image has been scanned. The event contains the repo, tag and digest of the scanned image, together with a summary of the trivy scan results. 

I would like to see this feature in Zot. We are using these webhook events for posting a summary of the scan result to a Mattermost channel. Example:

Image

_No response_

### Additional context

_No response_

rchincha:
Implements sending an HTTP event when Zot finishes an image scan.

Fixes #4251

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/search/cve/scan.go
+++ b/pkg/extensions/search/cve/scan.go
@@ -290,7 +290,7 @@ func (s *scanner) ScanImage(ctx context.Context, image string) (map[string]cvemo
 // manifests inside a multiarch index, so a tag cannot be required here. ok is false when
 // resolution failed; a warning has already been logged and the caller should skip both the
 // cache check and publishing.
-func (s *scanner) resolveDigestForEvent(ctx context.Context, image string) (repo, ref, digest string, ok bool) {
+func (s *scanner) resolveDigestForEvent(ctx context.Context, image string) (string, string, string, bool) {
 	repo, ref, isTag := zcommon.GetImageDirAndReference(image)
 
 	// When isTag is false, ref is already the digest ScanImage was called with.
--- a/pkg/extensions/search/cve/scan.go
+++ b/pkg/extensions/search/cve/scan.go
@@ -3,8 +3,14 @@ package cveinfo
 import (
 	"context"
 	"fmt"
+	"slices"
 	"sync"
 
+	godigest "github.com/opencontainers/go-digest"
+
+	zcommon "zotregistry.dev/zot/v2/pkg/common"
+	"zotregistry.dev/zot/v2/pkg/extensions/events"
+	cvemodel "zotregistry.dev/zot/v2/pkg/extensions/search/cve/model"
 	"zotregistry.dev/zot/v2/pkg/log"
 	mTypes "zotregistry.dev/zot/v2/pkg/meta/types"
 	reqCtx "zotregistry.dev/zot/v2/pkg/requestcontext"
@@ -190,7 +196,8 @@ func (st *scanTask) DoWork(ctx context.Context) error {
 
 	// We cache the results internally in the scanner
 	// so we can discard the actual results for now
-	if _, err := st.generator.scanner.ScanImage(ctx, image); err != nil {
+	_, err := st.generator.scanner.ScanImage(ctx, image)
+	if err != nil {
 		st.generator.log.Error().Err(err).Str("image", image).Msg("failed to perform scheduled cve scan for image")
 		st.generator.addError(st.digest, err)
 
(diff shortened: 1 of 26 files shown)


pkg/extensions/search/cve/scan.go
```diff
@@ -210,3 +215,120 @@ func (st *scanTask) String() string {
 func (st *scanTask) Name() string {
 	return "ScanTask"
 }
+
+// NewScannerWithEvent wraps scanner so that a successful ScanImage call publishes
+// an ImageScanned event via eventRecorder. metaDB is used to resolve tags for the
+// scanned digest.
+func NewScannerWithEvent(scanner Scanner, metaDB mTypes.MetaDB, eventRecorder events.Recorder,
```
vrajashkr: nit: I'm a tad bit worried that in future if we need another feature for scanner i.e. ScannerWithFeatureX, we will have a lot of these additional objects around.

We would have an independent ScannerWithFeatureX and a ScannerWithEventWithFeatureX, etc.
It might become hard to maintain if more such features come along.

Instead, I suggest having the eventRecorder for the usual scanner and as part of `ScanImage`, if the eventRecorder is not nil, attempt to publish the event.

pkg/extensions/search/cve/scan.go
```diff
@@ -210,3 +215,120 @@ func (st *scanTask) String() string {
...
+
+		matchingTags = append(matchingTags, tag)
+		if mediaType == "" {
+			mediaType = descriptor.MediaType
+		}
+	}
+
+	if len(matchingTags) == 0 {
+		s.log.Warn().Str("repository", repo).Str("digest", digest).
+			Msg("skipping image scanned event because no matching tag was found")
```
vrajashkr: I didn't quite understand this. Is tag mandatory for scanning?

test/blackbox/events_http_scan.bats
```diff
@@ -0,0 +1,160 @@
...
+
+    if ! command -v jq >/dev/null 2>&1; then
+        echo "you need to install jq as a prerequisite to running the tests" >&3
+        return 1
+    fi
+
+    if ! command -v docker >/dev/null 2>&1; then
+        echo "you need to install docker as a prerequisite to running the tests" >&3
+        return 1
+    fi
```
vrajashkr, line 23: nit: Could make this a loop

test/blackbox/events_http_scan.bats
```diff
@@ -0,0 +1,160 @@
...
+    # Use unique config name based on test file name and test run to avoid conflicts
+    export REGISTRY_NAME=$(basename "${BASH_SOURCE[0]}" .bats)-$(basename "${BATS_FILE_TMPDIR}")
+
+    # verify prerequisites are available
+    if ! verify_prerequisites; then
+        exit 1
+    fi
+
+    # Setup http server
+    http_server_port=$(get_free_port_for_service "http")
```
vrajashkr, line 36: I think these also need ports json changes.

test/blackbox/events_http_scan.bats
```diff
@@ -0,0 +1,160 @@
...
+    http_event_dir="${BATS_FILE_TMPDIR}/http_events"
+    http_server_start http_receiver_scan "${http_server_port}" "${http_event_dir}"
+    echo ${http_server_port} > ${BATS_FILE_TMPDIR}/http_server.port
+    wait_for_http_server $http_server_port
+
+    skopeo --insecure-policy copy --format=oci docker://ghcr.io/project-zot/golang:1.20 oci:${TEST_DATA_DIR}/golang:1.20
```
vrajashkr, line 42: Do we need insecure policy for copying to oci disk layout? The source is also ghcr so TLS would work.

vrajashkr (commented):
Looks good overall. Left a few comments.

test/blackbox/events_http_scan.bats
vrajashkr, line 1: Ah, I missed one. this file needs to be added to the test blackbox shell script to actually run it. (which explains why ports json change was also missing).

vrajashkr (changes requested):

pkg/storage/storage_test.go
```diff
@@ -350,6 +350,11 @@ func (c *captureImageEvents) ImageLintFailed(name, reference, digest, mediaType,
 	})
 }
 
+func (c *captureImageEvents) ImageScanned(name, reference, digest, mediaType string,
```
andaaron: This seems incomplete.

pkg/extensions/search/cve/scan.go
andaaron: I think we could implement the options pattern for the scanner,

test/blackbox/events_http_scan.bats
rchincha: The skopeo --insecure-policy global flag adopts a permissive, insecure signature verification policy that allows any image to be processed. It completely bypasses the need for a local configuration file (typically /etc/containers/policy.json). Use it before subcommands like copy or inspect to skip trust verification.

not TLS related.

pkg/storage/storage_test.go
rchincha: comment added.

test/blackbox/events_http_scan.bats
rchincha: fixed

pkg/extensions/search/cve/scan.go
rchincha: no, fixed

test/blackbox/helpers_redis.bash
```diff
@@ -2,7 +2,10 @@
 function redis_start() {
   local cname="$1" # container name
   local free_port="$2"
-  docker run -d --name ${cname} -p ${free_port}:6379 redis
+  # pinned to the ghcr mirror (not bare "redis") so this doesn't depend on reaching Docker Hub,
+  # matching every other blackbox helper container; see helpers_events.bash for what happens
+  # when a helper container isn't self-contained like this.
```
vrajashkr: IMO, it generally makes sense to use your own mirror image for tests to avoid docker hub rate limiting, etc. Given that, I think this comment could be removed.

```suggestion
```

test/blackbox/helpers_events.bash
```diff
@@ -37,69 +37,83 @@ function http_server_start() {
 
     echo "# starting HTTP server container '${cname}' on port ${port}..." >&3
 
+    # stdlib-only server (no pip install): a runtime package install here previously
+    # caused events_http/events_http_lint_failure/events_sink_failure/events_http_scan
+    # to all fail together whenever PyPI was slow/unreachable from the CI runner, even
+    # though every other blackbox helper container is a self-contained ghcr.io mirror.
```
vrajashkr: We probably don't need this comment.
```suggestion
```

test/blackbox/helpers_events.bash
```diff
@@ -37,69 +37,83 @@ function http_server_start() {
...
+            json.dump(event, f, indent=2)
+
+        self.send_response(200)
+        self.end_headers()
+
+    def log_message(self, format, *args):
+        pass  # silence per-request access logs, same as the original server
+
+HTTPServer((\"0.0.0.0\", 8080), Handler).serve_forever()
             " > app.py && python app.py
```
vrajashkr: I know this is existing code, but I think that this python script shouldn't be in the bash file.
I'd suggest creating a new py file with the python code and then mounting it into the container with a volume mount.

We could make our own image with this baked in, but I prefer the file mount approach IMO.

pkg/extensions/search/cve/scan.go
```diff
@@ -210,3 +217,128 @@ func (st *scanTask) String() string {
 func (st *scanTask) Name() string {
 	return "ScanTask"
 }
+
+// ScannerOption configures optional, cross-cutting behavior on a scanner
+// (e.g. event publishing). Additional scanner features should be added as
+// new options rather than new wrapper types.
```
vrajashkr:
```suggestion
// ScannerOption configures additional features and settings on a scanner.
```

vrajashkr (commented):
Some minor follow up comments.

test/blackbox/garbage_collect.bats
andaaron, line 1: The changes in this file should not be needed, since #4261 was merged to fix the same issues.

pkg/extensions/search/cve/scan_test.go
```diff
@@ -41,6 +42,47 @@ var (
...
 
+type scanEventCall struct {
+	name      string
+	reference string
+	digest    string
+	mediaType string
+	summary   events.ImageScanSummary
+}
+
+type scanEventRecorder struct {
```
andaaron: Why not move this to zot/pkg/test/mocks, maybe make it a bit more generic? This looks reusable.

pkg/extensions/search/cve/scan.go
```diff
@@ -210,3 +217,146 @@ func (st *scanTask) String() string {
...
+	metaDB        mTypes.MetaDB
+	eventRecorder events.Recorder
+	log           log.Logger
+}
+
+func (s *scanner) ScanImage(ctx context.Context, image string) (map[string]cvemodel.CVE, error) {
+	if s.eventRecorder == nil || s.metaDB == nil {
+		return s.Scanner.ScanImage(ctx, image)
+	}
+
```
andaaron: In case the event recorder is used this function calls some other nethods which makes calls to MetaDB, for information which is already available inside the s.Scanner.ScanImage() execution.

I propose to remove the extra logic to retrieve information from metadb here, and change the scanner interface to have Scanner.ScanImage() return a “scan result” struct that includes the resolved digest, mediaType, and information on cache status alongside cveMap. Then the wrapper can emit without extra metaDB calls.

test/blackbox/events_http_scan.bats
```diff
@@ -0,0 +1,168 @@
...
+    if ! verify_prerequisites; then
+        exit 1
+    fi
+
+    # Setup http server
+    http_server_port=$(get_free_port_for_service "http")
+    http_event_dir="${BATS_FILE_TMPDIR}/http_events"
+    http_server_start http_receiver_scan "${http_server_port}" "${http_event_dir}"
+    echo ${http_server_port} > ${BATS_FILE_TMPDIR}/http_server.port
+    wait_for_http_server $http_server_port
```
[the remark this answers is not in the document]
andaaron: I think this is a good fix to have.

andaaron (approved):

PR #4258, opened 2026-07-24
feat: add ztsd layer support for cve scans

### LINKED ISSUE #4255: [Bug]: CVE scanner rejects OCI images containing `application/vnd.oci.image.layer.v1.tar+zstd`
### zot version

ghcr.io/project-zot/zot-linux-amd64:latest

### Describe the bug

The CVE scanner rejects OCI images containing `application/vnd.oci.image.layer.v1.tar+zstd` layers before scanning starts. Since zstd-compressed OCI layers are part of the OCI Image Specification and increasingly common, the scanner should accept this media type and delegate decompression to the underlying image/scanner library instead of rejecting it up front.

Code in `pkg/extensions/search/cve/trivy/scanner.go`, zstd should be added to the `case` section:

```go
func (scanner Scanner) isManifestScanable(digestStr string) (bool, error) {
	if scanner.cache.Get(digestStr) != nil {
		return true, nil
	}

	manifestData, err := scanner.metaDB.GetImageMeta(godigest.Digest(digestStr))
	if err != nil {
		return false, err
	}

	for _, imageLayer := range manifestData.Manifests[0].Manifest.Layers {
		switch imageLayer.MediaType {
		case ispec.MediaTypeImageLayerGzip, ispec.MediaTypeImageLayer, string(regTypes.DockerLayer):
			continue
		default:
			return false, fmt.Errorf("%w: layer media type '%s'", zerr.ErrScanNotSupported, imageLayer.MediaType)
		}
	}

	return true, nil
}
```

### To reproduce

As example push `dhi.io/node:26-debian13-dev` into the zot registry. In the UI hoovering over the bug icon results in "Failed to scan".

The logs of the zot container shows:
```bash
image is not scannable
...
scanning is not supported for given media type:
application/vnd.oci.image.layer.v1.tar+zstd
```

### Expected behavior

Image should be scanned, AFAIK trivy is able to scan such images.

### Screenshots

Image

### Additional context

_No response_

rchincha:
Fixes issue #4255

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/search/cve/trivy/scanner.go
+++ b/pkg/extensions/search/cve/trivy/scanner.go
@@ -477,10 +477,14 @@ func (scanner Scanner) isManifestScanable(digestStr string) (bool, error) {
 
 	for _, imageLayer := range manifestData.Manifests[0].Manifest.Layers {
 		switch imageLayer.MediaType {
-		case ispec.MediaTypeImageLayerGzip, ispec.MediaTypeImageLayer, string(regTypes.DockerLayer):
+		case ispec.MediaTypeImageLayerGzip,
+			ispec.MediaTypeImageLayerZstd,
+			ispec.MediaTypeImageLayer,
+			string(regTypes.DockerLayer):
 			continue
 		default:
-			return false, fmt.Errorf("%w: layer media type '%s'", zerr.ErrScanNotSupported, imageLayer.MediaType)
+			return false, fmt.Errorf("%w: layer media type '%s'",
+				zerr.ErrScanNotSupported, imageLayer.MediaType)
 		}
 	}
 
@@ -494,7 +498,10 @@ func (scanner Scanner) isManifestDataScannable(manifestData mTypes.ManifestMeta)
 
 	for _, imageLayer := range manifestData.Manifest.Layers {
 		switch imageLayer.MediaType {
-		case ispec.MediaTypeImageLayerGzip, ispec.MediaTypeImageLayer, string(regTypes.DockerLayer):
+		case ispec.MediaTypeImageLayerGzip,
+			ispec.MediaTypeImageLayerZstd,
+			ispec.MediaTypeImageLayer,
+			string(regTypes.DockerLayer):
 			continue
 		default:
 			return false, fmt.Errorf("%w: layer media type '%s'", zerr.ErrScanNotSupported, imageLayer.MediaType)
--- a/pkg/extensions/search/cve/trivy/scanner.go
+++ b/pkg/extensions/search/cve/trivy/scanner.go
@@ -446,7 +446,7 @@ func (scanner Scanner) IsImageMediaScannable(repo, digestStr, mediaType string) 
 
 	if mediaType == ispec.MediaTypeImageManifest || //nolint:gocritic // not converting to switch-case
 		compat.IsCompatibleManifestMediaType(mediaType) {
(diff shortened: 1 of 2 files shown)


andaaron (dismissed):

pkg/extensions/search/cve/trivy/scanner.go
```diff
@@ -475,12 +475,20 @@ func (scanner Scanner) isManifestScanable(digestStr string) (bool, error) {
 		return false, err
 	}
 
+	if len(manifestData.Manifests) == 0 {
+		return false, zerr.ErrUnexpectedMediaType
```
andaaron: I think this should be more explicit, otherwise we'll not know what causes the error.

```suggestion
		return false, fmt.Errorf("%w:  manifest data has 0 manifests", zerr.ErrScanNotSupported)
```

pkg/extensions/search/cve/trivy/scanner.go
```diff
@@ -494,7 +502,10 @@ func (scanner Scanner) isManifestDataScannable(manifestData mTypes.ManifestMeta)
 
 	for _, imageLayer := range manifestData.Manifest.Layers {
 		switch imageLayer.MediaType {
-		case ispec.MediaTypeImageLayerGzip, ispec.MediaTypeImageLayer, string(regTypes.DockerLayer):
+		case ispec.MediaTypeImageLayerGzip,
+			ispec.MediaTypeImageLayerZstd,
+			ispec.MediaTypeImageLayer,
+			string(regTypes.DockerLayer):
```
vrajashkr: This logic is duplicated from above. I don't get why we have both `isManifestScanable` and `isManifestDataScannable`. They seem to be checking the same thing for layers.

That said, if we need to keep both, it would be good to extract the layer checking logic to a common function.

pkg/extensions/search/cve/trivy/scanner_internal_test.go
```diff
@@ -452,6 +452,18 @@ func TestImageScannable(t *testing.T) {
...
+		Layers([]Layer{{
+			MediaType: ispec.MediaTypeImageLayerZstd,
+			Digest:    ispec.DescriptorEmptyJSON.Digest,
+			Blob:      ispec.DescriptorEmptyJSON.Data,
+		}}).ImageConfig(validConfig).Build()
+
+	err = metaDB.SetRepoReference(context.Background(), "repo1", "valid-zstd", validZstdImage.AsImageMeta())
+	if err != nil {
+		panic(err)
+	}
```
vrajashkr, line 465: Do we need to panic here? Perhaps, just failing the test would suffice.

vrajashkr (commented):
Some comments on the changes - logic dedupe + failing test instead of panic.

pkg/extensions/search/cve/trivy/scanner.go
rchincha: one call is within a manifest and other is for manifest(s) inside an index

fixed now

pkg/extensions/search/cve/trivy/scanner.go
rchincha: fixed

vrajashkr (dismissed):
LGTM

andaaron (dismissed):

andaaron (approved):

PR #4260, opened 2026-07-25
fix(storage): check table existence before CreateTable in dynamodb cache

### LINKED ISSUE #4259: [Bug]: DynamoDB creation permission even if the describe permission are there
### zot version

v2.1.18-7bb211bcd4352b90f3e99752607fbd1f050bf7ca

### Describe the bug

Hello, 
I post a comment of #4066 but I'm not sure closed issue are monitored.  

Here is my IAM policy on AWS: 
```json
{
	"Statement": [
		{
			"Action": [
				"dynamodb:UpdateItem",
				"dynamodb:Scan",
				"dynamodb:Query",
				"dynamodb:PutItem",
				"dynamodb:PartiQLUpdate",
				"dynamodb:PartiQLSelect",
				"dynamodb:PartiQLInsert",
				"dynamodb:PartiQLDelete",
				"dynamodb:ListTagsOfResource",
				"dynamodb:GetItem",
				"dynamodb:DescribeTable",
				"dynamodb:DeleteItem",
				"dynamodb:ConditionCheckItem",
				"dynamodb:BatchWriteItem",
				"dynamodb:BatchGetItem"
			],
			"Effect": "Allow",
			"Resource": [ <REDACTED>],
		},
	]
}
```
Here is a test on my machine: 

```json

aws dynamodb describe-table --table-name <REDACTED>
{
		...
		"TableStatus": "ACTIVE",
		...
}
```

And here is the Zot error on startup:
> zot[217949]: Error: operation error DynamoDB: CreateTable, https response error StatusCode: 400, RequestID: <REDACTED>  api error AccessDeniedException: User: <REDACTED> is not authorized to perform: dynamodb:CreateTable on resource: <REDACTED>  because no identity-based policy allows the dynamodb:CreateTable action

Is there anything missing or an issue on the feature ? 

### To reproduce

See previously.

### Expected behavior

Zot start without complaining

### Screenshots

_No response_

### Additional context

_No response_

rchincha:
The storage-cache DynamoDB driver's NewTable always called CreateTable on startup and only tolerated the error when it happened to contain "Table already exists" (a ResourceInUseException, returned when the caller has CreateTable permission and the table exists). When the caller instead lacks dynamodb:CreateTable entirely -- even with full read/write/describe access to an existing table -- IAM denies the API call before DynamoDB ever checks the table's existence, so the driver got an AccessDeniedException it didn't recognize and zot failed to start.

Add tableExists (via DescribeTable, which the caller does have permission for) and only call CreateTable when the table is genuinely missing, mirroring the pattern already used by the metaDB DynamoDB driver in pkg/meta/dynamodb/dynamodb.go. Use errors.As against the typed SDK exceptions instead of matching on err.Error() substrings.

Added an internal regression test that intercepts only the CreateTable call against a real (localstack) DynamoDB endpoint to simulate the IAM policy from the report, proving it fails on the old code and passes with this fix.

Fixes https://github.com/project-zot/zot/issues/4259

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/storage/cache/dynamodb.go
+++ b/pkg/storage/cache/dynamodb.go
@@ -81,9 +81,22 @@ func (d *DynamoDBDriver) NewTable(tableName string) error {
 
 		// tableExists and CreateTable aren't atomic, so still tolerate a benign race where
 		// another zot instance created the table between the check above and this call.
+		// ResourceInUseException can also occur while a table is being deleted, so confirm
+		// the table actually exists before treating the error as benign.
 		var inUseErr *types.ResourceInUseException
-		if err != nil && !errors.As(err, &inUseErr) {
-			return err
+		if err != nil {
+			if !errors.As(err, &inUseErr) {
+				return err
+			}
+
+			exists, existsErr := d.tableExists(tableName)
+			if existsErr != nil {
+				return existsErr
+			}
+
+			if !exists {
+				return err
+			}
 		}
 	}
 
--- a/pkg/storage/cache/dynamodb.go
+++ b/pkg/storage/cache/dynamodb.go
@@ -33,8 +33,8 @@ type Blob struct {
 }
 
 // tableExists checks for the table via DescribeTable rather than attempting CreateTable
-// and inspecting the error. This allows callers that only have read/write item permissions on an
-// existing table (but not dynamodb:CreateTable) to work.
+// and inspecting the error. This allows callers that have dynamodb:DescribeTable (and read/write item
+// permissions) on an existing table, but not dynamodb:CreateTable, to work.
 func (d *DynamoDBDriver) tableExists(tableName string) (bool, error) {
 	_, err := d.client.DescribeTable(context.TODO(), &dynamodb.DescribeTableInput{
 		TableName: aws.String(tableName),
--- a/pkg/storage/cache/dynamodb.go
+++ b/pkg/storage/cache/dynamodb.go
@@ -2,8 +2,8 @@ package cache
 
 import (
 	"context"
+	"errors"
 	"slices"
-	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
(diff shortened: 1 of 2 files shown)


andaaron (dismissed):

pkg/storage/cache/dynamodb.go
```diff
@@ -32,29 +32,60 @@ type Blob struct {
 	OriginalBlobPath  string   `dynamodbav:"OriginalBlobPath,string"`
 }
 
+// tableExists checks for the table via DescribeTable rather than attempting CreateTable
+// and inspecting the error: callers that only have read/write item permissions on an
+// existing table (but not dynamodb:CreateTable) would otherwise get an AccessDeniedException
+// from CreateTable instead of ever finding out the table is already there. See #4259.
```
vrajashkr:
```suggestion
// tableExists checks for the table via DescribeTable rather than attempting CreateTable
// and inspecting the error. This allows callers that only have read/write item permissions on an
// existing table (but not dynamodb:CreateTable) to work.
```

pkg/storage/cache/dynamodb_internal_test.go
```diff
@@ -0,0 +1,87 @@
...
+	. "github.com/smartystreets/goconvey/convey"
+
+	"zotregistry.dev/zot/v2/pkg/log"
+	tskip "zotregistry.dev/zot/v2/pkg/test/skip"
+)
+
+// denyCreateTableTransport simulates an IAM policy that allows DescribeTable/GetItem/etc.
+// but denies dynamodb:CreateTable, by intercepting only CreateTable calls and returning
+// the same AccessDeniedException shape AWS returns; every other action is forwarded to
+// the real (localstack) endpoint. This reproduces https://github.com/project-zot/zot/issues/4259.
```
vrajashkr:
```suggestion
// the real (localstack) endpoint.
```

vrajashkr (commented):
Code LGTM. A couple of comment nits if you find them useful.

pkg/storage/cache/dynamodb.go
rchincha: fixed

pkg/storage/cache/dynamodb_internal_test.go
rchincha: fixed

andaaron (approved):

PR #4261, opened 2026-07-25
fix(ci): stabilize blackbox GC and HTTP events tests

andaaron:
Use a small alpine image in garbage_collect so pushes finish within gcDelay, and replace the Flask/pip HTTP sink with a stdlib server so events tests are not gated on PyPI install time.

### DIFF
--- a/pkg/extensions/sync/sync_test.go
+++ b/pkg/extensions/sync/sync_test.go
@@ -7702,9 +7702,8 @@ func TestSyncImageIndex(t *testing.T) {
 				dcm.StartAndWait(dctlr.Config.HTTP.Port)
 				defer dcm.StopServer()
 
-				// give it time to set up sync
-				t.Logf("waitsync(%s, %s)", dctlr.Config.Storage.RootDirectory, "index")
-				waitSync(dctlr.Config.Storage.RootDirectory, "index")
+				// Wait for SyncRepo to finish processing all tags for the "index" repo.
+				So(waitFinishedSyncingRepo(dctlr.Config.Log.Output, "index", 60*time.Second), ShouldBeTrue)
 
 				resp, err = resty.R().SetHeader("Content-Type", ispec.MediaTypeImageIndex).
 					Get(destBaseURL + "/v2/index/manifests/root")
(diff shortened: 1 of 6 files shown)


test/blackbox/helpers_events.bash
```diff
@@ -37,81 +37,96 @@ function http_server_start() {
...
+            \"body\": body,
+        }
+        with open(f\"/data/{counter}.json\", \"w\") as f:
+            json.dump(event, f, indent=2)
+        self.send_response(200)
+        self.end_headers()
+
+# Single-threaded so counter + /data/{n}.json stay deterministic under concurrent POSTs.
+HTTPServer((\"0.0.0.0\", 8080), Handler).serve_forever()
             " > app.py && python app.py
```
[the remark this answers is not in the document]
andaaron: Generated app.py starts with import base64 (no leading space) and parses cleanly. No change needed.

test/blackbox/helpers_events.bash
```diff
@@ -140,5 +156,10 @@ function wait_for_http_server() {
     done
 
     echo "# HTTP server failed to start on port ${port} after ${timeout}s" >&3
+    if [ -n "${cname}" ]; then
+        echo "# docker logs for '${cname}':" >&3
+        docker logs "${cname}" 2>&1 | head -50 >&3 || true
```
vrajashkr: I didn't quite catch why we only want the first 50 lines? Why not the full logs?

test/blackbox/helpers_events.bash
```diff
@@ -37,81 +37,96 @@ function http_server_start() {
...
     sleep 2
 
     # Check if container is running
-    if docker ps --format "table {{.Names}}" | grep -q "^${cname}$"; then
+    if docker ps --format "{{.Names}}" | grep -qx "${cname}"; then
         echo "# HTTP server container '${cname}' started successfully" >&3
     else
         echo "# WARNING: HTTP server container '${cname}' may not have started properly" >&3
-        docker logs "${cname}" 2>&1 | head -10 >&3
+        docker logs "${cname}" 2>&1 | head -20 >&3 || true
```
vrajashkr: I see that we previously only had the first 10 lines, but I'm curious why we didn't want the full logs.

test/blackbox/helpers_events.bash
vrajashkr, line 118: IMO, if it works just as well, I think I would suggest to have app.py be a separate file checked into repo that gets mounted into the http receiver container instead of being echoed into a file as part of BATS.

vrajashkr (commented):
Looks good overall. Left a few comments and questions.

test/blackbox/helpers_events.bash
rchincha: It was inline code and refactored into another inline code. So wrt this PR, probably ok - seems to be working right now.

vrajashkr (dismissed):

pkg/extensions/sync/sync_test.go
```diff
@@ -8507,3 +8506,44 @@ func waitSyncFinish(logPath string) bool {
...
+					continue
+				}
+
+				if entry.Message == "sync: finished syncing repo" && entry.Repo == repo {
+					return true
+				}
+			}
+		}
+
+		time.Sleep(10 * time.Millisecond)
```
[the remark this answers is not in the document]
vrajashkr: Zot logs within the scope of a test are not too large so IMO this is fine.

That said, sync takes a bit of time so 10ms is a bit on the shorter side. Maybe we can check once every 100ms?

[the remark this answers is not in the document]
andaaron: Pushed.

vrajashkr (approved):
LGTM

PR #4265, opened 2026-07-27
fix(ci): stabilize blackbox NATS events subscriber wait

andaaron:
Raise nats sub --wait from 5s to 60s and poll for dump files after push/delete so ImageUpdated is not lost under concurrent CI load.


rchincha (approved):

vrajashkr (approved):
LGTM

PR #4269, opened 2026-07-28
feat(sync): support RFC 8693 token exchange in oauth2 helper

### LINKED ISSUE #4267: [Feat]: RFC 8693 token exchange grant for the sync oauth2 credential helper

We run zot on-prem as a pull-through cache in front of Google Artifact Registry, authenticating with Workload Identity Federation so that no long-lived credential exists on disk anywhere.

The `oauth2` helper from #4147 gets us most of the way there — re-reading `assertionFile` on every refresh is exactly what a Kubernetes projected service account token needs. What is missing is the grant.

Reaching a security token service requires the RFC 8693 token-exchange grant, which sends `subject_token`, `subject_token_type`, `requested_token_type` and `audience`. `requestValues` can only send `assertion` (jwt-bearer) or `client_assertion` (client_credentials), so both existing grants are rejected by Google STS before the token is even inspected:

```
400 {"error":"invalid_request","error_description":"subject_token must be nonempty."}
```

PR #4147 does list Artifact Registry as a target, and that is fair: the `signingFile` path reaches it through `oauth2.googleapis.com` with the jwt-bearer grant. But that path requires zot to hold a service account private key. Workload Identity Federation exists precisely to remove that key, and it is only reachable through `sts.googleapis.com` with token exchange.

Add token exchange as a third grant in the same helper:

- three optional fields on `OAuth2HelperConfig` — `audience`, `subjectTokenType` and `requestedTokenType` — read only for this grant;
- one branch in `requestValues` that sends the assertion as `subject_token`;
- validation requiring `audience` for this grant, and rejecting the three fields for the others so a misconfiguration fails at load time.

```json
"credentialHelper": "oauth2",
"oauth2CredentialHelper": {
  "tokenURL": "https://sts.googleapis.com/v1/token",
  "assertionFile": "/var/run/secrets/tokens/registry-token",
  "grantType": "urn:ietf:params:oauth:grant-type:token-exchange",
  "audience": "//iam.googleapis.com/projects/N/locations/global/workloadIdentityPools/POOL/providers/PROVIDER",
  "scopes": ["https://www.googleapis.com/auth/cloud-platform"],
  "username": "oauth2accesstoken"
}
```

The token response shape is unchanged, so `fetchToken`, the token cache and the refresh machinery all stay as they are. Existing configurations are untouched: the new fields are optional and only read for a grant that cannot currently succeed.

These are plain RFC 8693 parameters, so this works against any compliant STS rather than adding a GCP-specific helper alongside `ecr`.

- A dedicated `gcp` helper: pulls in a vendor SDK to do what three standard parameters already cover.
- An external `docker-credential-*` helper via regclient's native `CredHelper`: needs an extra binary inside the distroless image and adds an exec surface, rejected for the same reasons given in #2650 and #4146.
- Impersonating a service account after the exchange: turns out to be unnecessary, Artifact Registry accepts the federated token directly.

### Additional context

This does not close #4146, which asked for the jwt-bearer helper that #4147 delivered.

I have this implemented and verified end to end: with a workload identity pool and a short-lived subject token on disk as the only credential, zot pulls a real image from Artifact Registry — matching manifest digest, byte-identical layer, persisted to local storage.

frauniki:
**What type of PR is this?**
feature

Fixes #4267

Adds the RFC 8693 token-exchange grant to the `oauth2` sync credential helper, so that sync can obtain a short-lived registry token from a security token service. Motivation and background are in #4267.

- `OAuth2HelperConfig` gains three optional fields — `audience`, `subjectTokenType` and `requestedTokenType` — read only for this grant.
- `requestValues` sends the assertion as `subject_token`, with `subject_token_type`, `requested_token_type` and `audience`. The two types default to the RFC 8693 JWT and access-token URNs.
- Validation requires `audience` for this grant and rejects the three fields under any other grant, so a misconfiguration fails at load time.
- New example config.

The token response shape is unchanged, so `fetchToken`, the token cache and the refresh machinery are untouched. The change is scoped away from `service.go` so it does not conflict with #4209.

Unit tests in `oauth2_credential_helper_test.go` cover the parameters sent, the token-type defaults and overrides, the subject token being re-read on every refresh, and both validation rules. Config loading tests in `root_test.go` cover the same rules.

Also verified end to end against a real workload identity pool and Google Artifact Registry: with only a short-lived subject token on disk, zot pulls a real image through as a pull-through cache — matching manifest digest, byte-identical layer, persisted to local storage.

**Automation added to e2e**:
No. The grant needs a live security token service to exercise meaningfully, so coverage is unit tests plus the manual run above.

No. The three fields are optional and only read when `grantType` is the token-exchange URN, which no existing configuration can be using successfully today.

Yes, additive only: three new optional keys under `oauth2CredentialHelper`.

**release-note**
The sync `oauth2` credential helper now supports the RFC 8693 token-exchange grant, enabling Workload Identity Federation against registries such as Google Artifact Registry with no long-lived credential on disk. New optional keys under `oauth2CredentialHelper`: `audience` (required for this grant), `subjectTokenType` and `requestedTokenType`.

### DIFF
--- a/examples/config-sync-oauth2-token-exchange.json
+++ b/examples/config-sync-oauth2-token-exchange.json
@@ -1,0 +1,41 @@
+{
+    "distSpecVersion": "1.1.1",
+    "storage": {
+        "rootDirectory": "/tmp/zot"
+    },
+    "http": {
+        "address": "0.0.0.0",
+        "port": "8080"
+    },
+    "log": {
+        "level": "debug"
+    },
+    "extensions": {
+        "sync": {
+            "downloadDir": "/tmp/zot",
+            "registries": [
+                {
+                    "urls": [
+                        "https://registry.example.com"
+                    ],
+                    "onDemand": true,
+                    "maxRetries": 5,
+                    "retryDelay": "2m",
+                    "credentialHelper": "oauth2",
+                    "oauth2CredentialHelper": {
+                        "tokenURL": "https://sts.example.com/v1/token",
+                        "assertionFile": "./examples/sync-oauth2-assertion.jwt",
+                        "grantType": "urn:ietf:params:oauth:grant-type:token-exchange",
+                        "audience": "//sts.example.com/pools/the-pool/providers/the-provider",
+                        "subjectTokenType": "urn:ietf:params:oauth:token-type:jwt",
+                        "requestedTokenType": "urn:ietf:params:oauth:token-type:access_token",
+                        "scopes": [
+                            "repository:pull"
+                        ],
+                        "username": "<token>"
+                    }
+                }
+            ]
+        }
+    }
+}
--- a/pkg/cli/server/root_test.go
+++ b/pkg/cli/server/root_test.go
@@ -283,6 +283,109 @@ func TestLoadConfigurationSyncCredentialHelper(t *testing.T) {
 		err := cli.LoadConfiguration(cfg, tmpfile)
 		So(err, ShouldNotBeNil)
 	})
+
+	Convey("the token-exchange grant loads with its RFC 8693 fields", t, func() {
+		content := `{
+			"storage": {"rootDirectory": "/tmp/zot"},
+			"http": {"address": "127.0.0.1", "port": "8080"},
+			"extensions": {
+				"sync": {
+					"registries": [
+						{
+							"urls": ["https://registry.example.com"],
+							"onDemand": true,
+							"credentialHelper": "oauth2",
+							"oauth2CredentialHelper": {
+								"tokenURL": "https://sts.example.com/v1/token",
+								"assertionFile": "/run/secrets/subject-token",
+								"grantType": "urn:ietf:params:oauth:grant-type:token-exchange",
+								"audience": "//sts.example.com/pools/the-pool/providers/the-provider",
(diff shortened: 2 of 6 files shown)


andaaron (dismissed):

rchincha:
@frauniki 

does it make sense to do a little more input validation?

strings.TrimSpace(...) for non-empty-after-trim checks
net/url.Parse(...) + scheme check (u.Scheme == "urn") for basic URN shape validation - so that subjectTokenType and requestedTokenType are `urn:` etc

otherwise lgtm

frauniki:
@rchincha 
Trim checks added in cf337ef3 — a whitespace-only `tokenURL`, assertion source or `audience` now fails at config load.

For the token types I used an absolute-URI check rather than `urn`: RFC 8693 section 3 permits other schemes, so pinning `urn` would lock out a provider using e.g. `https://schemas.example.com/legacy-token`. It still rejects `jwt`.

No scheme check on the audience — Google's workload identity audience is `//iam.googleapis.com/...`, which parses with no scheme, so that check would reject the case this PR is for. RFC 8693 section 2.1 calls it a logical name rather than a URI.

andaaron (approved):

matheuscscp:
My post-mortem review:  LGTM! :rocket:

PR #4270, opened 2026-07-28
fix(sync): refresh helper credentials on every sync entry point

### LINKED ISSUE #4268: [Bug]: sync credential helpers are never refreshed for periodic (pollInterval) sync
### zot version

main @ 037de220

### Describe the bug

`refreshRegistryTemporaryCredentials` is only called from `SyncImage`, the on-demand path (`pkg/extensions/sync/service.go:344`). The periodic paths — `SyncRepo`, `GetNextRepo` and `SyncReferrers` — never call it.

The result is that for a registry configured with `pollInterval` rather than `onDemand`, whatever credentials the helper returned at startup are used forever. Both built-in helpers hand out expiring tokens, so periodic sync eventually fails: quickly for the `oauth2` helper, after the token lifetime for `ecr`.

### To reproduce

1. Configuration: a sync registry with `credentialHelper: "oauth2"`, `pollInterval: 2s`, `content: [{"prefix": "**"}]` and `onDemand: false`, pointing at a token endpoint that returns a short `expires_in`.
2. Client tool used: none needed; count the requests arriving at the token endpoint.
3. Seen error: the token endpoint is hit exactly once, at startup, and never again — even though the scheduler is running and successfully syncing images the whole time.

For contrast, the same helper and registry with `onDemand: true` refreshes on every pull. I measured one hit at startup and zero further hits over 14 seconds of polling in the periodic case, versus three additional hits for three on-demand pulls.

### Expected behavior

Periodic sync should refresh expiring credentials the same way on-demand sync does, so that a `pollInterval`-driven mirror keeps working past the first token expiry.

### Additional context

I have not sent a patch because the fix interacts with locking: `refreshRegistryTemporaryCredentials` ends by calling `service.init()`, which takes `clientLock`, and calling it from the scheduler goroutine needs care about lock ordering. #4209 is already open against that same function. Happy to work on it once that lands, if the behaviour is indeed unintended rather than a deliberate trade-off.

frauniki:
**What type of PR is this?**
bug

Fixes #4268

`refreshRegistryTemporaryCredentials` was only called from `SyncImage`, so a registry driven by `pollInterval` kept using whatever credentials the helper returned at startup. Both built-in helpers hand out expiring tokens, so periodic sync eventually starts failing against the upstream.

This calls it from `GetNextRepo`, `SyncRepo` and `SyncReferrers` as well.

Each call is placed before any read lock is taken. Since #4272, `refreshRegistryTemporaryCredentials` acquires `clientLock` for reading and then for writing, so calling it while a read lock is held would deadlock. `SyncReferrers` in particular took `clientLock.RLock()` as its first statement, so the refresh goes above it.

One correction to the issue as I filed it: `SyncReferrers` is an on-demand path (`on_demand.go`, `routes.go`), not a periodic one. So the gap is slightly wider than "periodic sync only" — an `onDemand`-only deployment serving `GET /v2/<name>/referrers/<digest>` was also using stale credentials.

Added `TestCredentialRefreshOnEverySyncEntryPoint` in `sync_internal_test.go`, using a counting `CredentialHelper` against an unroutable upstream so each entry point fails fast after the refresh has run. It asserts that `GetNextRepo`, `SyncRepo`, `SyncReferrers` and `SyncImage` all refresh expiring credentials, that valid credentials are left alone, that nothing happens when no helper is configured, and that a refresh that fails is logged rather than returned to the caller. Passes under `-race`.

Also measured against a real zot with the `oauth2` helper and a token endpoint returning a short `expires_in`, configured with `pollInterval: 2s` and `onDemand: false`. Before the change the token endpoint was hit exactly once, at startup, over 14 seconds of polling while the scheduler was successfully syncing images. After the change it is hit on every poll cycle.

**Note on #4272**:

Rebased on top of it, and the merge-order question I raised earlier no longer applies: the data race in `refreshRegistryTemporaryCredentials` is fixed on main. Its fast path also keeps this cheap — when nothing has expired a refresh costs a read lock and one `AreCredentialsValid` call per host, so calling it per repo on every poll cycle is not a concern.

No.

No, other than credentials now being refreshed when they should have been.

**release-note**
Sync now refreshes credential-helper tokens for periodic (`pollInterval`) sync and for on-demand referrer syncs, not just for on-demand image pulls. Previously those paths kept using the credentials obtained at startup and failed once the token expired.

### DIFF
--- a/pkg/extensions/sync/service.go
+++ b/pkg/extensions/sync/service.go
@@ -312,6 +312,12 @@ func (service *BaseService) getNextRepoFromCatalog(lastRepo string) string {
 func (service *BaseService) GetNextRepo(lastRepo string) (string, error) {
 	var err error
 
+	/* Refresh before taking the read lock: a refresh reinitializes the client under the
+	write lock, which a held read lock would deadlock against. */
+	if err := service.refreshRegistryTemporaryCredentials(); err != nil {
+		service.log.Error().Err(err).Msg("failed to refresh credentials")
+	}
+
 	if len(service.repositories) == 0 {
 		service.clientLock.RLock()
 		service.repositories, err = service.remote.GetRepositories(context.Background())
@@ -385,6 +391,12 @@ func (service *BaseService) SyncImage(ctx context.Context, repo, reference strin
 func (service *BaseService) SyncReferrers(ctx context.Context, repo string,
 	subjectDigestStr string, referenceTypes []string,
 ) error {
+	/* Refresh before taking the read lock: a refresh reinitializes the client under the
+	write lock, which a held read lock would deadlock against. */
+	if err := service.refreshRegistryTemporaryCredentials(); err != nil {
+		service.log.Error().Err(err).Msg("failed to refresh credentials")
+	}
+
 	service.clientLock.RLock()
 	defer service.clientLock.RUnlock()
 
@@ -467,6 +479,10 @@ func (service *BaseService) SyncReferrers(ctx context.Context, repo string,
 func (service *BaseService) SyncRepo(ctx context.Context, repo string) error {
 	service.log.Info().Str("repo", repo).Str("registry", service.remote.GetHostName()).
 		Msg("sync: syncing repo")
+
+	if err := service.refreshRegistryTemporaryCredentials(); err != nil {
+		service.log.Error().Err(err).Msg("failed to refresh credentials")
+	}
 
 	var err error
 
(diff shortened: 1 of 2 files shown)


frauniki:
@rchincha @andaaron — could one of you take a look at this when you get a chance?

No reviewer was assigned automatically here. CODEOWNERS gives `/pkg/extensions/sync/` to `@peusebiu`, but that account no longer exists on GitHub, so the rule matches nobody and the top-level owners are not consulted for changes confined to that directory. #4209 is in the same position: it touches only that directory, has no human reviewer, and has been open since 10 July. It might be worth refreshing CODEOWNERS.

On the change itself: it is closely related to #4209. That PR fixes the data race in `refreshRegistryTemporaryCredentials`, and this one adds three more callers of it. The description covers how the two interact and why I do not think this needs to wait on it, but I am happy to be told otherwise on merge order.

andaaron (dismissed):

andaaron:
Thank you for the PR, I opened another https://github.com/project-zot/zot/pull/4272 to include the fix reported in https://github.com/project-zot/zot/pull/4209

andaaron:
@frauniki, please rebase and resolve the conflict.

andaaron (approved):

PR #4274, opened 2026-07-28
fix: mask event headers when event config is present

andaaron:


vrajashkr (approved):
LGTM

PR #4283, opened 2026-07-30
test(gql): add more unit tests

vrajashkr:
**What type of PR is this?**
tests

None

- Adds some helpful unit tests for graphql things.

N/A

N/A - all unit tests

**Automation added to e2e**:
N/A

Nope

No

### DIFF
--- a/pkg/extensions/search/base_image_list_gql_test.go
+++ b/pkg/extensions/search/base_image_list_gql_test.go
@@ -235,3 +235,193 @@ func TestBaseImageListGql(t *testing.T) {
 		So(respImg.Vulnerabilities.MaxSeverity, ShouldEqual, "NONE")
 	})
 }
+
+func TestBaseImageListGqlAuthorization(t *testing.T) {
+	Convey("Base image list for a repo should only be accessible to authorized users", t, func() {
+		rootDir := t.TempDir()
+		adminUserName := "admin"
+		adminPassword := "admin123"
+		normalUserName := "user"
+		normalPassword := "user123"
+		adminUserLine := GetBcryptCredString(adminUserName, adminPassword)
+		normalUserLine := GetBcryptCredString(normalUserName, normalPassword)
+		htpasswdPath := MakeHtpasswdFileFromString(t, adminUserLine+"\n"+normalUserLine)
+
+		conf := config.New()
+		conf.HTTP.Port = "0"
+		conf.HTTP.Auth.HTPasswd = config.AuthHTPasswd{Path: htpasswdPath}
+		conf.Storage.RootDirectory = rootDir
+		defaultVal := true
+
+		searchConfig := &extconf.SearchConfig{
+			BaseConfig: extconf.BaseConfig{Enable: &defaultVal},
+		}
+		conf.Extensions = &extconf.ExtensionConfig{
+			Search: searchConfig,
+		}
+
+		conf.HTTP.AccessControl = &config.AccessControlConfig{
+			Repositories: config.Repositories{
+				"admin/**": config.PolicyGroup{
+					Policies: []config.Policy{
+						{
+							Users:   []string{"admin"},
+							Actions: []string{"create", "read", "delete"},
+						},
+					},
+				},
+				"public/**": config.PolicyGroup{
+					Policies: []config.Policy{
+						{
+							Users:   []string{"admin"},
+							Actions: []string{"create", "read", "delete"},
+						},
+						{
+							Users:   []string{"user"},
+							Actions: []string{"read"},
+						},
+					},
+				},
+			},
+		}
+
+		ctlr := api.NewController(conf)
+		cm := NewControllerManager(ctlr)
+		baseURL := cm.StartAndWait()
+
+		adminImage1 := testImageCreateHelper(t, [][]byte{{10, 11, 10, 11}})
+		err := UploadImageWithBasicAuth(
+			adminImage1, baseURL, "admin/secret-base", "latest",
+			adminUserName, adminPassword,
+		)
+		So(err, ShouldBeNil)
+
+		publicImage1 := testImageCreateHelper(t, [][]byte{{10, 11, 10, 11}, {20, 21, 20, 21}})
+
+		err = UploadImageWithBasicAuth(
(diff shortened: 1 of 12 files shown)


pkg/extensions/search/base_image_list_gql_test.go
```diff
@@ -235,3 +235,193 @@ func TestBaseImageListGql(t *testing.T) {
...
+				expectedImages[key] = true
+			}
+		}
+
+		for _, found := range expectedImages {
+			So(found, ShouldBeTrue)
+		}
+
+		// non-admin user should be able to access the base image list for the admin image
+		// even though one of the base images is accessible to them.
```
andaaron, line 416: Comment says the non-admin “should be able to access the base image list for the admin image”, but the assertion expects empty results. Please flip the wording to “should not be able to access…”.

pkg/extensions/search/resolver.go
```diff
@@ -355,6 +389,19 @@ func getCVEDiffListForImages(
...
+		// This allows checking for repo permissions on it.
+		// If it is empty, the rest of the logic proceeds as usual.
+		if subtrahend.Repo != "" {
+			if ok, err := reqCtx.RepoIsUserAvailable(ctx, subtrahend.Repo); !ok || err != nil {
+				log.Info().Err(err).Str("repository", subtrahend.Repo).Bool("availability", ok).Str("component", "graphql").
+					Msg("repo user availability")
+
+				return &gql_generated.CVEDiffResult{}, nil //nolint:nilerr // don't give details to a potential attacker
+			}
+		}
```
andaaron: This is redundant. 

`minuendBaseImages, err := metaDB.FilterTags(ctx, mTypes.AcceptOnlyRepo(minuend.Repo),` and we already verified the minuend is available at `if ok, err := reqCtx.RepoIsUserAvailable(ctx, minuend.Repo); !ok || err != nil {`

pkg/extensions/search/resolver.go
```diff
@@ -1547,9 +1594,17 @@ func getImageList(ctx context.Context, repo string, metaDB mTypes.MetaDB, cveInf
 	}, nil
 }
 
-func getReferrers(metaDB mTypes.MetaDB, repo string, referredDigest string, artifactTypes []string,
+func getReferrers(ctx context.Context, metaDB mTypes.MetaDB, repo string, referredDigest string, artifactTypes []string,
 	log log.Logger,
 ) ([]*gql_generated.Referrer, error) {
+	// check if the user has access to the repo
+	if ok, err := reqCtx.RepoIsUserAvailable(ctx, repo); !ok || err != nil {
```
andaaron, line 1588: Maybe move this check inside metaDB.GetReferrersInfo(ctx)? Any specific reason to have it here?

pkg/extensions/search/cve/model/models.go
```diff
@@ -131,3 +133,38 @@ type ScanResult struct {
 	MediaType string
 	WasCached bool
 }
+
+type CveListForImageResponse struct {
```
andaaron: These GraphQL response envelopes are test/client unmarshalling DTOs, not CVE domain model. The same pattern already lives in pkg/common/model.go (ImageListResponse, ReferrersResp, ImageSummaryResult, …). Related copies also already exist as CveResult / ImgList / CVEResultForImage in pkg/extensions/search/cve/cve_test.go and as private types in pkg/cli/client/service.go.

Please move the existing cve_test.go helpers into pkg/common (alongside the other GQL response types), instead of growing a third set under cvemodel.

pkg/extensions/search/cve/model/models.go
```diff
@@ -131,3 +133,38 @@ type ScanResult struct {
...
+	CveListForImageSingle `json:"data"`
+
+	Errors []common.ErrorGQL `json:"errors"`
+}
+
+//nolint:tagliatelle // graphQL schema
+type CveListForImageSingle struct {
```
andaaron: CveListForImageSingle is the data envelope, not a “single CVE”. Elsewhere this layer is named after the query result (ImageList, ReferrersResult, cveData). Something like CveListForImageData / CVEListForImageResult would match existing naming better if these stay.

pkg/extensions/search/resolver.go
vrajashkr: Given the nature of the PR, the line of thought was that having it here would fix the primary issue for all the implementations of metaDB.

A follow-up PR could then go in to apply the same check inside each metaDB implementation.

pkg/extensions/search/cve/model/models.go
vrajashkr: Moving the cve_test.go models out to common is causing a lot of diff to show up. I can work on a separate PR for that.

andaaron: In that case, maybe move the ones here under pkg/common, and the imports in cve_test.go can change later?

pkg/extensions/search/base_image_list_gql_test.go
vrajashkr: Fixed. Thanks!

pkg/extensions/search/resolver.go
vrajashkr: I missed that. Thanks! Removed that code.

pkg/extensions/search/cve/model/models.go
vrajashkr: Done. I've moved them. I had to duplicate some of them temporarily to avoid importing cve in common.

pkg/extensions/search/cve/model/models.go
vrajashkr: You're right. Thanks. I've renamed it.

pkg/extensions/search/cve/model/models.go
andaaron: And I see pkg/cli/client/service.go also has structs for these responses.
We can consolidate in next PR.

andaaron (approved):

rchincha (approved):

pkg/extensions/search/cve/model/models.go
vrajashkr: https://github.com/project-zot/zot/pull/4287

PR #4289, opened 2026-08-01
fix(meta): don't 500 a manifest delete once metaDB signature cleanup …

### LINKED ISSUE #4286: error-level metadb noise for cosign v3 sigstore-bundle referrers (parse.go getCosignSignatureLayersInfo); DELETE of such referrers returns 500 though deletion succeeds
### zot version

v2.1.18 (Kubernetes, StatefulSet, local storage, retention + basic-auth ACL config)

### Describe the bug

cosign v3.x signs images as OCI 1.1 referrers carrying `application/vnd.dev.sigstore.bundle.v0.3+json` artifacts (no `sha256-<digest>.sig` tag, no `dev.cosignproject.cosign/signature` layer annotation). Two rough edges in zot's metadb when these are present:

**1. Error-level log noise per signed image (cosmetic).** Every metadb parse of a repo holding such a referrer emits ~2 error-level lines per signature:

```json
{"level":"error","message":"failed to get specific annotation of cosign signature","repository":"parascope-vault-collector","reference":"sha256:fea71caf17b2776a305a16fff61b54fb61303ecf9e5036f01362ba88c5dc400b","layerDigest":"sha256:e7c62a51d4e1993b50c4a33089c76f757069df01fe49f094015679977aabf38c","caller":"zotregistry.dev/zot/v2/pkg/meta/parse.go:395","func":"zotregistry.dev/zot/v2/pkg/meta.getCosignSignatureLayersInfo"}
```

`getCosignSignatureLayersInfo` appears to look for the legacy cosign layer annotation, which sigstore-bundle referrer layers legitimately do not carry. Everything works — the referrers API serves them, `cosign verify` succeeds against the referrer — so this looks purely cosmetic, but at error level it pollutes alerting pipelines that key on log severity. A signature-format-aware downgrade to debug/info (or handling the bundle format's annotations) would fix it.

**2. `DELETE /v2/<repo>/manifests/<referrer-digest>` returns HTTP 500 — but the deletion succeeds (functional wart).** Deleting a sigstore-bundle referrer manifest by digest (credential has delete permission) returns 500, yet a subsequent `GET /v2/<repo>/referrers/<subject-digest>` shows the referrer gone and the manifest is no longer retrievable. Correlated log line at the moment of each 500:

```json
{"level":"error","message":"failed to check if image is a signature or not","error":"image meta not found","component":"metadb","caller":"zotregistry.dev/zot/v2/pkg/meta/hooks.go:243","func":"zotregistry.dev/zot/v2/pkg/meta.OnDeleteManifest"}
```

So the storage delete goes through and the `OnDeleteManifest` metadb hook then fails to classify the just-deleted manifest ("image meta not found"), which surfaces as a 500 to the client despite the operation having taken effect. Clients doing cleanup have to treat 500 as "re-query to find out".

### To reproduce

1. Push any image: `crane append --oci-empty-base --new_layer layer.tar.gz --new_tag <zot>/repo:tag`
2. Sign with cosign v3 (v3.1.2 here): `cosign sign --yes --key k.key --tlog-upload=false --use-signing-config=false <zot>/repo@<digest>` → referrer lands, works.
3. Watch the zot log: the parse.go:395 error lines appear when metadb parses the repo (startup parse, and on access).
4. `curl -X DELETE <zot>/v2/repo/manifests/<referrer-manifest-digest>` with a delete-capable user → HTTP 500, hooks.go:243 error logged, but the referrer is actually gone on re-query.

### Expected behavior

1. No error-level logging for well-formed sigstore-bundle referrers (they are not malformed cosign signatures — they are the new format).
2. DELETE of a referrer manifest that succeeds returns 202, not 500.

### Additional context

Both behaviors observed on a production registry; happy to provide fuller logs or re-run repro steps against a scratch repo if useful.

rchincha:
…fails

Fixes #4286.

Two issues with cosign v3 sigstore-bundle referrers:

- getCosignSignatureLayersInfo logged at error level whenever a layer lacked the legacy cosign signature-key annotation. Sigstore-bundle referrers never carry that annotation by design, so every push of one produced spurious error-level noise. Downgraded to debug.

- DELETE on a signature referrer could return 500 even though the manifest was already gone from the image store. Root cause: BoltDB's RemoveRepoReference clears a digest's Signatures entry as soon as its last tag is removed (see the "foundTag" check), even if a referrer manifest pointing at that digest via Subject still exists, undeleted, in the store. Deleting that referrer afterward then finds nothing to update and fails with ErrImageMetaNotFound. This is deterministic, not a race - delete/replace the subject's tag, then delete its signature referrer, and it reproduces every time (also hit internally by retention's own GC pass, which removes untagged manifests before re-checking orphaned referrers). Since the actual deletion already succeeded by the time this metaDB bookkeeping runs, a failure there is now logged but no longer turns the response into a 500.

Also fixed a copy-pasted log message in OnDeleteManifest's DeleteSignature error branch that mislabeled this exact failure as "failed to check if image is a signature or not", which obscured the real cause while triaging this issue.

Tests: pkg/meta/hooks_test.go documents the metaDB-level mechanism directly; pkg/api/routes_test.go reproduces it end to end through the real HTTP handler (no mocked metaDB) and confirms the fix; a new blackbox test in annotations.bats signs a real image with cosign v3, deletes its tag, then deletes the referrer and asserts 202.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -1007,9 +1007,15 @@ func (rh *RouteHandler) DeleteManifest(response http.ResponseWriter, request *ht
 		err := meta.OnDeleteManifest(name, reference, mediaType, manifestDigest, manifestBlob,
 			rh.c.StoreController, rh.c.MetaDB, rh.c.Log)
 		if err != nil {
-			response.WriteHeader(http.StatusInternalServerError)
+			// The image store manifest may have been restored (or restore may itself have
+			// failed), so this must not report success either way.
+			if errors.Is(err, zerr.ErrManifestRestoreAttempted) {
+				rh.c.Log.Error().Err(err).Str("repository", name).Str("reference", reference).
+					Msg("failed to remove manifest from metadb; attempted to restore it in the image store")
+				response.WriteHeader(http.StatusInternalServerError)
 
-			return
+				return
+			}
 		}
 	}
 
--- a/pkg/api/routes_test.go
+++ b/pkg/api/routes_test.go
@@ -34,9 +34,13 @@ import (
 	"zotregistry.dev/zot/v2/pkg/api/config"
 	"zotregistry.dev/zot/v2/pkg/api/constants"
 	apiErr "zotregistry.dev/zot/v2/pkg/api/errors"
+	zcommon "zotregistry.dev/zot/v2/pkg/common"
+	"zotregistry.dev/zot/v2/pkg/extensions/monitoring"
 	"zotregistry.dev/zot/v2/pkg/log"
+	"zotregistry.dev/zot/v2/pkg/meta/boltdb"
 	mTypes "zotregistry.dev/zot/v2/pkg/meta/types"
 	reqCtx "zotregistry.dev/zot/v2/pkg/requestcontext"
+	"zotregistry.dev/zot/v2/pkg/storage/local"
 	storageTypes "zotregistry.dev/zot/v2/pkg/storage/types"
 	test "zotregistry.dev/zot/v2/pkg/test/common"
 	"zotregistry.dev/zot/v2/pkg/test/mocks"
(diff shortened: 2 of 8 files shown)


pkg/api/routes.go
```diff
@@ -1004,12 +1004,16 @@ func (rh *RouteHandler) DeleteManifest(response http.ResponseWriter, request *ht
...
+		// issue to flag, not a failed deletion to report to the client - concurrent metadb writers
+		// (e.g. retention's GC removing the same referrer's signature entry) can otherwise race
+		// this into an error even though there is nothing left to delete.
 		err := meta.OnDeleteManifest(name, reference, mediaType, manifestDigest, manifestBlob,
 			rh.c.StoreController, rh.c.MetaDB, rh.c.Log)
 		if err != nil {
-			response.WriteHeader(http.StatusInternalServerError)
-
-			return
+			rh.c.Log.Error().Err(err).Str("repository", name).Str("reference", reference).
```
vrajashkr: I agree that this is an error condition, but this goes back to Issue 1 reported in the issue - this would be an error log that is not actionable.

Instead, it would be ideal to check the error type. If it is an error emitted for a reason that zot can tolerate, then we can emit a debug log and skip. If not, emit an error log.

pkg/api/routes_test.go
```diff
@@ -3539,3 +3543,204 @@ func TestGetBlobMultipartReaderCloseError(t *testing.T) {
 	assert.Less(t, int64(len(body)), advertisedLen,
 		"a Close() error on the second range must truncate the body")
 }
+
+// TestDeleteManifestSucceedsDespiteMetaDBHookFailure covers issue #4286:
```
vrajashkr: Imo, this line should be removed. We probably don't want references to issues in the code base unless super necessary.

pkg/api/routes_test.go
```diff
@@ -3539,3 +3543,204 @@ func TestGetBlobMultipartReaderCloseError(t *testing.T) {
...
+// DeleteImageManifest (the actual deletion) has already succeeded by the time
+// OnDeleteManifest's metaDB step runs, so *any* failure there (forced here via a
+// mock; TestDeleteManifestSucceedsAfterSignatureSubjectOrphaned below reproduces
+// the real, deterministic trigger against real components) must not turn an
+// already-successful deletion into a 500 to the client.
```
vrajashkr:
```suggestion
// DeleteImageManifest (the actual deletion) has already succeeded by the time
// OnDeleteManifest's metaDB step runs, so any failure there should not return a client-facing error.
```

pkg/api/routes_test.go
```diff
@@ -3539,3 +3543,204 @@ func TestGetBlobMultipartReaderCloseError(t *testing.T) {
...
+		},
+		Subject: &ispec.Descriptor{
+			MediaType: ispec.MediaTypeImageManifest,
+			Digest:    subjectDigest,
+			Size:      10,
+		},
+	}
+
+	manifestJSON, err := json.Marshal(manifest)
+	require.NoError(t, err)
```
vrajashkr: nit: Are we not using go-convey and related assert helpers for this test?

pkg/api/routes_test.go
```diff
@@ -3539,3 +3543,204 @@ func TestGetBlobMultipartReaderCloseError(t *testing.T) {
...
+		"the manifest was already deleted from the image store; a metaDB-only failure must not surface as a 500")
+}
+
+// TestDeleteManifestSucceedsAfterSignatureSubjectOrphaned reproduces issue #4286's DELETE-500
+// end to end against real components (no mocked metaDB): BoltDB's RemoveRepoReference clears a
+// digest's Signatures entry as soon as its last tag is gone (see hooks_test.go's
+// TestOnDeleteManifest_OrphanedSignatureSubjectBreaksReferrerDeletion for the metaDB-level
+// mechanism), so deleting a signature referrer whose subject was already untagged always used
+// to 500 even though the referrer itself deletes cleanly. This is deterministic, not a race:
+// delete the subject's tag, then delete the referrer.
```
vrajashkr:
```suggestion
// BoltDB's RemoveRepoReference clears a
// digest's Signatures entry as soon as its last tag is gone so deleting a 
// signature referrer whose subject was already should not return an error
```

pkg/api/routes_test.go
```diff
@@ -3539,3 +3543,204 @@ func TestGetBlobMultipartReaderCloseError(t *testing.T) {
...
+	referrerBody, err := json.Marshal(referrer)
+	require.NoError(t, err)
+	referrerDigest := godigest.FromBytes(referrerBody)
+
+	statusCode = putReq(referrerDigest.String(), referrerBody)
+	require.Equal(t, http.StatusCreated, statusCode)
+
+	repoMeta, err := metaDB.GetRepoMeta(ctx, "repo")
+	require.NoError(t, err)
+	require.Contains(t, repoMeta.Signatures, subjectDigest.String(),
```
vrajashkr: nit: In this test as well, do we not want to use convey and its asserts with So?

pkg/extensions/search/search_test.go
```diff
@@ -6583,7 +6583,9 @@ func TestMetaDBWhenDeletingImages(t *testing.T) {
 				resp, err = resty.R().Delete(baseURL + "/v2/" + "repo1" + "/manifests/" + "signatureReference")
 				So(resp, ShouldNotBeNil)
 				So(err, ShouldBeNil)
-				So(resp.StatusCode(), ShouldEqual, http.StatusInternalServerError)
+				// the image is already gone from the image store at this point, so a metadb
+				// bookkeeping failure here is reported as a 202 (see issue #4286).
```
vrajashkr:
```suggestion
				// the image is already gone from the image store at this point, so a metadb
				// bookkeeping failure here is not exposed to the client.
```

pkg/extensions/search/search_test.go
```diff
@@ -6614,7 +6616,9 @@ func TestMetaDBWhenDeletingImages(t *testing.T) {
 					"sha256-343ebab94a7674da181c6ea3da013aee4f8cbe357870f8dcaf6268d5343c3474.sig")
 				So(resp, ShouldNotBeNil)
 				So(err, ShouldBeNil)
-				So(resp.StatusCode(), ShouldEqual, http.StatusInternalServerError)
+				// the image is already gone from the image store at this point, so a metadb
+				// bookkeeping failure here is reported as a 202 (see issue #4286).
```
vrajashkr:
```suggestion
				// the image is already gone from the image store at this point, so a metadb
				// bookkeeping failure here is not exposed to the client.
```

pkg/extensions/search/search_test.go
```diff
@@ -6652,7 +6656,9 @@ func TestMetaDBWhenDeletingImages(t *testing.T) {
 					"343ebab94a7674da181c6ea3da013aee4f8cbe357870f8dcaf6268d5343c3474.sig")
 				So(resp, ShouldNotBeNil)
 				So(err, ShouldBeNil)
-				So(resp.StatusCode(), ShouldEqual, http.StatusInternalServerError)
+				// the image is already gone from the image store at this point, so a metadb
+				// bookkeeping failure here is reported as a 202 (see issue #4286).
```
vrajashkr:
```suggestion
				// the image is already gone from the image store at this point, so a metadb
				// bookkeeping failure here is not exposed to the client.
```

pkg/meta/parse.go
```diff
@@ -399,10 +399,13 @@ func getCosignSignatureLayersInfo(
 			return layers, err
 		}
 
+		// The legacy cosign annotation is absent on sigstore-bundle referrers (and any other
+		// non-legacy cosign format) by design, not by error - SignatureKey is simply left
+		// empty below in that case. Debug, not Error: nothing here indicates a problem.
```
vrajashkr:
```suggestion
		// The legacy cosign annotation is absent on sigstore-bundle referrers (and any other
		// non-legacy cosign format) by design. SignatureKey is simply left
		// empty below in that case.
```

pkg/meta/hooks_test.go
```diff
@@ -468,3 +471,99 @@ func TestUpdateErrors(t *testing.T) {
...
+// TestOnDeleteManifest_OrphanedSignatureSubjectBreaksReferrerDeletion documents the exact,
+// deterministic mechanism behind issue #4286: RemoveRepoReference (called for the *subject*
+// image when its last tag is removed) unconditionally clears that digest's Signatures entry
+// once no tag references it anymore (see BoltDB's RemoveRepoReference, "foundTag" check) -
+// even though a referrer manifest pointing at it via Subject may still exist, undeleted, in
+// the image store. Deleting that referrer afterward then finds no Signatures entry to update.
+// This requires no concurrency to trigger: retention GC hits it internally within a single
+// CleanRepo pass (it removes untagged manifests before re-checking orphaned referrers), and a
+// plain sequential client "delete the tag, then delete the signature" also hits it - reproduced
+// live against a running zot with cosign v3.1.2 during triage of this issue.
```
vrajashkr: AI comments not good tbh.


```suggestion
// TestOnDeleteManifest_OrphanedSignatureReferrerDeletionSucceeds verifies that
// a signature referrer can be successfully deleted even after its subject's manifest has been untagged.
```

pkg/meta/hooks_test.go
```diff
@@ -468,3 +471,99 @@ func TestUpdateErrors(t *testing.T) {
...
+func TestOnDeleteManifest_OrphanedSignatureSubjectBreaksReferrerDeletion(t *testing.T) {
```
vrajashkr:
```suggestion
func TestOnDeleteManifest_OrphanedSignatureReferrerDeletionSucceeds(t *testing.T) {
```

vrajashkr (changes requested):

pkg/meta/hooks.go
```diff
@@ -240,8 +241,14 @@ func OnDeleteManifest(repo, reference, mediaType string, digest godigest.Digest,
 			SignatureType:   signatureType,
 		})
 		if err != nil {
-			log.Error().Err(err).Str("component", "metadb").
-				Msg("failed to check if image is a signature or not")
+			if errors.Is(err, zerr.ErrImageMetaNotFound) {
+				// Expected: a concurrent writer already cleared this entry.
```
andaaron: This is misleading. At least in the scenario in the bats test it is deterministic.

pkg/api/routes.go
```diff
@@ -1007,9 +1007,14 @@ func (rh *RouteHandler) DeleteManifest(response http.ResponseWriter, request *ht
 		err := meta.OnDeleteManifest(name, reference, mediaType, manifestDigest, manifestBlob,
 			rh.c.StoreController, rh.c.MetaDB, rh.c.Log)
 		if err != nil {
-			response.WriteHeader(http.StatusInternalServerError)
+			// A restore may have rolled back the delete, so this must not report success.
+			if errors.Is(err, zerr.ErrManifestRestoreAttempted) {
+				rh.c.Log.Error().Err(err).Str("repository", name).Str("reference", reference).
+					Msg("manifest deleted from image store but metadb update failed")
```
andaaron: ErrManifestRestoreAttempted is returned after a restore attempt, including when PutImageManifest succeeded and the blob is back. "manifest deleted from image store but metadb update failed" then describes the opposite of store state.

andaaron (approved):

test/blackbox/annotations.bats
```diff
@@ -189,6 +189,58 @@ function teardown_file() {
...
+@test "delete cosign sigstore-bundle referrer returns 202 without error-level logs" {
+    zot_port=`cat ${BATS_FILE_TMPDIR}/zot.port`
+
+    # Dedicated image content so this digest starts with zero referrers.
+    cat > ${BATS_FILE_TMPDIR}/Dockerfile.delete-referrer-test<<EOF
+FROM public.ecr.aws/t0x7q1g8/centos:7
+CMD ["/bin/sh", "-c", "echo 'delete cosign referrer test'"]
+EOF
+    run podman build -f ${BATS_FILE_TMPDIR}/Dockerfile.delete-referrer-test -t 127.0.0.1:${zot_port}/annotations:delete-referrer-test . --format oci
+    [ "$status" -eq 0 ]
```
vrajashkr, line 201: Just curious - is a fresh image build really required here? I'm wondering if pushing a ready image would suffice.

vrajashkr (commented):
One question. Rest of the code LGTM.

vrajashkr (approved):

PR #4295, opened 2026-08-04
feat(ci): also use floci for CI tests

vrajashkr:
**What type of PR is this?**
CI feature

N/A

- Adds duplicates of the current workflows where localstack is used, but uses floci instead.
- floci is MIT licensed and a very popular alternative to localstack.
- The aim is to run these in parallel for some time and then switchover completely by deleting these temporary workflows and modifying the main ones.

CI tests should pass

**Automation added to e2e**:
N/A

No

No

### DIFF
--- a/.github/actions/setup-floci/action.yaml
+++ b/.github/actions/setup-floci/action.yaml
@@ -29,8 +29,7 @@ runs:
         docker pull "$IMAGE_NAME"
         docker run -d \
           --name floci \
-          -p '4566:4566' \
-          -e 'FLOCI_STORAGE_MODE=persistent' "$IMAGE_NAME"
+          -p '4566:4566' "$IMAGE_NAME"
 
         # Wait for floci to be healthy
         attempt=1
@@ -65,6 +64,6 @@ runs:
       env:
         AWS_ACCESS_KEY_ID: fake
         AWS_SECRET_ACCESS_KEY: fake
-        IMAGE_NAME: ghcr.io/project-zot/floci/floci:1.5.34
+        IMAGE_NAME: ghcr.io/project-zot/ci-images/floci:1.5.34
         MAX_ATTEMPTS: 30
         WAIT_PERIOD_SECONDS: 1
--- a/.github/actions/setup-floci/action.yaml
+++ b/.github/actions/setup-floci/action.yaml
@@ -1,0 +1,70 @@
+name: 'Setup floci service'
+description: 'Download & run floci container (does not create S3 buckets; tests or workflows should create resources they need)'
+inputs:
+  # inputs for https://github.com/docker/login-action
+  username:
+    description: 'Username used to log against the github registry'
+    required: false
+    default: ${{ github.actor }}
+  password:
+    description: 'Password or personal access token used to log against the github registry'
+    required: false
+    default: ${{ github.token }}
+runs:
+  using: "composite"
+  steps:
+    - uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
+      with:
+        registry: ghcr.io
+        username: ${{ inputs.username }}
+        password: ${{ inputs.password }}
+    - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+      with:
+        python-version: '3.11'
+    - shell: bash
+      run: |
+        pip install --upgrade pyopenssl
+        pip install 'awscli-local[ver1]'
+
+        docker pull "$IMAGE_NAME"
+        docker run -d \
+          --name floci \
+          -p '4566:4566' \
+          -e 'FLOCI_STORAGE_MODE=persistent' "$IMAGE_NAME"
+
+        # Wait for floci to be healthy
+        attempt=1
+        ready=0
+        while [[ $ready -eq 0 && $attempt -le $MAX_ATTEMPTS ]]; do
+          echo "Attempt: $attempt"
+          response=$(curl -s -w "\n%{http_code}" "http://localhost:4566/health" || echo "000")
+          http_code=$(echo "$response" | tail -n 1)
+          json_body=$(echo "$response" | sed '$d')
+
+          if [[ "$http_code" -eq 200 ]]; then
+            s3_status=$(echo "$json_body" | jq -r '.services.s3 // empty' 2>/dev/null || true)
+            dynamodb_status=$(echo "$json_body" | jq -r '.services.dynamodb // empty' 2>/dev/null || true)
+            if [[ "$s3_status" == "running" && "$dynamodb_status" == "running" ]]; then
+              echo "S3 and DynamoDB are ready."
+              ready=1
+            else
+              echo "S3: $s3_status DynamoDB: $dynamodb_status. Continue waiting."
+            fi
+          else
+            echo "Status was $http_code. Continue waiting."
+          fi
+
+          sleep $WAIT_PERIOD_SECONDS
+          attempt=$((attempt + 1))
+        done
+        if [[ $ready -eq 0 ]]; then
+          echo "Floci failed to be ready in time."
+          docker logs floci || true
+          exit 1
+        fi
+      env:
+        AWS_ACCESS_KEY_ID: fake
+        AWS_SECRET_ACCESS_KEY: fake
+        IMAGE_NAME: ghcr.io/project-zot/floci/floci:1.5.34
+        MAX_ATTEMPTS: 30
+        WAIT_PERIOD_SECONDS: 1
--- a/.github/actions/setup-floci/action.yaml
+++ b/.github/actions/setup-floci/action.yaml
@@ -1,0 +1,54 @@
+name: 'Setup floci service'
+description: 'Download & run floci container (does not create S3 buckets; tests or workflows should create resources they need)'
+inputs:
+  # inputs for https://github.com/docker/login-action
+  username:
+    description: 'Username used to log against the github registry'
+    required: false
+    default: ${{ github.actor }}
+  password:
+    description: 'Password or personal access token used to log against the github registry'
+    required: false
+    default: ${{ github.token }}
+  persistent_mode:
+    description: 'Whether to use persistent storage mode for floci ("0" for no, "1" for yes)'
+    required: false
+    default: "0"
+runs:
+  using: "composite"
+  steps:
+    - uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
+      with:
+        registry: ghcr.io
+        username: ${{ inputs.username }}
+        password: ${{ inputs.password }}
+    - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+      with:
+        python-version: '3.11'
+    - shell: bash
+      run: |
+        pip install --upgrade pyopenssl
+        pip install 'awscli-local[ver1]'
+
+        # install floci-cli to manage floci
+        curl -fsSL "https://github.com/floci-io/floci-cli/raw/refs/tags/${FLOCI_INSTALLER_VERSION_TAG}/installer/install.sh" | sh
+        floci version
+
+        if [[ "1" == "${{ inputs.persistent_mode }}" ]]; then
+          mkdir -p ./floci-data
+          floci start --image "${IMAGE_NAME}" --services "${SERVICES_TO_START}" --persist ./floci-data
+        else
+          floci start --image "${IMAGE_NAME}" --services "${SERVICES_TO_START}"
+        fi
+        for svc in ${SERVICES_TO_START//,/ }; do
+          echo "Waiting for ${svc}"
+          # Default timeout is 30 seconds
+          floci wait --service "${svc}"
+        done
+      env:
+        AWS_ACCESS_KEY_ID: fake
+        AWS_SECRET_ACCESS_KEY: fake
+        IMAGE_NAME: ghcr.io/project-zot/ci-images/floci:1.5.34
+        FLOCI_INSTALLER_VERSION_TAG: 0.2.0
+        FLOCI_CLI_VERSION: 0.2.0
+        SERVICES_TO_START: 's3,dynamodb,secretsmanager'
--- a/.github/actions/teardown-floci/action.yaml
+++ b/.github/actions/teardown-floci/action.yaml
@@ -1,0 +1,8 @@
+name: 'Stop floci'
+description: 'Stop floci'
+runs:
+  using: "composite"
+  steps:
+    - shell: bash
+      run: |
+        floci stop --remove || true
--- a/.github/actions/teardown-floci/action.yaml
+++ b/.github/actions/teardown-floci/action.yaml
@@ -1,0 +1,10 @@
+name: 'Stop floci'
+description: 'Stop floci container'
+runs:
+  using: "composite"
+  steps:
+    - shell: bash
+      if: always()
+      run: |
+        docker stop floci || true
+        docker rm -f floci || true
--- a/.github/workflows/ecosystem-tools-with-floci.yaml
+++ b/.github/workflows/ecosystem-tools-with-floci.yaml
@@ -1,0 +1,167 @@
+name: "Ecosystem client tools with floci"
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches: [main]
+  release:
+    types:
+      - published
+
+permissions: read-all
+
+jobs:
+  floci-client-tools:
+    name: floci - Check client tools
+    runs-on: cncf-ubuntu-32-128-x86
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+        with:
+          cache: false
+          check-latest: true
+          go-version: 1.26.x
+      - name: Install dependencies
+        run: |
+          cd $GITHUB_WORKSPACE
+          go install github.com/swaggo/swag/cmd/swag@v1.16.2
+          go mod download
+          sudo apt-get update
+          sudo apt-get install -y libgpgme-dev libassuan-dev libbtrfs-dev \
+            libdevmapper-dev pkg-config rpm uidmap haproxy jq valkey-tools whois \
+            npm
+          # install skopeo
+          git clone -b v1.12.0 https://github.com/containers/skopeo.git
+          cd skopeo
+          make bin/skopeo
+          sudo cp bin/skopeo /usr/bin
+          skopeo -v
+          # install cri-o (for crictl)
+          OS=xUbuntu_22.04
+          CRIO_VERSION=1.26
+          curl -fsSL https://download.opensuse.org/repositories/isv:/kubernetes:/addons:/cri-o:/prerelease:/main:/build/deb/Release.key | sudo gpg --dearmor -o /etc/apt/keyrings/cri-o-apt-keyring.gpg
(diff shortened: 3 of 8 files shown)


.github/actions/teardown-floci/action.yaml
```diff
@@ -0,0 +1,8 @@
+name: 'Stop floci'
+description: 'Stop floci container'
+runs:
+  using: "composite"
+  steps:
+    - shell: bash
+      if: always()
+      run: docker stop floci && docker rm floci
```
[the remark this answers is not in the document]
vrajashkr: Fixed

.github/actions/setup-floci/action.yaml
```diff
@@ -0,0 +1,61 @@
...
+          sleep $WAIT_PERIOD_SECONDS
+          attempt=$((attempt + 1))
+        done
+        echo "Startup complete"
+      env:
+        AWS_ACCESS_KEY_ID: fake
+        AWS_SECRET_ACCESS_KEY: fake
+        IMAGE_NAME: ghcr.io/project-zot/floci/floci:1.5.34
+        MAX_ATTEMPTS: 30
+        WAIT_PERIOD_SECONDS: 1
```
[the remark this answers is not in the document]
vrajashkr: Not a problem

.github/actions/setup-floci/action.yaml
```diff
@@ -0,0 +1,69 @@
...
+          if [[ "$http_code" -eq 200 ]]; then
+            s3_status=$(echo "$json_body" | jq -r '.services.s3 // empty' 2>/dev/null || true)
+            dynamodb_status=$(echo "$json_body" | jq -r '.services.dynamodb // empty' 2>/dev/null || true)
+            if [[ "$s3_status" == "running" && "$dynamodb_status" == "running" ]]; then
+              echo "S3 and DynamoDB are ready."
+              ready=1
+            else
+              echo "S3: $s3_status DynamoDB: $dynamodb_status. Continue waiting."
+            fi
+          else
```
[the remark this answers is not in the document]
vrajashkr: /health works just fine in local testing and in CI.

.github/actions/setup-floci/action.yaml
```diff
@@ -0,0 +1,69 @@
...
+        ready=0
+        while [[ $ready -eq 0 && $attempt -le $MAX_ATTEMPTS ]]; do
+          echo "Attempt: $attempt"
+          response=$(curl -s -w "\n%{http_code}" "http://localhost:4566/health" || echo "000")
+          http_code=$(echo "$response" | tail -n 1)
+          json_body=$(echo "$response" | sed '$d')
+
+          if [[ "$http_code" -eq 200 ]]; then
+            s3_status=$(echo "$json_body" | jq -r '.services.s3 // empty' 2>/dev/null || true)
+            dynamodb_status=$(echo "$json_body" | jq -r '.services.dynamodb // empty' 2>/dev/null || true)
```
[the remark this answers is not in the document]
vrajashkr: This has been replaced with floci cli

.github/workflows/nightly-with-floci.yaml
```diff
@@ -0,0 +1,331 @@
+name: 'Nightly jobs with floci'
+on:
+  pull_request:
```
vrajashkr: This was kept on PR to exercise the nightly runs in the PR before it actually runs nightly. I'll change this after review.

rchincha: You probably also want to add

on:
  workflow_dispatch:

vrajashkr: Thanks! I've added that and also made it a 12 hour cron.

.github/workflows/ecosystem-tools-with-floci.yaml
```diff
@@ -0,0 +1,160 @@
...
+          if find /tmp/zot-ft-logs/redis -name '*.log' -print0 | xargs -0 cat | grep 'cannot proxy an already proxied request'; then
+            echo "detected multi-hop in Redis tests"
+            exit 1
+          else
+            exit 0
+          fi
+        continue-on-error: true
+      - name: Clean up Redis scale-out logs
+        run: |
+          rm -rf /tmp/zot-ft-logs/redis
```
andaaron, line 154: Original ecosystem-tools.yaml uploads ./zb-results/ after Redis checks; the floci twin jumps straight to teardown. That drops useful perf/debug artifacts during the comparison period.

.github/workflows/gc-stress-test-with-floci.yaml
```diff
@@ -0,0 +1,271 @@
...
+  pull_request:
+    branches: [main]
+  release:
+    types:
+      - published
+
+permissions: read-all
+
+jobs:
+  floci-gc-referrers-stress-local:
```
andaaron: The `local` tests don't use floci. These jobs don't make sense unless we do a drop-in replacement by renaming this file to replace the old jobs

.github/workflows/gc-stress-test-with-floci.yaml
```diff
@@ -0,0 +1,271 @@
...
+            minio.make_bucket('zot-storage')
+            print(f'{minio.list_buckets()}')
+            EOF
+
+      - name: Run zb
+        id: bench
+        run: |
+            make binary
+            make bench
+            ./bin/zot-linux-amd64 serve test/gc-stress/config-gc-referrers-bench-s3-minio.json &
```
andaaron, line 73: This is still using minio in the config file. I guess we do want to use minio in some testing. What is the goal for this file?

examples/aws-secrets-manager-bearer-auth-floci.sh
```diff
@@ -0,0 +1,503 @@
...
+
+    log_info "Cleaning up..."
+
+    # Stop zot
+    if [ -f "${ZOT_PID_FILE}" ]; then
+        kill "$(cat "${ZOT_PID_FILE}")" 2>/dev/null || true
+        rm -f "${ZOT_PID_FILE}"
+    fi
+
+    if [ "$USE_REAL_AWS" = true ]; then
```
andaaron, line 188: So in case of floci we leave the secret around?

examples/aws-secrets-manager-bearer-auth-floci.sh
```diff
@@ -0,0 +1,503 @@
...
+# Options:
+#   --use-real-aws    Use real AWS Secrets Manager (requires ~/.aws/credentials)
+#   --skip-build      Skip building zot (reuse existing binary)
+#   --keep-resources  Don't clean up resources on exit
+#   --region REGION   AWS region (default: us-east-1)
+#   --help            Show this help message
+#
+# Prerequisites:
+#   go, crane, jq, curl
+#   For floci mode (default): docker
```
andaaron: Header says floci mode needs only docker; the script does not start floci, requires aws, and expects something already on :4566. Update the header to match (aws CLI + running floci).

.github/workflows/sync-3rdparty-images.yaml
```diff
@@ -80,3 +80,25 @@ jobs:
...
+      contents: read
+      packages: write
+    strategy:
+      matrix:
+        floci_version:
+          - "1.5.34"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Log in to GitHub Docker Registry
+        uses: docker/login-action@af1e73f918a031802d376d3c8bbc3fe56130a9b0 # v4.4.0
```
andaaron:
```suggestion
        uses: docker/login-action@371161bbe7024a29a25c5e19bfcbc0804fe9ad2c # v4.5.2
```

.github/workflows/nightly-with-floci.yaml
andaaron, line 1: Nightly artifact names collide with localstack nightly — .github/workflows/nightly-with-floci.yaml:231-233, :260-261, :304-306, :332-333

Names like zot-scale-out-dynamodb-logs and zb-cloud-scale-out-perf-results-${{ github.sha }} match nightly.yaml. Ecosystem already uses -floci suffixes; nightly should too so parallel runs are distinguishable.

I guess the question is if we want to drop-in replace the old file.

.github/workflows/test-with-floci.yaml
```diff
@@ -0,0 +1,105 @@
...
+  pull_request:
+    branches: [main]
+  release:
+    types:
+      - published
+
+permissions: read-all
+
+jobs:
+  floci-test-run-minimal:
```
andaaron, line 15: This file is not ready as a drop-in replacement. Some jobs are missing.

.github/actions/setup-floci/action.yaml
```diff
@@ -0,0 +1,54 @@
...
+          echo "Waiting for ${svc}"
+          # Default timeout is 30 seconds
+          floci wait --service "${svc}"
+        done
+      env:
+        AWS_ACCESS_KEY_ID: fake
+        AWS_SECRET_ACCESS_KEY: fake
+        IMAGE_NAME: ghcr.io/project-zot/ci-images/floci:1.5.34
+        FLOCI_INSTALLER_VERSION_TAG: 0.2.0
+        FLOCI_CLI_VERSION: 0.2.0
```
andaaron, line 53: This seems unused

vrajashkr: `FLOCI_CLI_VERSION` is used by the floci installer script to set the version of the CLI to be installed.

https://github.com/floci-io/floci-cli/blob/main/installer/install.sh#L5

.github/workflows/test-with-floci.yaml
vrajashkr: Correct. That is intentional. Tests not using floci were removed. The old file will be updated to use floci and this workflow file deleted should we choose to move forward with floci.

.github/workflows/nightly-with-floci.yaml
vrajashkr: Thanks! Good catch. I totally missed these. I've updated the names.

The intention is to delete these and update the old workflow files if we decide to stick with floci. They won't be replacing the old files except for the secretsmanager script.

.github/workflows/sync-3rdparty-images.yaml
vrajashkr: Updated to v4.6.0 to match the other logins in the workflow.

examples/aws-secrets-manager-bearer-auth-floci.sh
vrajashkr: Thanks! Addressed.

examples/aws-secrets-manager-bearer-auth-floci.sh
vrajashkr: Thanks! Since floci may stick around (in CI it doesn't), it makes sense to delete the secret. I've made that change.

.github/workflows/gc-stress-test-with-floci.yaml
vrajashkr: Yes, this still uses minio for storage, but the cache is set to use dynamoDB which floci provides for this test.

.github/workflows/gc-stress-test-with-floci.yaml
vrajashkr: Thanks! I've removed the 2 local tests. A drop-in is not expected for the workflow files. They will be deleted later and the actual files updated should we choose to continue using floci.

.github/workflows/ecosystem-tools-with-floci.yaml
vrajashkr: Addressed. Added the upload back with a different artifact name.

.github/workflows/nightly-with-floci.yaml
```diff
@@ -0,0 +1,334 @@
...
+
+            killall --wait -r zot-*
+
+            # clean zot storage
+            sudo rm -rf /tmp/zot
+        env:
+          AWS_ACCESS_KEY_ID: fake
+          AWS_SECRET_ACCESS_KEY: fake
+        continue-on-error: true
+      - uses: ./.github/actions/teardown-floci
```
andaaron, line 165: The inner `if: always()` in `.github/actions/teardown-floci/action.yaml` only applies after the composite has been invoked. If an earlier workflow step fails, GitHub skips this `uses:` step (default `if: success()`), so teardown never runs and floci can be left on shared CNCF runners.

Please add `if: always()` on each caller teardown step (and azurite where used). Example of the same fix on the localstack workflows: https://github.com/project-zot/zot/pull/4341

We need to apply this fix in multiple places.

.github/workflows/gc-stress-test-with-floci.yaml
```diff
@@ -0,0 +1,187 @@
...
+        with:
+          name: gc-referrers-bench-s3-floci
+          path: /tmp/gc-referrers-bench-s3.log
+          if-no-files-found: error
+
+      - name: Check on failures
+        if: steps.bench.outcome != 'success'
+        run: |
+          cat /tmp/gc-referrers-bench-s3.log
+          exit 1
```
andaaron, line 101: `Check on failures` exits 1 when the bench failed, so the following `teardown-floci` step is skipped. Please move teardown before that check (or keep it after with `if: always()`).

Same pattern shown in https://github.com/project-zot/zot/pull/4341 for `gc-stress-test.yaml`.

.github/workflows/gc-stress-test-with-floci.yaml
```diff
@@ -0,0 +1,187 @@
...
+        with:
+          name: gc-bench-s3-floci
+          path: /tmp/gc-bench-s3.log
+          if-no-files-found: error
+
+      - name: Check on failures
+        if: steps.bench.outcome != 'success'
+        run: |
+          cat /tmp/gc-bench-s3.log
+          exit 1
```
andaaron, line 189: `Check on failures` exits 1 when the bench failed, so the following `teardown-floci` step is skipped. Please move teardown before that check (or keep it after with `if: always()`).

Same pattern shown in https://github.com/project-zot/zot/pull/4341 for `gc-stress-test.yaml`.

.github/workflows/nightly-with-floci.yaml
```diff
@@ -0,0 +1,334 @@
...
+        if: ${{ steps.scale.outcome != 'success' || steps.multihop.outcome != 'success' }}
+        run: |
+          exit 1
+      - name: Upload zb test results zip as build artifact
+        if: steps.scale.outcome == 'success'
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: zb-cloud-scale-out-redis-results-floci-${{ github.sha }}
+          path: ./zb-results/
+      - uses: ./.github/actions/teardown-floci
```
andaaron: On scale/multihop failure the job exits at `fail job if error`, so teardown is skipped. The dynamodb job in this same file already tears down earlier (before the fail step) — please align redis with that, and/or add `if: always()` on the teardown step.

Example reorder + `if: always()`: https://github.com/project-zot/zot/pull/4341 (`nightly.yaml` cloud-scale-out jobs).

.github/workflows/ecosystem-tools-with-floci.yaml
```diff
@@ -0,0 +1,166 @@
...
+          fi
+        continue-on-error: true
+      - name: Clean up DynamoDB scale-out logs
+        run: |
+          rm -r /tmp/zot-ft-logs
+      - name: Fail job if DynamoDB error
+        if: ${{ steps.dynamodb_scale.outcome != 'success' || steps.dynamodb_multihop.outcome != 'success' }}
+        run: |
+          echo "DynamoDB scale-out tests failed"
+          exit 1
```
andaaron, line 121: If DynamoDB scale-out fails, this `exit 1` skips Redis and the final teardown. Please put `if: always()` on `teardown-floci` so cleanup still runs (Redis still won't run after DynamoDB fail — that's fine / same as the original — but floci should not be left behind).

Example: https://github.com/project-zot/zot/pull/4341 (`ecosystem-tools.yaml`).

.github/workflows/ecosystem-tools-with-floci.yaml
```diff
@@ -0,0 +1,166 @@
...
+        run: |
+          echo "Redis scale-out tests failed"
+          exit 1
+      - name: Upload zb test results zip as build artifact
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        if: always()
+        with:
+          name: zb-cloud-scale-out-functional-results-floci-${{ github.sha }}
+          path: ./zb-results/
+      - uses: ./.github/actions/teardown-floci
```
andaaron, line 166: If DynamoDB scale-out fails, this `exit 1` skips Redis and the final teardown. Please put `if: always()` on `teardown-floci` so cleanup still runs (Redis still won't run after DynamoDB fail — that's fine / same as the original — but floci should not be left behind).

Example: https://github.com/project-zot/zot/pull/4341 (`ecosystem-tools.yaml`).

.github/actions/setup-floci/action.yaml
```diff
@@ -0,0 +1,54 @@
...
+      run: |
+        pip install --upgrade pyopenssl
+        pip install 'awscli-local[ver1]'
+
+        # install floci-cli to manage floci
+        curl -fsSL "https://github.com/floci-io/floci-cli/raw/refs/tags/${FLOCI_INSTALLER_VERSION_TAG}/installer/install.sh" | sh
+        floci version
+
+        if [[ "1" == "${{ inputs.persistent_mode }}" ]]; then
+          mkdir ./floci-data
```
andaaron: `mkdir ./floci-data` fails if the directory already exists (e.g. reused self-hosted workspace). Please use `mkdir -p ./floci-data`.

examples/aws-secrets-manager-bearer-auth-floci.sh
```diff
@@ -0,0 +1,508 @@
...
+    done
+
+    if [ "$USE_REAL_AWS" = true ]; then
+        if ! command -v aws &> /dev/null; then
+            missing="$missing aws"
+        fi
+    else
+        if ! command -v docker &> /dev/null; then
+            missing="$missing docker"
+        fi
```
andaaron: The header correctly says floci must already be running on :4566 and lists aws CLI — but local-mode `check_prerequisites` still requires `docker`, which this script never uses. Please drop the docker check (keep aws).

andaaron (commented):
Found several issues common with localstack while reviewing this. See also https://github.com/project-zot/zot/pull/4341.

examples/aws-secrets-manager-bearer-auth-floci.sh
vrajashkr: Addressed.

.github/actions/setup-floci/action.yaml
vrajashkr: Addressed.

.github/workflows/ecosystem-tools-with-floci.yaml
vrajashkr: Addressed.

.github/workflows/ecosystem-tools-with-floci.yaml
vrajashkr: Addressed.

.github/workflows/nightly-with-floci.yaml
vrajashkr: Addressed.

.github/workflows/gc-stress-test-with-floci.yaml
vrajashkr: Addressed.

.github/workflows/gc-stress-test-with-floci.yaml
vrajashkr: Addressed.

.github/workflows/nightly-with-floci.yaml
vrajashkr: Addressed.

andaaron (approved):

PR #4308, opened 2026-08-06
fix(events): only log successful publish when all sinks accept the event

### LINKED ISSUE #4303: Event recorder logs "published successfully" even when a sink returns NACK/undelivered
### zot version

main @ `7eb7c25`

### Describe the bug

In [`pkg/extensions/events/events.go` L43-53](https://github.com/project-zot/zot/blob/7eb7c25e1e50d251b7220ba00d24e7109bcd2f9f/pkg/extensions/events/events.go#L43-L53), `eventRecorder.publish` iterates sinks and logs `failed to publish event` on NACK/undelivered, but then unconditionally logs `event published successfully` after the loop regardless of per-sink outcome:

```go
for _, sink := range r.sinks {
    if response := sink.Emit(event); cloudevents.IsNACK(response) || cloudevents.IsUndelivered(response) {
        r.log.Error().Err(response).Msg("failed to publish event")
    }
}
r.log.Info().Msgf("event published successfully: %s", event.Type())
```

With a failing (or mixed) sink, the same event produces contradictory `failed to publish` + `published successfully` log lines. Operators tailing logs or alerting on the success message can miss a real delivery failure.

### To reproduce

Configure an event sink that rejects delivery (returns NACK/undelivered), emit any event, and observe both the error and the unconditional success line for the same event.

### Expected behavior

The success line should only be logged when no sink failed. Track per-sink failure and log success only if all sinks accepted (keep iterating so remaining sinks still receive the event).

### Additional context

Found while running [Ito](https://ito.ai) (AI code review, free for open source) against recently merged PRs — full analysis: https://app.ito.ai/share/aae12f98-53b5-4aaf-973e-af1177eadb63.

evanmarshall:
Fixes #4303 (as requested in https://github.com/project-zot/zot/issues/4303#issuecomment-5207404584).

### What

`eventRecorder.publish` logged `event published successfully` unconditionally after iterating sinks, even when a sink returned NACK or the event was undelivered — producing contradictory `failed to publish event` + `event published successfully` lines for the same event. Operators tailing logs or alerting on the success message could miss a real delivery failure.

### How

Track per-sink failure while iterating (so remaining sinks still receive the event), and only log the success line when every sink accepted the event.

### Testing

Added `TestEventsSinkFailure` covering:
- a failing sink logs the error and does **not** log success
- with a failing + healthy sink, the healthy sink still receives the event and success is not logged
- all sinks accepting still logs success

```
$ go test -tags events -run 'TestEventsSinkFailure|TestEvents$|TestEventSinkMissing' -count=1 ./pkg/extensions/events/
ok  	zotregistry.dev/zot/v2/pkg/extensions/events	0.355s
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/extensions/events/events.go
+++ b/pkg/extensions/events/events.go
@@ -42,13 +42,19 @@ func (r eventRecorder) closeSinks() error {
 
 func (r eventRecorder) publish(event *cloudevents.Event) {
 	go func() {
+		succeeded := true
+
 		for _, sink := range r.sinks {
 			if response := sink.Emit(event); cloudevents.IsNACK(response) || cloudevents.IsUndelivered(response) {
 				r.log.Error().Err(response).Msg("failed to publish event")
+
+				succeeded = false
 			}
 		}
 
-		r.log.Info().Msgf("event published successfully: %s", event.Type())
+		if succeeded {
+			r.log.Info().Msgf("event published successfully: %s", event.Type())
+		}
 	}()
 }
 
--- a/pkg/extensions/events/events_test.go
+++ b/pkg/extensions/events/events_test.go
@@ -3,10 +3,13 @@
 package events_test
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


pkg/extensions/events/events.go
```diff
@@ -42,13 +42,19 @@ func (r eventRecorder) closeSinks() error {
...
 		for _, sink := range r.sinks {
 			if response := sink.Emit(event); cloudevents.IsNACK(response) || cloudevents.IsUndelivered(response) {
 				r.log.Error().Err(response).Msg("failed to publish event")
+
+				succeeded = false
 			}
 		}
 
-		r.log.Info().Msgf("event published successfully: %s", event.Type())
+		if succeeded {
```
rchincha, line 55: Maybe the semantics should be success if at least one of them succeeded. The ones that failed will log the msg.

Another option could also be to just log every success at debug log level within the loop.

rchincha:
Superseded by https://github.com/project-zot/zot/pull/4334

Issue #4319, opened 2026-08-12
[Bug]: goroutine leak in sync extension when metrics extension is disabled


ghouscht:
### zot version

Observed on v2.1.13 and reproduced on main for this issue

### Describe the bug

Hi there,
We use zot and the `sync` extension (on-demand sync) in edge environments. In several environments we have seen zot using >10GB of memory (RSS) despite having a `GOMEMLIMIT=500MiB` set. I managed to obtain a coredump (can't share it) and with some help from Claude I think I found the root cause of the high memory usage being a goroutine leak in zot.

1. When a sync completes it calls `DestinationRegistry.CommitAll` which calls `getImageStoreFromImageReference` [here](https://github.com/project-zot/zot/blob/f1f7dd9b97a1bf67abeace25ed0073ed4f01b66a/pkg/extensions/sync/destination.go#L98)
2. `getImageStoreFromImageReference` calls `getImageStore` [here](https://github.com/project-zot/zot/blob/f1f7dd9b97a1bf67abeace25ed0073ed4f01b66a/pkg/extensions/sync/destination.go#L433)
3. `getImageStore` calls `monitoring.NewMetricsServer` [here](https://github.com/project-zot/zot/blob/f1f7dd9b97a1bf67abeace25ed0073ed4f01b66a/pkg/extensions/sync/destination.go#L442)
4. The `NewMetricsServer` call then spawns a `metricsServer.Run` goroutine (and subsequently another one) that never exits

Please note we built zot on our own **without** the `metrics` extension! The coredump in question was showing >3.3 million goroutines from where 99% were stuck in the `metricsServer.Run` function

### To reproduce

`config.yaml`
```yaml
storage:
  rootDirectory: "/tmp/zot-data"
http:
  address: "0.0.0.0"
  port: "55000"
  compat: ["docker2s2"]
log:
  level: "info"
extensions:
  metrics:
    enable: true
    prometheus:
      path: /metrics
  sync:
    enable: true
    registries:
      - urls: ["https://registry-1.docker.io"]
        onDemand: true
```
Binary built from `main` (f1f7dd9b97a1bf67abeace25ed0073ed4f01b66a) with:
```shell
make binary EXTENSIONS=sync,profile
```

1. Run zot:  `bin/zot-darwin-arm64 serve config.yaml`
2. Get a baseline goroutine profile e.g. using `go tool pprof -http=:8080 http://localhost:55000/v2/_zot/pprof/goroutine`
[...]

PR #4320, opened 2026-08-12
fix: stop metrics goroutine leak in sync without metrics build

andaaron:
Temp sync ImageStores called NewMetricsServer under !metrics, which starts Run() loops that were never stopped (issue #4319). Use a nop MetricServer for those stores, Stop CLI-scoped servers, and switch test stubs to the nop.

### DIFF
--- a/pkg/extensions/sync/destination.go
+++ b/pkg/extensions/sync/destination.go
@@ -439,7 +439,6 @@ func getImageStoreFromImageReference(repo string, imageReference ref.Ref, log lo
 }
 
 func getImageStore(rootDir string, log log.Logger) storageTypes.ImageStore {
-	metrics := monitoring.NewMetricsServer(false, log)
-
-	return local.NewImageStore(rootDir, false, false, log, metrics, nil, nil, nil, nil)
+	// Temp sync layouts are short-lived and don't need an active metrics server.
+	return local.NewImageStore(rootDir, false, false, log, monitoring.NewNopMetricServer(), nil, nil, nil, nil)
 }
--- a/pkg/api/authn_test.go
+++ b/pkg/api/authn_test.go
@@ -2272,9 +2272,7 @@ func TestAPIKeysGeneratorErrors(t *testing.T) {
 
 func TestCookiestoreCleanup(t *testing.T) {
 	log := log.NewTestLogger()
-	metrics := monitoring.NewMetricsServer(true, log)
-
-	defer metrics.Stop() // Clean up metrics server to prevent resource leaks
+	metrics := monitoring.NewNopMetricServer()
 
 	authCfgTestCases := []struct {
 		name string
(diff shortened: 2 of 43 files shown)


pkg/extensions/sync/destination.go
```diff
@@ -439,7 +439,7 @@ func getImageStoreFromImageReference(repo string, imageReference ref.Ref, log lo
 }
 
 func getImageStore(rootDir string, log log.Logger) storageTypes.ImageStore {
-	metrics := monitoring.NewMetricsServer(false, log)
-
-	return local.NewImageStore(rootDir, false, false, log, metrics, nil, nil, nil, nil)
+	// Temp sync layouts are short-lived; NewMetricsServer(!metrics) starts
+	// background goroutines that are never stopped and leak per CommitAll.
```
vrajashkr:
```suggestion
	// Temp sync layouts are short-lived and don't need an active metrics server.
```

vrajashkr (commented):
Changes look good to me. One nit on the comment in sync.

vrajashkr:
Do we have tests in other places that verify that when there is a real metrics server for (some of) the tests, the metrics are updated correctly? (in both metrics enabled and disabled cases).

If not, might be something useful to add in a subsequent PR. That said, the project coverage doesn't seem to have changed much so I take that those lines are already covered so we should be good.

andaaron:
> Do we have tests in other places that verify that when there is a real metrics server for (some of) the tests, the metrics are updated correctly? (in both metrics enabled and disabled cases).
> 
> If not, might be something useful to add in a subsequent PR. That said, the project coverage doesn't seem to have changed much so I take that those lines are already covered so we should be good.

Pushed new tests along with the existing ones:

Prometheus (metrics build)
TestMetricHelpersUpdateWhenEnabled in pkg/extensions/monitoring/monitoring_test.go drives IncDownloadCounter, IncUploadCounter, IncHTTPConnRequests, ObserveHTTPMethodLatency, IncGCRuns, IncGCDeleted, and SetSchedulerWorkers against both NewMetricsServer(true) and NewMetricsServer(false), and asserts Prometheus series move only when active. Existing integration coverage remains in TestExtensionMetrics / TestGCMetrics (scrape /metrics).

Minimal (!metrics build)
Same helper matrix in pkg/extensions/monitoring/minimal_test.go via ReceiveMetrics(). Plus existing pkg/storage/gc/gc_metrics_internal_test.go (IncGCDeleted path) and exporter tests in pkg/exporter/api/controller_test.go.

Leak fix itself
TestGetImageStoreDoesNotLeakMetricsGoroutines in pkg/extensions/sync/destination_internal_test.go asserts sync temp stores do not leave (*metricServer).Run goroutines.

So both “active updates correctly” and “inactive does not record” are covered for the main helpers on both implementations; the sync nop path has an explicit non-leak check.

rchincha (dismissed):

Issue #4215, opened 2026-07-10
[Feat]: Improve scale-out


rchincha:
## Summary

The scale-out docs state that registry sync, including on-demand and periodic sync, is compatible with scale-out clustering: repo names are hashed to a cluster member and only that member performs sync.

Current source mostly follows that model:

- dist-spec routes use `ClusterProxy`, which hashes the repo name and forwards to the owning member.
- periodic sync skips repos not owned by the local cluster member.
- on-demand sync is triggered from manifest/referrers routes and uses an in-process `requestStore` to coalesce duplicate requests on one member.

Open issues and PRs show several gaps before zot is reliable as a scale-out / multi-replica pull-through cache, especially for Kubernetes/containerd and shared remote storage.

## Related issues

- #4184: on-demand mirror repeatedly resyncs Docker manifest-list images such as `registry.k8s.io/pause:3.10.1`, causing high latency, repeated downloads, and failures.
- #4081: `GET /v2/<repo>/tags/list` only returns locally cached tags with on-demand sync.
- #4128: need an option to serve local immutable tags without contacting upstream.
- #4160: by-digest on-demand pulls create untagged local manifests, interacting poorly with retention/GC.
- #4187: support containerd `?ns=<registry>` as a sync/mirror selector.
- #2964: global image-store locking hurts performance under GC/retention/high concurrency.

## Relevant PRs in flight

- #4185: fixes Docker manifest-list OCI digest prediction for the #4184 resync loop.
- #4196: kind/containerd regression test for #4184.
- #4099: Redis-backed distributed coordination for on-demand sync.
- #4063 / #3778: blob streaming for on-demand sync.
- #4137: common blobstore refactor.
- #2968: per-repo/per-digest locking.
- #4195: safer dedupe rebuild/delete behavior with remote storage.
- #4166: repo-local blob read correctness with dedupe.
- #3128: sync architecture filtering.

### On-demand sync correctness

- [ ] Land/finalize #4185 and verify #4184 is fixed.
- [ ] Keep #4196 or equivalent e2e coverage.
- [ ] Address remaining first-pull latency and repeated layer-download concerns.
- [ ] Reconcile #4063 / #3778 and choose one blob-streaming approach.
- [ ] Define behavior for Docker media types with and without `preserveDigest`.

### Scale-out / multi-replica coordination

- [ ] Land or redesign #4099 so concurrent pulls trigger at most one upstream sync.
[...]

PR #4322, opened 2026-08-13
feat(sync): make upstream request concurrency/rate configurable

yinsenyan:
**What type of PR is this?**

feature

Related to #4215 (scale-out as a Kubernetes/containerd pull-through cache). This PR
addresses one concrete bottleneck in that scenario rather than closing the umbrella issue.

Sync uses a single shared `RegClient`, and regclient throttles requests per upstream
host to `ReqConcurrent = 3` by default. When one zot proxies many concurrent on-demand
pulls of *different* images from a *single* upstream (the typical pull-through-cache
deployment: one upstream Harbor, many Kubernetes node containerd clients), this cap
serializes requests to that host. Small manifest lookups then queue behind large blob
transfers, which shows up as slow manifest responses even when there is plenty of
bandwidth — a head-of-line-blocking symptom that is not tunable today.

This PR exposes two optional per-registry knobs so operators can raise the cap:

| field | type | meaning | unset default (regclient) |
|-------|------|---------|---------------------------|
| `reqConcurrent` | `*int` | max in-flight requests to the upstream and its mirrors | 3 |
| `reqPerSec` | `*float64` | max request rate (req/s) to the upstream and its mirrors | 0 (unlimited) |

Both are applied to the upstream `hostConfig` in `newClient()`; mirrors inherit them via
the existing `hostConfig` value copy. The fields are pointers so they are fully optional —
existing configs are unaffected and keep regclient's defaults. `validateSync()` rejects
non-positive values.

Repro: point one zot at a single upstream with `onDemand: true`, then drive concurrent
on-demand pulls of many different images (e.g. many k8s nodes pulling at once). With the
default cap of 3, manifest requests for small images stall behind in-flight blob copies to
the same host. Raising `reqConcurrent` removes the serialization.

- Unit tests added:
  - `pkg/extensions/sync/sync_internal_test.go` — `TestNewClientReqConcurrentReqPerSec`:
    verifies defaults are kept when unset, the upstream host picks up overrides, and mirrors
    inherit them.
  - `pkg/cli/server/root_test.go` — verify rejects `reqConcurrent: 0` and `reqPerSec: 0`,
    and accepts positive values.
  - Both suites pass locally (`go test -tags <ext> ./pkg/extensions/sync/`,
    `./pkg/cli/server/`).
- Manual end-to-end: built a linux/amd64 image and deployed it to a dev cluster as an
  on-demand pull-through cache against a single upstream. Startup `configuration settings`
  log confirms the values are parsed and loaded into the sync registry config
  (`"ReqConcurrent":30,"ReqPerSec":100`), config validation passes, and concurrent
  on-demand pulls of different images succeed.

**Automation added to e2e**:

No. Covered by unit tests above.

No. The new fields are optional pointers; when unset, behavior is identical to today
(regclient defaults). Downgrading simply ignores the fields.

Yes — two new optional sync registry config fields, documented in `examples/README.md`
and `examples/config-sync.json`.

```release-note
sync: add optional per-registry `reqConcurrent` and `reqPerSec` settings to configure the
concurrency and request rate limits used when syncing from an upstream registry (and its
mirrors). When unset, regclient's defaults are kept (concurrency 3, rate unlimited).
```

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1399,8 +1399,8 @@ Configure each registry sync:
 				"maxRetries": 5,                    # maxRetries in case of temporary errors (default: no retries)
 				"retryDelay": "1s",                 # initial HTTP retry delay; mandatory when using maxRetries
 				"maxRetryDelay": "30s",             # max HTTP retry backoff; optional, defaults to retryDelay (fixed interval). Set higher than retryDelay for exponential backoff.
-				"reqConcurrent": 30,                # max concurrent in-flight requests to this upstream and its mirrors (default: 3); raise it when one zot proxies many concurrent on-demand pulls from a single upstream
-				"reqPerSec": 100,                   # max request rate (requests/second) to this upstream and its mirrors (default: unlimited)
+				"reqConcurrent": 30,                # max concurrent in-flight requests per host, applied independently to this upstream and to each of its mirrors, not shared across them (default: 3); raise it when one zot proxies many concurrent on-demand pulls from a single upstream
+				"reqPerSec": 100,                   # max request rate (requests/second) per host, applied independently to this upstream and to each of its mirrors, not shared across them (default: unlimited)
 				"onlySigned": true,                 # sync only signed images (either notary or cosign)
 				"content":[                         # which content to periodically pull, also it's used for filtering ondemand images, if not set then periodically polling will not run
 					{
--- a/examples/README.md
+++ b/examples/README.md
@@ -1399,6 +1399,8 @@ Configure each registry sync:
 				"maxRetries": 5,                    # maxRetries in case of temporary errors (default: no retries)
 				"retryDelay": "1s",                 # initial HTTP retry delay; mandatory when using maxRetries
 				"maxRetryDelay": "30s",             # max HTTP retry backoff; optional, defaults to retryDelay (fixed interval). Set higher than retryDelay for exponential backoff.
+				"reqConcurrent": 30,                # max concurrent in-flight requests to this upstream and its mirrors (default: 3); raise it when one zot proxies many concurrent on-demand pulls from a single upstream
+				"reqPerSec": 100,                   # max request rate (requests/second) to this upstream and its mirrors (default: unlimited)
 				"onlySigned": true,                 # sync only signed images (either notary or cosign)
 				"content":[                         # which content to periodically pull, also it's used for filtering ondemand images, if not set then periodically polling will not run
 					{
--- a/examples/config-sync.json
+++ b/examples/config-sync.json
@@ -76,7 +76,9 @@
 					"maxRetries": 5,
 					"retryDelay": "1s",
 					"maxRetryDelay": "30s",
-					"syncTimeout": "10m"
+					"syncTimeout": "10m",
+					"reqConcurrent": 30,
+					"reqPerSec": 100
 				},
 				{
 					"urls": [
--- a/pkg/cli/server/root.go
+++ b/pkg/cli/server/root.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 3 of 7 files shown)


examples/README.md
```diff
@@ -1398,6 +1398,8 @@ Configure each registry sync:
 				"maxRetries": 5,                    # maxRetries in case of temporary errors (default: no retries)
 				"retryDelay": "1s",                 # initial HTTP retry delay; mandatory when using maxRetries
 				"maxRetryDelay": "30s",             # max HTTP retry backoff; optional, defaults to retryDelay (fixed interval). Set higher than retryDelay for exponential backoff.
+				"reqConcurrent": 30,                # max concurrent in-flight requests to this upstream and its mirrors (default: 3); raise it when one zot proxies many concurrent on-demand pulls from a single upstream
```
andaaron: Each host gets its own copy of the limits. One upstream plus one mirror with reqConcurrent: 30 is 30 in-flight per host, not 30 total. Worth one clarifying sentence.

Same goes for code comments.

pkg/cli/server/root.go
```diff
@@ -1712,6 +1712,22 @@ func validateSync(config *config.Config, logger zlog.Logger) error {
...
+				return fmt.Errorf("%w: %s", zerr.ErrBadConfig, msg)
+			}
+
+			if regCfg.ReqPerSec != nil && *regCfg.ReqPerSec <= 0 {
+				msg := "reqPerSec must be greater than 0"
+				logger.Error().Err(zerr.ErrBadConfig).Int("id", regID).Interface("extensions.sync.registries[id]",
+					extensionsConfig.Sync.Registries[regID]).Msg(msg)
+
+				return fmt.Errorf("%w: %s", zerr.ErrBadConfig, msg)
+			}
```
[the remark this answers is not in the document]
andaaron: The concern is valid.
But I suggest someting more like:
```
func invalidOptionalPositiveFloat(v *float64) bool {
	return v != nil && (*v <= 0 || math.IsNaN(*v) || math.IsInf(*v, 0))
}
if invalidOptionalPositiveFloat(regCfg.ReqPerSec) {
	msg := "reqPerSec must be a finite value greater than 0"
	logger.Error().Err(zerr.ErrBadConfig).Int("id", regID).Interface("extensions.sync.registries[id]",
		extensionsConfig.Sync.Registries[regID]).Msg(msg)
	return fmt.Errorf("%w: %s", zerr.ErrBadConfig, msg)
}
```

pkg/cli/server/root.go
andaaron, line 1699: Let's fix the gocyclo lint failure.

I suggest somethign similar to:
```
func validateSync(config *config.Config, logger zlog.Logger) error {
	extensionsConfig := config.CopyExtensionsConfig()
	if extensionsConfig == nil || extensionsConfig.Sync == nil {
		return nil
	}

	for regID, regCfg := range extensionsConfig.Sync.Registries {
		if err := validateSyncRegistry(config, regID, regCfg, logger); err != nil {
			return err
		}
	}

	return nil
}

func validateSyncRegistry(config *config.Config, regID int, regCfg syncconf.RegistryConfig, logger zlog.Logger) error {
	// existing retry / oauth2 / preserveDigest / content checks...
	if err := validateSyncReqLimits(regID, regCfg, logger); err != nil {
		return err
	}
	// ...
	return nil
}
```

pkg/extensions/config/sync/config.go
```diff
@@ -69,6 +69,14 @@ type RegistryConfig struct {
...
 	SyncTimeout            time.Duration  // overall HTTP client timeout for all sync operations
 	ResponseHeaderTimeout  time.Duration  `yaml:"-"` // response header timeout; set in root.go
+	// ReqConcurrent caps the number of in-flight requests to the upstream registry (and each of its
+	// mirrors). When unset it defaults to regclient's default (3). Raising it helps when a single zot
+	// proxies many concurrent on-demand pulls of different images from one upstream, where the default
+	// causes head-of-line blocking.
+	ReqConcurrent *int
+	// ReqPerSec caps the request rate (requests/second) to the upstream registry (and each of its
+	// mirrors). When unset it defaults to regclient's default (0, i.e. unlimited).
+	ReqPerSec *float64
```
andaaron: Each host gets its own copy of the limits. One upstream plus one mirror with reqConcurrent: 30 is 30 in-flight per host, not 30 total. Behavior contradicts comments.

yinsenyan:
Thanks for the review! All addressed in `3298e1ea`:

1. **gocyclo** — Split `validateSync` into `validateSyncRegistry` + `validateSyncReqLimits` + `validateSyncContent`, as suggested.
2. **Non-finite `reqPerSec`** — Added the `invalidOptionalPositiveFloat` helper (`math.IsNaN`/`IsInf`) plus a YAML test for `.nan`/`.inf`/`-.inf`.
3. **Per-host semantics** — Clarified in `config.go`, `service.go` and `examples/README.md` that the caps are per-host (upstream and each mirror), not a shared total.

`gofmt`/`go vet` clean; sync + `pkg/cli/server` verify tests pass locally.

pkg/extensions/config/sync/config.go
```diff
@@ -69,6 +69,16 @@ type RegistryConfig struct {
...
+	// ReqConcurrent caps the number of in-flight requests per upstream host. The limit is applied
+	// independently to the upstream registry and to each of its mirrors (regclient copies it onto every
+	// host), so it is a per-host cap, not a shared total across hosts. When unset it defaults to
+	// regclient's default (3). Raising it helps when a single zot proxies many concurrent on-demand pulls
+	// of different images from one upstream, where the default causes head-of-line blocking.
+	ReqConcurrent *int
+	// ReqPerSec caps the request rate (requests/second) per upstream host, applied independently to the
+	// upstream registry and to each of its mirrors (a per-host cap, not a shared total across hosts).
+	// When unset it defaults to regclient's default (0, i.e. unlimited).
+	ReqPerSec *float64
```
rchincha, line 82: Do we really need this to be a float?

The practical use case is for values between 0 and 1, in which case may this should reqs per millisec. If int instead, simplifies input validation.

andaaron (dismissed):
LGTM.

Please look at the comment from @rchincha and rebase.

pkg/extensions/config/sync/config.go
yinsenyan: Good point on the validation simplification. I kept it as `float64` for two reasons:

1. It mirrors regclient's own type — `config.Host.ReqPerSec` is a `float64` — so the value is passed straight through without a unit/type conversion on our side.
2. Fractional rates like `0.5` (one request every two seconds) are a fairly common way to express a gentle limit, and `float64` expresses them naturally.

That said, if you'd prefer `int`, I'm happy to change it — just let me know and I'll push that version.

examples/config-sync.json
```diff
@@ -76,7 +76,9 @@
 					"maxRetries": 5,
 					"retryDelay": "1s",
 					"maxRetryDelay": "30s",
-					"syncTimeout": "10m"
+					"syncTimeout": "10m",
+					"reqConcurrent": 30,
+					"reqPerSec": 100
```
vrajashkr: nit: It might be easier to reason about if these configurations were named `maxConcurrentRequests` and `maxRequestsPerSec`

pkg/extensions/config/sync/config.go
rchincha: | That said, if you'd prefer int, I'm happy to change it — just let me know and I'll push that version.

No leave it as is (as float) - reads more naturally.

andaaron (approved):

vrajashkr (approved):
LGTM. Thanks for the PR.

PR #4331, opened 2026-08-18
fix(cli): stop test server leaks and initialize Trivy logger safely

andaaron:
Make HotReloader stoppable, shut down CLI test servers after each case, and route Trivy logs through zot's slog handler to avoid DeferredHandler data races.

This is to fix the race in https://github.com/project-zot/zot/actions/runs/32109063736/job/95624416506?pr=4330

### DIFF
--- a/pkg/cli/server/extensions_test.go
+++ b/pkg/cli/server/extensions_test.go
@@ -317,11 +317,7 @@ func TestServeExtensions(t *testing.T) {
 	defer func() { os.Args = oldArgs }()
 
 	Convey("config file with no extensions", t, func(c C) {
-		logPath := MakeTempFilePath(t, "zot-log.txt")
-
-		tmpFile := t.TempDir()
-
-		content := fmt.Sprintf(`{
+		content := `{
 			"storage": {
 				"rootDirectory": "%s"
 			},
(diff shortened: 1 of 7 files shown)


pkg/cli/server/extensions_test.go
```diff
@@ -317,11 +317,7 @@ func TestServeExtensions(t *testing.T) {
 	defer func() { os.Args = oldArgs }()
 
 	Convey("config file with no extensions", t, func(c C) {
-		logPath := MakeTempFilePath(t, "zot-log.txt")
-
-		tmpFile := t.TempDir()
-
-		content := fmt.Sprintf(`{
+		content := `{
```
vrajashkr, line 320: nit: Since these configs are no longer `Sprintf`ed, they are more of a template than "content". Perhaps a rename to "confTemplate" might help make it easier to reason about.

vrajashkr (commented):
Looks good overall. Just one nit.

pkg/cli/server/extensions_test.go
andaaron: I agree in principle. But there are so many of these (including others not covered in this PR), I don't think it's worth to rename.

vrajashkr (approved):

PR #4332, opened 2026-08-18
fix(meta): release the repo quota slot when a repo is emptied

Aluchir:
**What type of PR is this?**

bug

**Which issue does this PR fix:**

None filed. Repro steps and output are below.

**What does this PR do / Why do we need it:**

storage.maxRepos is enforced against metaDB.CountRepos(), but no code path removed a repo's meta record on delete, and GC removed repo layouts without touching metadb. The count therefore never decreased: once a registry reached maxRepos, pushes creating new repositories were rejected until restart, and a layout GC reaped left a stale key that let a later push to the same name skip the limit.

Repository-level removal now happens in one place, ImageStore.RemoveIdleRepository, which drops the layout and the meta record under the same store lock once a repo holds no manifests, no blob uploads and no blobs:

- OnDeleteManifest calls it with a zero max blob age, so a repo emptied by manifest deletes is reclaimed immediately, orphan blobs included, instead of waiting for GC or a restart.
- cleanRepo calls it with the configured GC delay after blob reaping, so GC removes repos  it emptied, and also blobless empty layouts, which the old removeRepo predicate (len(gcBlobs) > 0) never removed.
- CleanupRepo no longer deletes repo directories, so every layout delete goes through RemoveIdleRepository and cannot skip metadb.

Because the layout and the meta record go together, _catalog and the quota count cannot diverge: the catalog never lists a repo the quota ignores, and the quota never counts a repo whose layout is gone.

**If an issue # is not available please add repro steps and logs showing the issue:**

Start a registry with a small quota:

```json
{
  "distSpecVersion": "1.1.1",
  "storage": { "rootDirectory": "/tmp/zot-quota", "gc": false, "maxRepos": 2 },
  "http": { "address": "127.0.0.1", "port": "15000" }
}
```

1. Push images to repo1 and repo2. Both succeed.
2. Push to repo3. Correctly refused: 429 {"errors":[{"code":"TOOMANYREQUESTS","detail":{"limit":"2"}}]}
3. Delete every manifest in repo1 by digest. Each returns 202.
4. Push to repo3 again. Still refused with 429, though only one repo now holds content.

The count never decreases, so once a registry is at its limit no new repository can ever be created. Observed directly against a Redis metadata backend, where the quota count is HLEN <keyprefix>:RepoMeta:

- before fix, after deleting repo 'alpha': HLEN zot:RepoMeta = 2   fields: alpha beta
- after  fix, after deleting repo 'alpha': HLEN zot:RepoMeta = 1   fields: beta

**Testing done on this change:**

Unit and integration tests added; all of them fail without the change (verified by stubbing RemoveIdleRepository to a no-op and watching every key test fail).

- pkg/storage/imagestore: RemoveIdleRepository removes an idle repo and its meta record, keeps repos with manifests, in-progress uploads or blobs younger than the max age, tolerates a nil metadb and a failed meta delete.
- pkg/storage/gc: GC deletes the meta record together with the layout, removes blobless empty layouts, and removes neither on a dry run.
- pkg/meta: an emptied repo loses layout and meta; a repo holding another tagged manifest, or only an untagged one, keeps both.
- pkg/api: partial delete keeps the slot, full delete releases it and drops the repo from _catalog, the emptied name is reusable and counts as a new repo under the limit, the release survives a restart, and the path holds with GC enabled.
- Manual testing against a running registry, driving the full OCI push/delete flow over HTTP: repeated fill-to-limit / delete / push-new-name cycles across the boltdb and Redis metadata backends, with GC disabled and enabled, and maxRepos of 2 and 3.
- Existing suites: pkg/storage/..., pkg/meta/... (all three backends) and pkg/api green under -race with the full extension tag set; golangci-lint v2.12.2 with the CI tag set reports 0 issues.

**Automation added to e2e:**

Yes, two cases in test/blackbox/quota.bats (wired into test/blackbox/ci.sh): deleting a repo at the limit frees a slot for a new repo, asserting the catalog no longer lists the deleted repo, and the quota is enforced again once that freed slot is taken.

No schema, config or API change. One behaviour change to note: a repository whose last manifest is deleted now disappears from _catalog immediately, where previously its empty layout remained listed until GC or restart. On upgrade, stale meta records for layouts already deleted are dropped by the existing startup reparse. On downgrade the previous behaviour resumes, with emptied repos holding their slot until a restart.

Yes. Deleting the last manifest in a repository now removes the repository immediately, its _catalog entry and its storage.maxRepos quota slot are released together, instead of the slot staying occupied until the registry restarted.

```release-note
Deleting the last manifest in a repository removes the repository: its storage layout, catalog entry and 
storage.maxRepos quota slot are released together. GC also removes emptied repositories (including 
blobless ones) together with their metadata, so a reaped repository no longer leaves a stale record 
that could skip the quota on re-push.
```

### DIFF
--- a/pkg/api/quota_test.go
+++ b/pkg/api/quota_test.go
@@ -6,7 +6,9 @@ import (
 	"net/http"
 	"sync"
 	"testing"
+	"time"
 
+	ispec "github.com/opencontainers/image-spec/specs-go/v1"
 	. "github.com/smartystreets/goconvey/convey"
 	"gopkg.in/resty.v1"
 
@@ -133,6 +135,226 @@ func TestQuotaConcurrency(t *testing.T) {
 
 			So(created, ShouldBeLessThanOrEqualTo, 5)
 			So(rejected, ShouldBeGreaterThanOrEqualTo, 5)
+		})
+	})
+}
+
+// startQuotaServerAt starts a quota-enforcing registry on a given root directory, so a test can restart
+// the server over the same storage.
+func startQuotaServerAt(t *testing.T, maxRepos int, rootDir string, gc bool) (string, func()) {
+	t.Helper()
+
+	conf := config.New()
+	conf.HTTP.Port = "0"
+	conf.Storage.RootDirectory = rootDir
+	conf.Storage.MaxRepos = maxRepos
+	// Dedupe keeps its cache open past Shutdown, blocking a restart on the same dir.
+	conf.Storage.Dedupe = false
+
+	if gc {
+		conf.Storage.GC = true
+		conf.Storage.GCDelay = 1 * time.Second
+		conf.Storage.GCInterval = 2 * time.Second
+	}
+
+	ctlr := api.NewController(conf)
+	ctlrManager := test.NewControllerManager(ctlr)
+	baseURL := ctlrManager.StartAndWait()
+
+	return baseURL, func() { ctlrManager.StopServer() }
+}
+
+func manifestDigest(t *testing.T, baseURL, repo, ref string) string {
+	t.Helper()
+
+	resp, err := resty.R().
+		SetHeader("Accept", ispec.MediaTypeImageManifest).
+		Get(baseURL + "/v2/" + repo + "/manifests/" + ref)
+	So(err, ShouldBeNil)
+
+	return resp.Header().Get("Docker-Content-Digest")
+}
+
+// uploadNewRepo pushes a complete image into a new repo.
+func uploadNewRepo(t *testing.T, baseURL, repo string) error {
+	t.Helper()
+
+	return UploadImage(CreateRandomImage(), baseURL, repo, "v1")
+}
+
+// catalogRepos lists the repositories the catalog currently reports.
+func catalogRepos(t *testing.T, baseURL string) []string {
+	t.Helper()
+
+	resp, err := resty.R().Get(baseURL + "/v2/_catalog")
+	So(err, ShouldBeNil)
+	So(resp.StatusCode(), ShouldEqual, http.StatusOK)
+
+	var catalog struct {
+		Repositories []string `json:"repositories"`
+	}
+
+	So(json.Unmarshal(resp.Body(), &catalog), ShouldBeNil)
+
+	return catalog.Repositories
+}
+
+// pushNewRepoStatus reports how the quota answers a new-repo push. Only a refusal is meaningful: the
+// manifest is sent without blobs, so an allowed push fails later in the handler. Use uploadNewRepo to
+// assert acceptance.
+func pushNewRepoStatus(t *testing.T, baseURL, repo string) int {
+	t.Helper()
+
+	img := CreateRandomImage()
+
+	manifestBody, err := json.Marshal(img.Manifest)
+	So(err, ShouldBeNil)
+
+	resp, err := resty.R().
+		SetHeader("Content-Type", ispec.MediaTypeImageManifest).
+		SetBody(manifestBody).
+		Put(baseURL + "/v2/" + repo + "/manifests/v1")
+	So(err, ShouldBeNil)
+
+	return resp.StatusCode()
+}
+
+// TestQuotaSlotReleasedOnDelete checks that a partial delete keeps the slot, a full delete releases it,
+// and the remaining images stay pullable.
+func TestQuotaSlotReleasedOnDelete(t *testing.T) {
+	Convey("Given a registry at its repo limit", t, func() {
+		rootDir := t.TempDir()
+		baseURL, stop := startQuotaServerAt(t, 3, rootDir, false)
+		defer stop()
+
+		first := CreateRandomImage()
+		second := CreateRandomImage()
+
+		So(UploadImage(first, baseURL, "shared", "v1"), ShouldBeNil)
+		So(UploadImage(second, baseURL, "shared", "v2"), ShouldBeNil)
+		So(UploadImage(CreateRandomImage(), baseURL, "second", "v1"), ShouldBeNil)
+		So(UploadImage(CreateRandomImage(), baseURL, "third", "v1"), ShouldBeNil)
+
+		So(pushNewRepoStatus(t, baseURL, "overflow"), ShouldEqual, http.StatusTooManyRequests)
+
+		Convey("Deleting one tag of a two-tag repo releases nothing", func() {
+			resp, err := resty.R().Delete(baseURL + "/v2/shared/manifests/" + first.DigestStr())
+			So(err, ShouldBeNil)
+			So(resp.StatusCode(), ShouldEqual, http.StatusAccepted)
+
+			// The sibling tag is untouched, and still serves the same bytes.
+			So(manifestDigest(t, baseURL, "shared", "v2"), ShouldEqual, second.DigestStr())
+
+			// The repo still holds content, so it still holds its slot.
+			So(pushNewRepoStatus(t, baseURL, "overflow"), ShouldEqual, http.StatusTooManyRequests)
+
+			Convey("Deleting the last tag releases the slot, and the name is reusable", func() {
+				resp, err := resty.R().Delete(baseURL + "/v2/shared/manifests/" + second.DigestStr())
+				So(err, ShouldBeNil)
+				So(resp.StatusCode(), ShouldEqual, http.StatusAccepted)
+
+				// The layout goes with the meta record, so the catalog and the quota
+				// count cannot diverge.
+				So(catalogRepos(t, baseURL), ShouldNotContain, "shared")
+
+				Convey("A differently named repo can take the freed slot", func() {
+					So(uploadNewRepo(t, baseURL, "afterfree"), ShouldBeNil)
+				})
+
+				Convey("The emptied name can be reused, and round-trips intact", func() {
+					reused := CreateRandomImage()
+					So(UploadImage(reused, baseURL, "shared", "again"), ShouldBeNil)
+					So(manifestDigest(t, baseURL, "shared", "again"), ShouldEqual, reused.DigestStr())
+				})
+			})
+		})
+	})
+}
+
+// TestQuotaSlotStaysReleasedAcrossRestart checks the startup reparse does not hand back a slot that a
+// delete released.
+func TestQuotaSlotStaysReleasedAcrossRestart(t *testing.T) {
+	Convey("Given a repo emptied while at the limit", t, func() {
+		rootDir := t.TempDir()
+		baseURL, stop := startQuotaServerAt(t, 2, rootDir, false)
+
+		doomed := CreateRandomImage()
+		keeper := CreateRandomImage()
+
+		So(UploadImage(doomed, baseURL, "doomed", "v1"), ShouldBeNil)
+		So(UploadImage(keeper, baseURL, "keeper", "v1"), ShouldBeNil)
+		So(pushNewRepoStatus(t, baseURL, "extra"), ShouldEqual, http.StatusTooManyRequests)
+
+		resp, err := resty.R().Delete(baseURL + "/v2/doomed/manifests/" + doomed.DigestStr())
+		So(err, ShouldBeNil)
+		So(resp.StatusCode(), ShouldEqual, http.StatusAccepted)
+
+		stop()
+
+		Convey("The slot is still free after a restart, and the survivor is intact", func() {
+			restartedURL, stopAgain := startQuotaServerAt(t, 2, rootDir, false)
+			defer stopAgain()
+
+			So(manifestDigest(t, restartedURL, "keeper", "v1"), ShouldEqual, keeper.DigestStr())
+			So(uploadNewRepo(t, restartedURL, "extra"), ShouldBeNil)
+		})
+	})
+}
+
+// TestQuotaSlotReleaseWithGCEnabled runs the release path with GC active, since GC also writes to
+// metadata.
+func TestQuotaSlotReleaseWithGCEnabled(t *testing.T) {
+	Convey("Given a registry with GC enabled and at its limit", t, func() {
+		rootDir := t.TempDir()
+		baseURL, stop := startQuotaServerAt(t, 2, rootDir, true)
+		defer stop()
+
+		doomed := CreateRandomImage()
+		keeper := CreateRandomImage()
+
+		So(UploadImage(doomed, baseURL, "doomed", "v1"), ShouldBeNil)
+		So(UploadImage(keeper, baseURL, "keeper", "v1"), ShouldBeNil)
+
+		resp, err := resty.R().Delete(baseURL + "/v2/doomed/manifests/" + doomed.DigestStr())
+		So(err, ShouldBeNil)
+		So(resp.StatusCode(), ShouldEqual, http.StatusAccepted)
+
+		So(uploadNewRepo(t, baseURL, "extra"), ShouldBeNil)
+
+		// Let several GC cycles run over the emptied repo.
+		time.Sleep(6 * time.Second)
+
+		So(manifestDigest(t, baseURL, "keeper", "v1"), ShouldEqual, keeper.DigestStr())
+		So(pushNewRepoStatus(t, baseURL, "yetanother"), ShouldEqual, http.StatusTooManyRequests)
+	})
+}
+
+// TestQuotaRepushOfRemovedNameCounts checks that a repo whose delete released its slot is treated as
+// brand new when pushed again: it must consume a slot, not slip past the limit.
+func TestQuotaRepushOfRemovedNameCounts(t *testing.T) {
+	Convey("Given a repo emptied while at the limit", t, func() {
+		rootDir := t.TempDir()
+		baseURL, stop := startQuotaServerAt(t, 2, rootDir, false)
+		defer stop()
+
+		doomed := CreateRandomImage()
+
+		So(UploadImage(doomed, baseURL, "doomed", "v1"), ShouldBeNil)
+		So(UploadImage(CreateRandomImage(), baseURL, "keeper", "v1"), ShouldBeNil)
+		So(pushNewRepoStatus(t, baseURL, "extra"), ShouldEqual, http.StatusTooManyRequests)
+
+		resp, err := resty.R().Delete(baseURL + "/v2/doomed/manifests/" + doomed.DigestStr())
+		So(err, ShouldBeNil)
+		So(resp.StatusCode(), ShouldEqual, http.StatusAccepted)
+		So(catalogRepos(t, baseURL), ShouldNotContain, "doomed")
+
+		Convey("Re-pushing the removed name succeeds but consumes the freed slot", func() {
+			repushed := CreateRandomImage()
+			So(UploadImage(repushed, baseURL, "doomed", "v2"), ShouldBeNil)
+			So(manifestDigest(t, baseURL, "doomed", "v2"), ShouldEqual, repushed.DigestStr())
+
+			// Had the re-push slipped past the limit, a further new name would still fit.
+			So(pushNewRepoStatus(t, baseURL, "onemore"), ShouldEqual, http.StatusTooManyRequests)
 		})
 	})
 }
--- a/pkg/meta/hooks.go
+++ b/pkg/meta/hooks.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"slices"
+	"time"
 
 	godigest "github.com/opencontainers/go-digest"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
@@ -15,6 +16,7 @@ import (
 	"zotregistry.dev/zot/v2/pkg/log"
 	mTypes "zotregistry.dev/zot/v2/pkg/meta/types"
 	"zotregistry.dev/zot/v2/pkg/storage"
+	storageTypes "zotregistry.dev/zot/v2/pkg/storage/types"
 )
 
 // priorTagManifest records where MetaDB believed each tag pointed before a digest PUT with tag=
@@ -220,6 +222,10 @@ func OnDeleteManifest(repo, reference, mediaType string, digest godigest.Digest,
 	storeController storage.StoreController, metaDB mTypes.MetaDB, log log.Logger,
 ) error {
 	if zcommon.IsReferrersTag(reference) {
+		// The store already deleted the referrers tag entry, which may have emptied the
+		// index; still attempt idle release (a no-op while content remains).
+		releaseIdleRepository(repo, storeController.GetImageStore(repo), metaDB, log)
+
 		return nil
 	}
 
@@ -233,29 +239,29 @@ func OnDeleteManifest(repo, reference, mediaType string, digest godigest.Digest,
 		return err
 	}
 
-	manageRepoMetaSuccessfully := true
+	var metaErr error
 
 	if isSignature {
-		err = metaDB.DeleteSignature(repo, signedManifestDigest, mTypes.SignatureMetadata{
+		metaErr = metaDB.DeleteSignature(repo, signedManifestDigest, mTypes.SignatureMetadata{
 			SignatureDigest: digest.String(),
 			SignatureType:   signatureType,
 		})
-		if err != nil {
-			if errors.Is(err, zerr.ErrImageMetaNotFound) {
-				// Expected: RemoveRepoReference already deleted Signatures[signedManifestDigest]
-				// when the last reference to the signed manifest was removed.
-				log.Debug().Err(err).Str("component", "metadb").
-					Msg("signature meta already removed")
-			} else {
-				log.Error().Err(err).Str("component", "metadb").
-					Msg("failed to delete signature meta")
-			}
 
-			manageRepoMetaSuccessfully = false
+		switch {
+		case errors.Is(metaErr, zerr.ErrImageMetaNotFound):
+			// Expected: RemoveRepoReference already deleted Signatures[signedManifestDigest]
+			// when the last reference to the signed manifest was removed.
+			log.Debug().Err(metaErr).Str("component", "metadb").
+				Msg("signature meta already removed")
+
+			metaErr = nil
+		case metaErr != nil:
+			log.Error().Err(metaErr).Str("tag", reference).Str("repository", repo).Str("component", "metadb").
+				Msg("failed to delete signature meta")
 		}
 	} else {
-		err = metaDB.RemoveRepoReference(repo, reference, digest)
-		if err != nil {
+		metaErr = metaDB.RemoveRepoReference(repo, reference, digest)
+		if metaErr != nil {
 			log.Info().Str("component", "metadb").Msg("restoring image store")
 
 			// restore image store
@@ -265,22 +271,53 @@ func OnDeleteManifest(repo, reference, mediaType string, digest godigest.Digest,
 					Msg("failed to restore manifest to image store, database is not consistent")
 			}
 
-			manageRepoMetaSuccessfully = false
+			log.Info().Str("tag", reference).Str("repository", repo).Str("component", "metadb").
+				Msg("failed to delete image meta for tag in repo")
+
+			return fmt.Errorf("%w: %w", zerr.ErrManifestRestoreAttempted, metaErr)
 		}
 	}
 
-	if !manageRepoMetaSuccessfully {
-		log.Info().Str("tag", reference).Str("repository", repo).Str("component", "metadb").
-			Msg("failed to delete image meta for tag in repo")
+	// The store delete has already gone through even when signature meta cleanup failed, so the
+	// index may be empty; always attempt idle release (a no-op while manifests or blobs remain).
+	releaseIdleRepository(repo, imgStore, metaDB, log)
 
-		if !isSignature {
-			return fmt.Errorf("%w: %w", zerr.ErrManifestRestoreAttempted, err)
-		}
+	return metaErr
+}
 
-		return err
+// releaseIdleRepository removes a repo left empty by a manifest delete: the storage layout goes
+// first, and only once it is gone the meta record follows, so the repo stops counting towards
+// maxRepos immediately without _catalog and the quota count diverging. A zero max blob age
+// reclaims the orphan blobs the deletes left behind. Failures are logged, not returned: the
+// manifest delete already succeeded, and ParseStorage corrects a stale record on next start.
+func releaseIdleRepository(repo string, imgStore storageTypes.ImageStore, metaDB mTypes.MetaDB, log log.Logger) {
+	var lockLatency time.Time
+
+	imgStore.Lock(&lockLatency)
+	defer imgStore.Unlock(&lockLatency)
+
+	removed, err := imgStore.RemoveIdleRepository(repo, 0)
+	if err != nil {
+		log.Error().Err(err).Str("repository", repo).Str("component", "metadb").
+			Msg("failed to remove repo emptied by manifest deletes")
+
+		return
 	}
 
-	return nil
+	if !removed {
+		return
+	}
+
+	if err := metaDB.DeleteRepoMeta(repo); err != nil {
+		// the layout is already gone; ParseStorage drops the stale record on next start
+		log.Error().Err(err).Str("repository", repo).Str("component", "metadb").
+			Msg("removed repo layout but failed to delete its meta record")
+
+		return
+	}
+
+	log.Debug().Str("repository", repo).Str("component", "metadb").
+		Msg("removed repo emptied by manifest deletes")
 }
 
 // OnGetManifest is called when a manifest is downloaded. It increments the download counter on that manifest.
--- a/pkg/storage/gc/gc.go
+++ b/pkg/storage/gc/gc.go
@@ -211,6 +211,30 @@ func (gc GarbageCollect) cleanRepo(ctx context.Context, repo string) error {
 			return err
 		}
 
+		/* remove the repo layout once nothing is left: no manifests, no blobs, no uploads.
+		This is the repository-level analogue of the manifest pruning above; the meta record
+		is dropped right after, keeping metadb on the same lifetime as storage, so a reaped
+		repo also stops counting towards storage.maxRepos. Blobs younger than the GC delay
+		keep their grace period. This runs before deleteBlobUploads so that an upload not yet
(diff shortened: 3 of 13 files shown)


pkg/meta/hooks.go
```diff
@@ -280,9 +282,47 @@ func OnDeleteManifest(repo, reference, mediaType string, digest godigest.Digest,
...
+// drops the reference, so the record would otherwise keep counting towards maxRepos until the startup
+// reparse. Emptiness comes from the index, since untagged content leaves no tag behind.
+func deleteRepoMetaIfEmpty(repo string, imgStore storageTypes.ImageStore, metaDB mTypes.MetaDB, log log.Logger) {
+	indexContent, err := imgStore.GetIndexContent(repo)
+	if err != nil {
+		log.Debug().Err(err).Str("repository", repo).Str("component", "metadb").
+			Msg("failed to read index while checking whether repo is empty")
+
+		return
+	}
```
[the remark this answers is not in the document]
Aluchir: Confirmed the premise: GetIndexContent reads index.json straight from the driver with no locking of its own, and parse.go wraps its call in RLock/RUnlock. So the caller owns the lock and I was not taking it.

Before adding the lock I checked it cannot deadlock. Both callers of OnDeleteManifest reach it after DeleteImageManifest has returned and released its own lock: the route handler in pkg/api/routes.go, and the multi-tag rollback in pkg/meta/hooks.go. So an RLock here is safe.

Applied as suggested, with the read bracketed by RLock/RUnlock and the error handling left outside the critical section.

pkg/meta/hooks.go
```diff
@@ -2,6 +2,7 @@ package meta
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"slices"
```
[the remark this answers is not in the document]
Aluchir: Fixed, added alongside the locking change

pkg/api/quota_test.go
```diff
@@ -136,3 +138,173 @@ func TestQuotaConcurrency(t *testing.T) {
...
+		So(pushNewRepoStatus(t, baseURL, "extra"), ShouldEqual, http.StatusTooManyRequests)
+
+		resp, err := resty.R().Delete(baseURL + "/v2/doomed/manifests/" + doomed.DigestStr())
+		So(err, ShouldBeNil)
+		So(resp.StatusCode(), ShouldEqual, http.StatusAccepted)
+
+		stop()
+		time.Sleep(2 * time.Second)
+
+		Convey("The slot is still free after a restart, and the survivor is intact", func() {
```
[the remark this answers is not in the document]
Aluchir: Removed! It should go, though for a slightly different reason than stated.

Shutdown does close the metadata DB, so that was not the problem. The lock I originally hit was on cache.db, the dedupe cache, which is not released on shutdown. I later set Dedupe = false in the test helper for exactly that reason, which made the sleep redundant without my noticing.

Verified: the restart test now passes 12 of 12 runs without the sleep, and each run drops from about 3s to about 1s.

andaaron:
# Symmetric repository GC: storage and MetaDB

Proposal for how zot should drop a **repository** from MetaDB, so that
storage GC and metadata GC use the same lifetime. This is a response to
the `maxRepos` leak (emptied repos keep a `RepoMeta` key) and to PR
[#4332](https://github.com/project-zot/zot/pull/4332), which deletes
`RepoMeta` when `index.json` is empty while the layout is still on disk.

**Rule:** MetaDB must not forget a repository while storage still has its
layout, and must not keep a repository after storage has deleted that
layout. Request-path DELETE and GC-off must not invent a repo-level
delete that storage does not perform.

## Terminology

GC in zot works at two levels:

| Term | Storage | MetaDB |
|------|---------|--------|
| **Manifest-level** (per reference) | One entry in `index.json` (`index.Manifests[]`); unreferenced blobs deleted after `gcDelay` | One tag/digest/signature removed via `RemoveRepoReference` / `DeleteSignature` |
| **Repository-level** (whole name) | The OCI layout directory `rootDir/<repo>` (`index.json`, `oci-layout`, `blobs/`, …) | The `RepoMeta` record for that name (`DeleteRepoMeta`) |

## What exists today

Storage GC already has both levels. MetaDB only implements manifest-level
updates today; repository-level delete is missing from GC.

| Level | Storage | MetaDB |
|-------|---------|--------|
| **Manifest-level** | Drop a descriptor from `index.json`. Unreferenced blobs are deleted after `gcDelay`. | `RemoveRepoReference` / `DeleteSignature` on the same prune (GC) or on `OnDeleteManifest` (API). |
| **Repository-level** | `CleanupRepo(removeRepo=true)` deletes `rootDir/<repo>` when every blob in that repo was reaped and there is no blob upload. | **Missing.** `DeleteRepoMeta` is never called from GC. The only production caller is `ParseStorage`, and only for names already absent from disk, at startup. |

Request path (`pkg/api/routes.go`):

- **PUT** manifest: storage first (`PutImageManifest`), then
  `OnUpdateManifest`. Meta failure is a 500.
- **DELETE** manifest: storage first (`DeleteImageManifest`), then
  `OnDeleteManifest`. `DeleteImageManifest` rewrites `index.json` and may
  delete that manifest blob. It does **not** remove the repository
  directory (`pkg/storage/imagestore/imagestore.go`).

There is no “delete repository” HTTP API. With **GC off**, nothing in
zot deletes `rootDir/<repo>`. `_catalog` still lists the layout
(`ValidateRepo` only requires `index.json`, `oci-layout`, and on
filesystem `blobs/`).

Quota (`pkg/api/quota.go`) uses MetaDB in two steps on a new-repo PUT:

1. `GetRepoMeta(name)` — if the record exists, the name is treated as an
   **existing** repository and the push is allowed without counting.
2. Otherwise `CountRepos()` — the number of `RepoMeta` keys is compared to
   `storage.maxRepos`.

Catalog lists **layouts** on disk (`GetRepositories` / `ValidateRepo`).
Today those sets diverge in both directions: GC can delete a folder and
leave a stale key (quota bypass on reuse of that name), and DELETE can
empty a layout while the key remains (429 after delete).

## Principle: two-level GC, same triggers

Mirror storage exactly.

```
Request DELETE /v2/<name>/manifests/<ref>
  storage: update index.json (manifest-level)
  meta:    RemoveRepoReference / DeleteSignature (manifest-level)
  neither side deletes the repository (baseline)

GC CleanRepo (storage.gc true)
  1. Prune index entries (retention, dangling referrers, untagged, stale)
     storage: drop descriptors, PutIndexContent
     meta:    RemoveRepoReference / DeleteSignature   (already done)
  2. Delete unreferenced blobs older than gcDelay
  3. If every blob in the repo was reaped and no upload is in flight:
     storage: Delete(rootDir/<repo>)                  (already done)
     meta:    DeleteRepoMeta(repo)                    (add this)
```

GC-off is not a special meta path. It is “step 3 never runs,” on **both**
sides, unless the optional DELETE-path removal in §6 is implemented.

## Recommended change

### 1. Repository-level MetaDB delete next to folder delete (required)

In `deleteUnreferencedBlobs` / `CleanupRepo`, after a successful
directory delete (`removeRepo &&` the `storeDriver.Delete` of the repo
path succeeds), call `metaDB.DeleteRepoMeta(repo)`.

- Same lock already held by `cleanRepo`.
- Same `removeRepo` predicate as storage (`len(gcBlobs) > 0 &&
  len(gcBlobs) == len(allBlobs)`, no blob uploads).
- Dry-run GC must not delete meta (same as not writing `index.json` /
  blobs).
- `metaDB == nil` (GC without search/metadata) stays a no-op, as today.

This is the missing analogue of storage’s repository-level GC.

### 2. Do not `DeleteRepoMeta` from `OnDeleteManifest` alone (reject PR #4332’s hook)

A last-tag DELETE leaves an OCI layout. Catalog still lists it. GC may
still need the `RepoMeta` record for untagged / referrer / retention
decisions on the next cycle. Deleting **only** the MetaDB record on DELETE:

- Frees a quota slot while `_catalog` still shows the name.
- Lets a **different** new name pass `maxRepos` while the empty layout
  occupies disk.
- Lets a later PUT to the emptied name look like a brand-new repo
  (middleware: `GetRepoMeta` miss → count++).
- Races with a concurrent push if the index snapshot is unlocked before
  `DeleteRepoMeta`.

Manifest-level `RemoveRepoReference` on DELETE is already the correct
symmetric update at manifest level.

### 3. Quota middleware: no change if MetaDB tracks layouts

`pkg/api/quota.go` can stay as-is **once** repository-level meta is
removed at the same time as the layout (§1, and §6 when GC is off).

How the middleware behaves after that fix:

| Step | What it checks | Meaning |
|------|----------------|---------|
| `GetRepoMeta(name)` | Is there a MetaDB record for this name? | “Existing repo” — skip the limit (same as today). |
| `CountRepos()` | How many `RepoMeta` keys exist? | Compared to `maxRepos` for **new** names (same as today). |

The bug today is not the middleware logic; it is that `CountRepos` only
ever grows because nothing deletes keys on DELETE or in GC. After §1,
each key corresponds to a layout still on disk, so the existing
middleware counts the right thing.

| Event | Storage | MetaDB | Catalog | Quota |
|-------|---------|--------|---------|-------|
| Last manifest DELETE, GC off (baseline) | Empty layout remains | Key remains | Lists name | Slot held |
| Last manifest DELETE, GC on, blobs still inside `gcDelay` | Empty index, layers remain | Key remains | Lists name | Slot held |
| GC reaps last blobs and deletes folder (after §1) | Layout gone | `DeleteRepoMeta` | Name gone | Slot free |
| GC deletes folder today without §1 | Layout gone | **Stale key** | Name gone | Slot held; PUT to that name **skips** limit |

No need to count `GetRepositories()` on every PUT unless MetaDB is
unavailable.

### 4. `ParseStorage`: keep as-is; no new empty-index rule

`ParseStorage` already deletes meta only for names **absent from the
storage walk** (`getReposToBeDeleted` in `pkg/meta/parse.go`). An empty
`index.json` with the layout directory still present **is** a storage
repo: `parseRepo` runs, `ResetRepoReferences` may clear tags, but the
`RepoMeta` key remains.

**For the symmetric GC fix (§1): no change to `parse.go` is required.**
GC will call `DeleteRepoMeta` when it deletes the folder; startup
reconcile already heals the case where the folder is gone but the key
remains (upgrade, crash between storage delete and meta delete, operator
removed the directory).

**Do not add** “delete meta when `index.Manifests` is empty” to
`parseRepo`. That would drop keys for layouts storage still has — the
same asymmetry as PR #4332, delayed until restart. The PR’s claim that
upgrade “cleans already-emptied repos” via reparse is only true if the
**directory** is already gone, not if only the index is empty.

### 5. Close the storage hole so “empty and blobless” can finish under GC

`removeRepo` is `len(gcBlobs) > 0 && len(gcBlobs) == len(allBlobs)`.
If the index is empty and **there are no blobs at all**, `gcBlobs` is
empty, `removeRepo` is false, and `CleanupRepo` also requires
`count > 0`. The folder is never deleted; meta would never be deleted
either (symmetry preserved, but a zombie layout).

Fix storage first, then meta follows:

- After blob GC, if `index.Manifests` is empty, `GetAllBlobs` is empty,
  and there are no blob uploads, delete the repository directory
  (same predicate as successful `removeRepo`).
- Then `DeleteRepoMeta` (§1).

Until that storage fix exists, GC-on still leaves blobless empty layouts
forever; that is a storage bug, not a reason to special-case MetaDB.

### 6. Optional: immediate slot release with GC off (symmetric DELETE path)

PR #4332’s repro uses `"gc": false` and expects a new name to succeed
right after deleting every manifest in a repo. The **baseline** proposal
(§1–§5) does not do that: with GC off, the empty layout and `RepoMeta`
key both remain, so `maxRepos` still applies.

If that behavior is required, implement it **symmetrically** — remove the
layout and the meta record **together** on the DELETE path, not meta
alone.

**When:** after a successful manifest DELETE, if the repository has no
remaining work:

- `index.json` has no manifests (`len(index.Manifests) == 0`), and
- no blobs remain in `blobs/` (or only the empty layout files), and
- no blob upload is in progress for that repo.

**What:** one shared helper, e.g. `removeIdleRepository(repo)`, called
from `OnDeleteManifest` (after manifest-level meta update) **and** from
GC (§1, §5) when the same conditions hold:

```
removeIdleRepository(repo):
  storage: delete rootDir/<repo>   (index.json, oci-layout, blobs/, …)
  meta:    DeleteRepoMeta(repo)
```

**Why not PR #4332’s approach:** `#4332` calls `DeleteRepoMeta` when the
index is empty but leaves the directory. That frees quota while `_catalog`
would still list the name until something else removes the layout.

**Edge cases:**

- **Untagged referrers / signatures still in the index** — do not remove
  the repository; only manifest-level updates (same as today). Emptiness
  must be read from `index.json`, not from tag maps in MetaDB.
- **Leftover layers after last manifest DELETE** — `DeleteImageManifest`
  may leave orphan blobs until GC; with GC off, either keep the layout
  (slot held) or delete unreferenced blobs synchronously before removing
  the directory. Prefer: only `removeIdleRepository` when `GetAllBlobs`
  is empty.
- **Concurrent PUT to the same name** — hold the store lock from the
  emptiness check through directory delete and `DeleteRepoMeta`, or
  re-check under lock before removing.
- **GC on** — §6 and §1 converge: GC already deletes the folder when all
  blobs are reaped; §6 makes DELETE immediate when the repo is already
  idle, without waiting for `gcDelay`.

**Tests (in addition to § Tests):** GC off, `maxRepos=2`, fill two repos,
delete all manifests in one, push a third name → must succeed; `_catalog`
must not list the removed name; `GetRepoMeta` for the old name must fail;
re-push to the old name must count as a new repo under the limit.

## Comparison to PR #4332

| | PR #4332 (`DeleteRepoMeta` on empty index) | Baseline (§1–§5) | With §6 (GC off immediate release) |
|--|---------------------------------------------|-------------------|-------------------------------------|
| Trigger | Last index entry gone on DELETE | Folder gone in GC (+ ParseStorage heal) | Idle repo on DELETE **or** folder gone in GC |
| GC off | Slot frees; **layout remains** | Slot held; layout remains | Slot frees; **layout removed** |
| Catalog vs quota | Diverge | Aligned | Aligned |
| Symmetry with storage | Meta delete without layout delete | Matches | Matches (both removed together) |
| Ghost meta after folder delete | Unfixed | Fixed (§1) | Fixed |

## Tests

- **GC on (§1):** fill `maxRepos`, DELETE last manifest in one repo,
  confirm 429 until GC has reaped blobs and removed the directory; then a
  new name succeeds; `GetRepoMeta` for the old name is not found; survivor
  image still pulls.
- **GC off (baseline):** same DELETE; confirm 429 remains; `_catalog`
  still lists the emptied name; `GetRepoMeta` still succeeds.
- **GC off (§6):** after DELETE empties an idle repo, new name succeeds
  immediately; catalog and meta both gone for the old name.
- **GC on, blobless empty index (§5):** folder and meta both removed in
  one `CleanRepo`.
- **Restart:** folder already deleted on disk, stale `RepoMeta` →
  `ParseStorage` drops the key (existing behavior; no parse change).
- **Quota bypass:** folder deleted, meta left (old binary) → PUT to that
  name must not skip the limit after §1.

## Code touch points

- `pkg/storage/gc/gc.go` — after successful repo-dir delete,
  `DeleteRepoMeta` (§1); blobless empty layout (§5).
- `pkg/storage/imagestore/imagestore.go` — shared
  `removeIdleRepository` (or equivalent) for directory delete.
- `pkg/meta/hooks.go` — manifest-level updates on DELETE; optional call
  to `removeIdleRepository` (§6), **not** standalone `DeleteRepoMeta`.
- `pkg/meta/parse.go` — **unchanged** (§4).
- `pkg/api/quota.go` — **unchanged** once keys track layouts (§3).
- Docs for `storage.maxRepos`: baseline counts layouts; §6 documents
  immediate release when GC is off and the repo is fully idle.

Aluchir:
> # Symmetric repository GC: storage and MetaDB
> Proposal for how zot should drop a **repository** from MetaDB, so that storage GC and metadata GC use the same lifetime. This is a response to the `maxRepos` leak (emptied repos keep a `RepoMeta` key) and to PR [#4332](https://github.com/project-zot/zot/pull/4332), which deletes `RepoMeta` when `index.json` is empty while the layout is still on disk.
> 
> **Rule:** MetaDB must not forget a repository while storage still has its layout, and must not keep a repository after storage has deleted that layout. Request-path DELETE and GC-off must not invent a repo-level delete that storage does not perform.
> 
> ## Terminology
> GC in zot works at two levels:
> 
> ....
> ## Code touch points
> * `pkg/storage/gc/gc.go` — after successful repo-dir delete,
>   `DeleteRepoMeta` (§1); blobless empty layout (§5).
> * `pkg/storage/imagestore/imagestore.go` — shared
>   `removeIdleRepository` (or equivalent) for directory delete.
> * `pkg/meta/hooks.go` — manifest-level updates on DELETE; optional call
>   to `removeIdleRepository` (§6), **not** standalone `DeleteRepoMeta`.
> * `pkg/meta/parse.go` — **unchanged** (§4).
> * `pkg/api/quota.go` — **unchanged** once keys track layouts (§3).
> * Docs for `storage.maxRepos`: baseline counts layouts; §6 documents
>   immediate release when GC is off and the repo is fully idle.

You are right, and I verified it before touching anything: with the previous patch at maxRepos: 2, emptying a repo freed its quota slot while _catalog still listed it, and a third new name then pushed successfully — the registry exceeded its own limit. Deleting only the meta record was wrong. The hook is gone; the PR now implements the symmetric
design (§1, §5 and §6).

What changed, mapped to the proposal:

- §1 + §5 + §6 land as one shared helper, ImageStore.RemoveIdleRepository: once a repo holds no manifests and no in-progress uploads, remaining blobs (unreferenced by definition, the index is empty) older than a max age are swept, and if none remain the layout and the meta record are deleted under the same store lock.
- GC calls it from cleanRepo with the configured GC delay, after blob reaping and before deleteBlobUploads so an upload not yet old enough to reap still counts as in progress and keeps the repo, preserving the guard CleanupRepo used to apply. This also covers the blobless-empty case (§5): the old removeRepo predicate required len(gcBlobs) > 0, so such layouts could never be removed.
- The DELETE path calls it with a zero max age (§6): a repo emptied by manifest deletes is reclaimed immediately, orphan blobs included. The sweep is required here because DeleteImageManifest leaves config/layer blobs behind; without it the idle condition would never hold with GC off.
- CleanupRepo no longer deletes repo directories at all, so every layout delete funnels through the helper and cannot skip metadb. §3 (quota middleware) and §4 (ParseStorage) are untouched, as proposed.

On the §6 edge case you flagged: concurrent PUT to the same name: the store lock is held from the emptiness check through the directory delete and DeleteRepoMeta, so a push either lands before the check (repo not idle, kept) or after the removal (repo re-initialised).
The residual window is a push whose blob upload has finished but whose manifest PUT has not arrived while the repo's last manifest is being deleted; that manifest PUT fails with BLOB_UNKNOWN and the client re-uploads the blob. That is the same failure class GC already has with gcDelay: 0, and the reason the DELETE path cannot honour a delay: honouring it would mean the immediate-release case never fires.

Tests follow the proposal's matrix, and all of them fail without the change (verified by stubbing RemoveIdleRepository to a no-op):

- GC on: meta record deleted together with the layout; blobless empty layout removed, dry-run removes neither.
- GC off: fill maxRepos, delete all manifests in one repo, a new name succeeds immediately; _catalog does not list the removed name; GetRepoMeta for it is gone,  re-pushing the old name counts as a new repo under the limit.
- Restart: a released slot stays released (startup reparse does not hand it back).
- Untagged-content guards: a repo holding only an untagged manifest (e.g. a signature  referrer attached by digest) keeps layout and meta.

One deviation to flag: none of the request-path behaviour changes when metadb is disabled, OnDeleteManifest is only invoked with a metadb, and GC without one removes the layout exactly as before.

pkg/storage/types/types.go
```diff
@@ -58,7 +59,8 @@ type ImageStore interface { //nolint:interfacebloat
 	GetBlobPartial(repo string, digest godigest.Digest, mediaType string, from, to int64,
 	) (io.ReadCloser, int64, int64, error)
 	DeleteBlob(repo string, digest godigest.Digest) error
-	CleanupRepo(repo string, blobs []godigest.Digest, removeRepo bool) (int, error)
+	CleanupRepo(repo string, blobs []godigest.Digest) (int, error)
+	RemoveIdleRepository(repo string, maxBlobAge time.Duration, metaDB mTypes.MetaDB) (bool, error)
```
andaaron: Interface takes `metaDB mTypes.MetaDB` — storage/types should not depend on MetaDB. Drop the parameter; return `removed bool` only.

pkg/storage/imagestore/imagestore.go
```diff
@@ -1968,16 +1970,102 @@ func (is *ImageStore) CleanupRepo(repo string, blobs []godigest.Digest, removeRe
...
+		is.log.Error().Err(err).Str("repository", repo).Msg("failed to remove repo")
+
+		return false, err
+	}
+
+	if metaDB != nil {
+		if err := metaDB.DeleteRepoMeta(repo); err != nil {
+			is.log.Error().Err(err).Str("repository", repo).
+				Msg("removed repo layout but failed to delete its meta record, startup reparse will drop it")
 		}
```
andaaron: Implementation deletes meta inside the store. Keep layout logic; move `DeleteRepoMeta` to callers. Update the doc comment at **1982–1996** / **1940–1941** so it no longer says “layout plus meta record”.

pkg/meta/hooks.go
```diff
@@ -280,9 +282,36 @@ func OnDeleteManifest(repo, reference, mediaType string, digest godigest.Digest,
...
+			Msg("failed to remove repo emptied by manifest deletes")
+
+		return
+	}
+
+	if removed {
+		log.Debug().Str("repository", repo).Str("component", "metadb").
+			Msg("removed repo emptied by manifest deletes")
+	}
+}
```
andaaron, line 321: `releaseIdleRepository` should: lock → `RemoveIdleRepository(repo, 0)` → if `removed`, `metaDB.DeleteRepoMeta`. This is the right orchestration layer.

pkg/storage/gc/gc.go
```diff
@@ -211,6 +211,19 @@ func (gc GarbageCollect) cleanRepo(ctx context.Context, repo string) error {
...
+		no blobs, no uploads. This is the repository-level analogue of the manifest pruning above,
+		keeping metadb on the same lifetime as storage, so a reaped repo also stops counting
+		towards storage.maxRepos. Blobs younger than the GC delay keep their grace period.
+		This runs before deleteBlobUploads so that an upload not yet old enough to be reaped
+		still counts as in progress and keeps the repo, as CleanupRepo's guard used to. */
+		if _, err := gc.imgStore.RemoveIdleRepository(repo, gc.opts.Delay, gc.metaDB); err != nil {
+			gc.log.Error().Err(err).Str("module", "gc").Str("repository", repo).
+				Msg("failed to remove idle repo")
+
+			return err
```
andaaron: After `RemoveIdleRepository(repo, gc.opts.Delay)` without metaDB, if `removed`, call `gc.metaDB.DeleteRepoMeta`. Prefer log-on-meta-failure (layout already gone).

pkg/meta/hooks.go
andaaron: `ErrImageMetaNotFound` is marked expected but skips `releaseIdleRepository` (**285**). Call release on this path (and treat expected miss as non-fatal).

andaaron, line 230: `IsReferrersTag` returns before release; store may already be empty. Call `releaseIdleRepository` here too.

pkg/storage/imagestore/imagestore_test.go
```diff
@@ -183,8 +190,162 @@ func TestCleanupRepoFailsOnUnexpectedDeleteBlobError(t *testing.T) {
...
+
+	removeIdle := func(store storageTypes.ImageStore, repo string, maxBlobAge time.Duration,
+		metaDB mTypes.MetaDB,
+	) (bool, error) {
+		var lockLatency time.Time
+
+		store.Lock(&lockLatency)
+		defer store.Unlock(&lockLatency)
+
+		return store.RemoveIdleRepository(repo, maxBlobAge, metaDB)
```
andaaron: Stop passing `metaDB` into the store helper; assert layout only. Meta-delete assertions belong in hooks / GC tests (already largely there).

andaaron (commented):
## Problem with the current rewrite

`ImageStore.RemoveIdleRepository` takes a `MetaDB` and calls
`DeleteRepoMeta` after deleting the layout:

```go
RemoveIdleRepository(repo string, maxBlobAge time.Duration, metaDB mTypes.MetaDB) (bool, error)
```

That puts MetaDB into `pkg/storage` / `storage/types`. Storage and MetaDB
were meant to be separate: storage owns OCI layouts; MetaDB owns search /
quota / tags; hooks and GC already hold both and are the right place to
keep them in sync.

Passing MetaDB into the store also does **not** create a real
transaction. The caller already holds the **store** lock; MetaDB has its
own locking. Nesting `DeleteRepoMeta` inside `RemoveIdleRepository` only
hides orchestration inside the wrong package.

## Desired invariant (unchanged)

| Event | Storage | MetaDB |
|-------|---------|--------|
| Manifest DELETE (index may empty) | Update `index.json` (and optionally remove idle layout — §6) | Manifest-level update, then `DeleteRepoMeta` **iff** the layout was removed |
| GC empties a repo | Delete `rootDir/<repo>` when idle | `DeleteRepoMeta` for that name |
| Layout still present | Keep directory | Keep `RepoMeta` |

Never delete `RepoMeta` while the layout remains. Never leave `RepoMeta`
after the layout is gone (except the brief window until the orchestrator
finishes, or until `ParseStorage` on restart).

## Suggested API split

### Storage — layout only

```go
// ImageStore / storage/types
RemoveIdleRepository(repo string, maxBlobAge time.Duration) (removed bool, err error)
```

Responsibilities:

1. Require caller-held store lock (same as today).
2. If index non-empty, uploads in flight, or blobs younger than
   `maxBlobAge` remain → return `(false, nil)`.
3. Otherwise delete orphan blobs (subject to age), delete
   `rootDir/<repo>`, return `(true, nil)`.
4. **Do not** import or call MetaDB.

`CleanupRepo` stays blob-only (as in the PR). Repository directory
removal stays only in `RemoveIdleRepository`.

### Orchestrators — MetaDB after successful layout removal

**DELETE path** (`pkg/meta/hooks.go` — already the natural place):

```go
func releaseIdleRepository(repo string, imgStore storageTypes.ImageStore, metaDB mTypes.MetaDB, log log.Logger) {
    var lockLatency time.Time
    imgStore.Lock(&lockLatency)
    defer imgStore.Unlock(&lockLatency)

    removed, err := imgStore.RemoveIdleRepository(repo, 0)
    if err != nil {
        log.Error().Err(err).Str("repository", repo).Msg("failed to remove idle repo layout")
        return
    }
    if !removed {
        return
    }
    if err := metaDB.DeleteRepoMeta(repo); err != nil {
        // Layout is gone; ParseStorage drops the stale key on next start.
        log.Error().Err(err).Str("repository", repo).
            Msg("removed repo layout but failed to delete its meta record")
    }
}
```

**GC path** (`pkg/storage/gc/gc.go` — already has `gc.metaDB`):

```go
removed, err := gc.imgStore.RemoveIdleRepository(repo, gc.opts.Delay)
if err != nil {
    return err
}
if removed {
    if err := gc.metaDB.DeleteRepoMeta(repo); err != nil {
        gc.log.Error().Err(err).Str("module", "gc").Str("repository", repo).
            Msg("removed repo layout but failed to delete its meta record")
        // Prefer log-not-fail (same as today's helper): layout is already
        // gone; ParseStorage heals. Returning err here would abort the
        // rest of cleanRepo after a successful layout delete.
    }
}
```

Package dependency graph stays:

```
pkg/storage          →  no MetaDB
pkg/meta (hooks)     →  storage + MetaDB   (orchestration)
pkg/storage/gc       →  storage + MetaDB   (orchestration)
```

## Secondary fix: do not skip release on expected signature paths

Today `OnDeleteManifest` only calls `releaseIdleRepository` after
`manageRepoMetaSuccessfully`. For signatures, `ErrImageMetaNotFound` is
documented as **expected** (subject already removed), but the code sets
`manageRepoMetaSuccessfully = false` and returns **before** release.
With GC off, deleting the last leftover signature can leave layout +
quota slot.

Same class: `IsReferrersTag` early-returns before release even though
`DeleteImageManifest` may already have emptied the index.

Suggested control flow:

```go
func OnDeleteManifest(...) error {
    if zcommon.IsReferrersTag(reference) {
        // Store already deleted the referrers tag entry; still try idle release.
        releaseIdleRepository(repo, storeController.GetImageStore(repo), metaDB, log)
        return nil
    }
    // ... CheckIsImageSignature ...

    var metaErr error
    if isSignature {
        metaErr = metaDB.DeleteSignature(...)
        if errors.Is(metaErr, zerr.ErrImageMetaNotFound) {
            log.Debug()... // expected
            metaErr = nil  // do not treat as hard failure for release
        }
    } else {
        metaErr = metaDB.RemoveRepoReference(...)
        // on failure: restore store, return (no release)
    }

    if metaErr != nil && !isSignature {
        return fmt.Errorf("%w: %w", zerr.ErrManifestRestoreAttempted, metaErr)
    }
    // For unexpected signature DeleteSignature errors: still attempt release
    // if the store index is empty; log metaErr. Or return metaErr after release.
    releaseIdleRepository(...)
    return metaErr // nil for expected ImageMetaNotFound
}
```

Exact error policy for unexpected signature meta failures can stay
“return err to caller” as today, but **always** attempt
`releaseIdleRepository` when the store may already be empty (signature
and referrers paths). `RemoveIdleRepository` is a no-op when manifests
or young blobs remain.

Aluchir:
Thanks for the detailed writeup, agreed the split is cleaner. Pushed the rework:

- `RemoveIdleRepository(repo, maxBlobAge)` is layout-only now and just returns whether it removed anything, no MetaDB anywhere in pkg/storage.
- `DeleteRepoMeta` moved to the callers: `releaseIdleRepository` in hooks and the GC path call it only after a successful layout removal, and log instead of failing since ParseStorage drops the stale record on the next start anyway.
- The referrers-tag path and the expected `ErrImageMetaNotFound` signature path now attempt the release too, and the expected miss returns nil instead of surfacing as an error. Unexpected signature meta failures still attempt the release first and then propagate as before.
- Store tests assert layout only, the meta ordering assertions moved into the hooks tests.

One thing I ran into while extending the orphaned-referrer test: deleting a tag whose manifest is still the subject of a live referrer only untags it, the descriptor stays in the index by digest, so the old test never actually emptied the repo. I changed it to delete the subject by digest so the last delete really is the leftover signature referrer, and it now asserts the layout and the meta record are both gone afterwards.

pkg/storage/types/types.go
Aluchir: Dropped, the interface is RemoveIdleRepository(repo, maxBlobAge) (bool, error) now and storage/types no longer imports the meta types.

pkg/storage/imagestore/imagestore.go
Aluchir: Moved out, the store only touches the layout now, and both doc comments are updated so they no longer mention the meta record.

pkg/meta/hooks.go
Aluchir: Reworked to exactly that: lock, RemoveIdleRepository(repo, 0), then DeleteRepoMeta only when the layout was removed, logging on failure since ParseStorage heals the stale record on restart.

pkg/storage/gc/gc.go
Aluchir: Done, GC deletes the meta record itself after a successful removal now, and logs on failure instead of aborting the rest of cleanRepo.

pkg/meta/hooks.go
Aluchir: Fixed, the expected miss is non-fatal now: the release still runs and OnDeleteManifest returns nil for that case. Unexpected signature meta errors attempt the release first and then propagate as before.

Aluchir: Added the release on this path too. While writing a test for it I found that deleting a tag whose manifest is still the subject of a live referrer only untags it, the descriptor stays in the index by digest, so the test deletes the subject by digest to end up with the referrer as the actual last manifest.

pkg/storage/imagestore/imagestore_test.go
Aluchir: The helper no longer takes a metaDB and the store test asserts layout only. The ordering assertions, meta record deleted only after a successful removal, live in the hooks tests now.

pkg/storage/imagestore/imagestore_test.go
```diff
@@ -183,8 +189,123 @@ func TestCleanupRepoFailsOnUnexpectedDeleteBlobError(t *testing.T) {
...
 		}
 
-		count, err := store.CleanupRepo(repo, []godigest.Digest{digest}, false)
+		count, err := store.CleanupRepo(repo, []godigest.Digest{digest})
 		So(err, ShouldNotBeNil)
 		So(count, ShouldEqual, 0)
 	})
 }
+
+func TestRemoveIdleRepository(t *testing.T) {
```
andaaron: Should we add a test for sync folders?

andaaron:
@Aluchir, the code looks good atm, please rebase.

Aluchir:
Rebased on top of c35b61df. Also brought the patch coverage up for the codecov gate: added tests for the driver failure branches in RemoveIdleRepository, the GC meta-delete failure path, and the unexpected signature meta failure in OnDeleteManifest.

sbom.json
andaaron, line 1: Where is this used for, why is it committed?

signature.json
andaaron, line 1: Where is this used for, why is it committed?

sbom.json
Aluchir: Not used anywhere, these were fixtures left behind in the repo root by a local run of test/blackbox/garbage_collect.bats and got swept into the amend by mistake. Removed both, thanks for spotting it.

signature.json
Aluchir: Same as sbom.json, a leftover from a local blackbox test run, committed by accident. Removed.

sbom.json
Aluchir: Not used anywhere, these were fixtures left behind in the repo root by a local run of test/blackbox/garbage_collect.bats and got swept into the amend by mistake. Removed both, thanks for spotting it.

signature.json
Aluchir: same, removed

andaaron (approved):

PR #4334, opened 2026-08-19
fix(events): only log successful publish when all sinks accept the event 

rchincha:
**What type of PR is this?**

Supersedes https://github.com/project-zot/zot/pull/4308

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/events/events.go
+++ b/pkg/extensions/events/events.go
@@ -42,19 +42,20 @@ func (r eventRecorder) closeSinks() error {
 
 func (r eventRecorder) publish(event *cloudevents.Event) {
 	go func() {
-		succeeded := true
+		succeeded := 0
 
 		for _, sink := range r.sinks {
 			if response := sink.Emit(event); cloudevents.IsNACK(response) || cloudevents.IsUndelivered(response) {
 				r.log.Error().Err(response).Msg("failed to publish event")
 
-				succeeded = false
+				continue
 			}
+
+			succeeded++
 		}
 
-		if succeeded {
-			r.log.Info().Msgf("event published successfully: %s", event.Type())
-		}
+		// logged unconditionally, even on partial/zero success, so the per-sink failures above have a summary line
+		r.log.Info().Msgf("event published successfully: %s (%d out of %d sinks)", event.Type(), succeeded, len(r.sinks))
 	}()
 }
 
--- a/pkg/extensions/events/events.go
+++ b/pkg/extensions/events/events.go
@@ -42,13 +42,19 @@ func (r eventRecorder) closeSinks() error {
 
 func (r eventRecorder) publish(event *cloudevents.Event) {
 	go func() {
+		succeeded := true
+
 		for _, sink := range r.sinks {
 			if response := sink.Emit(event); cloudevents.IsNACK(response) || cloudevents.IsUndelivered(response) {
 				r.log.Error().Err(response).Msg("failed to publish event")
+
+				succeeded = false
 			}
 		}
 
-		r.log.Info().Msgf("event published successfully: %s", event.Type())
+		if succeeded {
+			r.log.Info().Msgf("event published successfully: %s", event.Type())
+		}
 	}()
 }
 
--- a/pkg/extensions/events/events.go
+++ b/pkg/extensions/events/events.go
@@ -54,15 +54,23 @@ func (r eventRecorder) publish(event *cloudevents.Event) {
 			succeeded++
 		}
 
-		// Always logged, even on failure, as a summary line; level and wording reflect the actual outcome.
-		switch {
-		case succeeded == len(r.sinks):
-			r.log.Info().Msgf("event published successfully: %s (%d out of %d sinks)", event.Type(), succeeded, len(r.sinks))
-		case succeeded > 0:
-			r.log.Warn().Msgf("event publish incomplete: %s (%d out of %d sinks)", event.Type(), succeeded, len(r.sinks))
-		default:
-			r.log.Warn().Msgf("event publish failed: %s (0 out of %d sinks)", event.Type(), len(r.sinks))
+		logEvent := r.log.Info()
+		msg := "event published successfully"
+
+		if succeeded < len(r.sinks) {
+			logEvent = r.log.Warn()
+			msg = "event publish incomplete"
+
+			if succeeded == 0 {
+				msg = "event publish failed"
+			}
 		}
+
+		logEvent.Str("eventType", event.Type()).
+			Int("totalSinks", len(r.sinks)).
+			Int("sinksSucceeded", succeeded).
+			Int("sinksFailed", len(r.sinks)-succeeded).
+			Msg(msg)
 	}()
 }
 
--- a/pkg/extensions/events/events.go
+++ b/pkg/extensions/events/events.go
@@ -54,8 +54,15 @@ func (r eventRecorder) publish(event *cloudevents.Event) {
 			succeeded++
 		}
 
-		// logged unconditionally, even on partial/zero success, so the per-sink failures above have a summary line
-		r.log.Info().Msgf("event published successfully: %s (%d out of %d sinks)", event.Type(), succeeded, len(r.sinks))
+		// Always logged, even on failure, as a summary line; level and wording reflect the actual outcome.
+		switch {
+		case succeeded == len(r.sinks):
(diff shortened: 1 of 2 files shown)


pkg/extensions/events/events.go
```diff
@@ -42,13 +42,20 @@ func (r eventRecorder) closeSinks() error {
...
+
+				continue
 			}
+
+			succeeded++
 		}
 
-		r.log.Info().Msgf("event published successfully: %s", event.Type())
+		// logged unconditionally, even on partial/zero success, so the per-sink failures above have a summary line
+		r.log.Info().Msgf("event published successfully: %s (%d out of %d sinks)", event.Type(), succeeded, len(r.sinks))
```
andaaron: The unconditional INFO line with "published successfully" still fires when every sink fails (0 out of n). That was the main operational concern in #4303 — alerts or dashboards keyed on that substring will still treat total failure as success.

Suggestion: only use "successfully" when succeeded == len(r.sinks). For partial/zero, something like:


```suggestion
      switch {
      case succeeded == len(r.sinks):
          r.log.Info().Msgf("event published successfully: %s (%d out of %d sinks)", event.Type(), succeeded, len(r.sinks))
      case succeeded > 0:
          r.log.Warn().Msgf("event publish incomplete: %s (%d out of %d sinks)", event.Type(), succeeded, len(r.sinks))
      default:
          r.log.Warn().Msgf("event publish failed: %s (0 out of %d sinks)", event.Type(), len(r.sinks))
      }
```

pkg/extensions/events/events.go
```diff
@@ -42,13 +42,20 @@ func (r eventRecorder) closeSinks() error {
...
 				r.log.Error().Err(response).Msg("failed to publish event")
+
+				continue
 			}
+
+			succeeded++
 		}
 
-		r.log.Info().Msgf("event published successfully: %s", event.Type())
+		// logged unconditionally, even on partial/zero success, so the per-sink failures above have a summary line
```
andaaron: "so the per-sink failures above have a summary line" — that's clear. Once we gate "successfully" on full success, worth updating this comment to mention the level/message split so future readers don't assume INFO always means success.

pkg/extensions/events/events_test.go
```diff
@@ -88,6 +122,63 @@ func TestEvents(t *testing.T) {
 	})
 }
 
+func TestEventsSinkFailure(t *testing.T) {
+	Convey("with a failing sink", t, func() {
+		buf := &syncBuffer{}
+		logger := log.NewLoggerWithWriter("debug", buf)
+
+		Convey("logs 0 out of 1 when the only sink fails", func() {
```
andaaron, line 152: This case asserts "0 out of 1 sinks" but doesn't check that "event published successfully" is absent. Given #4303 was filed specifically because that substring appeared on failure, I'd add:

`So(buf.String(), ShouldNotContainSubstring, "event published successfully")`

or assert the warn-level incomplete/failed message if we adopt the split above

pkg/extensions/events/events_test.go
```diff
@@ -88,6 +122,63 @@ func TestEvents(t *testing.T) {
...
+
+			logged := func() bool { return strings.Contains(buf.String(), "failed to publish event") }
+			for i := 0; i < 100 && !logged(); i++ {
+				time.Sleep(10 * time.Millisecond)
+			}
+			So(logged(), ShouldBeTrue)
+			So(buf.String(), ShouldContainSubstring, "0 out of 1 sinks")
+		})
+
+		Convey("remaining sinks still receive the event and count reflects partial success", func() {
```
andaaron, line 167: Same for the partial-success case (1 out of 2 sinks): good that delivery to the healthy sink is verified. Please also assert we don't log "event published successfully" here — partial delivery shouldn't look like full success to an operator or alert rule.

andaaron (dismissed):

pkg/extensions/events/events.go
```diff
@@ -42,13 +42,27 @@ func (r eventRecorder) closeSinks() error {
...
 				r.log.Error().Err(response).Msg("failed to publish event")
+
+				continue
 			}
+
+			succeeded++
 		}
 
-		r.log.Info().Msgf("event published successfully: %s", event.Type())
+		// Always logged, even on failure, as a summary line; level and wording reflect the actual outcome.
```
vrajashkr: nit: we probably don't need this comment line.

pkg/extensions/events/events_test.go
```diff
@@ -88,6 +122,83 @@ func TestEvents(t *testing.T) {
...
+			// wait for the summary line, not just the per-sink error: it's a separate, later log
+			// statement in the same goroutine, so the error alone doesn't guarantee it's written yet
+			logged := func() bool {
+				return strings.Contains(buf.String(), "failed to publish event") &&
+					strings.Contains(buf.String(), "event publish failed")
+			}
+			for i := 0; i < 100 && !logged(); i++ {
+				time.Sleep(10 * time.Millisecond)
+			}
+			So(logged(), ShouldBeTrue)
```
vrajashkr: We could possibly have a function like `waitForLogMessage` that asserts this. I see the logic is duplicated across all the tests below.

pkg/extensions/events/events.go
```diff
@@ -42,13 +42,27 @@ func (r eventRecorder) closeSinks() error {
...
+		switch {
+		case succeeded == len(r.sinks):
+			r.log.Info().Msgf("event published successfully: %s (%d out of %d sinks)", event.Type(), succeeded, len(r.sinks))
+		case succeeded > 0:
+			r.log.Warn().Msgf("event publish incomplete: %s (%d out of %d sinks)", event.Type(), succeeded, len(r.sinks))
+		default:
+			r.log.Warn().Msgf("event publish failed: %s (0 out of %d sinks)", event.Type(), len(r.sinks))
```
vrajashkr: Should these be keys in the log instead of being part of the message?

`{"eventType": TYPE, "totalSinks": len(r.sinks), "sinksSucceeded": succeeded, "sinksFailed": len(r.sinks}-succeeded, "msg": message prefix as per condition}`

vrajashkr (commented):
Some comments on the changes.

vrajashkr:
> Commits must have verified signatures.

I think the first commit is missing a signature.

pkg/extensions/events/events_test.go
```diff
@@ -88,6 +122,89 @@ func TestEvents(t *testing.T) {
...
+
+			recorder.RepositoryCreated("test", nil)
+
+			So(waitForLogMessage(buf, "failed to publish event", "event publish failed"), ShouldBeTrue)
+			So(buf.String(), ShouldContainSubstring, `"totalSinks":1`)
+			So(buf.String(), ShouldContainSubstring, `"sinksSucceeded":0`)
+			So(buf.String(), ShouldContainSubstring, `"sinksFailed":1`)
+			// zero success must not claim "successfully", and must be logged at Warn, not Info
+			So(buf.String(), ShouldNotContainSubstring, "published successfully")
+			So(buf.String(), ShouldContainSubstring, "event publish failed")
```
vrajashkr: This assert has already been checked in line 158. We can probably omit it.

pkg/extensions/events/events_test.go
```diff
@@ -88,6 +122,89 @@ func TestEvents(t *testing.T) {
...
+			ev := <-sink.store
+			So(ev.Type(), ShouldEqual, events.RepositoryCreatedEventType.String())
+
+			So(waitForLogMessage(buf, "failed to publish event", "event publish incomplete"), ShouldBeTrue)
+			So(buf.String(), ShouldContainSubstring, `"totalSinks":2`)
+			So(buf.String(), ShouldContainSubstring, `"sinksSucceeded":1`)
+			So(buf.String(), ShouldContainSubstring, `"sinksFailed":1`)
+			// partial success must not claim "successfully" either, only full success does
+			So(buf.String(), ShouldNotContainSubstring, "published successfully")
+			So(buf.String(), ShouldContainSubstring, "event publish incomplete")
```
vrajashkr: This assert has already been checked in line 178. We can probably omit it.

vrajashkr: Would be good to also assert that log level is warn here.

vrajashkr (commented):
Looks good overall. A few final comments.

andaaron (approved):

vrajashkr (commented):
LGTM. Thanks for addressing the comments.

PR #4335, opened 2026-08-19
fix(sync): add disableHTTP2 option to saturate high-bandwidth links

### LINKED ISSUE #4333: [Feat]: sync cannot saturate a link against an HTTP/2 upstream, every request shares one TCP connection

`reqConcurrent` (#4322) caps in-flight requests per host, which fixes head-of-line blocking in the throttle. But against an HTTP/2 upstream that concurrency does not become throughput: `net/http` multiplexes every request to a host onto a single TCP connection, so all of `reqConcurrent` shares one congestion window and a single loss event stalls every in-flight blob. Concurrency divides the one flow instead of adding flows, so raising `reqConcurrent` cannot help — past a point it makes each transfer slower.

Measured in production, syncing from a remote upstream over a long path with high RTTs: **~250 Mbps aggregate across 16 concurrent blob streams on a 20 Gbps NIC, with exactly one established TCP connection to the upstream** (`ss -tn`). That is the expected steady state for a single loss-limited TCP flow — but it means `reqConcurrent` may be inert against any upstream speaking h2, which we have encountered on a lot of servers.

On-demand sync feels it worst: large layers run out `syncTimeout` at a fraction of the available bandwidth, and since each attempt allocates a fresh `.sync` session that `CleanupImage` discards on failure, an image that misses its window once never converges.

We ruled out HTTP/2 flow control as the limiter, Go's bundled http2 already advertises a 1 GiB connection window and 4 MiB per stream (`transportDefaultConnFlow` / `transportDefaultStreamFlow`). It is the single congestion window.

A per-registry opt-out alongside the existing sync knobs, defaulting to `false`:

```json
{ "urls": ["https://registry.example.com"], "reqConcurrent": 16, "disableHTTP2": true }
```

When set, sync negotiates HTTP/1.1 only against that registry, so each in-flight request gets its own connection and `reqConcurrent` parallelizes across TCP flows rather than h2 streams on one flow.

Something we got wrong on the first attempt to test this: Clearing `Transport.TLSNextProto` is not enough. It unregisters the h2 RoundTripper without stopping the handshake from advertising `h2`, and the `http.DefaultTransport.Clone()` sync uses carries `NextProtos: ["h2","http/1.1"]`. The upstream then selects h2 on a connection the client speaks HTTP/1.1 on, and every request dies with `malformed HTTP response "\x00\x00..."`. ALPN must be pinned too; Go 1.24's `Transport.Protocols` does not fix this either. Repro in https://github.com/project-zot/zot/pull/4075#issuecomment-5333981571. We also raised `MaxIdleConnsPerHost` with it. `DefaultTransport` leaves it at 2, and we paid for fresh TCP + TLS handshakes on most requests.

Implemented with unit tests (default behaviour, forced HTTP/1.1, ALPN regression) and running in production for ~36 hours: ~250 Mbps → ~6.5 Gbps; on-demand pulls that used to time out complete in a couple minutes. Happy to open a draft PR with our experimental code!

Question first, since it decides the shape of the patch: #4075 and #4114 add an HTTP/1.1 fallback on h2 framing errors — a different trigger (a slow but healthy h2 connection never produces a framing error) but the same HTTP/1.1-only transport. Would you rather have `disableHTTP2` as an independent flag, or one setting that also drives that machinery (`http2: "auto" | "disabled"`), landing after one of those merges? Unless you want a second mechanism.

- **Raise `reqConcurrent` further** — the obvious one, and what led us here. The added concurrency lands on the same connection.
- **Tune h2 flow control** (`MaxUploadBufferPer*`, `MaxReadFrameSize`) — ruled out by the Go defaults above; the windows are nowhere near binding at these rates.
- **Keep h2 but force multiple connections**, via a custom `http2.Transport` connection pool or `StrictMaxConcurrentStreams` — N congestion windows without giving up multiplexing. Technically the nicest option and the one we would most like your opinion on. We skipped it because it means owning a connection pool and reconciling it with how regclient injects per-host TLS into a plain `*http.Transport`. Willing to pursue it if you prefer it to a flag.
- **Auto-detect and downgrade** — we found no signal separating "loss-limited path" from "slow upstream"; guessing wrong is worse than an opt-in.
- **Network tuning** (BBR, buffers, an accelerating proxy) — helps, but is outside most operators' control and raises the single-flow ceiling rather than removing it.
- **Rely on #4075/#4114** — triggers only on framing errors, which this never produces.

### Additional context

How we confirmed it rather than inferred it: `ss -tn` showed one established connection while 16 blob transfers were in flight; aggregate throughput did not move when `reqConcurrent` was raised; disabling h2 produced one connection per in-flight request and multi-Gbps throughput with nothing else changed.

Cross-references: #4322 (`reqConcurrent`, merged — this is the case it cannot help); #4075 / #4114 (HTTP/1.1 fallback, same transport, different trigger); #3778 / #4063 / #4149 (sync blob streaming).

The ALPN behaviour is arguably a Go bug of its own — `Transport.Protocols` set to HTTP/1-only does not strip `h2` from inherited ALPN, because `onceSetNextProtoDefaults` returns before `adjustNextProtos`.

rchincha:
net/http multiplexes every request to a host onto a single HTTP/2 connection, so reqConcurrent-style concurrency cannot open more TCP connections and all traffic shares one congestion window. Add a per-registry disableHTTP2 config option that forces HTTP/1.1 (pinning ALPN, not just clearing TLSNextProto) so concurrent requests each get their own connection, plus a maxIdleConnsPerHost override so those connections get pooled instead of closed after Go's default of 2.

Fixes https://github.com/project-zot/zot/issues/4333

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/extensions/config/sync/config.go
+++ b/pkg/extensions/config/sync/config.go
@@ -75,6 +75,12 @@ type RegistryConfig struct {
 	// host), so it is a per-host cap, not a shared total across hosts. When unset it defaults to
 	// regclient's default (3). Raising it helps when a single zot proxies many concurrent on-demand pulls
 	// of different images from one upstream, where the default causes head-of-line blocking.
+	//
+	// ReqConcurrent and DisableHTTP2 compose, and both are usually needed together: ReqConcurrent
+	// alone still funnels every request through one HTTP/2 connection (one TCP congestion window) if
+	// the upstream negotiates h2, while DisableHTTP2 alone still caps concurrency at ReqConcurrent's
+	// default of 3 connections. To actually open more than 3 parallel connections to an HTTP/2
+	// upstream, raise ReqConcurrent and set DisableHTTP2.
 	ReqConcurrent *int
 	// ReqPerSec caps the request rate (requests/second) per upstream host, applied independently to the
 	// upstream registry and to each of its mirrors (a per-host cap, not a shared total across hosts).
@@ -82,12 +88,18 @@ type RegistryConfig struct {
 	ReqPerSec *float64
 	// DisableHTTP2 forces HTTP/1.1 to the upstream, so each concurrent request opens its own TCP
 	// connection instead of sharing the single connection (and congestion window) HTTP/2 multiplexes
-	// requests onto. Useful for saturating high-bandwidth links where ReqConcurrent alone can't help
-	// because everything still funnels through one HTTP/2 connection.
+	// requests onto. See the ReqConcurrent doc comment above: the two settings compose, and raising
+	// ReqConcurrent is what actually lets more than 3 connections open.
 	DisableHTTP2 *bool
-	// MaxIdleConnsPerHost overrides the transport's per-host idle connection pool size (Go default: 2).
-	// Only useful alongside DisableHTTP2, so that concurrent HTTP/1.1 connections are pooled/reused
-	// instead of being closed once the pool is full.
+	// MaxIdleConnsPerHost overrides the transport's per-host idle connection pool size (Go default: 2),
+	// applied whether or not DisableHTTP2 is set (it also helps a plain-HTTP or already-HTTP/1.1
+	// upstream keep its connections pooled). An explicit value here always wins.
+	//
+	// When DisableHTTP2 is true and this is left unset, it defaults to ReqConcurrent's effective
+	// value (the configured value, or regclient's default of 3) instead of Go's default of 2 —
+	// otherwise DisableHTTP2 alone would still hit the handshake-per-request pattern this option
+	// exists to avoid, since concurrent HTTP/1.1 connections beyond 2 would get closed instead of
+	// pooled.
 	MaxIdleConnsPerHost *int
 }
 
--- a/pkg/extensions/config/sync/config.go
+++ b/pkg/extensions/config/sync/config.go
@@ -80,6 +80,15 @@ type RegistryConfig struct {
 	// upstream registry and to each of its mirrors (a per-host cap, not a shared total across hosts).
 	// When unset it defaults to regclient's default (0, i.e. unlimited).
 	ReqPerSec *float64
+	// DisableHTTP2 forces HTTP/1.1 to the upstream, so each concurrent request opens its own TCP
+	// connection instead of sharing the single connection (and congestion window) HTTP/2 multiplexes
+	// requests onto. Useful for saturating high-bandwidth links where ReqConcurrent alone can't help
+	// because everything still funnels through one HTTP/2 connection.
+	DisableHTTP2 *bool
+	// MaxIdleConnsPerHost overrides the transport's per-host idle connection pool size (Go default: 2).
+	// Only useful alongside DisableHTTP2, so that concurrent HTTP/1.1 connections are pooled/reused
+	// instead of being closed once the pool is full.
+	MaxIdleConnsPerHost *int
 }
 
 // OAuth2HelperConfig holds the options used by the "oauth2" credential helper,
--- a/pkg/extensions/sync/service.go
+++ b/pkg/extensions/sync/service.go
@@ -4,6 +4,7 @@ package sync
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"net/http"
@@ -987,8 +988,8 @@ func newClient(opts syncconf.RegistryConfig, credentials syncconf.CredentialsFil
 	}
 
 	// set TLS configuration
-	tls := getTLSConfigOption(urls[0], opts.TLSVerify)
-	hostConfig.TLS = tls
+	tlsConf := getTLSConfigOption(urls[0], opts.TLSVerify)
+	hostConfig.TLS = tlsConf
 
 	if opts.CertDir != "" {
 		clientCert, clientKey, regCert, err := getCertificates(opts.CertDir)
@@ -1060,6 +1061,28 @@ func newClient(opts syncconf.RegistryConfig, credentials syncconf.CredentialsFil
 	// which are separate component timeouts. Doesn't cover body transfer time, which is expected
 	// to be slow for large images.
 	transport.ResponseHeaderTimeout = opts.ResponseHeaderTimeout
+
+	// DisableHTTP2: net/http multiplexes every request to a host onto a single HTTP/2 connection,
+	// so all of it shares one TCP congestion window regardless of client-side concurrency. Forcing
+	// HTTP/1.1 makes each concurrent request open its own TCP connection instead. Clearing
+	// TLSNextProto alone isn't enough, since the server can still select "h2" during the TLS
+	// handshake; NextProtos must be pinned so ALPN negotiation itself excludes it.
+	if opts.DisableHTTP2 != nil && *opts.DisableHTTP2 {
+		transport.ForceAttemptHTTP2 = false
+		transport.TLSNextProto = map[string]func(string, *tls.Conn) http.RoundTripper{}
+
+		if transport.TLSClientConfig == nil {
+			// No fields set here diverge from http.DefaultTransport's TLS behavior (e.g. MinVersion
+			// stays at Go's secure default); this only exists as a place to pin NextProtos below.
+			transport.TLSClientConfig = &tls.Config{} //nolint:gosec
+		}
+
+		transport.TLSClientConfig.NextProtos = []string{"http/1.1"}
+	}
+
+	if opts.MaxIdleConnsPerHost != nil {
+		transport.MaxIdleConnsPerHost = *opts.MaxIdleConnsPerHost
+	}
 
 	// Use SyncTimeout for overall HTTP client timeout. This is the maximum time for the entire
 	// HTTP request, covering all stages: DialContext (connection establishment), TLSHandshakeTimeout
--- a/pkg/extensions/sync/service.go
+++ b/pkg/extensions/sync/service.go
@@ -949,6 +949,58 @@ func httpRetryDelayBounds(opts syncconf.RegistryConfig) (time.Duration, time.Dur
 	return delayInit, delayMax, true
 }
 
+// effectiveMaxIdleConnsPerHost returns the transport's MaxIdleConnsPerHost setting: the
+// configured override when set, or — when DisableHTTP2 is enabled — reqConcurrent, so the idle
+// pool is sized to match how many HTTP/1.1 connections can actually be in flight (otherwise it
+// stays at Go's default of 2, and connections beyond that get closed instead of pooled, undoing
+// most of DisableHTTP2's benefit under concurrency). Otherwise 0, meaning "leave it at the Go
+// default" (the same value the cloned http.DefaultTransport already carries).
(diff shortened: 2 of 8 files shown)


pkg/extensions/config/sync/config.go
```diff
@@ -80,6 +80,15 @@ type RegistryConfig struct {
 	// upstream registry and to each of its mirrors (a per-host cap, not a shared total across hosts).
 	// When unset it defaults to regclient's default (0, i.e. unlimited).
 	ReqPerSec *float64
+	// DisableHTTP2 forces HTTP/1.1 to the upstream, so each concurrent request opens its own TCP
```
andaaron, line 89: This needs the same operator-facing treatment #4322 gave reqConcurrent.

Please add disableHTTP2 (and maxIdleConnsPerHost) to examples/README.md
around the reqConcurrent bullet (~line 1402 on main) and to
examples/config-sync.json.

Also spell out that the two knobs compose: disableHTTP2 without raising
reqConcurrent still caps at regclient’s default of 3 connections, and
reqConcurrent without disableHTTP2 still shares one h2 congestion
window. The issue’s working config was both (`reqConcurrent: 16`,
`disableHTTP2: true`).

When disableHTTP2 is true and maxIdleConnsPerHost is omitted, the idle
pool becomes reqConcurrent, or 3 if that is also unset.
Raising reqConcurrent is still needed to actually open more than 3
connections.

pkg/extensions/config/sync/config.go
```diff
@@ -80,6 +80,15 @@ type RegistryConfig struct {
...
+	// connection instead of sharing the single connection (and congestion window) HTTP/2 multiplexes
+	// requests onto. Useful for saturating high-bandwidth links where ReqConcurrent alone can't help
+	// because everything still funnels through one HTTP/2 connection.
+	DisableHTTP2 *bool
+	// MaxIdleConnsPerHost overrides the transport's per-host idle connection pool size (Go default: 2).
```
andaaron: The comment says this is “only useful alongside DisableHTTP2”, but
newClient applies MaxIdleConnsPerHost unconditionally
(service.go:1083-1085). That’s actually fine — it also helps a
plain-HTTP or already-HTTP/1.1 upstream — so the comment is the part
that’s wrong.

The bigger issue is the reverse pairing: DisableHTTP2 with this unset
leaves MaxIdleConnsPerHost at Go’s default of 2, which is the
handshake storm called out in #4333. If someone sets only
`"disableHTTP2": true`, they will not get the production result from
the issue.

When disableHTTP2 is true and maxIdleConnsPerHost is unset, set the
idle pool to reqConcurrent if that is set, otherwise 3 (regclient’s
default ReqConcurrent). Explicit maxIdleConnsPerHost always wins,
including when disableHTTP2 is false.

pkg/extensions/config/sync/config.go
```diff
@@ -80,6 +80,15 @@ type RegistryConfig struct {
...
+	// Only useful alongside DisableHTTP2, so that concurrent HTTP/1.1 connections are pooled/reused
+	// instead of being closed once the pool is full.
+	MaxIdleConnsPerHost *int
```
andaaron, line 103: #4322 added validateSyncReqLimits for reqConcurrent / reqPerSec
(pkg/cli/server/root.go:1804-1821) plus verify tests in root_test.go
(non-positive rejected, positive accepted).

MaxIdleConnsPerHost has no equivalent. `0` is a confusing no-op (Go
treats 0 as DefaultMaxIdleConnsPerHost=2), and a negative value would
be passed straight into the transport.

Please reject <= 0 when the pointer is set, same helper style as
reqConcurrent, and add a verify test next to
Test verify with non-positive sync reqConcurrent
(pkg/cli/server/root_test.go:2007).

To be explicit: this validation is only for when the pointer is set.
disableHTTP2: true with maxIdleConnsPerHost omitted must still verify;
newClient fills the idle pool as on service.go:1070.

pkg/extensions/sync/service.go
```diff
@@ -1061,6 +1062,28 @@ func newClient(opts syncconf.RegistryConfig, credentials syncconf.CredentialsFil
 	// to be slow for large images.
 	transport.ResponseHeaderTimeout = opts.ResponseHeaderTimeout
 
+	// DisableHTTP2: net/http multiplexes every request to a host onto a single HTTP/2 connection,
+	// so all of it shares one TCP congestion window regardless of client-side concurrency. Forcing
+	// HTTP/1.1 makes each concurrent request open its own TCP connection instead. Clearing
+	// TLSNextProto alone isn't enough, since the server can still select "h2" during the TLS
+	// handshake; NextProtos must be pinned so ALPN negotiation itself excludes it.
+	if opts.DisableHTTP2 != nil && *opts.DisableHTTP2 {
```
andaaron: The three-way disable is correct and matches the #4075 ALPN write-up:
ForceAttemptHTTP2=false, TLSNextProto={}, NextProtos=["http/1.1"].

When DisableHTTP2 is true and MaxIdleConnsPerHost is unset, set the
idle pool to ReqConcurrent if set, otherwise 3 (regclient’s default
ReqConcurrent). Leaving it at Go’s default of 2 is the handshake
storm from #4333. Explicit MaxIdleConnsPerHost always wins.

pkg/extensions/sync/service.go
```diff
@@ -1061,6 +1062,28 @@ func newClient(opts syncconf.RegistryConfig, credentials syncconf.CredentialsFil
...
+		transport.ForceAttemptHTTP2 = false
+		transport.TLSNextProto = map[string]func(string, *tls.Conn) http.RoundTripper{}
+
+		if transport.TLSClientConfig == nil {
+			// No fields set here diverge from http.DefaultTransport's TLS behavior (e.g. MinVersion
+			// stays at Go's secure default); this only exists as a place to pin NextProtos below.
+			transport.TLSClientConfig = &tls.Config{} //nolint:gosec
```
andaaron: the nil TLSClientConfig branch is dead on http.DefaultTransport.Clone() (always non-nil; that’s the Codecov miss)

pkg/extensions/sync/service.go
```diff
@@ -1061,6 +1062,28 @@ func newClient(opts syncconf.RegistryConfig, credentials syncconf.CredentialsFil
...
+		}
+
+		transport.TLSClientConfig.NextProtos = []string{"http/1.1"}
+	}
+
+	if opts.MaxIdleConnsPerHost != nil {
+		transport.MaxIdleConnsPerHost = *opts.MaxIdleConnsPerHost
```
andaaron: Codecov miss: TestNewClientDisableHTTP2 never sets
MaxIdleConnsPerHost, so this assignment is uncovered and we don’t
know the value actually lands on the client.

newClient doesn’t return the *http.Transport, so the easiest coverage
is a small helper, e.g. applySyncTransportOptions(transport, opts),
that this block calls and a unit test can inspect (DisableHTTP2
NextProtos + MaxIdleConnsPerHost). That’s also a better place to put
the “default idle pool when disableHTTP2” logic.

Please don’t accept maxIdleConnsPerHost: 0/-1 here; validate in
validateSyncReqLimits.

pkg/extensions/sync/sync_internal_test.go
```diff
@@ -2116,6 +2116,107 @@ func TestNewClientReqConcurrentReqPerSec(t *testing.T) {
 	})
 }
 
+// drainNegotiatedProto discards any value left over from a previous request on the channel,
+// so a subtest only ever observes the protocol of the request it is about to issue.
+func drainNegotiatedProto(negotiatedProto chan string) {
```
andaaron, line 2122: Copilot’s remaining note still applies. The handler comment says a
HEAD can hit more than once (auth challenge), and the channel is
buffered to 8, but drainNegotiatedProto only discards a single
value. A leftover second proto from the first Convey can make the
HTTP/2 Convey assert on HTTP/1.1 (or vice versa) and flake.

Drain until empty:

    func drainNegotiatedProto(negotiatedProto chan string) {
        for {
            select {
            case <-negotiatedProto:
            default:
                return
            }
        }
    }

pkg/extensions/sync/sync_internal_test.go
```diff
@@ -2116,6 +2116,107 @@ func TestNewClientReqConcurrentReqPerSec(t *testing.T) {
...
+		defer server.Close()
+
+		host := strings.TrimPrefix(server.URL, "https://")
+
+		// httptest's TLS cert is self-signed, so skip verification to isolate this test
+		// from cert trust; DisableHTTP2/ALPN pinning is orthogonal to cert validation.
+		tlsVerifyOff := false
+		tlsVerify := &tlsVerifyOff
+
+		Convey("DisableHTTP2 forces HTTP/1.1 against an HTTP/2-capable server", func() {
```
andaaron, line 2162: This Convey is the only DisableHTTP2=true path and it doesn’t set
MaxIdleConnsPerHost, which is why service.go:1083-1085 is uncovered.

Please assert exactly:
- disableHTTP2 true, both unset → MaxIdleConnsPerHost == 3
- disableHTTP2 true, reqConcurrent 16, idle unset → 16
- maxIdleConnsPerHost 20 set → 20 (wins over the default)

The ManifestHead protocol assertion is still the right end-to-end
check; keep it.

test/blackbox/sync.bats
```diff
@@ -180,6 +185,42 @@ EOF
...
+    "log": {
+        "level": "debug"
+    },
+    "extensions": {
+        "sync": {
+            "registries": [
+                {
+                    "urls": [
+                        "http://localhost:${zot_port3}"
+                    ],
```
andaaron: This upstream is http://localhost, not https://. disableHTTP2 only
changes ALPN / HTTP/2 on TLS. Against HTTP this flag is a no-op, so
the test cannot fail if the pin is wrong, and it cannot pass for the
reason the option exists.

The comment on the test at 275-277 already admits protocol coverage
lives in the unit test. In that case the extra zot (port5 / fifth
process in setup_file:239-240) is cost with no signal — sync.bats
already starts four servers and sleeps in several tests.

Please either:
- drop the fifth instance and fold `"disableHTTP2": true` into
  zot_sync_per_config_file (still only proves config parse, but at
  least it doesn’t add a daemon), or
- talk HTTPS to an h2-capable upstream if you really want blackbox
  protocol coverage (the unit test already does that better).

I’d drop it.

test/blackbox/sync.bats
```diff
@@ -180,6 +185,42 @@ EOF
...
+                    "onDemand": false,
+                    "tlsVerify": false,
+                    "PollInterval": "10s",
+                    "disableHTTP2": true,
+                    "maxIdleConnsPerHost": 20,
```
andaaron: If you keep any blackbox coverage, putting disableHTTP2 +
maxIdleConnsPerHost on the existing periodic-sync zot is enough to
prove verify/load accepts the keys. A dedicated process + 20s sleep
doesn’t test throughput or ALPN.

andaaron (approved):

rchincha:
cc: @seniorquico

PR #4338, opened 2026-08-19
fix(meta): don't mass-delete metaDB repos on a suspiciously empty sto…

### LINKED ISSUE #4336: [Bug]: S3 + namespaced repos: /v2/_catalog empty while per-repo reads work; metaDB emptied on restart
## zot version

`v2.1.20`

## Describe the bug

With the S3 storage driver and repositories under a namespace prefix, repository enumeration returns nothing, while per-repository reads on the same repo as the same user work fine:

```console
$ curl -s -u user:*** https://registry.example.org/v2/myorg/myapp/tags/list
{"name":"myorg/myapp","tags":["sha-aaaa0001","sha-aaaa0002"]}

$ curl -s -u user:*** https://registry.example.org/v2/_catalog
{"repositories":[]}
```

The user holds `read` on `myorg/**`, which the first call demonstrates, so this is not authz filtering in `listStorageRepositories`.

The startup walk uses the same enumeration, so metaDB is rebuilt empty on every restart: the UI and `GlobalSearch` then show nothing, and a repo only reappears once it is pushed again. One repo here has stayed invisible for hours while intact in the bucket (verified against S3 directly).

Nothing is logged at any level: `GetRepositories()` converts `PathNotFoundError` into an empty slice with a nil error.

## To reproduce

1. Configuration: `storageDriver.name: s3` with `rootdirectory: /zot`, `storage.rootDirectory: /var/lib/registry`, `extensions.search.enable: true`, single replica, `remoteCache: false`.
2. Client: push an OCI image to `<registry>/myorg/myapp:<tag>` — kaniko v1.23.2 and `docker buildx imagetools create` both succeed (201).
3. `GET /v2/_catalog` returns `{"repositories":[]}` while `GET /v2/myorg/myapp/tags/list` returns 200 with tags. Restart zot and the web UI shows no images.

## Expected behavior

`GetRepositories()` should enumerate repositories under a namespace prefix, so metaDB is rebuilt correctly at startup.

Separately: swallowing `PathNotFoundError` into an empty-but-successful result makes unreachable or misread storage indistinguishable from an empty registry, at any log level.

## Additional context

metaDB is boltdb on a persistent volume that survives restarts, so entries are being replaced by the empty walk result rather than merely not added.

Happy to test a patch or gather more detail.

rchincha:
…rage walk

A storage walk that comes back empty while metaDB already has entries was treated as authoritative: parseStorage deleted every metaDB repo not in that (possibly wrong) result on every restart, and reported the wipe as a successful "complete" parse, which could lock the emptied state in via the fast-restart stamp on the next boot. Any transient or partial storage failure that made GetRepositories() come back empty (a swallowed PathNotFoundError, a scoped/misconfigured prefix, S3 unavailability) was therefore indistinguishable from a genuinely empty registry, and silently wiped out repos that were still present and reachable in storage.

Guard the deletion pass: when storage enumerates zero repos but metaDB has some, skip deleting and mark the parse incomplete instead, so the next restart retries rather than the wipe persisting. Also log the previously silent PathNotFoundError swallow in GetRepositories(), matching the existing log line in its sibling GetNextRepositories().

Fixes https://github.com/project-zot/zot/issues/4336

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/meta/maybe_parse_test.go
+++ b/pkg/meta/maybe_parse_test.go
@@ -204,4 +204,26 @@ func TestMaybeParseStorageGate(t *testing.T) {
 		err := meta.MaybeParseStorage(mock, store, false, "v1", logger)
 		So(err, ShouldBeNil)
 	})
+
+	Convey("a suspiciously empty storage walk is not stamped (issue #4336)", t, func() {
+		// metaDB already knows about a repo, but the storage walk comes back empty - the guard
+		// in parseStorage should skip deleting it and MaybeParseStorage must not lock that state
+		// in via the fast-restart stamp, or the next restart would take the fast path and skip
+		// reparsing, permanently hiding the repo.
+		store := storage.StoreController{DefaultStore: mocks.MockedImageStore{
+			GetRepositoriesFn: func() ([]string, error) { return []string{}, nil },
+		}}
+
+		mock := mocks.MetaDBMock{
+			GetAllRepoNamesFn: func() ([]string, error) { return []string{repo}, nil },
+			SetFastRestartStampFn: func(string) error {
+				t.Fatal("must not stamp when the storage walk looked suspiciously empty")
+
+				return nil
+			},
+		}
+
+		err := meta.MaybeParseStorage(mock, store, false, "v1", logger)
+		So(err, ShouldBeNil)
+	})
 }
--- a/pkg/meta/parse.go
+++ b/pkg/meta/parse.go
@@ -26,13 +26,14 @@ const (
 
 // parseStats tracks per-repo outcomes of a storage walk.
 type parseStats struct {
-	failedRepos  int // skipped on a StatIndex or ParseRepo error
-	partialRepos int // parsed, but a manifest blob was missing
+	failedRepos      int  // skipped on a StatIndex or ParseRepo error
+	partialRepos     int  // parsed, but a manifest blob was missing
+	suspectEmptyWalk bool // storage walk returned no repos while metaDB had some; deletion pass was skipped
 }
 
 // complete reports whether the walk fully populated the metaDB.
 func (s parseStats) complete() bool {
-	return s.failedRepos == 0 && s.partialRepos == 0
+	return s.failedRepos == 0 && s.partialRepos == 0 && !s.suspectEmptyWalk
 }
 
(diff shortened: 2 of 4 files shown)


andaaron:
The fix is overstated, #4336 has two failures, and we're waiting on details from the reporter with regards to the root cause:
1. **`_catalog` empty** while per-repo reads work — **unfixed** by #4338
2. **metaDB wiped on restart** — **mitigated** by #4338

The PR treats the problem **downstream**:

```text
Driver/walk misbehaves → GetRepositories returns [] + nil
                              ↓
                    parseStorage trusts it
                              ↓
              #4338: don't delete metaDB (mitigation)
```

The preferred fix is **upstream**:

```text
Driver/walk misbehaves → return real error (or distinguish empty vs failed)
                              ↓
                    GetRepositories propagates it
                              ↓
              _catalog → 500 / logs; parseStorage → err, no silent trust
```

The storage change in #4338 only adds a **debug** line when `PathNotFoundError` is already swallowed:

```go
if errors.As(err, &perr) {
    is.log.Debug().Err(err).Str("rootDir", dir).Msg("empty rootDir")
    return stores, nil  // still success at default log level
}
```

I suggest these directions:

Layer 1 — Driver / imagestore (root cause)
- Audit what S3 `Walk` / `List` return for: empty prefix, IAM denied, wrong `rootdirectory`, nested repos.
- Reserve `PathNotFoundError` → empty **only** for a provably uninitialized root (document the condition).
- Propagate other errors from `GetRepositories` / `GetNextRepositories` (consider **warn** at minimum for swallowed NotFound, not debug-only). We should first discuss additional error types we want to define and return.
- Add S3 test: push `myorg/myapp`, assert `GetRepositories()` and `_catalog` include it (see `docs/issue-4336-triage-checklist.md`).

Layer 2 — metaDB (defense in depth)
- Land #4338-style guard: don't mass-delete on suspicious empty walk. This should depend on the correct error type being returned from storage driver, right now the storage driver doesn't properly propagate error types.
- Optionally extend later: heuristics for partial mismatch (metaDB repos absent from a non-empty but incomplete storage walk).

Layer 3 — Triage
- Collect IAM List vs Get evidence from the reporter before declaring root cause (see triage checklist).

PR #4341, opened 2026-08-21
ci: always run emulator teardown after job failures

andaaron:
Composite teardown actions only run if the caller step runs. Add if: always() on teardown uses sites, and move teardown before intentional exit 1 checks so localstack/azurite/gcs cleanup still happens on shared runners.

### DIFF
--- a/.github/actions/teardown-localstack/action.yaml
+++ b/.github/actions/teardown-localstack/action.yaml
@@ -4,5 +4,4 @@ runs:
   using: "composite"
   steps:
     - shell: bash
-      if: always()
-      run: localstack stop
+      run: localstack stop || true
--- a/.github/workflows/ecosystem-tools.yaml
+++ b/.github/workflows/ecosystem-tools.yaml
@@ -164,3 +164,4 @@ jobs:
           name: zb-cloud-scale-out-functional-results-${{ github.sha }}
           path: ./zb-results/
       - uses: ./.github/actions/teardown-localstack
+        if: always()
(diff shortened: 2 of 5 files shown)


rchincha (dismissed):

rchincha:
Is it faster to just kill the container instead of cleaning up first and then kill the container?

andaaron:
> Is it faster to just kill the container instead of cleaning up first and then kill the container?

Can you elaborate on this? Which job and commands are gou referring to? Which cleanup?

.github/actions/teardown-azurite/action.yaml
```diff
@@ -4,4 +4,5 @@ runs:
   using: "composite"
   steps:
     - shell: bash
+      if: always()
```
vrajashkr: Is this required? There is only 1 step in this action so it will inherently always run when used by a workflow.

.github/actions/teardown-localstack/action.yaml
```diff
@@ -5,4 +5,4 @@ runs:
   steps:
     - shell: bash
       if: always()
```
vrajashkr: Similar question here, though it is existing code. Since this is the only step in this workflow, wouldn't it always run anyway? Wondering if we need this.

vrajashkr (commented):
Left a couple of questions. The other changes look fine to me.

.github/actions/teardown-azurite/action.yaml
andaaron: Fixed

.github/actions/teardown-localstack/action.yaml
andaaron: fixed

vrajashkr (commented):
LGTM. Thanks for addressing the comments.

PR #4346, opened 2026-08-23
feat(search): support Trivy ignore files

Jmainguy:
**What type of PR is this?**

feature

N/A — there is no existing issue for exposing Trivy's ignore-file option through zot configuration.

Expose Trivy's native ignore-file filtering through the CVE scanner configuration. The new optional `extensions.search.cve.trivy.ignoreFile` setting is passed to Trivy's `ReportOptions.IgnoreFile` for both the default image store and every routed sub-store.

This lets operators suppress findings using Trivy's existing filtering semantics instead of maintaining zot-specific filtering logic. Both plain `.trivyignore` files and the experimental structured `.trivyignore.yaml`/`.yml` format are supported by the embedded Trivy version. The example configuration and documentation describe the option and note that the file must be accessible to the zot process.

Before this change, `TrivyConfig` has no ignore-file field, and `getNewScanOptions` leaves `flag.ReportOptions.IgnoreFile` empty. An operator therefore cannot apply an existing Trivy ignore file to CVE results returned by zot.

Reproduction:

1. Run zot with search and CVE scanning enabled.
2. Push an image with a Trivy finding that must be suppressed after external verification.
3. Add a `.trivyignore` file to the zot container.
4. Observe that there is no zot configuration setting that passes this file to the embedded Trivy scanner, so the finding remains in search results.

After this change, configure:

```json
{
  "extensions": {
    "search": {
      "enable": true,
      "cve": {
        "trivy": {
          "ignoreFile": "/etc/zot/.trivyignore.yaml"
        }
      }
    }
  }
}
```

Production startup output confirmed the configured value was loaded:

```text
"Trivy":{"DBRepository":"ghcr.io/aquasecurity/trivy-db","JavaDBRepository":"ghcr.io/aquasecurity/trivy-java-db","IgnoreFile":"/etc/zot/.trivyignore.yaml","VulnSeveritySources":["auto"],"SBOM":null}
```

Automated tests:

```text
$ GOEXPERIMENT=jsonv2 go test github.com/aquasecurity/trivy/pkg/result -run 'TestParseIgnoreFile/happy_path_valid_(config|YAML_config)_file' -count=1
ok github.com/aquasecurity/trivy/pkg/result 0.032s

$ GOEXPERIMENT=jsonv2 go test -tags search ./pkg/extensions/search/cve/trivy -run '^TestIgnoreFileConfiguration$' -count=1
ok zotregistry.dev/zot/v2/pkg/extensions/search/cve/trivy 0.057s

$ GOEXPERIMENT=jsonv2 go test -tags 'sync,scrub,metrics,search,userprefs,mgmt,imagetrust,events' ./pkg/cli/server -run '^TestServeSearchEnabledDefaultCVEDB$' -count=1
ok zotregistry.dev/zot/v2/pkg/cli/server 28.313s

$ GOEXPERIMENT=jsonv2 go vet -tags search ./pkg/extensions/search/cve/trivy ./pkg/extensions/config

$ GOEXPERIMENT=jsonv2 golangci-lint run --build-tags 'sync,scrub,metrics,search,userprefs,mgmt,imagetrust,events' ./pkg/extensions/config/... ./pkg/extensions/search/cve/trivy/... ./pkg/cli/server/...
0 issues.
```

Manual integration testing used an amd64 image built from this commit and deployed to a production zot registry.

Plain ignore file test:

```text
/etc/zot/.trivyignore:
GO-2026-5932
```

The pod started cleanly, rescanned registry images, and the GraphQL `ImageListForCVE(id: "GO-2026-5932")` query returned:

```json
{"data":{"ImageListForCVE":{"Page":{"TotalCount":0,"ItemCount":0},"Results":[]}}}
```

Experimental YAML ignore file test:

```yaml
# /etc/zot/.trivyignore.yaml
vulnerabilities:
  - id: GO-2026-5932
```

The zot configuration was changed to `"ignoreFile": "/etc/zot/.trivyignore.yaml"`. The pod again started cleanly with zero restarts, rescanned images without ignore-file parse/filter errors, and the same GraphQL query returned:

```json
{"data":{"ImageListForCVE":{"Page":{"TotalCount":0,"ItemCount":0},"Results":[]}}}
```

The YAML form is additionally scoped to vulnerability findings, unlike the plain format which applies an ID across scanner categories.

**Automation added to e2e**:

No. Unit coverage verifies that zot passes the configured path to Trivy, configuration coverage verifies JSON decoding/default handling, and Trivy's parser tests cover both supported file formats. Manual integration testing covered the complete image scan and GraphQL result path for both formats.

No. The field is optional and its zero value preserves existing behavior. Upgrading does not change scanning unless an operator configures `ignoreFile`. Before downgrading to a zot version without this field, operators should remove the setting from their configuration.

Yes. Operators can configure Trivy ignore files for zot CVE scanning.

```release-note
The Trivy CVE scanner supports an optional `extensions.search.cve.trivy.ignoreFile` setting. The configured file is used for Trivy's native vulnerability filtering and may use either the plain `.trivyignore` format or the experimental structured `.trivyignore.yaml`/`.yml` format.
```

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1503,9 +1503,11 @@ A minimal configuration only sets how often the DB is refreshed; zot applies def
 
 To set those options explicitly (for example to mirror standalone Trivy’s `--vuln-severity-source` behavior), use a `trivy` object under `cve`:
 
-- [config-cve-trivy.json](config-cve-trivy.json) — shows optional `dbRepository`, `javaDBRepository`, `ignoreFile`, `vulnSeveritySources`, and `sbom`.
+- [config-cve-trivy.json](config-cve-trivy.json) — shows optional `dbRepository`, `javaDBRepository`, `vulnSeveritySources`, and `sbom`.
 
-`ignoreFile` specifies the path to a [Trivy ignore file](https://trivy.dev/docs/latest/configuration/filtering/#trivyignore) and supports the same plain-text and YAML formats as Trivy's `--ignorefile` option. The file must be accessible to the zot process.
+`ignoreFile` specifies the path to a [Trivy ignore file](https://trivy.dev/docs/latest/configuration/filtering/#trivyignore) and supports the same plain-text and YAML formats as Trivy's `--ignorefile` option. The file must exist and be readable when zot loads its configuration. Relative paths are resolved from zot's process working directory, not from the directory containing the zot configuration file, so use an absolute path in deployments, for example `"ignoreFile": "/etc/zot/.trivyignore"`. A sample [`.trivyignore`](.trivyignore) is included in this directory.
+
+Scan results are cached by manifest digest. Editing an ignore file does not invalidate existing cached results: newly ignored vulnerabilities can remain visible, and newly unignored vulnerabilities can remain hidden, until the CVE database refresh purges the scan cache or zot restarts. The minimum database `updateInterval` is two hours.
 
 `vulnSeveritySources` is a list of source names in priority order (for example `auto`, `nvd`, or vendor IDs such as `redhat`, `alpine`). If omitted, zot defaults it to `["auto"]`, consistent with the Trivy CLI. See [Trivy: severity selection](https://trivy.dev/docs/latest/scanner/vulnerability/#severity-selection).
 
--- a/examples/README.md
+++ b/examples/README.md
@@ -1503,7 +1503,9 @@ A minimal configuration only sets how often the DB is refreshed; zot applies def
 
 To set those options explicitly (for example to mirror standalone Trivy’s `--vuln-severity-source` behavior), use a `trivy` object under `cve`:
 
-- [config-cve-trivy.json](config-cve-trivy.json) — shows optional `dbRepository`, `javaDBRepository`, `vulnSeveritySources`, and `sbom`.
+- [config-cve-trivy.json](config-cve-trivy.json) — shows optional `dbRepository`, `javaDBRepository`, `ignoreFile`, `vulnSeveritySources`, and `sbom`.
+
+`ignoreFile` specifies the path to a [Trivy ignore file](https://trivy.dev/docs/latest/configuration/filtering/#trivyignore) and supports the same plain-text and YAML formats as Trivy's `--ignorefile` option. The file must be accessible to the zot process.
 
 `vulnSeveritySources` is a list of source names in priority order (for example `auto`, `nvd`, or vendor IDs such as `redhat`, `alpine`). If omitted, zot defaults it to `["auto"]`, consistent with the Trivy CLI. See [Trivy: severity selection](https://trivy.dev/docs/latest/scanner/vulnerability/#severity-selection).
 
--- a/examples/config-cve-trivy.json
+++ b/examples/config-cve-trivy.json
@@ -18,7 +18,6 @@
         "trivy": {
           "dbRepository": "ghcr.io/aquasecurity/trivy-db",
           "javaDBRepository": "ghcr.io/aquasecurity/trivy-java-db",
-          "ignoreFile": "/etc/zot/.trivyignore.yaml",
           "vulnSeveritySources": ["auto"]
         }
       }
--- a/examples/config-cve-trivy.json
+++ b/examples/config-cve-trivy.json
@@ -18,6 +18,7 @@
         "trivy": {
           "dbRepository": "ghcr.io/aquasecurity/trivy-db",
           "javaDBRepository": "ghcr.io/aquasecurity/trivy-java-db",
+          "ignoreFile": "/etc/zot/.trivyignore.yaml",
           "vulnSeveritySources": ["auto"]
         }
       }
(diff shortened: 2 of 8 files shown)


Jmainguy:
I came across a need for this, while trying to make everything green inside of Zot, and got stuck on GO-2026-5932, which led me to https://github.com/aquasecurity/trivy/discussions/10944 which suggested they had no intentions of fixing in trivy, leaving ignoring this finding as the only remaining option for me. With this patch, all my images now show as green in Zot, which makes me happy.

examples/README.md
```diff
@@ -1503,7 +1503,9 @@ A minimal configuration only sets how often the DB is refreshed; zot applies def
 
 To set those options explicitly (for example to mirror standalone Trivy’s `--vuln-severity-source` behavior), use a `trivy` object under `cve`:
 
-- [config-cve-trivy.json](config-cve-trivy.json) — shows optional `dbRepository`, `javaDBRepository`, `vulnSeveritySources`, and `sbom`.
+- [config-cve-trivy.json](config-cve-trivy.json) — shows optional `dbRepository`, `javaDBRepository`, `ignoreFile`, `vulnSeveritySources`, and `sbom`.
+
+`ignoreFile` specifies the path to a [Trivy ignore file](https://trivy.dev/docs/latest/configuration/filtering/#trivyignore) and supports the same plain-text and YAML formats as Trivy's `--ignorefile` option. The file must be accessible to the zot process.
```
andaaron: Please document the scan cache here.

scanManifest returns scanner.cache.Get(digest) (scanner.go:639) before runTrivy. Filter (scanner.go:373) is the only place Trivy re-reads IgnoreFile, so a cache hit never sees ignore-file edits.

Adding an ID fails noisy (CVE still shows until UpdateDB → cache.Purge at scanner.go:982). Removing an ID can leave that CVE hidden until the next DB refresh (updateInterval, 2h minimum) or a process restart (new scanner, empty cache).

A file watcher / mtime bust is follow-up. A short note on this paragraph is enough for this PR.

pkg/cli/server/extensions_test.go
```diff
@@ -976,6 +976,27 @@ func TestServeSearchEnabledDefaultCVEDB(t *testing.T) {
...
+					"search": {
+						"enable": true,
+						"cve": { "trivy": { "ignoreFile": "/etc/zot/.trivyignore.yaml" } }
+					}
+				}
+			}`
+		configPath := MakeTempFileWithContent(t, "zot-test.json", content)
+
+		err := cli.LoadConfiguration(cfg, configPath)
+		So(err, ShouldBeNil)
```
andaaron, line 998: This test currently asserts LoadConfiguration succeeds with ignoreFile set to /etc/zot/.trivyignore.yaml when that path is not on disk. That matches today’s code: applyDefaultValues / validateConfiguration never Stat the file.

Trivy’s library ParseIgnoreFile (v0.73 pkg/result/ignore.go) treats fs.ErrNotExist as empty ignores (debug log only). The CLI is stricter. If the operator set ignoreFile, zot should fail config load when the path is missing or unreadable, same as sessionKeysFile / LDAP / OpenID credential files.

Please add a Stat of a non-empty ignoreFile in LoadConfiguration or validateConfiguration, and replace this case with two tests:
- temp file present and readable → load succeeds, IgnoreFile equals that path
- path missing or unreadable → load fails

Do not parse Trivy ignore syntax at config load. YAML is experimental; format errors belong in Filter at scan time.

Watch the CI trap: make verify-config runs `zot verify` on every examples/config-*. If we Stat and leave examples/config-cve-trivy.json pointing at /etc/zot/.trivyignore.yaml, verify-config will fail. Point the example at a committed sample file, or drop the key from the verified JSON.

examples/config-cve-trivy.json
```diff
@@ -18,6 +18,7 @@
         "trivy": {
           "dbRepository": "ghcr.io/aquasecurity/trivy-db",
           "javaDBRepository": "ghcr.io/aquasecurity/trivy-java-db",
+          "ignoreFile": "/etc/zot/.trivyignore.yaml",
```
andaaron: The ignoreFile key is already here. Please also add a companion sample under examples/ (plain .trivyignore and/or .trivyignore.yaml), same pattern as examples/sync-oauth2-signing-key.pem.

Relative paths work, but only against the process CWD. zot copies the string into ReportOptions.IgnoreFile; Trivy os.Stat’s that string. It is not resolved relative to the zot config file. Prefer an absolute path in this example and say that in the README.

If config load starts Stating ignoreFile (see the extensions_test.go comment), this /etc/zot/.trivyignore.yaml value will break `make verify-config`. Either:
- point this field at the committed sample, or
- omit ignoreFile from this verified example and document the setting only in examples/README.md

Jmainguy:
@andaaron Thank you for giving this a review, I believe the code now reflects your previous requested changes, I am a big fan of the project, let me know if anything else needs to be added / change. I really appreciate your help

andaaron (approved):

andaaron:
> @andaaron Thank you for giving this a review, I believe the code now reflects your previous requested changes, I am a big fan of the project, let me know if anything else needs to be added / change. I really appreciate your help

Thank you as well

PR #4353, opened 2026-08-24
fix(sync): surface on-demand upstream errors

### LINKED ISSUE #4352: [Bug]: On-demand sync failures are masked as repository/manifest not found
### zot version

v2.1.20

### Summary

When an on-demand pull-through sync fails, Zot logs the underlying error but discards it:

```go
if errSync := routeHandler.c.SyncOnDemand.SyncImage(...); errSync != nil {
    log.Error().Err(errSync).Msg("failed to sync image")
}

return imgStore.GetImageManifest(name, reference)
```

If the image is not already cached, the subsequent local lookup fails and the client receives a misleading repository or manifest-not-found error.

For example, an upstream TLS failure is logged as:

```text
tls: failed to verify certificate: x509: certificate signed by unknown authority
```

But the pulling client receives:

```text
repository name not known to registry
```

This makes the failure difficult for an end user to diagnose because they generally do not have access to Zot logs.

### To reproduce

1. Configure an on-demand sync registry with `tlsVerify: true`.
2. Point it at an upstream registry whose certificate is signed by an untrusted CA.
3. Pull an uncached image through Zot.
4. Compare the client response with the Zot logs.

### Expected behavior

When on-demand synchronization fails and no cached manifest is available, Zot should return a sanitized representation of the synchronization failure to the client.

### Actual behavior

The synchronization error is only logged. The client receives the error from the subsequent local lookup, usually `NAME_UNKNOWN` or `MANIFEST_UNKNOWN`.

### Screenshots

_No response_

### Additional context

This is related to #4290 and #4292. PR #4292 preserved structured error details for sync diagnostics, but the on-demand manifest path still discards the returned sync error before generating the client response.

amold1:
Fixes #4352.

### What does this PR do?

Preserves and returns on-demand synchronization failures when the requested manifest is not available in the local cache.

Previously, `getImageManifest` logged errors returned by `SyncImage` and then discarded them. The subsequent local lookup caused clients to receive a misleading `NAME_UNKNOWN` or `MANIFEST_UNKNOWN` response.

This change classifies upstream failures as:

- TLS verification failure
- Authentication failure
- Connectivity or timeout failure
- Generic synchronization failure

These failures are returned as sanitized HTTP 502 responses. TLS errors include the certificate verification reason, while upstream URLs and arbitrary internal error details are not exposed.

If an upstream refresh fails but the manifest is already cached, Zot continues to serve the cached manifest.

### Testing

Added regression coverage verifying that:

- TLS failures return the certificate verification cause.
- Upstream URLs are not exposed.
- Authentication failures are classified.
- Timeouts are classified.
- Unrecognized errors return a generic message without leaking their contents.
- Cached manifests are served when an upstream refresh fails.

```text
GOEXPERIMENT=jsonv2 go test \
  -tags 'sync,scrub,metrics,search,lint,userprefs,mgmt,imagetrust,ui' \
  ./pkg/api \
  -run 'TestGetManifest(OnDemandSyncErrors|CheckInterval)$' \
  -count=1
```

### User-facing change

Before:

```text
repository name not known to registry
```

After, for an upstream certificate failure:

```text
upstream TLS verification failed: x509: certificate signed by unknown authority
```


andaaron:
I am not sure about this.
1. There're nothing the end user client side can do to fix sync configuration issues. Those can only be fixed by zot admin users, who already have access to the zot server side logs to troubleshoot.
2. Sending the failure details to the end user client side could be a vulnerability, as it could disclose information about the zot environment/setup to a person who could be looking for exploits.

rchincha:
OTOH, if the admin and user is the same, then makes it easier to debug.
Pls DO NOT merge this, yet.

amold1:
@rchincha @andaaron Thank you for taking a look at this and I appreciate your feedback.

I agree this PR may be too aggressive in exposing errors which it should not but I think responding to the request with `repository name not known to registry` type error is misleading.

I am ok with sending something like a `502 Bad Gateway` to signify that there was an issue between the registry and the upstream registry and then admins can take a look. That way at least users are not confused about the error. Currently one might think that the image doesn't exist on the repository at all(cache or upstream)

Let me know your thoughts

PR #4359, opened 2026-08-26
fix(compat): preserve Docker media types through MetaDB and search

### LINKED ISSUE #4355: [Bug]: Docker image format compatibility with Notation signed image
### zot version

v2.1.20

### Describe the bug

> ⚠️ Full disclaimer first - the issue was debugged with the help of Claude Sonnet 5.

Dear ZOT community,
While configuring ZOT with `http.compat: ["docker2s2"]` and the trust extension enabled ( `"trust": { "enable": true, "cosign": true, "notation": true }` ) I noticed that the images I pushed (via `oras copy -r` from our internal Artifactory) showed up in ZOT as Signed-by: "unknown". Using the notation CLI I was able to confirm that the images in ZOT are signed, still the badge in the UI being orange was unnerving. The images in question are signed by a commercial certificate from DigiCert.

### To reproduce

1. Configuration: the full config.json is below - the main part is to enable docker2s2 compatibility and the trust extension for notation.
2. Client tool used:
  - `oras` to copy the artifacts 
    - make sure to push an artifact that is a Docker-format manifest (application/vnd.docker.distribution.manifest.v2+json)
  - `curl` to query ZOT
  - Firefox to visit the ZOT UI.
3. Seen error - the UI showed an orange badge for an image that notation CLI confirmed to be signed by a verified authority. 
```JSON
{
  "distSpecVersion": "1.1.0",
  "storage": {
    "rootDirectory": "/var/lib/zot/data",
    "dedupe": true,
    "gc": true,
    "gcDelay": "1h"
  },
  "http": {
    "address": "0.0.0.0",
    "port": "5000",
    "tls": {
      "cert": "/etc/zot/tls/server.crt",
      "key": "/etc/zot/tls/server.key"
    },
   "compat": ["docker2s2"]
  },
  "log": {
    "level": "info",
    "output": "/var/log/zot/zot.log"
  },
  "extensions": {
    "search": { "enable": true },
    "ui": { "enable": true },
    "trust": {
      "enable": true,
      "cosign": true,
      "notation": true
    }
  }
}
```

### Expected behavior

Green badge showing:
- Tool: notation
- Signed-by: CN=E....,O=E....,L=P....,C=CZ

### Screenshots

_No response_

### Additional context

ZOT logs had no record indicating why the docker images weren't trusted. With the help of the Claude CLI and Sonnet 5 the issue was investigated in these steps:
1. more verbose logging was needed to understand what is going wrong. The file `zot/pkg/meta/boltdb/boltdb.go` was edited on line 1399 and the blank identifier was replaced by `err` to catch the error:
```Go
for _, layerInfo := range sigInfo.LayersInfo {
		author, date, isTrusted, err := imgTrustStore.VerifySignature(sigType, layerInfo.LayerContent,
				layerInfo.SignatureKey, manifestDigest, mConvert.GetImageMeta(&protoImageMeta), repo)
		if err != nil {
				bdw.Log.Error().Err(err).Str("repo", repo).Str("signatureType", sigType).
						Str("manifestDigest", manifestDigest.String()).
						Msg("failed to verify signature validity")
		}
```
2. Thanks to this change a more helpful error message was logged:
```JSON
{
  "time": "2026-08-25T11:44:01.862733393+02:00",
  "level": "error",
  "message": "failed to verify signature validity",
  "error": "content descriptor mismatch",
  "repo": ".....",
  "signatureType": "notation",
  "manifestDigest": "sha256:e31....",
  "caller": "zotregistry.dev/zot/v2/pkg/meta/boltdb/boltdb.go:1402",
  "func": "zotregistry.dev/zot/v2/pkg/meta/boltdb.(*BoltDB).UpdateSignaturesValidity.func1",
  "goroutine": 1
}
```
3. Claude kept digging further and further (I have minimal experience with the Go programming language), and to cut a long story short, it arrived at the following root cause: 

> The real bug turned out to be a media-type normalization bug that spans three layers of zot's own metadata handling:
> 1. `pkg/meta/convert/convert.go` (`GetImageManifestMeta`) — hardcoded every single-manifest image's cached MediaType to the OCI constant, discarding the actual wire format.
> 2. `pkg/meta/convert/convert_proto.go` (`GetProtoImageMeta/GetProtoImageManifestData`) — the BoltDB/Redis/DynamoDB proto-serialization layer had the same hardcoding, plus a switch that matched only the literal OCI string.
> 3. `pkg/extensions/search/{resolver.go,convert/metadb.go}` and `pkg/meta/common/common.go` — four more places doing the same exact-OCI-only switch, silently relying on the fact that MediaType was always wrong-but-consistent (OCI) before.

I forked the repo and added the changes in this commit https://github.com/jkovacs-eleveo/zot/commit/af297369a78822462d56a0f9d1bbf848d1660526. I have deployed the binary from this branch to our test environment and the images now have a nice green badge.

andaaron:
Docker schema2 manifest and manifest-list Content-Types were rewritten to OCI during metadata conversion and several OCI-only switches, so docker2s2 images lost their wire type in MetaDB, GraphQL, CLI, and signature validity checks (including Notation trust).

- Persist ingest/wire mediaType on ImageMeta and through proto conversion
- Prefer nested JSON mediaType when set; fill empty nested from descriptor
- Accept Docker types alongside OCI in MetaDB, search, CLI, and imagestore
- Log VerifySignature errors during signature validity updates

Closes #4355 

### DIFF
--- a/pkg/extensions/search/convert/metadb.go
+++ b/pkg/extensions/search/convert/metadb.go
@@ -12,6 +12,7 @@ import (
 
 	zerr "zotregistry.dev/zot/v2/errors"
 	zcommon "zotregistry.dev/zot/v2/pkg/common"
+	"zotregistry.dev/zot/v2/pkg/compat"
 	cveinfo "zotregistry.dev/zot/v2/pkg/extensions/search/cve"
 	"zotregistry.dev/zot/v2/pkg/extensions/search/gql_generated"
 	"zotregistry.dev/zot/v2/pkg/extensions/search/pagination"
@@ -389,10 +390,12 @@ type (
 
 func FullImageMeta2ImageSummary(ctx context.Context, fullImageMeta mTypes.FullImageMeta,
 ) (*gql_generated.ImageSummary, map[BlobDigest]int64, error) {
-	switch fullImageMeta.MediaType {
-	case ispec.MediaTypeImageManifest:
+	switch {
+	case fullImageMeta.MediaType == ispec.MediaTypeImageManifest ||
+		compat.IsCompatibleManifestMediaType(fullImageMeta.MediaType):
 		return ImageManifest2ImageSummary(ctx, fullImageMeta)
-	case ispec.MediaTypeImageIndex:
+	case fullImageMeta.MediaType == ispec.MediaTypeImageIndex ||
+		compat.IsCompatibleManifestListMediaType(fullImageMeta.MediaType):
 		return ImageIndex2ImageSummary(ctx, fullImageMeta)
 	default:
 		return nil, nil, zerr.ErrMediaTypeNotSupported
@@ -412,7 +415,7 @@ func ImageIndex2ImageSummary(ctx context.Context, fullImageMeta mTypes.FullImage
 		indexBlobs          = map[string]int64{}
 
 		indexDigestStr    = fullImageMeta.Digest.String()
-		indexMediaType    = ispec.MediaTypeImageIndex
+		indexMediaType    = fullImageMeta.MediaType
 		lastPullTimestamp = fullImageMeta.Statistics.LastPullTimestamp
 		pushTimestamp     = fullImageMeta.Statistics.PushTimestamp
 		pushedBy          = fullImageMeta.Statistics.PushedBy
@@ -430,7 +433,7 @@ func ImageIndex2ImageSummary(ctx context.Context, fullImageMeta mTypes.FullImage
 		imageManifestSummary, manifestBlobs, err := ImageManifest2ImageSummary(ctx, mTypes.FullImageMeta{
 			Repo:            fullImageMeta.Repo,
 			Tag:             fullImageMeta.Tag,
-			MediaType:       ispec.MediaTypeImageManifest,
+			MediaType:       imageManifest.Manifest.MediaType,
 			Digest:          imageManifest.Digest,
 			Size:            imageManifest.Size,
 			Manifests:       []mTypes.FullManifestMeta{imageManifest},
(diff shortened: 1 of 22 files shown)


pkg/extensions/search/convert/metadb.go
```diff
@@ -389,10 +390,12 @@ type (
 
 func FullImageMeta2ImageSummary(ctx context.Context, fullImageMeta mTypes.FullImageMeta,
 ) (*gql_generated.ImageSummary, map[BlobDigest]int64, error) {
-	switch fullImageMeta.MediaType {
-	case ispec.MediaTypeImageManifest:
+	switch {
+	case fullImageMeta.MediaType == ispec.MediaTypeImageManifest ||
```
vrajashkr, line 394: The condition of `fullImageMeta.MediaType == ispec.MediaTypeImageManifest ||
		compat.IsCompatibleManifestMediaType(fullImageMeta.MediaType)`

seems to be repeated quite a bit. I'm wondering if we can have a helper function for it. Something like `IsSupportedManifestMediaType` where OCI and docker (if docker support is enabled) are "supported media types".

Similar comment for the ManifestList for multi-arch.

vrajashkr (commented):
Partial review. Just one comment for now.

pkg/extensions/search/convert/metadb.go
andaaron: This is across the entire zot codebase. Can you open a separate issue to track it?

rchincha: If that is the case, yes, let's cleanup separately.
https://github.com/project-zot/zot/issues/4376

rchincha (approved):
we should make sure the ui issue (notation signatures, etc) also gets fixed with this.

PR #4363, opened 2026-08-27
fix(api)!: make blob HEAD/range reads repo-local by default

andaaron:
BREAKING CHANGE: with dedupe enabled, HEAD and ranged GET no longer
rematerialize digests from the store dedupe cache into the target
repository. Clients that depended on that HEAD side effect must use
POST `/v2/<name>/blobs/uploads/?mount=<digest>`, or set
storage.hydrateBlobOnRead to true (per store/subpath).

Default blob reads use StatBlob so OCI AtomicDelete (DELETE then HEAD → 404)
holds. POST `?mount=` still uses CheckBlob. When access control is enabled,
mount and opt-in hydrate paths require create on the destination and read on
a cache source via AccessController.can. Share writeBlobReadError for
HEAD/GET; document the flag; cover default, opt-in, and authz paths in
tests; run conformance with dedupe enabled.

### DIFF
--- a/examples/config-boltdb.json
+++ b/examples/config-boltdb.json
@@ -3,6 +3,7 @@
     "storage": {
         "rootDirectory": "/tmp/zot",
         "dedupe": true,
+        "hydrateBlobOnRead": false,
         "remoteCache": false
     },
     "http": {
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -1021,33 +1021,133 @@ func (rh *RouteHandler) DeleteManifest(response http.ResponseWriter, request *ht
 	response.WriteHeader(http.StatusAccepted)
 }
 
-// canMount checks if a user has read permission on cached blobs with this specific digest.
-// returns true if the user have permission to copy blob from cache.
-func canMount(userAc *reqCtx.UserAccessControl, imgStore storageTypes.ImageStore, digest godigest.Digest,
+// canMount reports whether the caller may materialize digest into destRepo from
+// the dedupe cache. That requires create on destRepo and read on at least
+// one repository that already holds the digest. Permissions are evaluated with
+// AccessController.can so policy CEL conditions apply. Call only when authz is
+// enabled.
+func (rh *RouteHandler) canMount(
+	request *http.Request, userAc *reqCtx.UserAccessControl, imgStore storageTypes.ImageStore,
+	digest godigest.Digest, destRepo string,
 ) (bool, error) {
-	canMount := true
+	authnMwCtx, err := reqCtx.GetAuthnMiddlewareContext(request.Context())
+	if err == nil && authnMwCtx != nil && authnMwCtx.AuthnType == BEARER {
+		return true, nil
+	}
 
-	if userAc != nil {
-		canMount = false
+	// Snapshot AC config and reuse the controller logger (avoid opening Log.Output per call).
+	accessControlConfig := rh.c.Config.CopyAccessControlConfig()
+	if accessControlConfig == nil {
+		accessControlConfig = &config.AccessControlConfig{}
+	}
 
-		repos, err := imgStore.GetAllDedupeReposCandidates(digest)
-		if err != nil {
-			return false, err
-		}
+	acCtrlr := &AccessController{
+		Config: accessControlConfig,
+		Log:    rh.c.Log,
+	}
 
-		if len(repos) == 0 {
-			canMount = false
-		}
+	digestRef := digest.String()
 
-		// check if user can read any repo which contain this blob
-		for _, repo := range repos {
-			if userAc.Can(constants.ReadPermission, repo) {
-				canMount = true
-			}
+	if ok, _ := acCtrlr.can(request, userAc, constants.CreatePermission, destRepo, digestRef); !ok {
+		return false, nil
+	}
+
+	repos, err := imgStore.GetAllDedupeReposCandidates(digest)
+	if err != nil {
+		return false, err
+	}
+
+	for _, repo := range repos {
+		if ok, _ := acCtrlr.can(request, userAc, constants.ReadPermission, repo, digestRef); ok {
+			return true, nil
 		}
 	}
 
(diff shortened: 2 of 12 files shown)


pkg/api/routes.go
```diff
@@ -1021,33 +1021,123 @@ func (rh *RouteHandler) DeleteManifest(response http.ResponseWriter, request *ht
...
+// AccessController.can so policy CEL conditions apply. Call only when authz is
+// enabled.
+func (rh *RouteHandler) canMount(
+	request *http.Request, userAc *reqCtx.UserAccessControl, imgStore storageTypes.ImageStore,
+	digest godigest.Digest, destRepo string,
 ) (bool, error) {
-	canMount := true
+	authnMwCtx, err := reqCtx.GetAuthnMiddlewareContext(request.Context())
+	if err == nil && authnMwCtx != nil && authnMwCtx.AuthnType == BEARER {
+		return true, nil
```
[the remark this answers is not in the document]
andaaron: out of scope

examples/config-boltdb.json
```diff
@@ -3,6 +3,7 @@
     "storage": {
         "rootDirectory": "/tmp/zot",
         "dedupe": true,
+        "crossRepoBlobMountOnRead": false,
```
rchincha: hydrateBlobOnRead

andaaron: Updated

pkg/api/routes.go
```diff
@@ -1083,55 +1173,15 @@ func (rh *RouteHandler) CheckBlob(response http.ResponseWriter, request *http.Re
...
-	} else {
-		var lockLatency time.Time
-
-		imgStore.RLock(&lockLatency)
-		defer imgStore.RUnlock(&lockLatency)
-
-		ok, blen, _, err = imgStore.StatBlob(name, digest)
-	}
-
+	ok, blen, err := rh.resolveBlobPresence(request, imgStore, name, digest)
```
rchincha, line 1192: Something is not right here.
There are two cases here right?
1. an explicit mount where you have to check src and dest rbac
2. a hydrateBlobPath, which doesn't have an explicit src and only dest rbac needs to be checked?

andaaron: Both paths end up calling CheckBlob, which can hard-link from the store’s dedupe cache into the destination, so they share the same authz bar via canMount:

- Explicit mount (POST ...?mount=): create on dest + read on at least one repo that already holds the digest.
Hydrate on read (hydrateBlobOnRead on HEAD / ranged GET): same checks. There is no from= query param, but the source is still implicit — GetAllDedupeReposCandidates(digest) — and we require read on one of those before rematerializing.
- Dest-only for hydrate would not be enough: middleware already grants read on the destination for HEAD/GET, while hydrate is a write into that repo from cache content that may belong to other repositories. Skipping source-read would re-open cross-repo reuse without permission on a holding repo.

So resolveBlobPresence → userMayMountBlob → canMount is intentional for both cases; hydrate is not dest-RBAC-only.

andaaron: With hydrate rematerialization, blob reads and mounts need:

### `HEAD /v2/<dest>/blobs/<digest>` (hydrate rematerialize)

| Where | Permission |
| --- | --- |
| **dest** | `read` (implemented in authz middleware) **and** `create` (`implemented in canMount`) |
| **src** (a repo that already holds the digest) | `read` (`implemented in canMount`) |

### `GET /v2/<dest>/blobs/<digest>` (incl. ranged GET; hydrate rematerialize)

| Where | Permission |
| --- | --- |
| **dest** | `read` (implemented in authz middleware) **and** `create` (`implemented in canMount`) |
| **src** (a repo that already holds the digest) | `read` (`implemented in canMount`) |

Same path as HEAD: both go through `resolveBlobPresence` → `userMayMountBlob` → `canMount`.

### `POST /v2/<dest>/blobs/uploads/?mount=<digest>` (explicit mount)

| Where | Permission |
| --- | --- |
| **dest** | `create` (implemented in authz middleware) **and** `create` again via `implemented in canMount`, this is a duplication, but here for coonsistency |
| **src** (a repo that already holds the digest) | `read` (`implemented in canMount`) |

In zot ACL terms that is `create`, not a generic `write`.

### Nuances

- Default (`hydrateBlobOnRead` false): `HEAD` / `GET` only need `read` on dest (`StatBlob`); no source check.
- Hydrate on but rematerialize not allowed (no create / no source read): `HEAD` / `GET` still work for digests already in dest via the `StatBlob` fallback; they just will not pull from the cache.
- Explicit `POST ?mount=` always uses `canMount` when authz is enabled; it does not depend on `hydrateBlobOnRead`.

pkg/extensions/extension_mgmt_internal_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAuthMarshalJSONNilHTPasswd(t *testing.T) {
+	t.Parallel()
+
+	buf, err := json.Marshal(Auth{APIKey: true})
```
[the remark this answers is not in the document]
andaaron: This was supposed to be pushed in a different PR.

pkg/api/routes.go
```diff
@@ -1349,7 +1409,7 @@ func (c *byteCountingWriter) Write(p []byte) (int, error) {
 // layer's metadata path (e.g. a deleted blob) would historically have
 // produced a 4xx; under this design they too truncate. The 16-range
 // cap and coalesceRanges already bound the worst case, and the eager
```
rchincha, line 1411: // cap and coalesceRanges already bound the worst case, and the eager
// existence check in GetBlob (resolveBlobPresence — StatBlob by
// default, or CheckBlob when hydrateBlobOnRead permits rematerialization)
// still rejects the obvious "blob does not exist" case before we get here.

^ more accurate?

PR #4365, opened 2026-08-28
fix(api): only apply failDelay when credentials are sent

### LINKED ISSUE #4361: [Bug]: If you log in fast, the delayed `/v2/` probe causes a spontaneous logout.
### zot version

v2.1.18

### Describe the bug

Logging into the zot UI quickly can result in being logged out approximately five seconds later. Waiting about six seconds on the login page before submitting the same credentials avoids the problem.

This appears to be a race between:

1. An unauthenticated `GET /v2/` initiated by the login page.
2. A successful authenticated `GET /v2/` initiated when the user submits the    login form.

When the initial unauthenticated request completes after the successful login, its delayed `401` response expires the newly created `session` cookie. The UI's global `401` handler then calls `POST /zot/auth/logout` and redirects to `/login`.

### To reproduce

1. Configure zot with:
   - UI enabled
   - htpasswd authentication
   - an authentication failure delay of approximately five seconds
2. Open `/login` in a fresh browser session.
3. Immediately submit valid credentials, before the initial unauthenticated
   `/v2/` request completes. Storing the password in the browser is what allows the fast login here.
4. Optional: Click around, e.g., search for a repository or open one from the list.
5. Observe that the UI redirects to `/login` about five seconds after the
   initial page load.

As a control, reload `/login`, wait at least six seconds, and then submit the same credentials. The session remains authenticated.

Reproduced in Firefox 154, but it appears timing-dependent rather than browser-specific.

### Expected behavior

Don't get logged out randomly (or, what feels like randomly).

### Screenshots

_No response_

### Additional context

_No response_

andaaron:
UI login probes GET /v2/ with X-ZOT-API-CLIENT but no Authorization header. Delaying those 401s widened a race with a later successful login: the stale response could trigger logout after navigation.

Limit authFail sleep to requests that actually supplied credentials, preserving brute-force protection for failed basic auth.

Fixes project-zot/zot#4361

### DIFF
--- a/pkg/api/authn.go
+++ b/pkg/api/authn.go
@@ -817,7 +817,10 @@ func getRelyingPartyArgs(cfg *config.Config, provider string, hashKey, encryptKe
 }
 
 func authFail(w http.ResponseWriter, r *http.Request, realm string, delay int) {
-	if !isAuthorizationHeaderEmpty(r) || hasSessionHeader(r) {
+	// Only delay when credentials were supplied. UI probes (X-ZOT-API-CLIENT without
+	// Authorization) and expired-session checks must not sleep, or a stale 401 can race
+	// with a later successful login
+	if !isAuthorizationHeaderEmpty(r) {
 		time.Sleep(time.Duration(delay) * time.Second)
 	}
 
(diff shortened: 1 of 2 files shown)


rchincha (approved):
what a corner case!

andaaron:
Superseded by https://github.com/project-zot/zot/pull/4368

PR #4374, opened 2026-08-30
fix(gc): treat missing blobs as GC-eligible instead of aborting repo GC

### LINKED ISSUE #4271: [Bug]: GC aborts the whole repo (and blocks remaining repo tasks) when index.json references a missing blob
## What happened

On zot **v2.1.5** (linux-amd64, S3 storage driver, `gc: true`, retention keepTags policies, dedupe+remoteCache/redis), one repository's `index.json` referenced a manifest digest whose blob did not exist in storage — the digest of empty content (`sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`). The data arrived via a bucket-level migration from a docker `registry:2` layout, so it never passed zot's push validation.

Every GC pass then failed like this:

```
"failed to stat blob"                       error="blob not found" repository=<repo> digest=sha256:e3b0c4...
"failed to determine GC delay"              error="blob not found"
"failed to run GC for /zot/<repo>"          error="blob not found"
"failed to execute task"  component=scheduler
```

Observed impact: not only was that repo never collected — **repos scheduled after it were not garbage-collected either**. In our estate GC reached only 3 of 16 repos before dying, on every pass, indefinitely. Storage reclamation was silently disabled estate-wide while the registry kept growing (~90 GiB of collectable data accumulated). After we manually uploaded the missing (empty) blob, the next pass completed all 16 repos and reclaimed the space.

## Why

`pkg/storage/gc/gc.go`: `isBlobOlderThan()` returns the `StatBlob` error, `gcManifest()` propagates it, and `cleanRepo()` aborts — one dangling reference poisons the entire repo's GC and, via the task error, subsequent repo tasks in the pass. The propagation is unchanged on current `main` (`isBlobOlderThan` → `return false, err`).

## Expected behavior

A missing blob is precisely the thing GC exists to clean up around — it should be treated as "nothing to preserve" for that manifest (or logged and skipped, e.g. counted as a dangling-reference warning), letting GC continue with the remaining manifests and repos. A single corrupt/dangling entry — which bucket migrations, partial writes, or eventual-consistency blips can all produce — should degrade one manifest's handling, not permanently disable storage reclamation for the whole registry with no surfaced health signal.

Related (closed) #4226 is the same failure class on GCS: one unexpected storage object silently stops GC enumeration.

## Reproduction sketch

1. Create an S3-backed repo layout where `index.json` contains a manifest descriptor whose digest has no corresponding `blobs/sha256/<digest>` file (bucket-level write, bypassing the API).
2. Enable `gc` with any retention policy.
3. Start zot; observe the log chain above and that repos after the failing one are not collected.

## zot version

v2.1.5 (error propagation identical on main at time of filing)

marogz:
**What type of PR is this?**

bug

Fixes #4271

When a repo's `index.json` references a blob digest that no longer exists in storage, `isBlobOlderThan()` in `pkg/storage/gc/gc.go` propagates the `StatBlob` error all the way up through `gcManifest()` and `cleanRepo()`, aborting GC for that entire repo.

This PR changes `isBlobOlderThan()` to distinguish between a genuine "not found" error (the referenced blob truly doesn't exist, so it is treated as GC-eligible), and any other error (e.g. transient S3/network errors) that still fails closed as before, since we can't safely tell those apart from a real "missing" blob, and we don't want to risk deleting live index rows during transient storage issues.

To make the `PathNotFoundError` detection possible with `errors.As`, `pkg/storage/imagestore/imagestore.go` was also updated so that `ErrBlobNotFound` is wrapped together with the underlying driver error instead of discarding it. Previously the underlying error was swallowed, which made it impossible for callers to check whether the root cause was specifically a missing path.

As a side effect of this richer error wrapping, a few existing tests that asserted on the exact error string/value (`ShouldEqual "blob not found"` / `ShouldEqual zerr.ErrBlobNotFound`) were relaxed to `ShouldContainSubstring` / `errors.Is(...)` checks, since the error message now includes the wrapped underlying error.

N/A, see #4271.

Unit tests updated/added:
- `pkg/storage/gc/gc_internal_test.go`: `TestGarbageCollectWithMockedImageStore` — updated the "removeReferrersWithMissingSubject" case to reflect that a missing referrer blob is now treated as GC-eligible (stale referrer entry gets removed) instead of aborting.
- `pkg/storage/local/local_test.go`: `TestGarbageCollectErrors` — updated the "Missing untagged manifest blob" case to assert `CleanRepo` now succeeds instead of erroring when the underlying blob file was removed from disk.
- `pkg/storage/gcs/gcs_test.go`, `pkg/storage/scrub_test.go`, `pkg/test/oci-utils/oci_layout_test.go`: relaxed exact-string/exact-value error assertions to account for the richer wrapped error message.

Manual test runs:
```
$ make test-extended
...
ok      zotregistry.dev/zot/v2/pkg/storage      (cached)        coverage: 7.2% of statements in ./...
ok      zotregistry.dev/zot/v2/pkg/storage/azure        (cached)        coverage: 0.5% of statements in ./...
ok      zotregistry.dev/zot/v2/pkg/storage/cache        (cached)        coverage: 1.2% of statements in ./...
ok      zotregistry.dev/zot/v2/pkg/storage/common       (cached)        coverage: 3.1% of statements in ./...
?       zotregistry.dev/zot/v2/pkg/storage/constants    [no test files]
ok      zotregistry.dev/zot/v2/pkg/storage/gc   57.674s coverage: 8.9% of statements in ./...
ok      zotregistry.dev/zot/v2/pkg/storage/gcs  (cached)        coverage: 0.7% of statements in ./...
ok      zotregistry.dev/zot/v2/pkg/storage/imagestore   (cached)        coverage: 0.8% of statements in ./...
ok      zotregistry.dev/zot/v2/pkg/storage/local        59.307s coverage: 7.0% of statements in ./...
ok      zotregistry.dev/zot/v2/pkg/storage/s3   (cached)        coverage: 2.7% of statements in ./...
?       zotregistry.dev/zot/v2/pkg/storage/types        [no test files]
ok      zotregistry.dev/zot/v2/pkg/test/auth    (cached)        coverage: 0.4% of statements in ./...
        zotregistry.dev/zot/v2/pkg/test/azurite         coverage: 0.0% of statements
ok      zotregistry.dev/zot/v2/pkg/test/common  2.871s  coverage: 4.2% of statements in ./...
ok      zotregistry.dev/zot/v2/pkg/test/image-utils     (cached)        coverage: 9.1% of statements in ./...
        zotregistry.dev/zot/v2/pkg/test/inject          coverage: 0.0% of statements
        zotregistry.dev/zot/v2/pkg/test/mocks           coverage: 0.0% of statements
ok      zotregistry.dev/zot/v2/pkg/test/oci-utils       (cached)        coverage: 10.8% of statements in ./...
ok      zotregistry.dev/zot/v2/pkg/test/signature       (cached)        coverage: 6.1% of statements in ./...
ok      zotregistry.dev/zot/v2/pkg/test/skip    (cached)        coverage: 0.0% of statements in ./...
ok      zotregistry.dev/zot/v2/pkg/test/tls     (cached)        coverage: 0.6% of statements in ./...
ok      zotregistry.dev/zot/v2/swagger  (cached)        coverage: 0.0% of statements in ./...
...

$ go test ./pkg/storage/gc/... -run "TestGarbageCollectWithMockedImageStore" -v
...
188 total assertions
--- PASS: TestGarbageCollectWithMockedImageStore (0.02s)
PASS
ok      zotregistry.dev/zot/v2/pkg/storage/gc   0.042s

$ go test ./pkg/storage/local/... -run "TestGarbageCollectErrors" -v
...
104 total assertions
--- PASS: TestGarbageCollectErrors (2.10s)
PASS
ok      zotregistry.dev/zot/v2/pkg/storage/local        2.137s
```
**Automation added to e2e**:

No. The fix should be fully covered by the unit tests listed above.

No. This only changes in-process GC error handling.

Yes — GC behavior changes for repos with dangling index entries.

```release-note
Garbage collection no longer aborts an entire repository's cleanup when `index.json` references a blob that is missing from storage. Such entries are now treated as GC-eligible and removed, while transient stat errors still fail closed as before to avoid risking deletion of live data.
```

Assisted-by opencode (claude-sonnet)

### DIFF
--- a/pkg/storage/gc/gc.go
+++ b/pkg/storage/gc/gc.go
@@ -1099,10 +1099,20 @@ func isBlobOlderThan(imgStore types.ImageStore, repo string,
 ) (bool, error) {
 	_, _, modtime, err := imgStore.StatBlob(repo, digest)
 	if err != nil {
-		// Fail closed: ImageStore.StatBlob maps any underlying Stat failure (including
-		// transient S3/network errors) to ErrBlobNotFound, so treating "missing" as
-		// GC-eligible would risk deleting live index rows during storage blips.
-		// Stale prune can still remove truly absent blobs later when CleanRepo succeeds.
+		var pathNotFoundErr driver.PathNotFoundError
+		if errors.As(err, &pathNotFoundErr) {
+			// Blob is missing from storage but referenced in index.json - treat as
+			// GC-eligible so the stale index entry gets removed.
+			log.Warn().Err(err).Str("module", "gc").Str("repository", repo).Str("digest", digest.String()).
+				Msg("blob missing from storage, treating as GC-eligible to clean up stale index entry")
+
+			return true, nil
+		}
+
+		// Fail closed for other errors (transient S3/network issues): we can't
+		// distinguish them from permanently missing blobs, so don't risk deleting
+		// live index rows during storage blips. Stale prune can still remove truly
+		// absent blobs later when CleanRepo succeeds.
 		log.Error().Err(err).Str("module", "gc").Str("repository", repo).Str("digest", digest.String()).
 			Msg("failed to stat blob")
 
--- a/pkg/storage/gc/gc_internal_test.go
+++ b/pkg/storage/gc/gc_internal_test.go
@@ -1565,20 +1565,31 @@ func TestGarbageCollectWithMockedImageStore(t *testing.T) {
 			So(hasTag, ShouldBeFalse)
 		})
 
-		Convey("removeReferrersWithMissingSubject aborts when StatBlob reports missing", func() {
-			// Match main: StatBlob errors (including ErrBlobNotFound from ImageStore's
-			// error collapsing) fail closed in isBlobOlderThan.
+		Convey("removeReferrersWithMissingSubject treats missing referrer blob as GC-eligible", func() {
+			// When a referrer's blob is missing from storage, isBlobOlderThan returns (true, nil)
+			// so the stale referrer entry gets removed.
 			missingSubject := godigest.FromString("missing-subject")
 			cosignTag := "sha256-" + missingSubject.Encoded() + ".sig"
-			missingDigest := godigest.FromString("missing-cosign-blob")
+			referrerDigest := godigest.FromString("referrer-blob")
+
+			cosignManifest := ispec.Manifest{
+				MediaType: ispec.MediaTypeImageManifest,
+				Subject: &ispec.Descriptor{
+					MediaType: ispec.MediaTypeImageManifest,
+					Digest:    missingSubject,
+					Size:      100,
+				},
+				ArtifactType: zcommon.ArtifactTypeCosign,
+			}
+			cosignManifestBlob, _ := json.Marshal(cosignManifest)
 
 			parentIndex := ispec.Index{
 				MediaType: ispec.MediaTypeImageIndex,
 				Manifests: []ispec.Descriptor{
 					{
 						MediaType: ispec.MediaTypeImageManifest,
-						Digest:    missingDigest,
-						Size:      10,
+						Digest:    referrerDigest,
+						Size:      int64(len(cosignManifestBlob)),
 						Annotations: map[string]string{
 							ispec.AnnotationRefName: cosignTag,
 						},
@@ -1588,10 +1599,16 @@ func TestGarbageCollectWithMockedImageStore(t *testing.T) {
 
 			imgStore := mocks.MockedImageStore{
 				GetBlobContentFn: func(repo string, digest godigest.Digest) ([]byte, error) {
+					if digest == referrerDigest {
+						return cosignManifestBlob, nil
+					}
 					return nil, driver.PathNotFoundError{Path: digest.String(), DriverName: "local"}
 				},
 				StatBlobFn: func(repo string, digest godigest.Digest) (bool, int64, time.Time, error) {
-					return false, -1, time.Time{}, driver.PathNotFoundError{Path: digest.String(), DriverName: "local"}
+					if digest == referrerDigest {
+						return false, -1, time.Time{}, driver.PathNotFoundError{Path: digest.String(), DriverName: "local"}
+					}
+					return true, 100, time.Now(), nil
 				},
 			}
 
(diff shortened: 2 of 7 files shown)


andaaron:
Thanks for the PR — the abort-before-stale-prune path for [#4271](https://github.com/project-zot/zot/issues/4271) is a known gap, but I don't think we should change GC eligibility on `StatBlob` failure yet.

On `main`, `isBlobOlderThan` (`pkg/storage/gc/gc.go`) is intentionally fail-closed:

```go
// Fail closed: ImageStore.StatBlob maps any underlying Stat failure (including
// transient S3/network errors) to ErrBlobNotFound, so treating "missing" as
// GC-eligible would risk deleting live index rows during storage blips.
// Stale prune can still remove truly absent blobs later when CleanRepo succeeds.
```

Treating `PathNotFound` as GC-eligible assumes the driver error means “permanently gone.” On object storage that is not always true — the same `PathNotFound` surface can appear for empty prefixes / eventual-consistency blips, so we can delete live `index.json` rows during a storage blip.

Before we change the expected behavior for [#4271](https://github.com/project-zot/zot/issues/4271), we need an independent PR that:

1. Reviews what each storage driver (`local` / S3 / GCS / Azure) actually returns from `Stat` / `Walk` / `List` for missing vs transient vs empty-prefix cases.
2. Classifies those errors in a way callers can reliably distinguish (without collapsing everything to bare `ErrBlobNotFound`).
3. Documents which errors are safe to soft-skip / prune vs fail closed.

Once that classification exists, we can revisit how GC should handle dangling `index.json` references (soft-continue + stale prune, prune-first, or a carefully scoped GC-eligible path) without guessing from driver errors alone.

Until then, we can't merge this PR, as there is a risk of data loss if storage is temporarily unavailable.

andaaron:
Follow-up after more GC work landed on `main`.

Thanks again for the PR. The imagestore idea of preserving the underlying driver error under `ErrBlobNotFound` is still useful, but I still don't think we should merge the GC behavior change here: treating `PathNotFound` as GC-eligible in `isBlobOlderThan` (delete the index row because Stat said missing). That assumes the driver error means "permanently gone," which is not reliably true on object storage (eventual-consistency / empty-prefix surfaces). Fail-closed on Stat remains intentional.

This PR is no longer needed for [#4271](https://github.com/project-zot/zot/issues/4271)
On current `main`, the abort-before-cleanup path is already handled without flipping eligibility:

- [#4180](https://github.com/project-zot/zot/pull/4180) — inventory-based `removeStaleManifestEntries`
- [#4410](https://github.com/project-zot/zot/pull/4410) — soft-continue on age-check / orphan `StatBlob` failure so `CleanRepo` reaches that prune

`isBlobOlderThan` stays fail-closed; callers skip the row and continue.

Will close this PR as superseded for the #4271 symptom.

We will open a separate PR for improved storage/GC error handling (preserving driver causes, classifying missing vs transient vs empty-prefix per backend, and documenting what callers may soft-skip vs fail closed). Any further work on that front should land there rather than by flipping GC eligibility on `PathNotFound` in this change set.

PR #4375, opened 2026-08-30
fix(scanner): synchronize trivy scans

vrajashkr:
**What type of PR is this?**
bug

- Multiple callers could trigger multiple scans of the same image.

- Synchronizes trivy scans to prevent multiple simultaneous requests for an uncached digest's CVEList from triggering multiple re-scans of the same digest. Only 1 caller runs the scan while others wait for the result.

Unit Tests

**Automation added to e2e**:
N/A

No

No

### DIFF
--- a/pkg/extensions/search/cve/trivy/scanner.go
+++ b/pkg/extensions/search/cve/trivy/scanner.go
@@ -32,6 +32,7 @@ import (
 	godigest "github.com/opencontainers/go-digest"
 	"github.com/opencontainers/image-spec/specs-go"
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"golang.org/x/sync/singleflight"
 	_ "modernc.org/sqlite"
 
 	zerr "zotregistry.dev/zot/v2/errors"
@@ -178,6 +179,10 @@ type Scanner struct {
 	javaDBRepositoryRef name.Reference
 	vulnSeveritySources []dbTypes.SourceID
 	sbomOptions         sbomOptions
+
+	// SingleFlightGroup is used to prevent multiple requests
+	// running a trivy scan for the same digest repeatedly.
+	scanSingleFlightGroup *singleflight.Group
 }
 
 type sbomOptions struct {
@@ -279,16 +284,17 @@ func NewScanner(storeController storage.StoreController,
 	cveController.SubCveConfig = subCveConfig
 
 	return &Scanner{
-		log:                 log,
-		metaDB:              metaDB,
-		cveController:       cveController,
-		storeController:     storeController,
-		dbLock:              &sync.Mutex{},
-		cache:               cvecache.NewCveCache(cacheSize, log),
-		dbRepositoryRef:     dbRepositoryRef,
-		javaDBRepositoryRef: javaDBRepositoryRef,
-		vulnSeveritySources: sevSources,
-		sbomOptions:         sbomOpts,
+		log:                   log,
+		metaDB:                metaDB,
+		cveController:         cveController,
+		storeController:       storeController,
+		dbLock:                &sync.Mutex{},
+		cache:                 cvecache.NewCveCache(cacheSize, log),
+		dbRepositoryRef:       dbRepositoryRef,
+		javaDBRepositoryRef:   javaDBRepositoryRef,
+		vulnSeveritySources:   sevSources,
+		sbomOptions:           sbomOpts,
+		scanSingleFlightGroup: &singleflight.Group{},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/extensions/search/cve/trivy/scanner.go
```diff
@@ -32,6 +32,7 @@ import (
 	godigest "github.com/opencontainers/go-digest"
 	"github.com/opencontainers/image-spec/specs-go"
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"golang.org/x/sync/singleflight"
```
andaaron, line 35: go.mod / lint — CI lint fails on “Check go.mod and go.sum are up to date”. The PR imports golang.org/x/sync/singleflight (pkg/extensions/search/cve/trivy/scanner.go:35, :157, :266) but does not touch go.mod. On main, golang.org/x/sync is // indirect; a direct import needs go mod tidy to promote it.

pkg/extensions/search/cve/trivy/scanner.go
```diff
@@ -641,98 +647,120 @@ func (scanner Scanner) scanManifest(ctx context.Context, repo, digest string) (m
 		return cachedMap, true, nil
 	}
 
-	cveidMap := map[string]zcommon.CVE{}
-	image := repo + "@" + digest
+	// prevent multiple requests running trivy multiple times
+	result, err, _ := scanner.scanSingleFlightGroup.Do(digest, func() (any, error) {
```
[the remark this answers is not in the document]
andaaron: We need to solve this.

pkg/extensions/search/cve/trivy/scanner.go
```diff
@@ -641,98 +647,120 @@ func (scanner Scanner) scanManifest(ctx context.Context, repo, digest string) (m
...
-	report, sbom, err := scanner.runTrivy(ctx, opts)
-	scanner.dbLock.Unlock()
-	if sbom != nil && sbom.filePath != "" {
-		defer os.Remove(sbom.filePath)
-	}
+		// Double check the cache under flight group lock to prevent a race
+		// where a caller just sees a cache miss before the cache is updated.
+		// In this case, the caller initiates a fresh flight even though a scan just finished up.
+		// This avoids a double scan of the image.
+		if cachedMap := scanner.cache.Get(digest); cachedMap != nil {
```
andaaron, line 688: Double-check cache hit still returns wasCached=false

pkg/extensions/search/cve/trivy/scanner.go
vrajashkr: Whoops, missed this. This change is already in mainline now as the other PR with singleflight has been merged.

pkg/extensions/search/cve/trivy/scanner.go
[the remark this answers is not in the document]
vrajashkr: I've used a similar strategy as the other PR where the context used for scan is derived from parent with cancellation removed.

pkg/extensions/search/cve/trivy/scanner.go
vrajashkr: This should be fixed now. New type `cacheableScanResult` informs the outer code whether this was cached from the double-check.

pkg/extensions/search/cve/trivy/scanner.go
```diff
@@ -949,38 +976,65 @@ func getNVDReference(references []string) (string, bool) {
...
+				}
 
-			maps.Copy(indexCveIDMap, manifestCveIDMap)
+				maps.Copy(indexCveIDMap, manifestCveIDMap)
+			}
 		}
+
+		scanner.cache.Add(digest, indexCveIDMap)
+
+		return indexCveIDMap, nil
```
andaaron, line 1028: Here we return a map, while on the successful cache hit path we return. cacheableScanResult.
Caller expects a cacheableScanResult. not a map

pkg/extensions/search/cve/trivy/scanner.go
```diff
@@ -673,98 +679,119 @@ func (scanner Scanner) scanManifest(ctx context.Context, repo, digest string) (m
...
+		scanner.dbLock.Unlock()
+		if sbom != nil && sbom.filePath != "" {
+			defer os.Remove(sbom.filePath)
+		}
 
-	if err != nil { //nolint: wsl
-		return cveidMap, false, err
-	}
+		if err != nil {
+			return cveidMap, err
```
andaaron, line 706: Here we return a map, while on the successful cache hit path we return. cacheableScanResult.
Caller expects a cacheableScanResult. not a map

pkg/extensions/search/cve/trivy/scanner.go
andaaron: We need test to catch these issues.

Maybe we can extract the logic called in scanSingleFlightGroup.Do() in separate named functions, where can type check.

andaaron:
Superseded by https://github.com/project-zot/zot/pull/4440

PR #4379, opened 2026-08-31
fix(sync): dedupe on-demand sync with singleflight

### LINKED ISSUE #4345: On-demand sync: requestStore entry is poisoned forever if the original caller's request is abandoned before syncImage completes
### zot version

v2.1.17 (reproduced), confirmed still present in v2.1.20 by source inspection

### Describe the bug

`BaseOnDemand.SyncImage` (`pkg/extensions/sync/on_demand.go`) stores a per-`(repo,reference)` unbuffered channel in a `sync.Map` (`requestStore`) and spawns `syncImage` to do the work, ending with a blocking send: `syncResult <- err`. The only reader is the original caller's `err := <-syncResult` inside `SyncImage` itself.

If the original HTTP request that triggered the on-demand sync is abandoned — e.g. the client (kubelet, containerd) hits its own timeout and disconnects — while `syncImage` is still running or about to send, the goroutine performing `syncResult <- err` blocks forever: nothing is left to receive on that channel. Because `SyncImage`'s `defer onDemand.requestStore.Delete(req)` only runs once its own `<-syncResult` read returns, and that read's owning goroutine has itself already returned when the client disconnected upstream, the `requestStore` entry for that exact `(repo, reference)` is **never removed**.

Every subsequent request for that same image digest then hits `LoadOrStore` -> `loaded == true`, gets the same stale channel, and blocks forever waiting on it too — even though the underlying sync succeeded and the image is fully present in local storage. The only way out we found is clearing Zot's in-memory state (e.g. restarting the pod), since `requestStore` is not persisted.

This is a **different bug** from #3344 (fixed in #3348, shipped v2.1.8). That fix made sure a channel gets `close()`d by whichever goroutine receives from it, so a set of *concurrent* waiters don't hang relative to each other. It does not address the case here: there is no other concurrent waiter, no receiver ever existed for `syncResult`, and the send itself never returns.

We noticed `context.WithoutCancel` was later added (#3537) specifically so the sync's own upstream work isn't aborted by client disconnection — but that change only protects the *sync* from being interrupted; it does not protect the *result-delivery channel* from having no receiver.

### To reproduce

Not consistently reproducible on demand, but the shape that triggered it for us: a scale-out cluster deployment (3 members, Zot's native `cluster` config, one member proxies to the repo's owning member for a request it doesn't itself own) with `onDemand: true` pull-through sync from a remote registry, and a client (Kubernetes kubelet doing an image pull) whose own request timeout is shorter than a worst-case on-demand sync (e.g. `http.readTimeout`/`writeTimeout` of 300s used by the cluster proxy hop). When the proxied request is slow enough to hit that 300s ceiling and gets canceled, the owning member's in-flight `syncImage` for that digest is left running with its result channel permanently un-received. All later pulls of that exact digest then hang too.

Logs from the owning cluster member (`zot-1`), showing the sync actually succeeding well before a later request still hangs on it:

```
{"time":"...T09:34:40.043Z","message":"trying to get updated image by syncing on demand","repo":"datarecs/platform-agent","reference":"sha256:cea4037..."}
{"time":"...T09:34:40.043Z","message":"sync: syncing image","remote":"ghcr.io","repo":"datarecs/platform-agent","reference":"sha256:cea4037..."}
{"time":"...T09:34:44.444Z","message":"pushing synced local image to local registry","repository":"datarecs/platform-agent"}
...
{"time":"...T09:39:58.619Z","message":"trying to get updated image by syncing on demand","repo":"datarecs/platform-agent","reference":"sha256:cea4037..."}
{"time":"...T09:39:58.619Z","message":"image already demanded, waiting on channel","repo":"datarecs/platform-agent","reference":"sha256:cea4037..."}
```

The second request (5+ minutes after the first had already committed) still hangs waiting on the first's channel, and is eventually killed by the client's own timeout — reproduced twice independently, on two different tenant pulls of the same image digest, hours apart, both self-resolving only after enough time/restarts passed.

### Expected behavior

An abandoned/canceled original caller should not permanently poison the `requestStore` entry for that `(repo, reference)`. Once `syncImage`'s work is done (success or error), the entry should be removed and the result discarded if nobody is listening, rather than blocking forever on an unbuffered send. A buffered channel (size 1) for `syncResult`, or moving the `requestStore.Delete` into `syncImage`'s own `defer` (rather than relying on the original `SyncImage` caller's line ordering) would fix it.

### Additional context

We believe `manifestCheckInterval` (#4328, not yet in a tagged release as of v2.1.20) would reduce exposure to this bug by avoiding redundant on-demand checks for an already-synced manifest, but it doesn't address the underlying unbuffered-channel-with-abandoned-receiver hazard directly — a future request for a *different*, not-yet-synced digest could still trigger it.

Happy to provide the full pod logs / goroutine dump if useful — we didn't capture a `pprof` goroutine dump at the time (would do so if it recurs, per the profiling approach mentioned in #3344).

andaaron:
Replace hand-rolled sync.Map + channel coordination in BaseOnDemand with
golang.org/x/sync/singleflight for SyncImage and SyncReferrers.

Fixes #4345. When the original HTTP caller disconnected before reading the
result channel, the worker could block on send and leave a stale
requestStore entry, causing later pulls for the same digest to hang on
"already demanded". singleflight dedupes in-flight work by key, ties
cleanup to sync completion (not channel receive), and returns the same
error to all waiters.

Expose on-demand sync through ext.SyncOnDemand (sync.OnDemand in
common.go) instead of *BaseOnDemand. EnableSyncExtension returns the
interface so failure paths yield a true nil; guard assignment and
isSyncOnDemandEnabled with != nil instead of fmt.Sprintf.

requestStore is kept only for background-retry dedup. Add tests for
concurrent SyncImage/SyncReferrers performing a single upstream sync.

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -49,7 +49,7 @@ type Controller struct {
 	Metrics         monitoring.MetricServer
 	EventRecorder   events.Recorder
 	CveScanner      ext.CveScanner
-	SyncOnDemand    SyncOnDemand
+	SyncOnDemand    ext.SyncOnDemand
 	RelyingParties  map[string]rp.RelyingParty
 	CookieStore     *CookieStore
 	HTPasswd        *HTPasswd
@@ -594,8 +594,9 @@ func (c *Controller) StartBackgroundTasks() {
 		c.Log.Error().Err(err).Msg("failed to start sync extension")
 	}
 
-	// Only set SyncOnDemand if sync is actually enabled
-	if extensionsConfig.IsSyncEnabled() {
+	// EnableSyncExtension returns ext.SyncOnDemand (an interface), so failure/disabled
+	// paths yield a true nil interface — safe for isSyncOnDemandEnabled's != nil check.
+	if extensionsConfig.IsSyncEnabled() && syncOnDemand != nil {
 		c.SyncOnDemand = syncOnDemand
 	}
 
@@ -641,10 +642,4 @@ func RunGCTasks(conf *config.Config, storeController storage.StoreController, me
 			}
 		}
 	}
-}
-
-type SyncOnDemand interface {
-	SyncImage(ctx context.Context, repo, reference string) error
-	SyncReferrers(ctx context.Context, repo string, subjectDigestStr string, referenceTypes []string) error
-	ShouldCheckUpstreamManifest(repo, reference string) bool
 }
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -3015,12 +3015,8 @@ func isSyncOnDemandEnabled(ctlr *Controller) bool {
 	}
 
 	extensionsConfig := ctlr.Config.CopyExtensionsConfig()
-	if extensionsConfig.IsSyncEnabled() &&
-		fmt.Sprintf("%v", ctlr.SyncOnDemand) != fmt.Sprintf("%v", nil) {
-		return true
-	}
 
-	return false
+	return extensionsConfig.IsSyncEnabled() && ctlr.SyncOnDemand != nil
 }
 
 func eventContextFromRequest(r *http.Request) *events.EventContext {
--- a/pkg/extensions/sync/on_demand.go
+++ b/pkg/extensions/sync/on_demand.go
@@ -8,6 +8,8 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/sync/singleflight"
+
 	zerr "zotregistry.dev/zot/v2/errors"
 	"zotregistry.dev/zot/v2/pkg/common"
 	"zotregistry.dev/zot/v2/pkg/log"
@@ -22,16 +24,19 @@ type request struct {
 }
 
 /*
-BaseOnDemand tracks requests that can be an image/signature/sbom.
+BaseOnDemand tracks on-demand image/referrer sync requests.
 
-It keeps track of all parallel requests, if two requests of same image/signature/sbom comes at the same time,
-process just the first one, also keep track of all background retrying routines.
+Concurrent SyncImage/SyncReferrers calls for the same key are deduplicated with
+singleflight (one upstream sync, shared result). requestStore tracks in-flight
+background retry goroutines so at most one retry runs per service/key.
 */
 type BaseOnDemand struct {
 	services []Service
-	// map[request]chan err
-	requestStore *sync.Map
-	log          log.Logger
+	// background retry dedup: map[request]struct{}
+	requestStore   *sync.Map
+	imageFlight    singleflight.Group
+	referrerFlight singleflight.Group
+	log            log.Logger
 }
 
 func NewOnDemand(log log.Logger) *BaseOnDemand {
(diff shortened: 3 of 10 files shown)


pkg/api/routes.go
```diff
@@ -2975,12 +2975,8 @@ func isSyncOnDemandEnabled(ctlr *Controller) bool {
 	}
 
 	extensionsConfig := ctlr.Config.CopyExtensionsConfig()
-	if extensionsConfig.IsSyncEnabled() &&
-		fmt.Sprintf("%v", ctlr.SyncOnDemand) != fmt.Sprintf("%v", nil) {
-		return true
-	}
 
-	return false
+	return extensionsConfig.IsSyncEnabled() && ctlr.SyncOnDemand != nil
```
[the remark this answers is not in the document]
andaaron: This is theoretical today — no path returns (*BaseOnDemand)(nil), nil. Belt-and-suspenders: if syncOnDemand != nil { c.SyncOnDemand = syncOnDemand }. Not a regression.

Moreover, if we use the suggested code, it races in the tests.

pkg/api/controller.go
```diff
@@ -594,8 +594,9 @@ func (c *Controller) StartBackgroundTasks() {
 		c.Log.Error().Err(err).Msg("failed to start sync extension")
 	}
 
-	// Only set SyncOnDemand if sync is actually enabled
-	if extensionsConfig.IsSyncEnabled() {
+	// EnableSyncExtension returns ext.SyncOnDemand (an interface), so failure/disabled
+	// paths yield a true nil interface — safe for isSyncOnDemandEnabled's != nil check.
```
vrajashkr, line 598: I didn't really understand this comment. I'm wondering whether we need it.

pkg/extensions/sync/on_demand.go
```diff
@@ -56,71 +61,57 @@ func (onDemand *BaseOnDemand) ShouldCheckUpstreamManifest(repo, reference string
 	return true
 }
 
-func (onDemand *BaseOnDemand) SyncImage(ctx context.Context, repo, reference string) error {
-	req := request{
-		repo:      repo,
-		reference: reference,
-	}
+func onDemandKey(repo, reference string) string {
+	return repo + "\x00" + reference
```
vrajashkr, line 65: Is there a reason to specifically use `"\x00"` as the separator. I'm wondering if an ordinary hyphen or underscore might work here.

vrajashkr (commented):
Looks good overall. Left a couple of small questions.

pkg/api/controller.go
andaaron: The comment is trying to explain why EnableSyncExtension now returns ext.SyncOnDemand (an interface) instead of *sync.BaseOnDemand.

On main, isSyncOnDemandEnabled used fmt.Sprintf("%v", ctlr.SyncOnDemand) != fmt.Sprintf("%v", nil) to detect a set handler. With singleflight, that became unsafe: formatting walked into *BaseOnDemand while another request mutated imageFlight/referrerFlight (data race in TestOnDemandPullsReferrersOnce).

Replacing that with ctlr.SyncOnDemand != nil is only reliable if the interface is a true nil on failure/disabled paths. Returning *BaseOnDemand can create a typed nil ((*BaseOnDemand)(nil) stored in an interface), which is non-nil and would still enter the sync path. Returning ext.SyncOnDemand keeps disabled/error returns as a real nil interface, so the != nil check and this assignment guard are safe.

pkg/extensions/sync/on_demand.go
andaaron: \x00 isn’t allowed in OCI repository names or references, so it can’t appear inside either half of the key. A hyphen/underscore can appear in both repo and reference, which would allow collisions (e.g. "a-b"+"c" vs "a"+"b-c"). NUL avoids that without needing escaping.

vrajashkr (approved):
LGTM

PR #4380, opened 2026-08-31
fix(api): enforce dedupe mount auth for bearer and unify via UserAc

andaaron:
Traditional bearer validated only the target-repo JWT scope and never populated UserAc, so canMount() saw empty permissions as allow-all. Map token access claims to UserAc glob patterns at authn time, wire config authz through a request-scoped PermissionEvaluator (ac.can), and use CanOnResource in canMount so handlers need not branch on auth mode.

- Add BearerAuthorizer.Authenticate and UserAccessControlFromBearerAccess
- Attach permission evaluator in BaseAuthzHandler; add CanOnResource
- Gate mount/check paths via shouldCheckMountSourceAccess
- Add bearer, CEL, and canMount unit/integration tests

### DIFF
--- a/pkg/api/authz.go
+++ b/pkg/api/authz.go
@@ -253,6 +253,16 @@ func (ac *AccessController) updateUserAccessControl(httpReq *http.Request, userA
 		ac.getGlobPatterns(mkER(constants.DetectManifestCollisionPermission)))
 }
 
+// attachPermissionEvaluator wires config-based policy evaluation into userAc so
+// handlers can use CanOnResource without constructing AccessController.
+func (ac *AccessController) attachPermissionEvaluator(httpReq *http.Request, userAc *reqCtx.UserAccessControl) {
+	userAc.SetPermissionEvaluator(func(action, repository, reference string) bool {
+		ok, _ := ac.can(httpReq, userAc, action, repository, reference)
+
+		return ok
+	})
+}
+
 // getAuthnMiddlewareContext builds ac context(allowed to read repos and if user is admin) and returns it.
 func (ac *AccessController) getAuthnMiddlewareContext(authnType string, request *http.Request) context.Context {
 	amwCtx := reqCtx.AuthnMiddlewareContext{
@@ -644,6 +654,7 @@ func BaseAuthzHandler(ctlr *Controller) mux.MiddlewareFunc {
 			}
 
 			aCtlr.updateUserAccessControl(request, userAc)
+			aCtlr.attachPermissionEvaluator(request, userAc)
 			userAc.SaveOnRequest(request)
 
 			next.ServeHTTP(response, request) //nolint:contextcheck
--- a/pkg/api/authz_internal_test.go
+++ b/pkg/api/authz_internal_test.go
@@ -651,6 +651,47 @@ func TestSetGlobPatternsOrderIndependence(t *testing.T) {
 	assert.True(t, uac2.IsAdmin())
 }
 
+func TestCanOnResourceUsesPermissionEvaluator(t *testing.T) {
+	t.Parallel()
+
+	cfg := &config.AccessControlConfig{
+		Repositories: config.Repositories{
+			"**": config.PolicyGroup{
+				Policies: []config.Policy{
+					{
+						Users:   []string{"limited"},
+						Actions: []string{constants.ReadPermission, constants.CreatePermission},
+						Conditions: []config.Condition{
+							{
+								Expression: `req.action != "read" || !req.repository.startsWith("private/")`,
+								Message:    "read under private/ not permitted",
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	programs, err := CompileAccessControl(cfg)
+	assert.NoError(t, err)
+	cfg.StoreCompiledConditions(programs)
+
+	uac := reqCtx.NewUserAccessControl()
+	uac.SetUsername("limited")
+	uac.SetIsAdmin(false)
+	uac.SetGlobPatterns(constants.ReadPermission, map[string]bool{"**": true})
+
+	// Optimistic glob patterns allow read on private/src.
+	assert.True(t, uac.Can(constants.ReadPermission, "private/src"))
+
+	ac := &AccessController{Config: cfg, Log: log.NewLogger("debug", "")}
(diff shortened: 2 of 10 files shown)


rchincha:
Is there a bearer token test for `pull`-only scope + hydrateOnBlobRead working together? now that we have hydrateOnBlobRead merged.

andaaron:
> Is there a bearer token test for `pull`-only scope + hydrateOnBlobRead working together? now that we have hydrateOnBlobRead merged.
> 
> 

Well pull only should not hidrate unless you have read permission on at least one other repo containing the blob, and write permission (in addition to read) on the repo you are pulling from.

rchincha:
> > Is there a bearer token test for `pull`-only scope + hydrateOnBlobRead working together? now that we have hydrateOnBlobRead merged.
> 
> Well pull only should not hidrate unless you have read permission on at least one other repo containing the blob, and write permission (in addition to read) on the repo you are pulling from.

So what would be the token scope then?

andaaron:
> > > Is there a bearer token test for `pull`-only scope + hydrateOnBlobRead working together? now that we have hydrateOnBlobRead merged.
> > 
> > Well pull only should not hidrate unless you have read permission on at least one other repo containing the blob, and write permission (in addition to read) on the repo you are pulling from.
> 
> So what would be the token scope then?

Pull from both repos and push on the one you want to pull from

rchincha:
> > > > Is there a bearer token test for `pull`-only scope + hydrateOnBlobRead working together? now that we have hydrateOnBlobRead merged.
> > > 
> > > 
> > > Well pull only should not hidrate unless you have read permission on at least one other repo containing the blob, and write permission (in addition to read) on the repo you are pulling from.
> > 
> > 
> > So what would be the token scope then?
> 
> Pull from both repos and push on the one you want to pull from

Do we have a test for this?

andaaron:
> > > > > Is there a bearer token test for `pull`-only scope + hydrateOnBlobRead working together? now that we have hydrateOnBlobRead merged.
> > > > 
> > > > 
> > > > Well pull only should not hidrate unless you have read permission on at least one other repo containing the blob, and write permission (in addition to read) on the repo you are pulling from.
> > > 
> > > 
> > > So what would be the token scope then?
> > 
> > 
> > Pull from both repos and push on the one you want to pull from
> 
> Do we have a test for this?

Yes — added in the latest push.

## Integration test: `TestBearerAuthorizationHydrateOnRead`

`pkg/api/controller_test.go` — bearer auth, dedupe, and `HydrateBlobOnRead` enabled. A blob is uploaded to `bearer/src`, then hydrate is attempted on `bearer/dest`.

### Denied (no hydrate)

| Case | Token scope | Expected |
|------|-------------|----------|
| Challenge-scoped dest pull only | `repository:bearer/dest:pull` (from WWW-Authenticate on HEAD) | HEAD and ranged GET → 404; blob not materialized on disk |
| Dest pull + push, no src | `repository:bearer/dest:pull,push` | HEAD and ranged GET → 404 |

### Allowed (hydrate + mount)

| Case | Token scope | Expected |
|------|-------------|----------|
| Hydrate on read | `repository:bearer/src:pull` + `repository:bearer/dest:pull,push` | HEAD → 200, ranged GET → 206; blob present under dest repo |
| Dedupe mount POST | `repository:bearer/src:pull` + `repository:bearer/dest:push` | POST `?mount=` → 201 |

So the allowed scope is what we discussed: **pull on the source repo that already holds the digest, and pull + push on the destination repo** (push maps to create, which `canMount` requires on dest; pull on dest is needed for the HEAD/range read itself).

`TestBearerAuthorizationMountBlob` in the same file still covers the mount path without hydrate; the new test covers hydrate-on-read specifically.

## Unit tests

`pkg/api/routes_internal_test.go` — `TestUserMayMountBlobGate` adds:

- dest **pull only** → remount denied
- dest **pull + push**, no src pull → remount denied
- src pull + dest push → remount allowed

Same gate as integration: `userMayMountBlob` → `shouldCheckMountSourceAccess` → `canMount` → `CanOnResource`.

---

*Copy everything below the line for the GitHub comment.*

---

Yes — added `TestBearerAuthorizationHydrateOnRead` in `pkg/api/controller_test.go` (latest push).

It runs with traditional bearer auth, dedupe, and `HydrateBlobOnRead` enabled. A blob is seeded in `bearer/src`, then we try to read it from `bearer/dest`.

**Denied (no hydrate):**

- Challenge-scoped **dest pull only** (`repository:bearer/dest:pull` from the HEAD challenge) → HEAD and ranged GET return 404; nothing materialized under dest.
- **Dest pull + push, no src pull** → same (404 on HEAD/range).

**Allowed:**

- Token with `repository:bearer/src:pull` and `repository:bearer/dest:pull,push` → HEAD 200, ranged GET 206, blob on disk under dest.
- Token with `repository:bearer/src:pull` and `repository:bearer/dest:push` → dedupe mount POST returns 201.

So the scope is: pull on the repo that already has the blob, and pull + push on the repo you are reading into (push → create on dest for `canMount`; pull on dest for the read request).

Unit coverage for the same gate is in `TestUserMayMountBlobGate` in `routes_internal_test.go` (dest pull only; dest pull+push without src; src pull + dest push).

rchincha:
We now have two (previously only one) return status codes based on the following. Maybe we should add a bats ecosystem tools test to cover this also if there isn't one already.

201 Created (Mount Succeeded)

202 Accepted (Mount Failed, New Upload Session Started)

andaaron:
> We now have two (previously only one) return status codes based on the following. Maybe we should add a bats ecosystem tools test to cover this also if there isn't one already.
> 
> 201 Created (Mount Succeeded)
> 
> 202 Accepted (Mount Failed, New Upload Session Started)

The 201/202 split isn’t new — it’s spec behavior and was already implemented and tested for config-based authz (TestAuthorizationMountBlob). What’s new is that bearer mount requests now participate in that same gate, so dest-only tokens can hit the existing 202 fallback instead of incorrectly getting 201.

WTR bats tests: we don't have a clean way to test that most of it is internal to docker or other tools which don't report status codes of individual requests. Also, a bearer system would need specific token 
I can experiment with bats tests for blob mount in a separate PR for basic auth.

rchincha (approved):

PR #4385, opened 2026-09-01
feat(events): configurable CloudEvents source and type namespace

Aluchir:
**What type of PR is this?**

feature

No existing issue.

The CloudEvents source and every event type are compile-time constants, so each delivery announces `zotregistry.dev` with types under `zotregistry.*`. The CloudEvents spec intends `source` to identify the producer, but an operator running zot inside a larger platform cannot present their own identity, and a receiver has no way to tell one deployment, or one tenant, from another: the value is identical for every zot in existence.

This adds two optional fields to the events extension config:

- `source` overrides the CloudEvents source attribute on every event the instance publishes.
- `typePrefix` replaces the leading namespace of each event type, so `image.updated` is published as `<typePrefix>.image.updated`.

Both default to empty, which keeps the built-in values, so upgrading changes nothing until an operator sets them. `NewRecorder` keeps its signature; the override goes through a new `NewRecorderWithIdentity`. A type outside the built-in namespace passes through untouched.

**How to verify it**

Unit tests cover the default fallbacks, the source override, the type renaming, and the passthrough of non-namespaced types. `examples/config-events.json` shows the new fields and passes `zot verify`.

### DIFF
--- a/examples/config-events.json
+++ b/examples/config-events.json
@@ -13,6 +13,8 @@
   "extensions": {
     "events": {
       "enable": true,
+      "source": "registry.example.com/tenant-a",
+      "typePrefix": "com.example.registry",
       "sinks": [{
           "type": "nats",
           "address": "nats://127.0.0.1:4222",
--- a/pkg/cli/server/root.go
+++ b/pkg/cli/server/root.go
@@ -592,6 +592,22 @@ func validateMetricsConfig(cfg *extconf.ExtensionConfig) error {
 	return nil
 }
 
+func validateEventsConfig(cfg *extconf.ExtensionConfig) error {
+	eventsCfg := cfg.GetEventsConfig()
+	if eventsCfg == nil {
+		return nil
+	}
+
+	// The prefix replaces the leading segment of dot-separated event types,
+	// so it must not create empty segments in the rendered type.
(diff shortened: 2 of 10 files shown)


rchincha:
```
// in pkg/cli/server/root.go, alongside validateMetricsConfig
func validateEventsConfig(extensionsConfig *config.ExtensionConfig) error {
	eventsCfg := extensionsConfig.GetEventsConfig()
	if eventsCfg == nil {
		return nil
	}

	if strings.HasSuffix(eventsCfg.TypePrefix, ".") || strings.Contains(eventsCfg.TypePrefix, "..") {
		return fmt.Errorf("%w: events typePrefix must not contain a trailing or double dot", zerr.ErrBadConfig)
	}

	return nil
}
```

^ some additional input validation?

Aluchir:
@rchincha added, thanks. validateEventsConfig sits next to validateMetricsConfig and rejects a typePrefix with a leading, trailing or double dot, since any of those would render event types with empty segments. Wired into validateExtensionsConfig and covered with tests for the three rejected shapes plus a valid one.

rchincha (approved):

andaaron (approved):

PR #4387, opened 2026-09-01
test(blackbox): add mount authorization tests with curl and regctl

andaaron:
Add pushpull_mount.bats to verify dedupe mount behavior under per-user access control: curl asserts 202 when mount is denied and 201 when allowed; regctl blob copy covers denied (202 in trace log, copy fails) and allowed (copy + blob head succeed) paths.

### DIFF
--- a/test/blackbox/helpers_pushpull_mount.bash
+++ b/test/blackbox/helpers_pushpull_mount.bash
@@ -1,0 +1,224 @@
+# Helpers for blob mount authorization blackbox tests (curl + regctl).
+
+load helpers_pushpull_authn
+
+MOUNT_USER1=mountuser1
+MOUNT_PASS1=mountSecret1
+MOUNT_USER2=mountuser2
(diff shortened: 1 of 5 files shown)


test/blackbox/helpers_pushpull_mount.bash
```diff
@@ -0,0 +1,180 @@
...
+    local zot_port=${3}
+    local zot_htpasswd_file=${4}
+    local log_file=${5}
+
+    cat >"${zot_config_file}" <<EOF
+{
+  "distSpecVersion":"1.1.1",
+  "storage":{
+    "dedupe": true,
+    "gc": false,
```
rchincha, line 42: Should we also add tests for hydrateOnBlobRead set?

andaaron: done

rchincha (approved):

PR #4391, opened 2026-09-02
fix(api): don't 500 on GET manifest when download-stats update fails

### LINKED ISSUE #4390: [Bug]: GET manifest returns 500 when the download-stats metaDB update fails (spurious ImagePullBackOff under concurrent pull bursts)
### Description

When zot is used as a pull-through cache with a shared redis metaDB (`storage.remoteCache: true`, `cacheDriver.name: redis`), a `GET /v2/<name>/manifests/<ref>` can return **HTTP 500 even though the manifest was successfully retrieved and verified**.

The root cause is that the download-statistics bookkeeping hook is treated as fatal. In [`pkg/api/routes.go`](https://github.com/project-zot/zot/blob/main/pkg/api/routes.go) (`GetManifest`):

```go
if rh.c.MetaDB != nil {
    err := meta.OnGetManifest(name, reference, mediaType, content, rh.c.StoreController, rh.c.MetaDB, rh.c.Log)
    if err != nil && !errors.Is(err, zerr.ErrImageMetaNotFound) && !errors.Is(err, zerr.ErrRepoMetaNotFound) {
        response.WriteHeader(http.StatusInternalServerError)
        return
    }
}
```

`OnGetManifest` → `MetaDB.UpdateStatsOnDownload` only updates best-effort bookkeeping: it increments `DownloadCount` and sets `LastPullTimestamp`. In the redis metaDB implementation this whole update runs under a **per-repo distributed lock** (`withRSLocks` / redsync). Under a burst of concurrent pulls of the *same* repo — e.g. a large fan-out job whose dozens of worker pods pull the same image at once — the per-repo lock is contended, redsync exhausts its retries, and `UpdateStatsOnDownload` returns a lock error such as:

```
failed to acquire redis lock ... lock already taken, locked nodes: [0]
```

The manifest read is turned into a 500. Since the manifest itself was fine, this manifests as **spurious kubelet `ImagePullBackOff`** that later "self-heals" once the herd subsides.

### Impact

- Correct, already-verified manifests are discarded and returned as 500.
- Under real fan-out workloads (distributed training / batch jobs) this reliably produces `ImagePullBackOff` on many pods at once.
- The failing operation is purely cosmetic bookkeeping — `DownloadCount` feeds search/GraphQL stats, `LastPullTimestamp` feeds retention (`pulledWithin`), and on lock loss there is by definition a concurrent winner that already updated the timestamp.

### Observed

- zot v2.1.16, 3 replicas, shared redis metaDB.
- Exact 1:1 correlation in a 14s window: N × `statusCode:500` ↔ N × `failed to acquire redis lock`, client UA `containerd/...`, all from the same repo being pulled by many pods simultaneously.
- Each affected request hung ~5–8s (redsync retry) before the 500.

### Expected behavior

A failure to update download statistics must **not** fail the client's manifest read. The manifest has already been retrieved and verified before the hook runs; a bookkeeping/lock error should be logged (warn) and the manifest served.

### Suggested fix

Don't treat `OnGetManifest` errors as fatal in `GetManifest`; log and continue. (Happy to open a PR — I have a minimal two-hunk patch that does exactly this and downgrades the log level in `pkg/meta/hooks.go`.)

### Version

- zot `v2.1.16` (also present on `main` as of filing).

yinsenyan:
## What / why

Fixes #4390.

`GetManifest` in `pkg/api/routes.go` treats the download-statistics bookkeeping hook (`meta.OnGetManifest`) as fatal: any error other than the two "not found" sentinels turns into an HTTP 500. But the manifest has **already been retrieved and verified** by the time this hook runs — `OnGetManifest` → `MetaDB.UpdateStatsOnDownload` only increments `DownloadCount` and sets `LastPullTimestamp`.

In the redis metaDB implementation that update runs under a per-repo distributed lock (`withRSLocks` / redsync). Under a burst of concurrent pulls of the same repo (e.g. a large fan-out job whose worker pods all pull the same image at once), the lock is contended, redsync exhausts its retries, and `UpdateStatsOnDownload` returns a lock error. That was being turned into a 500, discarding a perfectly good manifest and causing spurious kubelet `ImagePullBackOff`.

## Change

- `pkg/api/routes.go`: don't fail the client's manifest read on an `OnGetManifest` error. The manifest is served regardless; the bookkeeping failure is logged at warn.
- `pkg/meta/hooks.go`: downgrade the now-non-fatal `UpdateStatsOnDownload` failure log from `Error` to `Warn`.

## Impact / safety

The failing operation is best-effort bookkeeping only:
- `DownloadCount` feeds search/GraphQL stats.
- `LastPullTimestamp` feeds retention (`pulledWithin`); on a lock loss there is by definition a concurrent winner that already updated the timestamp.

No effect on serving correctness, GC, dedupe, sync, or data integrity.

See #4390 for the full root-cause analysis and observed 1:1 correlation between the 500s and the redis lock failures.

### DIFF
--- a/pkg/api/routes.go
+++ b/pkg/api/routes.go
@@ -578,11 +578,15 @@ func (rh *RouteHandler) GetManifest(response http.ResponseWriter, request *http.
 	}
 
 	if rh.c.MetaDB != nil {
-		err := meta.OnGetManifest(name, reference, mediaType, content, rh.c.StoreController, rh.c.MetaDB, rh.c.Log)
-		if err != nil && !errors.Is(err, zerr.ErrImageMetaNotFound) && !errors.Is(err, zerr.ErrRepoMetaNotFound) {
-			response.WriteHeader(http.StatusInternalServerError)
-
-			return
+		// OnGetManifest only updates best-effort download bookkeeping (download count and
+		// last-pull timestamp). The manifest has already been retrieved and verified above, so a
+		// failure here must not fail the client's read. In particular, a burst of concurrent
+		// pulls of the same repo (e.g. a large fan-out job) contends the per-repo metaDB lock and
+		// UpdateStatsOnDownload returns a lock error; turning that into a 500 discards a perfectly
+		// good manifest and causes spurious ImagePullBackOff. Log it and serve the manifest.
+		if err := meta.OnGetManifest(name, reference, mediaType, content, rh.c.StoreController, rh.c.MetaDB, rh.c.Log); err != nil {
+			rh.c.Log.Warn().Err(err).Str("name", name).Str("reference", reference).
+				Msg("failed to update download stats; serving manifest anyway")
 		}
 	}
 
--- a/pkg/meta/hooks.go
+++ b/pkg/meta/hooks.go
@@ -343,7 +343,9 @@ func OnGetManifest(name, reference, mediaType string, body []byte,
 
 	err = metaDB.UpdateStatsOnDownload(name, reference)
 	if err != nil {
-		log.Error().Err(err).Str("repository", name).Str("reference", reference).
+		// Best-effort bookkeeping: the caller serves the manifest regardless of this error
+		// (e.g. a contended per-repo metaDB lock under a concurrent pull burst), so log at warn.
+		log.Warn().Err(err).Str("repository", name).Str("reference", reference).
 			Msg("failed to update stats on download image")
 
 		return err


pkg/api/routes.go
```diff
@@ -578,11 +578,15 @@ func (rh *RouteHandler) GetManifest(response http.ResponseWriter, request *http.
...
-			response.WriteHeader(http.StatusInternalServerError)
-
-			return
+		// OnGetManifest only updates best-effort download bookkeeping (download count and
+		// last-pull timestamp). The manifest has already been retrieved and verified above, so a
+		// failure here must not fail the client's read. In particular, a burst of concurrent
+		// pulls of the same repo (e.g. a large fan-out job) contends the per-repo metaDB lock and
+		// UpdateStatsOnDownload returns a lock error; turning that into a 500 discards a perfectly
+		// good manifest and causes spurious ImagePullBackOff. Log it and serve the manifest.
+		if err := meta.OnGetManifest(name, reference, mediaType, content, rh.c.StoreController, rh.c.MetaDB, rh.c.Log); err != nil {
```
andaaron, line 587: Old code skipped ErrImageMetaNotFound / ErrRepoMetaNotFound without logging. Now every such miss warns (twice). Keep serving the manifest, but either restore the silent skip for those two, or log them at Debug.

andaaron (commented):
Can you maybe add a test for the new behavior?

andaaron:
Superseded by https://github.com/project-zot/zot/pull/4403

PR #4405, opened 2026-09-06
perf(storage): read each manifest once per repository walk and prefetch them concurrently

BinSquare:
Overwriting a multi-arch tag and every GC pass read each manifest and index in the repository from storage one round-trip at a time, and a GC pass re-read the same blobs in each of its walks, so on a repository with thousands of tagged builds a single push or collection took minutes. Each repository-wide walk now parses a blob once through a memo scoped to that walk and prefetches the walk's blobs concurrently, so storage stays authoritative between walks while the walk itself runs from memory.

### DIFF
--- a/pkg/storage/common/common.go
+++ b/pkg/storage/common/common.go
@@ -427,6 +427,145 @@ func UpdateIndexWithPrunedImageManifests(imgStore storageTypes.ImageStore, index
 	return nil
 }
 
+// walkReadParallelism bounds concurrent blob reads when a repository-wide walk
+// warms its memo. Remote drivers are latency-bound, so a modest fan-out
+// collapses a long sequential walk without swamping the backend.
+const walkReadParallelism = 32
+
+// concurrentReader is implemented by stores whose reads are safe to issue from
+// many goroutines at once. Anything else, such as a test double, is read
+// sequentially with unchanged semantics.
+type concurrentReader interface {
+	ConcurrentReadSafe() bool
+}
+
+// WalkMemo holds the manifests and indexes one repository-wide walk has parsed,
+// so a digest referenced many times within that walk is read from storage once.
+// It lives for a single walk: a GC pass or one manifest update. Across walks,
+// storage stays authoritative, so a blob altered or removed between them is
+// seen. A nil *WalkMemo reads through without memoising.
+type WalkMemo struct {
+	mu        sync.Mutex
+	indexes   map[godigest.Digest]ispec.Index
+	manifests map[godigest.Digest]ispec.Manifest
+}
+
+// NewWalkMemo returns an empty memo for one walk.
+func NewWalkMemo() *WalkMemo {
+	return &WalkMemo{
+		indexes:   map[godigest.Digest]ispec.Index{},
+		manifests: map[godigest.Digest]ispec.Manifest{},
+	}
+}
+
+// Index is GetImageIndex through the memo.
+func (m *WalkMemo) Index(imgStore storageTypes.ImageStore, repo string, digest godigest.Digest, log zlog.Logger,
+) (ispec.Index, error) {
+	if m == nil {
+		return GetImageIndex(imgStore, repo, digest, log)
+	}
+
+	m.mu.Lock()
+	idx, ok := m.indexes[digest]
+	m.mu.Unlock()
+
+	if ok {
+		return idx, nil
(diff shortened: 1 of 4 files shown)


andaaron:
This is a very interesting PR, I need to review carefully.

My concern is it won't react well if storage mutates outside zot API. I.E. someone uses skopeo copy or a similar tool to copy a new image into local zot storage instead of using the zot API.
Theoretically we support running zot on top of an already exisitng OCI layout, and there may be users which interract with that layout independently of what zot does. So there may be cases where content may be deleted outside of zot control and zot reports it being present based on cache.

BinSquare:
> This is a very interesting PR, I need to review carefully.
> 
> My concern is it won't react well if storage mutates outside zot API. I.E. someone uses skopeo copy or a similar tool to copy a new image into local zot storage instead of using the zot API. Theoretically we support running zot on top of an already exisitng OCI layout, and there may be users which interract with that layout independently of what zot does. So there may be cases where content may be deleted outside of zot control and zot reports it being present based on cache.

The implementation currently only exists for the memo during a single walk for one operation, the GC pass. 

based on my undersanding, for the case where content is deleted and it's stored in the memo. It would be corrected next pass. 

I think the trade comes down to: over-retain (# of deleted blobs) by one gc cycle vs eliminating another read of (~all blobs). 

From my perspective, seems reasonable trade since the # of all blobs can be huge, while # of deleted blobs during the gc memo is minimal in comparison.

pkg/storage/common/common.go
```diff
@@ -427,6 +427,145 @@ func UpdateIndexWithPrunedImageManifests(imgStore storageTypes.ImageStore, index
...
+		if !isIndex && !isManifest {
+			continue
+		}
+
+		m.mu.Lock()
+		_, haveIdx := m.indexes[desc.Digest]
+		_, haveMan := m.manifests[desc.Digest]
+		m.mu.Unlock()
+
+		if (isIndex && haveIdx) || (isManifest && haveMan) {
```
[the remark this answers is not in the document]
andaaron: So the issue seems to be in case a second goroutine starts for the same digest before the 1st one finishes writing it to `indexes` or `manifests`.
Maybe deduplicate the digests in descs or use singleflight to ensure there's one call made per digest.

pkg/storage/common/common.go
```diff
@@ -427,6 +427,145 @@ func UpdateIndexWithPrunedImageManifests(imgStore storageTypes.ImageStore, index
...
+// NewWalkMemo returns an empty memo for one walk.
+func NewWalkMemo() *WalkMemo {
+	return &WalkMemo{
+		indexes:   map[godigest.Digest]ispec.Index{},
+		manifests: map[godigest.Digest]ispec.Manifest{},
+	}
+}
+
+// Index is GetImageIndex through the memo.
+func (m *WalkMemo) Index(imgStore storageTypes.ImageStore, repo string, digest godigest.Digest, log zlog.Logger,
```
andaaron, line 462: Method names should be actions, they should contain verbs.

pkg/storage/imagestore/walk_memo_test.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		t.Fatalf("tags: %v", err)
+	}
+
+	sort.Strings(tags)
+
+	return tags
+}
+
+func sameTags(before, after []string) bool {
```
andaaron, line 70: Isn't there a standard library function we can use instead https://cs.opensource.google/go/go/+/go1.27.1:src/slices/slices.go;l=20?

pkg/storage/common/common.go
```diff
@@ -427,6 +427,145 @@ func UpdateIndexWithPrunedImageManifests(imgStore storageTypes.ImageStore, index
...
+type concurrentReader interface {
+	ConcurrentReadSafe() bool
+}
+
+// WalkMemo holds the manifests and indexes one repository-wide walk has parsed,
+// so a digest referenced many times within that walk is read from storage once.
+// It lives for a single walk: a GC pass or one manifest update. Across walks,
+// storage stays authoritative, so a blob altered or removed between them is
+// seen. A nil *WalkMemo reads through without memoising.
+type WalkMemo struct {
```
andaaron, line 447: I think WalkMemo deserves its own file in this package.

Issue #4323, opened 2026-08-13
[Feat]: Stream on-demand pulls to the client to avoid first-pull timeouts (ImagePullBackOff)


yinsenyan:
### Is your feature request related to a problem? Please describe.

When zot is used as an on-demand pull-through cache (`extensions.sync` with `onDemand: true`),
the **first** pull of an uncached image frequently fails on the client with a timeout, even
though the pull eventually succeeds on retry. This is a poor experience at scale (many
Kubernetes nodes pulling through a single zot from a single upstream).

Root cause — on-demand sync is synchronous behind the manifest request, and streams nothing
to the client while it runs:

- `getImageManifest()` calls `SyncOnDemand.SyncImage(...)` and only returns the manifest
  **after** the sync finishes — `pkg/api/routes.go` (the `getImageManifest` → `SyncImage`
  call).
- `SyncImage` blocks on a channel until the background `syncImage` goroutine has copied the
  **entire** image (manifest + config + all layers) into local storage
  (`ImageCopy` upstream→temp, then `CommitAll` temp→permanent) — `pkg/extensions/sync/on_demand.go`.
- For the whole duration of that copy, the client's `GET /v2/<repo>/manifests/<ref>` request
  receives **zero bytes**. The connection sits idle.

For a large image (or a slow/distant upstream), that idle period exceeds the client's
timeout. With kubelet this means the image pull hits its timeout and the pod goes into
`ImagePullBackOff`; the next attempt (after backoff) succeeds because the data is now cached.
A regular registry does not exhibit this: it already holds the blobs, so `GET /manifests`
returns immediately and `GET /blobs/<digest>` streams bytes continuously, keeping the
connection active.

Note: zot already detaches the sync context so a client disconnect does **not** abort the
in-progress copy —
`syncCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), timeout)` in
`syncImage` (`pkg/extensions/sync/on_demand.go`), with the comment *"ensure sync completes
even if HTTP client disconnects … prevents Kubernetes timeout/retries from aborting
in-progress image downloads."* That change makes the eventual retry succeed, but it is a
mitigation of the symptom, not the cause: **the first client request is still guaranteed to
stall with no data and time out on large images.**

This is the same underlying problem reported in #3359 ("On-demand sync causes image pull
timeouts during first request"), which was closed as *not planned* / `no-issue-activity`
rather than after a design decision, and is related to #2505 (background contexts for pulls).
[...]

PR #4407, opened 2026-09-06
feat(sync): stream image blobs to clients during on-demand sync

### LINKED ISSUE #4386: [Feat]: Sync - Reuse existing layers across tags

We are using Zot for a pull-through cache in an on-prem cluster. Many of our images share a large (~10GB) set of base layers, with only small changes in the later layers.

Pulling two different tags always downloads the full 10GB image, even if the only difference is the last ~100kB layer.

Zot should reuse previously downloaded layers, even when the tag is different.

_No response_

### Additional context

I'm happy to work on a PR for this feature.

rchincha:
Add opt-in streaming for on-demand sync (extensions.sync.registries[].stream): a client pulling a not-yet-synced image gets its manifest and blobs forwarded as they arrive from upstream, instead of waiting for the full image to be committed to local storage first. The real sync into local storage keeps running in the background and singleflight-dedupes concurrent requests for the same reference.

Also fixes on-demand sync re-downloading blobs that are already present locally under a different tag/repo (fixes #4386): predictOCIDigest now collects the manifest tree's blob digests so already-local blobs can be reused instead of re-fetched from upstream.

This is a rewrite of #3778, addressing its review feedback: the chunked reader now respects the caller's buffer size instead of always attempting fixed-size chunks, blob paths are built from a properly parsed digest rather than a hash of the digest string, and cleanup runs reliably on every manifest-casting/layer-retrieval failure path.

Streaming requires onDemand, preserveDigest, and a TLS-verified upstream (bytes may reach a client before the blob's digest is fully verified, so there's no room for an unauthenticated upstream identity or a manifest whose digest could change on commit). A per-registry maxConcurrentStreams caps concurrent streamed blobs; once hit, requests fall back to the ordinary non-streaming on-demand path.

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


rchincha:
cc: @vrajashkr 
based on your PR
Pls. review and check if the TTFB tests still show improvements, etc

rchincha:
https://github.com/project-zot/zot/issues/4323

rchincha:
https://github.com/rchincha/zot/actions/runs/34790528668/job/103814509774

A new streaming nightly test
20GB image streamed to 8 concurrent clients.

pkg/api/routes.go
```diff
@@ -1662,6 +1723,117 @@ func (rh *RouteHandler) GetBlob(response http.ResponseWriter, request *http.Requ
...
+	}
+
+	return true
+}
+
+// streamBlobRangeToClient is streamBlobToClient's Range-request counterpart: it attaches to
+// digest's active stream and serves a single-range 206 response sourced from bytes already on
+// disk plus new bytes as they continue arriving from upstream.
+//
+// Only a single range is supported - multi-range (multipart/byteranges) against a live, still-
```
vrajashkr, line 1778: Is this alright for OCI clients? If they send multi-range, it could break the flow when streaming is enabled. If typical OCI clients are always going to send single ranges, then it should be fine.

pkg/api/routes_blob_stream_test.go
```diff
@@ -0,0 +1,410 @@
+//go:build sync && scrub && metrics && search && lint && userprefs && mgmt && imagetrust && ui
```
vrajashkr, line 1: I think this is probably not needed. API is available in minimal.

pkg/extensions/sync/blob_reuse.go
```diff
@@ -0,0 +1,99 @@
...
+
+		return err
+	}
+
+	if err := dest.Close(); err != nil {
+		return err
+	}
+
+	return os.Rename(tempPath, destPath)
+}
```
vrajashkr, line 99: This looks expensive IMO and increases the IO needed to stream because we have to try and work around regclient's requirement of having the blob available on disk at the temp ref.

IMO, we should find a way to override regclient's blob ref check with something faster like a metaDB lookup. i.e. Skip the copy if the blob is known available (present in metadb) or Skip the copy if the blob Stat in the main image store succeeds.

kwohlfahrt: FWIW, I took a similar approach in #4420 (just the seed feature, no streaming). It copies into the temporary store, though tries to hard-link first to avoid I/O if possible. If the backing store is on object storage or a different disk, it must copy.

Unfortunately `regclient` does not allow overriding the check as far as I can see. Also, the temp copy helps avoid races with GC - if we check and find a blob present, it could be GC'd before the manifest that owns it is pushed, and then we'd have an incomplete image in the store. The staging temporary copy helps avoid that (at the cost of an extra local copy).

IMO, while copying isn't optimal, it is still an improvement to copy from the local store than upstream.

rchincha: The main use case this feature is solving is frequent publishing of large images and downstream clients also have low latency requirement of picking up published images as soon as they are published. When the image is large, they wait timeout for the image to be mirrored first.

Also that this feature is behind a feature gate and enabling this requires large amount of local storage which should be cleaned up once images land inside zot's image storage and then retention policies take over.

pkg/api/routes_blob_stream_test.go
rchincha: sync is not available in minimal right?

PR #4441, opened 2026-09-18
chore: fix dependabot alerts

rchincha:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```


andaaron:
Please fix lint issues.

pkg/storage/gc/gc.go
```diff
@@ -1153,8 +1152,7 @@ func isBlobOlderThan(imgStore types.ImageStore, repo string,
 }
 
 func getSubjectFromCosignTag(tag string) godigest.Digest {
-	alg := strings.Split(tag, "-")[0]
-	encoded := strings.Split(tag, "-")[1]
+	alg, encoded, _ := strings.Cut(tag, "-")
```
vrajashkr, line 1155: Is it guaranteed that the tag that will come here is validated to be in the right format?
strings.Cut returns before, after, and found. In case the tag has 2 hyphens in it, the "after" will be invalid.

If the tag is validated/always generated by code, then no problems here.

vrajashkr (approved):
LGTM. Left one question.

